PR #1, opened 2021-03-09
Lower default sampling frequency to 10

Jongy:
## Description
Affects perf & async-profiler. Python profiling is already capped at 10.

## Motivation and Context
100 is too high for truly continuous.


liorgorb (dismissed):

Jongy:
@d3dave can you please check what the problem is? I the failure remained after re-running, so it's not about flakiness.

liorgorb:
Oh, could it be that sampling rate is too low, and sometimes you don't get the expected symbol in the fg? Need to check...

DanielShaulov (approved):

PR #3, opened 2021-03-11
Update perf to a newer release

DanielShaulov:
## Description
We released a new version of [perf](https://github.com/Granulate/linux/releases/tag/v5.6g2-perf), updating `build.sh` to download it
## Motivation and Context
This perf update will mean that the root stack is the process name instead of the thread name, which is much clearer


Jongy (approved):
LGTM

PR #5, opened 2021-03-11
python: Select processes to profile based on "python*" in their exe's basename

Jongy:
## Description
This is more accurate than "pgrep python", because a Python script exec'd directly won't
have "python" in its comm (hence the different PYTHON_KNOWN_PROCS we pgrep'd for).
This new approach is more accurate, and it can be taken to further to actually finding
processes with ".*python.*" in their /proc/pid/maps file.

## Motivation and Context
Catch more processes we used to miss.

## How Has This Been Tested?
By running the existing tests + I'll add a test for a Python script with a shebang.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/gprofiler/python.py
+++ b/gprofiler/python.py
@@ -9,7 +9,7 @@ from threading import Event
 from typing import Dict
 
 from .exceptions import StopEventSetException, ProcessStoppedException
-from .utils import pgrep, run_process, resource_path
+from .utils import pgrep_exe, run_process, resource_path
 
 logger = logging.getLogger(__name__)
 
@@ -18,7 +18,6 @@ PYTHON_PROFILER_MAX_FREQUENCY = 10
 
 class PythonProfiler:
     BLACKLISTED_PYTHON_PROCS = ["unattended-upgrades", "networkd-dispatcher", "supervisord", "tuned"]
-    PYTHON_KNOWN_PROCS = ["python", "uwsgi", "gunicorn", "celery", "ipython", "python3"]
 
     def __init__(self, frequency: int, duration: int, stop_event: Event, storage_dir: str):
         self._frequency = min(frequency, PYTHON_PROFILER_MAX_FREQUENCY)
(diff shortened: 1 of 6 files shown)


gprofiler/python.py
```diff
@@ -69,7 +68,7 @@ def profile_process(self, pid: int, cmdline: str):
 
     def find_python_processes_to_profile(self) -> Dict[str, str]:
         filtered_procs = {}
-        for process in pgrep("^{}$".format("|".join(self.PYTHON_KNOWN_PROCS))):
+        for process in pgrep_exe(r"^.+/python[^/]+$"):
```
d3dave, line 71:
```suggestion
        for process in pgrep_exe(r"^.+/python[^/]*$"):
```

d3dave (changes requested):

gprofiler/python.py
Jongy: Yes, makes sense

Jongy:
If these pass now, I guess it's good enough as a test proving my fix

d3dave:
No reason to keep the original `pgrep`. Using `exe` for Java should work better as well.

Jongy:
> No reason to keep the original `pgrep`. Using `exe` for Java should work better as well.

Fair - will do

Jongy:
Actually @d3dave I think it doesn't run if the filename isn't `java`. So it doesn't matter. Java won't ever be run with shebang (I think? no textual/script files) so unless someone adds a binfmt for `.class` files, I guess it's irrelevant.

d3dave:
> I think the `comm` (by which we `pgrep`) will always equal the basename of the Java exe.

@Jongy All the more reason to get rid of the `comm` logic then.

d3dave (approved):

Jongy:
@d3dave added a comment

d3dave (approved):
Awesome, thanks 👍

PR #6, opened 2021-03-14
Improve Dockerfile ignored files

Jongy:
## Description
I noticed that some untracked files were copied into the image. Additionally, `__pycache__` and some `*.pyc`s.
This PR copies files more strictly. It's not perfect yet - perfect would be https://stackoverflow.com/q/51693245, or otherwise to maintain the full list of desired files in `.dockerignore`, which isn't worth the hassle, IMO.

## Motivation and Context
Prevent unwanted files from reaching the image.

## How Has This Been Tested?
I built the image, then verified the profiler manages to run and all its needed files are found.

## Types of changes

## Checklist:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,4 @@
 /bin/
 /build/
-__pycache__
-
+**/__pycache__
+**/*.pyc
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,8 @@ RUN ./build.sh
 COPY requirements.txt requirements.txt
 RUN pip3 install --no-cache-dir -r requirements.txt
 
-COPY . .
+COPY LICENSE.md MANIFEST.in README.md setup.py ./
+COPY gprofiler gprofiler
 
 STOPSIGNAL SIGINT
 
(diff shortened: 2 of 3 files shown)


leonid-granulate:
I think generally is not the best idea to build locally for production. The code may contain more not tracked files, changes that didn't pass CI and also without CR. Even manual pipe line

d3dave (approved):

Jongy:
> I think generally is not the best idea to build locally for production. The code may contain more not tracked files, changes that didn't pass CI and also without CR. Even manual pipe line

Agree. Currently we don't have CD so until we get to set it up - I suggest we build by `git clone ... && cd gprofiler && docker build`, to ensure the directory is really clean. I think it's better than to actually clone in the Dockerfile (`RUN git clone`) because it allows for easier local building of the image.

leonid-granulate:
> > I think generally is not the best idea to build locally for production. The code may contain more not tracked files, changes that didn't pass CI and also without CR. Even manual pipe line
> 
> Agree. Currently we don't have CD so until we get to set it up - I suggest we build by `git clone ... && cd gprofiler && docker build`, to ensure the directory is really clean. I think it's better than to actually clone in the Dockerfile (`RUN git clone`) because it allows for easier local building of the image.

Yes it could be a temporary solution, anyway add a manual action to Github is easy as well .. I can do that if needed

Jongy:
> anyway add a manual action to Github is easy as well .. I can do that if needed

I suggest we add it on tags

PR #9, opened 2021-03-21
python: Select processes to profile based on python/libpython in their maps

Jongy:
## Description

In 3e9381bc4ff7 I suggested this change as an improvement, but it's actually required
because we currently miss executables which load libpython, such as uwsgi.

This change means that we'll prefer the Python samples from *any* process which has
libpython loaded, even if it doesn't even run Python code now. We might need to address
that later, by making smarter decisions on the source for our samples.

Fixes: 3e9381bc4ff7

## Motivation and Context

To enable profiling of processes whose basename isn't `python`, such as `uwsgi`.

## How Has This Been Tested?
1. By running `uwsgi` on my box and seeing it "get caught" in the profiler.
2. By the new test `test_python_select_by_libpython`.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
## Checklist:

### DIFF
--- a/gprofiler/python.py
+++ b/gprofiler/python.py
@@ -11,7 +11,7 @@ from typing import Dict
 
 from .merge import parse_collapsed
 from .exceptions import StopEventSetException, ProcessStoppedException
-from .utils import pgrep_exe, run_process, resource_path
+from .utils import pgrep_maps, run_process, resource_path
 
 logger = logging.getLogger(__name__)
 
@@ -70,7 +70,7 @@ class PythonProfiler:
 
     def find_python_processes_to_profile(self) -> Dict[str, str]:
         filtered_procs = {}
-        for process in pgrep_exe(r"^.+/python[^/]*$"):
+        for process in pgrep_maps(r"^.+/(?:lib)?python[^/]*$"):
             try:
                 if process.pid == os.getpid():
                     continue
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -105,5 +105,10 @@ def pgrep_exe(match: str) -> Iterator[Process]:
     return (process for process in psutil.process_iter() if pattern.match(process.exe()))
 
 
+def pgrep_maps(match: str) -> Iterator[Process]:
+    pattern = re.compile(match)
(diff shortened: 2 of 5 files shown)


Jongy:
About the performance overhead of iterating over all processes and reading all `/proc/pid/maps` files: I `perf`ed gprofiler's Python process (includes all child `perf`/`py-spy`s) on my machine (400~ processes). Ran `perf record` for a few mins, with gprofiler running with `-d 10 -c` (so it actually re-runs the selection logic every once in a short while).

Kernel time inside the implementation of `/proc/pid/maps` accounts for about 3.5% of the all gprofiler samples. Usermode `python` samples have increased as well, roughly by 5%.

Not negligible - but I still prefer this solution (for its correctness) over what we had before #5.

d3dave (approved):
lgtm

Jongy:
> About the performance overhead of iterating over all processes and reading all `/proc/pid/maps` files: I `perf`ed gprofiler's Python process (includes all child `perf`/`py-spy`s) on my machine (400~ processes). Ran `perf record` for a few mins, with gprofiler running with `-d 10 -c` (so it actually re-runs the selection logic every once in a short while).
> 
> Kernel time inside the implementation of `/proc/pid/maps` accounts for about 3.5% of the all gprofiler samples. Usermode `python` samples have increased as well, roughly by 5%.
> 
> Not negligible - but I still prefer this solution (for its correctness) over what we had before #5.
> 
> It _is_ negligible when using `-d 60`, which is the current default

I was confused - `-d ...` is not enough, I also needed to add `-i` to get it to run more than once per minute. `-i` accepts minutes so I modified the code manually to use 5s interval, and ran with `-d 5`, then measured the performance. It is visible but still not very very prominent. Good enough for now

Jongy:
Ran again - gprofiler with default settings, in continuous mode; `perf record -a -F 99` alongside. For 20~ mins. Concluding: `python3` is quite low (0.4%), `py-spy` is *very* low (<0.1%) and `perf` is about 1% ~. Will need to optimize `perf` itself, eventually :)

Good enough.

PR #12, opened 2021-03-29
utils: Provide a more performant implementation of pgrep_maps()

Jongy:
## Description
Provide a faster implementation of `pgrep_maps` from https://github.com/Granulate/gprofiler/pull/9.

According to the tests I conducted in that PR, the old `pgrep_maps` wasn't the core issue in gprofiler. However, as the benchmarks here show, it is very slow (compared to what we could get):
```
root@0728b99acd88:/# cat bench.py 
from __future__ import annotations
import subprocess
import sys
from psutil import Process
import re
import psutil

if sys.argv[1] == "new":
    def pgrep_maps(match: str) -> List[Process]:
        # this is much faster than iterating over processes' maps, comparing to psutil.
        result = subprocess.run(
            f"grep -lE '{match}' /proc/*/maps", stdout=subprocess.PIPE, stderr=subprocess.DEVNULL, shell=True, text=True
        )
        # might get 2 (which 'grep' exits with, if some files were unavailable, because processes have exited)
        assert result.returncode in (0, 2)

        processes: List[Process] = []
        for line in result.stdout.splitlines():
            assert line.startswith("/proc/") and line.endswith("/maps")
            pid = int(line[len("/proc/") : -len("/maps")])
            processes.append(Process(pid))

        return processes
else:
    def pgrep_maps(match: str) -> Iterator[Process]:
        pattern = re.compile(match)
        return (process for process in psutil.process_iter() if any(pattern.match(m.path) for m in process.memory_maps()))

for i in range(5):
    print(list(pgrep_maps(r"^.+/(?:lib)?python[^/]*$")))

root@0728b99acd88:/# time python bench.py old
....

real	0m29.349s
user	0m25.708s
sys	0m3.636s
root@0728b99acd88:/# time python bench.py new
....

real	0m0.400s
user	0m0.069s
sys	0m0.333s
```

(I redacted the printed stuff, in both cases it's the same set of processes)

Sometimes it's better to microbenchmark instead of checking the total effect, I guess (as I did in #9 )

## Motivation and Context
Reducing gprofiler's overhead.

## How Has This Been Tested?
I ran it locally, and ran the tests.

## Types of changes
## Checklist:

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+max-line-length = 120
+# options for compatibility with black - see https://black.readthedocs.io/en/stable/compatible_configs.html#flake8
+extend-ignore = E203, W503
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -105,9 +105,21 @@ def pgrep_exe(match: str) -> Iterator[Process]:
     return (process for process in psutil.process_iter() if pattern.match(process.exe()))
 
 
-def pgrep_maps(match: str) -> Iterator[Process]:
-    pattern = re.compile(match)
-    return (process for process in psutil.process_iter() if any(pattern.match(m.path) for m in process.memory_maps()))
+def pgrep_maps(match: str) -> List[Process]:
+    # this is much faster than iterating over processes' maps with psutil.
+    result = subprocess.run(
+        f"grep -lP '{match}' /proc/*/maps", stdout=subprocess.PIPE, stderr=subprocess.DEVNULL, shell=True
+    )
+    # might get 2 (which 'grep' exits with, if some files were unavailable, because processes have exited)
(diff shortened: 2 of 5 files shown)


Jongy:
@d3dave I was thinking perhaps to avoid `grep -P`, as I'm not sure if it's consistent across operating systems (thinking of a future, standalone execution of gprofiler, without containers). Do you have any idea?

Jongy:
> @d3dave I was thinking perhaps to avoid `grep -P`, as I'm not sure if it's consistent across operating systems (thinking of a future, standalone execution of gprofiler, without containers). Do you have any idea?

Tested on alpine - after `apk add grep` it works just fine (well, it's the same GNU grep). I guess it's okay.

d3dave (approved):

PR #20, opened 2021-04-08
Check for availability of "nsenter" before attempting to use it

Jongy:
## Description
Check whether `nsenter` is installed or not, and raise a descriptive error if not.

## Motivation and Context
Makes it more more clear that you need to install `nsenter`, when compared to `FileNotFoundError: [Errno 2] No such file or directory: 'nsenter'`.

## How Has This Been Tested?
Ran the container normally - it works fine & profiles Java, because `nsenter` exists.

Then I added some logic to remove `/usr/bin/nsenter` and I got the nice exception.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,9 @@ To upload results, you will have to register and generate a token on the website
 
 ## Requirements
 gProfiler works on **Linux** and requires **Python 3.6+** to run.
+
+The `nsenter` program needs to be installed for Java profiling. For Debian/Ubuntu, install the `util-linux` package.
+
 It can produce specialized stack traces for the following runtimes:
 * Java runtimes (version 7+) based on the HotSpot JVM,
 including the Oracle JDK and other builds of OpenJDK like AdoptOpenJDK and Azul Zulu.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -8,6 +8,9 @@ To upload results, you will have to register and generate a token on the website
 
 ## Requirements
 gProfiler works on **Linux** and requires **Python 3.6+** to run.
+
+The `nsenter` program needs to be installed for Java profiling. For Debian/Ubuntu, install the `util-linux` package.
```
leonid-granulate, line 12: Maybe it worth to mention that `util-linux` in ubuntu 14.04 not include `nsenter`. Or delete the section about Debian/Ubuntu

Jongy: Until we provide a solution for that (either by bringing that `nsenter` with us, or by not using it) - I'll let the users handle the directions as given in https://askubuntu.com/questions/439056/why-there-is-no-nsenter-in-util-linux

leonid-granulate (approved):

PR #25, opened 2021-04-10
Log system info upon startup

Jongy:
## Description
Log some generic system info upon startup, so we have some clue on where did the profier run when inspecting its logs.

## Motivation and Context
Helps debugging the profiler using its logs.

## How Has This Been Tested?
Manually.

* Tested the container on my box.
* Tested standalone on my box (it doesn't try to `setns` when not needed)
* Also tested `get_libc_version()` on Alpine.

## Types of changes

## Checklist:

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -7,6 +7,7 @@ import datetime
 import logging
 import logging.config
 import logging.handlers
+import platform
 import os
 import sys
 import time
@@ -19,6 +20,7 @@ from threading import Event
 
 import configargparse
 from requests import RequestException, Timeout
+import psutil
 
 from . import __version__
 from . import merge
@@ -243,12 +245,21 @@ def verify_preconditions():
     return True
 
 
+def log_system_info():
+    uname = platform.uname()
+    logger.info(f"Kernel uname release: {uname.release}")
+    logger.info(f"Kernel uname version: {uname.version}")
+    logger.info(f"Total CPUs: {os.cpu_count()}")
+    logger.info(f"Total RAM: {psutil.virtual_memory().total / (1 << 30):.2f} GB")
+
+
 def main():
     args = parse_cmd_args()
     setup_logger(logging.DEBUG if args.verbose else logging.INFO, args.log_file)
 
     try:
         logger.info(f"Running gprofiler (version {__version__})...")
+        log_system_info()
 
         if not verify_preconditions():
             sys.exit(1)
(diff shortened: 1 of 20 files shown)


gprofiler/main.py
```diff
@@ -243,12 +245,21 @@ def verify_preconditions():
     return True
 
 
+def log_system_info():
+    uname = platform.uname()
+    logger.info(f"Kernel uname release: {uname.release}")
+    logger.info(f"Kernel uname version: {uname.version}")
+    logger.info(f"Total CPUs: {os.cpu_count()}")
+    logger.info(f"Total RAM: {psutil.virtual_memory().total / (1 << 30):.2f} GB")
+
```
liorgorb, line 254: Let's add `platform.linux_distribution()`.

liorgorb (changes requested):

gprofiler/main.py
Jongy: Good idea. But no such function

Jongy: There's https://pypi.org/project/distro/

Jongy: But that'd require escaping the container :/ Otherwise we'll just get Ubuntu 20.04.

Easiest would be `nsenter -t 1 -m cat /etc/os-release`

liorgorb: `/etc/os-release` isn't enough :(

`distro` does it like this:
```
* The os-release file /etc/os-release, if present.
* The output of the lsb_release command, if available.
* The distro release file (/etc/*(-|_)(release|version)), if present.
```
By the way I'm not sure how important the linux distro is if we're running inside a container...? But probably better safe than sorry.

Jongy: 1. I'm thinking ahead, for the pyinstaller version
2. It's important also when we're a container (for example - if OS is alpine and there's some Java running on the host, we need to use musl-based async-profiler, which we currently don't employ. Distro name in the logs will explain that)

liorgorb: What I meant was that in the host we can just use `distro`... I do agree that it would be better to just cover all cases.

Jongy: I'll go with this:
1. If `is_same_ns(1, "mnt")` - just run `distro`.
2. If not, `fork()` then `setns` and then run `distro`. Then I need to pipe back the information.

Jongy: Done

liorgorb (approved):

PR #35, opened 2021-04-21
Feature/stacks container name

alonbentamar:
## Description
The V2 profile upload API now sends the container name for stacks from processes that are inside a Docker container 
An example profile file:
```
#{"containers": ["amazing_lamport", "datadog_1"], "hostname": "ip-10-0-0-9", "container_names_disabled": false}
;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];default_idle_call_[k];arch_cpu_idle_[k];native_safe_halt_[k] 51
;swapper;secondary_startup_64_[k];x86_64_start_kernel_[k];x86_64_start_reservations_[k];start_kernel_[k];arch_call_rest_init_[k];rest_init_[k];cpu_startup_entry_[k];do_idle_[k];default_idle_call_[k];arch_cpu_idle_[k];native_safe_halt_[k] 46
datadog_1;process-agent;[/opt/datadog-agent/embedded/bin/process-agent];[/opt/datadog-agent/embedded/bin/process-agent];_start;_start;_start 1
;py-spy;process_vm_readv 1
amazing_lamport;python;<module> (<string>) 100
```
## Checklist:


gprofiler/docker_client.py
```diff
@@ -0,0 +1,80 @@
+import logging
```
Jongy, line 1: Add license to all new files

gprofiler/docker_client.py
```diff
@@ -0,0 +1,80 @@
...
+                           '"-v /var/run/docker.sock:/var/run/docker.sock". Otherwise, please open a new issue here: '
+                           'https://github.com/Granulate/gprofiler/issues/new')
+            self._client = None
+
+        self._pid_to_container_name_cache: Dict[int, str] = {}
+        self._container_id_to_name_cache: Dict[str, str] = {}
+
+    def reset_cache(self):
+        self._pid_to_container_name_cache = {}
+        self._container_id_to_name_cache = {}
```
Jongy, line 29: Container IDs are unique enough to be cached forever, I think.

gprofiler/docker_client.py
```diff
@@ -0,0 +1,80 @@
...
+        if container_name is None:
+            self._pid_to_container_name_cache[pid] = ''
+            return f';{stack_line}'
+        self._pid_to_container_name_cache[pid] = container_name
+        return f'{container_name};{stack_line}'
+
+    def _get_process_container_name(self, pid: int) -> Optional[str]:
+        try:
+            container_id = self._get_process_container_id(pid)
+            if not container_id:
```
Jongy, line 50:
```suggestion
            if container_id is None:
```

gprofiler/docker_client.py
```diff
@@ -0,0 +1,80 @@
...
+        self._container_id_to_name_cache: Dict[str, str] = {}
+
+    def reset_cache(self):
+        self._pid_to_container_name_cache = {}
+        self._container_id_to_name_cache = {}
+
+    def get_container_names(self) -> List[str]:
+        return list(self._container_id_to_name_cache.values())
+
+    def get_stack_line_with_container_name(self, stack_line: str, pid: int) -> str:
```
Jongy, line 34: The logic of embedding the name into the line repeats multiple times here. Instead, I think a nicer API would be if this function will return the container name or `''` give a PID, and the caller will do the formatting.

gprofiler/docker_client.py
```diff
@@ -0,0 +1,80 @@
...
+                           'by running the Docker run command with the following argument: '
+                           '"-v /var/run/docker.sock:/var/run/docker.sock". Otherwise, please open a new issue here: '
+                           'https://github.com/Granulate/gprofiler/issues/new')
+            self._client = None
+
+        self._pid_to_container_name_cache: Dict[int, str] = {}
+        self._container_id_to_name_cache: Dict[str, str] = {}
+
+    def reset_cache(self):
+        self._pid_to_container_name_cache = {}
```
Jongy, line 28: Can use `.clear()`

gprofiler/docker_client.py
sobolron: I'd reset them from time to time just to make sure we don't have a 9 billion keys map.

sobolron (approved):
I've got nothing to add bsically.
Make sure you fix @Jongy comments.

gprofiler/docker_client.py
Jongy: Can use `functools.lru_cache` on the translation function.

ezean21 (commented):
can we add hostname in first line like: 
#{"containers": ["datadog_1", "dazzling_wescoff"], "hostname": "any-hostname-example-001"}

gprofiler/merge.py
```diff
@@ -94,7 +100,11 @@ def merge_perfs(perf_all: Iterable[Mapping[str, str]], process_perfs: Mapping[in
...
             for stack, count in process_stacks.items():
-                full_stack = ";".join([process_names[pid], stack])
-                new_samples[full_stack] += round(count * ratio)
-
-    return "\n".join((f"{stack} {count}" for stack, count in new_samples.items()))
+                stack_line = ";".join([process_names[pid], stack])
+                stack_line_with_container_name = docker_client.get_stack_line_with_container_name(stack_line, pid)
+                new_samples[stack_line_with_container_name] += round(count * ratio)
+    container_names = docker_client.get_container_names()
+    docker_client.reset_cache()
```
talsaiag, line 107: please include only `get_container_names` that currently have stack sampled from within them - not all in the cache / running now
UNLESS
are you resetting the cache every time? @Jongy is this legit performance wise or should we prolong the values across runs?

gprofiler/merge.py
```diff
@@ -94,7 +100,11 @@ def merge_perfs(perf_all: Iterable[Mapping[str, str]], process_perfs: Mapping[in
...
+    output = [f"#{json.dumps({'containers': container_names})}"]
```
talsaiag, line 108: as @ezean21 asked, can you please also include the `hostname` in the awesome metadata dict?
thanks!

gprofiler/merge.py
```diff
@@ -73,7 +76,8 @@ def parse_perf_script(script: str):
             logger.exception(f"Error processing sample: {sample}")
 
 
-def merge_perfs(perf_all: Iterable[Mapping[str, str]], process_perfs: Mapping[int, Mapping[str, int]]) -> str:
+def merge_perfs(perf_all: Iterable[Mapping[str, str]], process_perfs: Mapping[int, Mapping[str, int]],
```
talsaiag, line 79: the format of the output file has changed (and now each sample includes the container-name);
@Jongy - should we strip the container names from the file when generating an svg file on the remote?
(please bare in mind that the logic for extracting a meaningful name from the k8s-pods's containers is not present here and only exists in the backend)

talsaiag (changes requested):

gprofiler/merge.py
Jongy: > please include only get_container_names that currently have stack sampled from within them - not all in the cache / running now

We attempt to get the container ID only for PIDs present in the stackcollapsed output, so that should be alright.

> are you resetting the cache every time? @Jongy is this legit performance wise or should we prolong the values across runs?

Resetting the PIDs cache is okay (and desired). Reading `/proc/pid/cgroup` per PID should be okay.

Resetting the container ID <-> name cache should be avoided (like I mentioned earlier). We need to measure the impact, but I presume that querying the Docker daemon for tens/hundreds of container names every interval is *not* desired.

Jongy: Actually, @alonbentamar I have a better suggestion: you will manage an ID <-> name cache, but instead of querying Docker for every missing mapping, you will query just *once* via `_client.containers.list(...)` and update the cache with all results. This way we avoid multiple Docker API queries.

gprofiler/merge.py
Jongy: In the remote - we need to discuss.

Locally - I think it's nice to keep, but we need @alonbentamar to show an example of how it looks (it might be too long and be rendered badly; in that case we should remove)

gprofiler/main.py
```diff
@@ -20,13 +20,14 @@
...
-from . import merge
-from .client import APIClient, APIError, GRANULATE_SERVER_HOST, DEFAULT_UPLOAD_TIMEOUT
-from .java import JavaProfiler
-from .perf import SystemProfiler
-from .python import PythonProfiler
-from .utils import is_root, run_process, get_iso8061_format_time, resource_path, log_system_info, TEMPORARY_STORAGE_PATH
+from gprofiler.docker_client import DockerClient
+from gprofiler import __version__
+from gprofiler import merge
+from gprofiler.client import APIClient, APIError, GRANULATE_SERVER_HOST, DEFAULT_UPLOAD_TIMEOUT
```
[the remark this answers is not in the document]
Jongy: Need to add `isort` :)

gprofiler/merge.py
talsaiag: this PR should not get in till we have fully supported in the backend as well @alonbentamar @ezean21 

regarding the remote - lets talk :)

talsaiag:
@alonbentamar please pull master after https://github.com/Granulate/gprofiler/pull/37 and update the command line there as well to include the new mount

gprofiler/docker_client.py
alonbentamar: Pretty much the same code as I need to save the current list of container names but did it anyway 🤷

alonbentamar:
> Can we add hostname in first line like:
> #{"containers": ["datadog_1", "dazzling_wescoff"], "hostname": "any-hostname-example-001"}
> @talsaiag wdyt?

Talked about this, and decided it's not needed

gprofiler/python.py
```diff
@@ -11,13 +11,13 @@
 from threading import Event
 from typing import List, Mapping, Optional, Union
 
-from gprofiler.merge import parse_collapsed
-from gprofiler.exceptions import StopEventSetException, ProcessStoppedException
-from gprofiler.utils import pgrep_maps, run_process, resource_path
+from psutil import Process
```
talsaiag, line 14: double line

gprofiler/docker_client.py
```diff
@@ -22,43 +27,45 @@ def __init__(self):
...
+            return self._get_container_name(container_id)
         except Exception:
             logger.warning(f'Could not get a container name for PID {pid}', exc_info=True)
             return None
 
+    @functools.lru_cache(4096)
+    def _get_container_name(self, container_id):
+        container = self._client.containers.get(container_id)
+        container_name = container.name
+        self._current_container_names.add(container_name)
```
talsaiag, line 66: calling this function with the same `container_id` in a new run where the `lru_cache` is not empty might cause it to skip `self._current_container_names.add(container_name)` which should be called - move outside?

gprofiler/merge.py
```diff
@@ -42,6 +42,54 @@ def parse_one_collapsed(collapsed: str) -> Mapping[str, int]:
     return dict(stacks)
 
 
+def parse_many_collapsed(text: str) -> Mapping[int, Mapping[str, int]]:
```
talsaiag: duplicated

gprofiler/main.py
```diff
@@ -20,13 +20,14 @@
...
-from . import __version__
-from . import merge
-from .client import APIClient, APIError, GRANULATE_SERVER_HOST, DEFAULT_UPLOAD_TIMEOUT
-from .java import JavaProfiler
-from .perf import SystemProfiler
-from .python import get_python_profiler
-from .utils import is_root, run_process, get_iso8061_format_time, resource_path, log_system_info, TEMPORARY_STORAGE_PATH
+from gprofiler.docker_client import DockerClient
+from gprofiler import __version__
+from gprofiler import merge
```
talsaiag, line 25: merge both lines?

talsaiag (changes requested):

gprofiler/docker_client.py
alonbentamar: Changed logic, so this is irrelevant

gprofiler/docker_client.py
```diff
@@ -59,12 +59,21 @@ def _safely_get_process_container_name(self, pid: int) -> Optional[str]:
...
-    def _get_container_name(self, container_id):
-        container = self._client.containers.get(container_id)
-        container_name = container.name
-        self._current_container_names.add(container_name)
-        return container_name
+    def _get_container_name(self, container_id) -> Optional[str]:
+        if container_id in self._container_id_to_name_cache:
+            return self._container_id_to_name_cache[container_id]
+
+        self.refresh_container_names_cache()
```
talsaiag, line 69: I would not cause a `cache.clear()` here (when calling `refresh_container_names_cache`) and only append to the list of known container-ids, since a container might die during this and I would still want its name :)

talsaiag (changes requested):

gprofiler/merge.py
sobolron: > In the remote - we need to discuss.
> 
> Locally - I think it's nice to keep, but we need @alonbentamar to show an example of how it looks (it might be too long and be rendered badly; in that case we should remove)

It will be too long... for sure. Just think about a k8s container...
I think I'd prefer stripping it down.

gprofiler/docker_client.py
sobolron: @Jongy Please resolve your conversations when you finish going over fixes..

gprofiler/__init__.py
```diff
@@ -1 +1,5 @@
-__version__ = "0.0.7"
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+__version__ = "0.0.8"
```
Jongy, line 5: We will release the next version probably before merging this PR, so you can drop it. Also, I think it's better in general if we bump the version in separate PRs (because after merging this & before releasing a new version, we will want to run some extra testing on the version before publishing it)

gprofiler/main.py
```diff
@@ -77,15 +79,15 @@ def _generate_output_files(
         flamegraph_path = base_filename + ".html"
         flamegraph = (
             Path(resource_path("flamegraph/flamegraph_template.html"))
-            .read_text()
```
Jongy, line 80: Which `black` version are you using? Why is it changing everything o.O

Let's ping it? (Can pin your current mypy/flake8/... as well)

gprofiler/merge.py
```diff
@@ -118,7 +124,10 @@ def merge_perfs(perf_all: Iterable[Mapping[str, str]], process_perfs: Mapping[in
...
             for stack, count in process_stacks.items():
-                full_stack = ";".join([process_names[pid], stack])
-                new_samples[full_stack] += round(count * ratio)
-
-    return "\n".join((f"{stack} {count}" for stack, count in new_samples.items()))
+                stack_line = ";".join([docker_client.get_container_name(pid), process_names[pid], stack])
+                new_samples[stack_line] += round(count * ratio)
+    container_names = docker_client.container_names
+    docker_client.reset_cache()
+    output = [f"#{json.dumps({'containers': container_names, 'hostname': socket.gethostname()})}"]
```
Jongy, line 131: What's the hostname doing here? Let's discuss it in Slack

gprofiler/merge.py
Jongy: k can strip it

gprofiler/docker_client.py
```diff
@@ -0,0 +1,101 @@
...
+        self._refresh_container_names_cache()
+        if container_id not in self._container_id_to_name_cache:
+            self._container_id_to_name_cache[container_id] = None
+            return None
+        container_name = self._container_id_to_name_cache[container_id]
+        self._current_container_names.add(container_name)
+        return container_name
+
+    def _refresh_container_names_cache(self):
+        self._container_id_to_name_cache.clear()
```
Jongy, line 78: Can explain why it's legit to drop the cache ("we are refreshing it with all currently executing container .... so all entries not to be re-added are certainly stale")

gprofiler/merge.py
Jongy: Good now.

gprofiler/docker_client.py
Jongy: If a container died here, we won't see its ID in `/proc/pid/cgroup`. The only chance for us to see a container ID who's eliminated from the cache here is if it died between the time we read its `/proc/pid/cgroup` to the time we tried to get its name by ID. Pretty rare race IMO.

gprofiler/docker_client.py
```diff
@@ -0,0 +1,101 @@
...
+        running_containers = self._client.containers.list()
+        for container in running_containers:
+            self._container_id_to_name_cache[container.id] = container.name
+
+    @staticmethod
+    def _get_process_container_id(pid: int) -> Optional[str]:
```
Jongy, line 84: See `get_process_container_id` in utils.py

gprofiler/merge.py
alonbentamar: @talsaiag and @ezean21 asked for it. The hostname will also show up in the filename that's saved to S3.

gprofiler/main.py
alonbentamar: I used PyCharm reformatting

gprofiler/docker_client.py
alonbentamar: My code seems a bit better, but I added :/lxc/

Jongy: Merge the implementations then

alonbentamar: Removing utils implementation then

gprofiler/main.py
Jongy: We use `black` and linting fails because yours doesn't conform :/ configure your pycharm to use it.

alonbentamar:
@talsaiag @Jongy re-CR?

talsaiag (approved):
LGTM @Jongy?

Jongy:
Fix linting and I'll take another look

Jongy:
Also, please make it configurable. Like `--no-docker` or `--no-container-names`

alonbentamar:
> Also, please make it configurable. Like `--no-docker` or `--no-container-names`
@Jongy 

So the container name should be empty, i.e. every stack line will start with `;`?

Jongy:
> > Also, please make it configurable. Like `--no-docker` or `--no-container-names`
> 
> @Jongy
> So the container name should be empty, i.e. every stack line will start with `;`?

I think it's nicer without - this will allow using our `.col` outputs as standard stackcollapse format (`comm; stack1; stack2; ...`)

Jongy:
You can run `ln -s "../../lint.sh" "$(git rev-parse --show-toplevel)/.git/hooks/pre-commit`. I'll also add to the README

alonbentamar:
> > > Also, please make it configurable. Like `--no-docker` or `--no-container-names`
> > 
> > 
> > @Jongy
> > So the container name should be empty, i.e. every stack line will start with `;`?
> 
> I think it's nicer without - this will allow using our `.col` outputs as standard stackcollapse format (`comm; stack1; stack2; ...`)

I'm already handling the `.col` files by removing any added chars. I'll keep the `;` at the start when sending to the performance studio for consistency

alonbentamar:
> You can run `ln -s "../../lint.sh" "$(git rev-parse --show-toplevel)/.git/hooks/pre-commit`. I'll also add to the README

Tried it and it completely failed so I gave up and ran the commands in `lint.sh` individually 🤷

alonbentamar:
@Jongy that was quite a pain the CI is finally working (BTW, it's a bad idea to have blackbox tests in the CI like this, they are slow and inconsistent - I had to run them a few times until they worked)
Can you re-CR?

gprofiler/docker_client.py
```diff
@@ -0,0 +1,106 @@
...
+        # standard Docker uses /docker/container-id
+        # k8s uses /kubepods/{burstable,besteffort}/uuid/container-id
+        try:
+            with open(f"/proc/{pid}/cgroup", 'r') as cgroup_file:
+                cgroup = cgroup_file.read()
+        except FileNotFoundError:
+            # The process died before we got to this point
+            return None
+        for line in cgroup.split():
+            if any(s in line for s in (':/docker/', ':/ecs/', ':/kubepods', ':/lxc/')):
```
Jongy, line 100: Can remove `lxc` from here (it's not Docker anyway, so you won't be able to use its ID to match names)

gprofiler/merge.py
```diff
@@ -118,7 +130,20 @@ def merge_perfs(perf_all: Iterable[Mapping[str, str]], process_perfs: Mapping[in
...
-    return "\n".join((f"{stack} {count}" for stack, count in new_samples.items()))
+                container_name = _get_container_name(pid, docker_client, should_determine_container_names)
+                stack_line = ";".join([container_name, process_names[pid], stack])
+                new_samples[stack_line] += round(count * ratio)
+    container_names = docker_client.container_names
+    docker_client.reset_cache()
+    profile_metadata = {
+        'containers': container_names,
+        'hostname': socket.gethostname(),
+        'container_names_disabled': not should_determine_container_names,
```
Jongy, line 141:
```suggestion
        'container_names_enabled': should_determine_container_names,
```

Why the negative?

gprofiler/main.py
```diff
@@ -281,6 +304,13 @@ def parse_cmd_args():
 
     parser.add_argument("-v", "--verbose", action="store_true", default=False, dest="verbose")
     parser.add_argument("--log-file", action="store", type=str, dest="log_file", default=DEFAULT_LOG_FILE)
+    parser.add_argument(
+        "--disable-container-names",
+        action="store_true",
+        dest="disable_container_names",
+        default=False,
+        help="The gProfiler won't gather the container names of processes that run in " "containers",
```
Jongy, line 312:
```suggestion
        help="gProfiler won't gather the container names of processes that run in " "containers",
```

Also, the string is fucked up (2 stray `"`)

gprofiler/main.py
```diff
@@ -80,6 +89,8 @@ def __init__(self, frequency: int, duration: int, output_dir: str, flamegraph: b
             self._frequency, self._duration, self._stop_event, self._temp_storage_dir.name
         )
         self.initialize_python_profiler()
+        self._docker_client = DockerClient()
+        self._should_determine_container_names = should_determine_container_names
```
Jongy, line 93:
```suggestion
        self._include_container_names = include_container_names
```

Simpler name :)

gprofiler/main.py
```diff
@@ -287,6 +310,13 @@ def parse_cmd_args():
 
     parser.add_argument("-v", "--verbose", action="store_true", default=False, dest="verbose")
     parser.add_argument("--log-file", action="store", type=str, dest="log_file", default=DEFAULT_LOG_FILE)
+    parser.add_argument(
+        "--disable-container-names",
+        action="store_true",
+        dest="disable_container_names",
+        default=False,
+        help="gProfiler won't gather the container names of processes that run in " "containers",
```
Jongy, line 318:
```suggestion
        help="gProfiler won't gather the container names of processes that run in containers",
```

gprofiler/main.py
```diff
@@ -327,6 +349,14 @@ def parse_cmd_args():
         dest="log_rotate_backup_count",
         default=DEFAULT_LOG_BACKUP_COUNT,
     )
+    parser.add_argument("--log-file", action="store", type=str, dest="log_file", default=DEFAULT_LOG_FILE)
```
Jongy, line 352: These were already added ^^

Jongy (approved):

alonbentamar:
@talsaiag want to go over it again? There were some changes since your last review

sobolron:
Any reason this is not merged? @alonbentamar

alonbentamar:
> Any reason this is not merged? @alonbentamar
@sobolron 
Waiting for Saiag to respond (and I need to fix the conflicts)

Jongy (approved):

PR #42, opened 2021-04-26
python: PyPerf: Faster initialization + timeouts in snapshot()

Jongy:
## Description
I see no reason to run the PyPerf bootstrap checks again right after running `PythonEbpfProfiler.test`. It only slows down the boostrap of gProfiler.

I added timeouts in `_dump()` (called every `snapshot()`) so we can catch dead/hanging PyPerfs.

## Motivation and Context
Faster bootstrap of gProfiler.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/gprofiler/exceptions.py
+++ b/gprofiler/exceptions.py
@@ -18,14 +18,12 @@ class CalledProcessError(subprocess.CalledProcessError):
     def __str__(self):
         if self.returncode and self.returncode < 0:
             try:
-                return f"Command '{self.cmd}' died with {signal.Signals(-self.returncode)!r}."
+                base = f"Command '{self.cmd}' died with {signal.Signals(-self.returncode)!r}."
             except ValueError:
-                return f"Command '{self.cmd}' died with unknown signal {-self.returncode}."
+                base = f"Command '{self.cmd}' died with unknown signal {-self.returncode}."
         else:
-            return (
-                f"Command '{self.cmd}' returned non-zero exit status {self.returncode}. "
-                f"stderr: {self.stderr}, stdout: {self.stdout}"
-            )
+            base = f"Command '{self.cmd}' returned non-zero exit status {self.returncode}. "
+        return base + f"stderr: {self.stderr}, stdout: {self.stdout}"
 
 
 class ProgramMissingException(Exception):
--- a/gprofiler/exceptions.py
+++ b/gprofiler/exceptions.py
@@ -23,7 +23,7 @@ class CalledProcessError(subprocess.CalledProcessError):
                 base = f"Command '{self.cmd}' died with unknown signal {-self.returncode}."
         else:
             base = f"Command '{self.cmd}' returned non-zero exit status {self.returncode}. "
-        return base + f"stderr: {self.stderr}, stdout: {self.stdout}"
+        return f"{base}\nstdout: {self.stdout}\nstderr: {self.stderr}"
 
 
 class ProgramMissingException(Exception):
--- a/gprofiler/python.py
+++ b/gprofiler/python.py
@@ -9,7 +9,7 @@ import signal
 import glob
 from pathlib import Path
 from threading import Event
-from typing import List, Mapping, Optional, Union
+from typing import Callable, List, Mapping, Optional, Union
 from subprocess import Popen
 
 from psutil import Process
@@ -24,11 +24,19 @@ logger = logging.getLogger(__name__)
 class PythonProfilerBase:
     MAX_FREQUENCY = 100
 
-    def __init__(self, frequency: int, duration: int, stop_event: Optional[Event], storage_dir: str):
+    def __init__(
+        self,
+        frequency: int,
+        duration: int,
+        stop_event: Optional[Event],
+        storage_dir: str,
+        reinitialize_profiler: Callable[[], None],
+    ):
         self._frequency = min(frequency, self.MAX_FREQUENCY)
         self._duration = duration
         self._stop_event = stop_event or Event()
         self._storage_dir = storage_dir
+        self._reinitialize_profiler = reinitialize_profiler
         logger.info(f"Initializing Python profiler (frequency: {self._frequency}hz, duration: {duration}s)")
 
     def start(self):
@@ -132,8 +140,15 @@ class PythonEbpfProfiler(PythonProfilerBase):
     dump_poll_retries = 50  # 50 retries * dump_poll_interval = 5 seconds
     poll_timeout = 10  # seconds
 
-    def __init__(self, frequency: int, duration: int, stop_event: Optional[Event], storage_dir: str):
-        super().__init__(frequency, duration, stop_event, storage_dir)
+    def __init__(
+        self,
+        frequency: int,
+        duration: int,
+        stop_event: Optional[Event],
+        storage_dir: str,
+        reinitialize_profiler: Callable[[], None],
+    ):
+        super().__init__(frequency, duration, stop_event, storage_dir, reinitialize_profiler)
         self.process = None
         self.output_path = Path(self._storage_dir) / "py.col.dat"
 
@@ -218,6 +233,7 @@ class PythonEbpfProfiler(PythonProfilerBase):
             global _profiler_class
             _profiler_class = PySpyProfiler
             logger.warn("PyPerf dead/not responding, killing it and reverting to py-spy")
+            self._reinitialize_profiler()
 
             process = self.process  # save it
             self._terminate()
@@ -265,9 +281,9 @@ def determine_profiler_class(storage_dir: str, stop_event: Event):
 
 
 def get_python_profiler(
-    frequency: int, duration: int, stop_event: Event, storage_dir: str
+    frequency: int, duration: int, stop_event: Event, storage_dir: str, reinitialize_profiler: Callable[[], None]
 ) -> Union[PythonEbpfProfiler, PySpyProfiler]:
     global _profiler_class
     if _profiler_class is None:
         _profiler_class = determine_profiler_class(storage_dir, stop_event)
-    return _profiler_class(frequency, duration, stop_event, storage_dir)
+    return _profiler_class(frequency, duration, stop_event, storage_dir, reinitialize_profiler)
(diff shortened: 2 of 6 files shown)


gprofiler/python.py
```diff
@@ -188,35 +199,26 @@ def start(self):
...
                 break
 
-            if self._stop_event.wait(0.1):
+            if self._stop_event.wait(self.dump_poll_interval):
                 raise StopEventSetException()
+        else:
+            if self.process.poll() is not None:
+                self._pyperf_error(self.process)
+            else:
+                raise Exception("PyPerf is not responding!")
```
d3dave, line 221: Then kill it?

Jongy: I'm not sure what to do here. We can kill it, and what? Restart it? I don't wanna get into a restart loop of PyPerf due to an error.

At this point I want someone to inspect the exception & PyPerf's output (perhaps I can add the output here then)

d3dave: No, just kill it and revert to py-spy

Jongy: Sounds like a plan

Jongy: Done

gprofiler/exceptions.py
```diff
@@ -18,14 +18,12 @@ class CalledProcessError(subprocess.CalledProcessError):
...
             except ValueError:
-                return f"Command '{self.cmd}' died with unknown signal {-self.returncode}."
+                base = f"Command '{self.cmd}' died with unknown signal {-self.returncode}."
         else:
-            return (
-                f"Command '{self.cmd}' returned non-zero exit status {self.returncode}. "
-                f"stderr: {self.stderr}, stdout: {self.stdout}"
-            )
+            base = f"Command '{self.cmd}' returned non-zero exit status {self.returncode}. "
+        return base + f"stderr: {self.stderr}, stdout: {self.stdout}"
```
d3dave, line 26:
```suggestion
        return f"{base}\nstdout: {self.stdout}\nstderr: {self.stderr}"
```

gprofiler/python.py
```diff
@@ -127,10 +136,19 @@ def snapshot(self) -> Mapping[int, Mapping[str, int]]:
 class PythonEbpfProfiler(PythonProfilerBase):
     PYPERF_RESOURCE = "python/pyperf/PyPerf"
     dump_signal = signal.SIGUSR2
+    dump_poll_interval = 0.1  # seconds
+    dump_poll_retries = 50  # 50 retries * dump_poll_interval = 5 seconds
```
d3dave, line 140: Instead of fixing the number of iterations fix the time limit and measure the loop, like `poll_process`.

gprofiler/python.py
```diff
@@ -260,9 +281,9 @@ def determine_profiler_class(storage_dir: str, stop_event: Event):
...
 
 def get_python_profiler(
-    frequency: int, duration: int, stop_event: Event, storage_dir: str
+    frequency: int, duration: int, stop_event: Event, storage_dir: str, reinitialize_profiler: Callable[[], None]
 ) -> Union[PythonEbpfProfiler, PySpyProfiler]:
     global _profiler_class
     if _profiler_class is None:
         _profiler_class = determine_profiler_class(storage_dir, stop_event)
-    return _profiler_class(frequency, duration, stop_event, storage_dir)
+    return _profiler_class(frequency, duration, stop_event, storage_dir, reinitialize_profiler)
```
d3dave, line 289: `reinitialize_profiler` isn't specific to a profiler class. I think it's better to store it as a module global.

d3dave (changes requested):

gprofiler/exceptions.py
Jongy: Better indeed.

gprofiler/python.py
Jongy: Okie

gprofiler/python.py
Jongy: Doesn't really matter IMO (the timeout is only so that we have a timeout, idc if it's 5s or 10s).

But agree it looks better, changing

gprofiler/python.py
```diff
@@ -188,39 +213,39 @@ def start(self):
...
+        self.process = start_process(cmd)
+        # no need to check for success here - we're already past calling test(), PyPerf has been already proved
+        # working.
 
     def _dump(self) -> Path:
         assert self.process is not None, "profiling not started!"
         self.process.send_signal(self.dump_signal)
+        end_time = time.monotonic() + self.dump_timeout
         # important to not grab the transient data file
         while True:
```
d3dave, line 225: Fairly generic logic. I think it fits in `utils.py` alongside `poll_process` as `poll_files` or similar.

d3dave (changes requested):

Jongy:
The sleeps proved helpful. So, it's a race. I'll change `start()` to wait until the file is ready...

gprofiler/python.py
Jongy: K will try to move it.

Jongy: Done

Jongy:
> The sleeps proved helpful. So, it's a race. I'll change `start()` to wait until the file is ready...

Done. PTAL @d3dave

gprofiler/utils.py
```diff
@@ -85,19 +85,26 @@ def start_process(cmd: Union[str, List[str]], **kwargs) -> Popen:
     return popen
 
 
-def poll_process(process, timeout, stop_event):
-    timefn = time.monotonic
-    endtime = timefn() + timeout
+def wait_event(timeout: float, stop_event: Event, condition: Callable[[], bool]) -> bool:
```
d3dave, line 88: I think raising a `TimeoutError` on timeout is clearer than returning a boolean.

gprofiler/python.py
```diff
@@ -188,40 +212,45 @@ def start(self):
...
+        output_files = self._glob_output()
         # All the snapshot samples should be in one file
         assert len(output_files) == 1
         return Path(output_files[0])
 
+    def _dump(self) -> Path:
+        assert self.process is not None, "profiling not started!"
+        self.process.send_signal(self.dump_signal)
+
+        path = self._wait_for_output_file(self.dump_timeout)
```
d3dave, line 238: Along with the above, it's clearer if we explicitly catch a `TimeoutError` here.

tests/test_libpython.py
```diff
@@ -35,6 +35,7 @@ def test_python_select_by_libpython(
     Tests that profiling of processes running Python, whose basename(readlink("/proc/pid/exe")) isn't "python".
     (for example, uwsgi). We expect to select these because they have "libpython" in their "/proc/pid/maps".
     """
-    profiler = get_python_profiler(1000, 1, Event(), str(tmp_path))
-    process_collapsed = profiler.snapshot()
+    profiler = get_python_profiler(1000, 1, Event(), str(tmp_path), lambda: None)
+    with profiler:
```
d3dave, line 39: Can merge these two lines

d3dave (changes requested):

gprofiler/python.py
```diff
@@ -188,40 +212,45 @@ def start(self):
...
-            self.process = process
-        else:
-            # process terminated, must be an error:
-            self._check_output(process, self.output_path)
-            raise RuntimeError("_check_output did not raise!")
+        self.process = start_process(cmd)
+        # wait until the transient data file appears - because once returning from here, PyPerf may
+        # be polled via snapshot() and we need it to finish installing its signal handler.
+        if not wait_event(self.poll_timeout, self._stop_event, lambda: os.path.exists(self.output_path)):
+            raise TimeoutError("PyPerf didn't boot")
```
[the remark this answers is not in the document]
Jongy: Yup I forgot it. Will add

tests/test_libpython.py
Jongy: kk

Jongy: kk

gprofiler/python.py
Jongy: Aight

gprofiler/utils.py
Jongy: Aight

gprofiler/python.py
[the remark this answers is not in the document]
Jongy: It is weird tho, despite `__exit__` not being called, PyPerf does exit if I simulate this state

Jongy:
Looks good.

gprofiler/python.py
```diff
@@ -204,39 +228,47 @@ def start(self):
...
+
+        try:
+            return self._wait_for_output_file(self.dump_timeout)
+        except TimeoutError:
+            # error flow :(
+            if _reinitialize_profiler is not None:
+                logger.warning("Reverting to py-spy")
+                global _profiler_class
+                _profiler_class = PySpyProfiler
+                _reinitialize_profiler()
```
d3dave, line 265: `_pyperf_error` raises, and this may raise as well, so to make this safe I think the easiest option is to place one in a `try` and the other in a `finally`. Doesn't matter which comes first, I think.

tests/test_libpython.py
```diff
@@ -35,6 +35,6 @@ def test_python_select_by_libpython(
     Tests that profiling of processes running Python, whose basename(readlink("/proc/pid/exe")) isn't "python".
     (for example, uwsgi). We expect to select these because they have "libpython" in their "/proc/pid/maps".
     """
-    profiler = get_python_profiler(1000, 1, Event(), str(tmp_path))
-    process_collapsed = profiler.snapshot()
+    with get_python_profiler(1000, 1, Event(), str(tmp_path), lambda: None) as profiler:
```
d3dave, line 38: To avoid `lambda: None` you can make it optional. It handles it as it is right now anyway.

d3dave (changes requested):

tests/test_libpython.py
Jongy: k

gprofiler/python.py
Jongy: Can't really raise now (since py-spy c'tor does nothing). That's also why I put it first.
But agree that the 2 are unrelated and should happen regardless the possible failure of the others.

d3dave (approved):

PR #45, opened 2021-04-28
PyInstaller: Improvements & robustness

Jongy:
## Description
A bunch of commits that'll make it more robust.

1. Install UPX so PyInstaller uses it and the size goes down a bit.
2. Fail PyInstaller build if any of our `gprofiler.*` 
3. Avoid `COPY . .` so build is cached for non-code changes.
4. Include DSOs needed by our resources when calling staticx. This step also ensures all DSO dependencies are resolved, and fails the build if they are not. Currently all executables are static, only PyPerf is dynamic ~~and misses some DSOs in `centos:7`, so build fails.~~

~~I plan to add a few more commits of improvements here (stuff for staticx as well)~~

## Motivation and Context
Easier development & less chances to inject errors.

## How Has This Been Tested?
Locally.

I tried to build with `SyntaxError`s and now the build fails (instead of passing, and producing a broken binary which results in `ImportError`) - this tests the PyInstaller Analysis checks.

## Types of changes

## Checklist:

### DIFF
--- a/gprofiler/python.py
+++ b/gprofiler/python.py
@@ -160,6 +160,20 @@ class PythonEbpfProfiler(PythonProfilerBase):
             raise CalledProcessError(process.returncode, process.args, stdout, stderr)
 
     @classmethod
+    def get_pyperf_cmd(cls) -> List[str]:
+        pyperf = resource_path(cls.PYPERF_RESOURCE)
+        staticx_dir = os.getenv("STATICX_BUNDLE_DIR")
+        # are we running under staticx?
+        if staticx_dir is not None:
+            # STATICX_BUNDLE_DIR is where staticx has extracted all of the libraries it had collected
+            # earlier.
+            # see https://github.com/JonathonReinhart/staticx#run-time-information
+            # we run PyPerf with the same ld.so whose libc it was compiled against.
+            return [f"{staticx_dir}/ld-2.17.so", "--library-path", staticx_dir, pyperf]
+        else:
+            return [pyperf]
+
+    @classmethod
     def test(cls, storage_dir: str, stop_event: Optional[Event]):
         test_path = Path(storage_dir) / ".test"
         for f in glob.glob(f"{str(test_path)}.*"):
@@ -168,7 +182,7 @@ class PythonEbpfProfiler(PythonProfilerBase):
         # Run the process and check if the output file is properly created.
         # Wait up to 10sec for the process to terminate.
         # Allow cancellation via the stop_event.
-        cmd = [resource_path(cls.PYPERF_RESOURCE), "--output", str(test_path), "-F", "1", "--duration", "1"]
+        cmd = cls.get_pyperf_cmd() + ["--output", str(test_path), "-F", "1", "--duration", "1"]
         process = start_process(cmd)
         try:
             poll_process(process, cls.poll_timeout, stop_event)
@@ -180,8 +194,7 @@ class PythonEbpfProfiler(PythonProfilerBase):
 
     def start(self):
         logger.info("Starting profiling of Python processes with PyPerf")
-        cmd = [
-            resource_path(self.PYPERF_RESOURCE),
+        cmd = self.get_pyperf_cmd() + [
             "--output",
             str(self.output_path),
             "-F",
--- a/gprofiler/python.py
+++ b/gprofiler/python.py
@@ -169,7 +169,7 @@ class PythonEbpfProfiler(PythonProfilerBase):
             # earlier.
             # see https://github.com/JonathonReinhart/staticx#run-time-information
             # we run PyPerf with the same ld.so whose libc it was compiled against.
-            return [f"{staticx_dir}/ld-2.17.so", "--library-path", staticx_dir, pyperf]
+            return [f"{staticx_dir}/.staticx.interp", "--library-path", staticx_dir, pyperf]
         else:
             return [pyperf]
 
--- a/pyi.Dockerfile
+++ b/pyi.Dockerfile
@@ -38,7 +38,16 @@ COPY --from=bcc-builder /bcc/bcc/licenses gprofiler/resources/python/pyperf/lice
 COPY --from=bcc-builder /bcc/bcc/NOTICE gprofiler/resources/python/pyperf/
 
 COPY . .
-RUN pyinstaller pyinstaller.spec
+
+# run PyInstaller and make sure no 'gprofiler.*' modules are missing.
+# see https://pyinstaller.readthedocs.io/en/stable/when-things-go-wrong.html
+# from a quick look I didn't see how to tell PyInstaller to exit with an error on this, hence
+# this check in the shell.
+RUN pyinstaller pyinstaller.spec \
+    && echo \
+    && test -f build/pyinstaller/warn-pyinstaller.txt \
+    && if grep 'gprofiler\.' build/pyinstaller/warn-pyinstaller.txt ; then echo 'PyInstaller failed to pack gProfiler code! See lines above. Make sure to check for SyntaxError as this is often the reason.'; exit 1; fi;
+
 # staticx packs dynamically linked app with all of their dependencies, it tries to figure out which dynamic libraries are need for its execution
 # in some cases, when the application is lazily loading some DSOs, staticx doesn't handle it.
 # libcgcc_s is such a library, so we pass it manually to staticx.
--- a/pyi.Dockerfile
+++ b/pyi.Dockerfile
@@ -71,11 +71,10 @@ RUN pyinstaller pyinstaller.spec \
 COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # staticx packs dynamically linked app with all of their dependencies, it tries to figure out which dynamic libraries are need for its execution
 # in some cases, when the application is lazily loading some DSOs, staticx doesn't handle it.
-# libcgcc_s is such a library, so we pass it manually to staticx.
-# additionally, we use list_needed_libs.sh to list the dynamic dependencies of *all* of our resources,
+# we use list_needed_libs.sh to list the dynamic dependencies of *all* of our resources,
 # and make staticx pack them as well.
 # using scl here to get the proper LD_LIBRARY_PATH set
-RUN source scl_source enable devtoolset-8 llvm-toolset-7 && libs=$(./scripts/list_needed_libs.sh) && staticx $libs -l /usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so dist/gprofiler dist/gprofiler
+RUN source scl_source enable devtoolset-8 llvm-toolset-7 && libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gprofiler dist/gprofiler
 
 FROM scratch AS export-stage
 
--- a/pyi.Dockerfile
+++ b/pyi.Dockerfile
@@ -37,12 +37,13 @@ COPY --from=bcc-builder /bcc/bcc/LICENSE.txt gprofiler/resources/python/pyperf/
 COPY --from=bcc-builder /bcc/bcc/licenses gprofiler/resources/python/pyperf/licenses
 COPY --from=bcc-builder /bcc/bcc/NOTICE gprofiler/resources/python/pyperf/
 
-COPY . .
+COPY gprofiler gprofiler
 
 # run PyInstaller and make sure no 'gprofiler.*' modules are missing.
 # see https://pyinstaller.readthedocs.io/en/stable/when-things-go-wrong.html
 # from a quick look I didn't see how to tell PyInstaller to exit with an error on this, hence
 # this check in the shell.
+COPY pyi_build.py pyinstaller.spec .
 RUN pyinstaller pyinstaller.spec \
     && echo \
     && test -f build/pyinstaller/warn-pyinstaller.txt \
--- a/pyi.Dockerfile
+++ b/pyi.Dockerfile
@@ -49,10 +49,11 @@ RUN pyinstaller pyinstaller.spec \
     && test -f build/pyinstaller/warn-pyinstaller.txt \
     && if grep 'gprofiler\.' build/pyinstaller/warn-pyinstaller.txt ; then echo 'PyInstaller failed to pack gProfiler code! See lines above. Make sure to check for SyntaxError as this is often the reason.'; exit 1; fi;
 
+COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # staticx packs dynamically linked app with all of their dependencies, it tries to figure out which dynamic libraries are need for its execution
 # in some cases, when the application is lazily loading some DSOs, staticx doesn't handle it.
 # libcgcc_s is such a library, so we pass it manually to staticx.
-RUN staticx -l /usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so dist/gprofiler dist/gprofiler
+RUN libs=$(./scripts/list_needed_libs.sh) && staticx $libs -l /usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so dist/gprofiler dist/gprofiler
 
 FROM scratch AS export-stage
 
--- a/pyi.Dockerfile
+++ b/pyi.Dockerfile
@@ -1,22 +1,40 @@
-# TODO: copied from the main Dockerfile... too bad we don't have includes.
-FROM ubuntu:20.04 as bcc-builder
+# Centos 7 image is used to grab an old version of `glibc` during `pyinstaller` bundling.
+# This will allow the executable to run on older versions of the kernel, eventually leading to the executable running on a wider range of machines.
+FROM centos:7 AS build-stage
 
-RUN apt-get update
+# bcc part
+# TODO: copied from the main Dockerfile... but modified a lot. we'd want to share it some day.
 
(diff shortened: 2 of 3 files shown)


Jongy:
@d3dave here we can solve the libraries issue. The output on my box now:
```
#30 [build-stage 19/19] RUN libs=$(./scripts/list_needed_libs.sh) && staticx $libs -l /usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so dist/gprofiler dist/gprofiler
#30 sha256:6a731ce1e9fee76d9a5c86caa07a15aeec97826936ae8617a964f0077cc24566
#30 0.349 missing libs/symbols for binary gprofiler/resources/python/pyperf/PyPerf
#30 0.350 gprofiler/resources/python/pyperf/PyPerf: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by gprofiler/resources/python/pyperf/PyPerf)
#30 0.350 gprofiler/resources/python/pyperf/PyPerf: /lib64/libm.so.6: version `GLIBC_2.29' not found (required by gprofiler/resources/python/pyperf/PyPerf)
#30 0.350 gprofiler/resources/python/pyperf/PyPerf: /lib64/libc.so.6: version `GLIBC_2.22' not found (required by gprofiler/resources/python/pyperf/PyPerf)
#30 0.350 gprofiler/resources/python/pyperf/PyPerf: /lib64/libstdc++.so.6: version `CXXABI_1.3.8' not found (required by gprofiler/resources/python/pyperf/PyPerf)
#30 0.350 gprofiler/resources/python/pyperf/PyPerf: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.22' not found (required by gprofiler/resources/python/pyperf/PyPerf)
#30 0.350 gprofiler/resources/python/pyperf/PyPerf: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by gprofiler/resources/python/pyperf/PyPerf)
#30 0.350 gprofiler/resources/python/pyperf/PyPerf: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by gprofiler/resources/python/pyperf/PyPerf)
#30 0.350 gprofiler/resources/python/pyperf/PyPerf: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by gprofiler/resources/python/pyperf/PyPerf)
#30 0.350 gprofiler/resources/python/pyperf/PyPerf: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by gprofiler/resources/python/pyperf/PyPerf)
#30 0.350 	linux-vdso.so.1 =>  (0x00007ffc177c3000)
#30 0.350 	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f332a815000)
#30 0.350 	libdl.so.2 => /lib64/libdl.so.2 (0x00007f332a611000)
#30 0.350 	libtinfo.so.6 => not found
#30 0.350 	libz.so.1 => /lib64/libz.so.1 (0x00007f332a3fb000)
#30 0.350 	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f332a0f3000)
#30 0.350 	libm.so.6 => /lib64/libm.so.6 (0x00007f3329df1000)
#30 0.350 	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f3329bdb000)
#30 0.350 	libc.so.6 => /lib64/libc.so.6 (0x00007f332980d000)
#30 0.350 	/lib64/ld-linux-x86-64.so.2 (0x00007f332e70b000)
```

I think that the easiest way would be to build bcc on an older machine. If we can build on the same `centos:7` that'd be best. Trying it now.

Jongy:
Okay. It works now. It has grown a bit out of hand - I wasn't expecting it to require that much effort, I admit.

Since PyPerf itself doesn't require running old old glibc, technically it can be compiled in a newer system, we just gotta bundle everything with it, or make it static.

Alternatively, we can switch to another method (from staticx to appiamge/....). I am not very pleased with the solution, but it does work now and I don't think it's fragile.

pyi.Dockerfile
```diff
@@ -31,18 +49,33 @@ RUN python3 -m pip install -r dev-requirements.txt
...
-RUN pyinstaller pyinstaller.spec
+COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # staticx packs dynamically linked app with all of their dependencies, it tries to figure out which dynamic libraries are need for its execution
 # in some cases, when the application is lazily loading some DSOs, staticx doesn't handle it.
 # libcgcc_s is such a library, so we pass it manually to staticx.
-RUN staticx -l /usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so dist/gprofiler dist/gprofiler
+# additionally, we use list_needed_libs.sh to list the dynamic dependencies of *all* of our resources,
+# and make staticx pack them as well.
+# using scl here to get the proper LD_LIBRARY_PATH set
+RUN source scl_source enable devtoolset-8 llvm-toolset-7 && libs=$(./scripts/list_needed_libs.sh) && staticx $libs -l /usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so dist/gprofiler dist/gprofiler
```
leonid-granulate, line 78: `libs=$(./scripts/list_needed_libs.sh)` already bring the `libgcc_s.so`, so there is no need to pass it again.
I've checked that the files are identical by running `cmp /lib64/libgcc_s.so.1 /usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so`
`scripts/list_needed_libs.sh` do a great work :)

```suggestion
RUN source scl_source enable devtoolset-8 llvm-toolset-7 && libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gprofiler dist/gprofiler
```

leonid-granulate (changes requested):

pyi.Dockerfile
```diff
@@ -1,26 +1,44 @@
...
+    python3 \
+    flex \
+    bison \
+    zlib-devel.x86_64 \
+    ncurses-devel \
+    elfutils-libelf-devel
 
 WORKDIR /bcc
 
 RUN git clone --depth 1 https://github.com/Granulate/bcc.git && cd bcc && git reset --hard 119d71bf9681182759eb76d40660c0ec19f3fc42
```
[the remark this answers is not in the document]
Jongy: It is tagged, I think. But I prefer hardcoding the commit because tags/branches may change, but this commit refers to exactly one revision, forever. I prefer it for hardcoding versions.

pyi.Dockerfile
Jongy: This is my output:
```
#29 0.353 found needed lib: /lib64/libpthread.so.0
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangFrontend.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangSerialization.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangDriver.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangParse.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangSema.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangCodeGen.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangAnalysis.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangRewrite.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangEdit.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangAST.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangLex.so.5
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/libclangBasic.so.5
#29 0.353 found needed lib: /lib64/librt.so.1
#29 0.353 found needed lib: /lib64/libdl.so.2
#29 0.353 found needed lib: /lib64/libtinfo.so.5
#29 0.353 found needed lib: /lib64/libz.so.1
#29 0.353 found needed lib: /lib64/libstdc++.so.6
#29 0.353 found needed lib: /lib64/libm.so.6
#29 0.353 found needed lib: /lib64/libgcc_s.so.1
#29 0.353 found needed lib: /lib64/libc.so.6
#29 0.353 found needed lib: /opt/rh/llvm-toolset-7/root/usr/lib64/../lib64/libLLVM-5.0.so
#29 0.353 found needed lib: /lib64/libffi.so.6
#29 0.353 found needed lib: /lib64/libc.so.6
#29 0.353 found needed lib: /lib64/libdl.so.2
#29 0.353 found needed lib: /lib64/libpthread.so.0
#29 0.353 found needed lib: /lib64/librt.so.1
#29 0.353 found needed lib: /lib64/libstdc++.so.6
#29 0.353 found needed lib: /lib64/libm.so.6
#29 0.353 found needed lib: /lib64/libgcc_s.so.1
#29 0.353 found needed lib: /lib64/libc.so.6
```

Doesn't seem like the same file - running on a clean `centos:7`:
```
[root@d922e9e023b8 /]# ls -l /lib64/libgcc_s.so.1
lrwxrwxrwx 1 root root 28 Nov 13 01:53 /lib64/libgcc_s.so.1 -> libgcc_s-4.8.5-20150702.so.1
[root@d922e9e023b8 /]# ls -l /lib64/libgcc_s.so.1^C
[root@d922e9e023b8 /]# ls /usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so
ls: cannot access /usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so: No such file or directory
```

I can try removing the manual `libgcc` reference and see if it works. Remind me, what exactly wasn't working for which we added this libgcc?

leonid-granulate: I think it was a steps to support `alpine` and `ubuntu 14.04`.
Anyway i compared the files in `pyi.Dockerfile` based container.
```bash
[root@4c04088634dd app]# cmp /lib64/libgcc_s.so.1 /usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so 
[root@4c04088634dd app]# 
```
So I think you can safely delete it.

pyi.Dockerfile
```diff
@@ -1,26 +1,44 @@
-# TODO: copied from the main Dockerfile... too bad we don't have includes.
-FROM ubuntu:20.04 as bcc-builder
+# Centos 7 image is used to grab an old version of `glibc` during `pyinstaller` bundling.
+# This will allow the executable to run on older versions of the kernel, eventually leading to the executable running on a wider range of machines.
+FROM centos:7 AS build-stage
```
d3dave, line 3: Any reason to not use holy-build-box?

scripts/list_needed_libs.sh
```diff
@@ -0,0 +1,40 @@
...
+for f in $BINS ; do
+    set +e
+    ldd_output=$(ldd $f 2>&1)
+    ret=$?
+    set -e
+
+    if [ $ret -eq 0 ]; then
+        if [[ "$ldd_output" == *" not found"* ]]; then
+            >&2 echo "missing libs/symbols for binary $f"
+            >&2 cat <<< "$ldd_output"
```
d3dave, line 26:
```suggestion
            >&2 echo "$ldd_output"
```

scripts/list_needed_libs.sh
```diff
@@ -0,0 +1,40 @@
...
+# (staticx knows to pack the libraries used by the executable we're packing. it doesn't know
+# which executables are to be used by it)
+
+BINS=$(find gprofiler/resources -executable -type f)
+
+libs=
+
+for f in $BINS ; do
+    set +e
+    ldd_output=$(ldd $f 2>&1)
```
d3dave, line 19:
```suggestion
    IFS= ldd_output="$(ldd $f 2>&1)"
```

gprofiler/python.py
```diff
@@ -159,6 +159,20 @@ def _check_output(cls, process, output_path):
...
+    def get_pyperf_cmd(cls) -> List[str]:
+        pyperf = resource_path(cls.PYPERF_RESOURCE)
+        staticx_dir = os.getenv("STATICX_BUNDLE_DIR")
+        # are we running under staticx?
+        if staticx_dir is not None:
+            # STATICX_BUNDLE_DIR is where staticx has extracted all of the libraries it had collected
+            # earlier.
+            # see https://github.com/JonathonReinhart/staticx#run-time-information
+            # we run PyPerf with the same ld.so whose libc it was compiled against.
+            return [f"{staticx_dir}/ld-2.17.so", "--library-path", staticx_dir, pyperf]
```
d3dave, line 172:
```suggestion
            return [f"{staticx_dir}/.staticx.interp", "--library-path", staticx_dir, pyperf]
```

d3dave (changes requested):

pyi.Dockerfile
Jongy: Hmm. Weird tho because one is `libgcc_s-4.8.5-20150702.so.1` and the other is `/usr/lib/gcc/x86_64-redhat-linux/4.8.2/libgcc_s.so`, *4.8.2*...

But alright - will remove and see if CI passes on alpine :)

scripts/list_needed_libs.sh
Jongy: I used `cat` because `echo` got rid of line ends for me. Tested now again and it looks alright. Weird

gprofiler/python.py
Jongy: Nice

pyi.Dockerfile
Jongy: No reason. If I were to rewrite it, then maybe. HBB has caused me some headaches in other projects so I preferred to use mainline CentOS.

scripts/list_needed_libs.sh
Jongy: Will switch to it

scripts/list_needed_libs.sh
Jongy: Why is the `IFS=` needed? To avoid any mashup if the output contains `"`?

scripts/list_needed_libs.sh
d3dave: Because it was split with `$IFS`

Jongy: I tried on my shell now:
```
$ cat a
a
b
c
$ x=$(cat a)
$ echo "$x"
a
b
c
$ cat <<< "$x"
a
b
c
```

without touching `IFS`

d3dave: Ah right, it's because of the quoting.

pyi.Dockerfile
Jongy: Seems to pass :)

scripts/list_needed_libs.sh
d3dave: Not needed apparently because bash preserves the value when it's quoted.

pyi.Dockerfile
Jongy: Hmm no alpine tests. But those that we have seem to pass..

scripts/list_needed_libs.sh
Jongy: kk

Jongy:
Anything else here? @d3dave @leonid-granulate

leonid-granulate (approved):
Clean from my side.

gprofiler/python.py
```diff
@@ -159,6 +159,20 @@ def _check_output(cls, process, output_path):
...
+    def get_pyperf_cmd(cls) -> List[str]:
+        pyperf = resource_path(cls.PYPERF_RESOURCE)
+        staticx_dir = os.getenv("STATICX_BUNDLE_DIR")
+        # are we running under staticx?
+        if staticx_dir is not None:
+            # STATICX_BUNDLE_DIR is where staticx has extracted all of the libraries it had collected
+            # earlier.
+            # see https://github.com/JonathonReinhart/staticx#run-time-information
+            # we run PyPerf with the same ld.so whose libc it was compiled against.
+            return [f"{staticx_dir}/.staticx.interp", "--library-path", staticx_dir, pyperf]
```
d3dave, line 172: why not `patchelf` when running staticx?

Jongy: No reason. I'll still need to pass `LD_LIBRARY_PATH` tho so it doesn't allow getting rid of the code here (unless we patch that as well?)

d3dave: Nope. Use `--set-rpath '$ORIGIN/../..../' --force-rpath`. Passing `LD_LIBRARY_PATH` would mess up any programs spawned by pyperf (because they would inherit it).

Jongy: Well, any program spawned by PyPerf should get its `LD_LIBRARY_PATH` as well, to use any libs that bcc might want (and in general, that's the only directory any program should be loaded libraries from).

I'll try the `patchelf`.

Jongy: We can modify staticx to do that (it already does it for the main executable)

Jongy: Talked IRL - `$ORIGIN` won't do due to:

> Since the path is determined at runtime, that's not possible (unless we hardcode it by modifying staticx)

(and they're not in the same directory)

d3dave (approved):

PR #48, opened 2021-04-29
Check for root privileges before attempting to open /var/log/gprofiler/gprofiler.log

d3dave:
## Description
When executed as a regular user, opening the default log file will fail. Check for root privileges first.
## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -268,15 +268,15 @@ def parse_cmd_args():
     return args
 
 
-def verify_preconditions():
+def verify_preconditions(args):
     if not is_root():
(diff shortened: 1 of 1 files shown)


gprofiler/main.py
```diff
@@ -268,15 +268,15 @@ def parse_cmd_args():
     return args
 
 
-def verify_preconditions():
+def verify_preconditions(args):
```
Jongy, line 271: Why accepts `args`?

d3dave: no more

Jongy (approved):

PR #52, opened 2021-05-01
Implement a global gProfiler lock

Jongy:
## Description
Implement a global gProfiler lock, to ensure only 1 gProfiler is run at a time.

## Motivation and Context
gProfiler is a system-wide profiler, so no more than 1 instance should run concurrently.

Besides the unnecessary performance "cost" of running 2 simultaneously, there's also a certain risk with profilers that are not fully external in their operation: for example, async-profiler is loaded into target processes and gProfiler manages it by sending commands with `jattach`. 2 gProfilers trying to profile the same Java process could potentially get their commands mixed, resulting, for example, in async-profiler kept in profiling due to a missed `stop` command (this itself should be fixed in the `java.py` profiler logic - it should handle such unexpected cases on itself and abort profiling gracefully; but we should add this additional layer of protection with the global lock).

## How Has This Been Tested?
Tested locally - ran gProfiler in a container (w/o `--network=host`) and then tried to run another one:
* container with `--network=host`
* container without `--network=host`
* standalone executable

All refused to execute.

## Types of changes
## Checklist:


gprofiler/utils.py
```diff
@@ -261,6 +266,30 @@ def get_run_mode() -> str:
         return "local_python"
 
 
+# we can't setns(CLONE_NEWNS) in a multithreaded program (unless we unshare(CLONE_NEWNS) before)
+# so, we start a new thread, unshare() & setns() it, get our needed information and then stop this thread
+# (so we don't keep unshared threads running around)
+# for other namespace types, we use this function to execute callbacks without changing the namespaces for the
+# core threads.
+def run_in_init_ns(nstype: str, callaback: Callable[[], None]) -> None:
```
guybortnikov, line 274:
```suggestion
def run_in_init_ns(nstype: str, callback: Callable[[], None]) -> None:
```

guybortnikov (changes requested):
Looks good, small typo fix and can be merged.

guybortnikov (approved):

gprofiler/main.py
```diff
@@ -27,7 +27,15 @@
 from .java import JavaProfiler
 from .perf import SystemProfiler
 from .python import get_python_profiler
-from .utils import is_root, run_process, get_iso8061_format_time, resource_path, log_system_info, TEMPORARY_STORAGE_PATH
+from .utils import (
```
talsaiag, line 30: @Jongy wtf is this formatting?

Jongy: Default `black` settings?

talsaiag: lol looks ugly (but effective git diff wise, since each is at its own line)

Jongy: I don't really mind how it looks tbh, as long as it just works (and yeah, the git diff is nice)

PR #53, opened 2021-05-01
Add --rotating-output to allow saving only latest output files

Jongy:
## Description
Add `--rotating-output` which will tell gProfiler to keep only the latest output files. Additionally, `last_profile.col` and `last_flamegraph.html` were added, to point to those latest output files. They are always accessible and will always point to the latest files.

In the future we can add more sophisticated rotation (e.g: all profiles from the last day; or X last profiles).

## Motivation and Context
Running with local output (`-o`) alongside upload mode (`-u`) is sometimes convenient, to be able to view the last results from the last profiling sessions locally & immediately. However, `-o` creates a (currently unlimited) number of files, which is problematic.

Using `--rotating-output` helps for that cause.

## How Has This Been Tested?
Locally.

* Without `--rotating-output`, we keep getting files as we had before this PR.
* With `--rotating-output`, we have just the 2 last files.
* With `--no-flamegraph`, we have just the collapsed stacks file.
* And the `last_*` files always point to the last output, in any case.

## Types of changes

## Checklist:


gprofiler/main.py
```diff
@@ -103,16 +107,28 @@ def _generate_output_files(
...
         base_filename = os.path.join(self._output_dir, "profile_{}".format(end_ts))
+
         collapsed_path = base_filename + ".col"
         Path(collapsed_path).write_text(collapsed_data)
+
+        # point last_profile.col at the new file; and possibly, delete the previous one.
+        last_profile = os.path.join(self._output_dir, "last_profile.col")
+        prev_profile = Path(last_profile).resolve()
+        atomically_symlink(os.path.basename(collapsed_path), last_profile)
+        if self._rotating_output:
```
leonid-granulate, line 122:
```suggestion
        if self._rotating_output and prev_profile.name != Path(last_profile).name:
```
Otherwise if `self_rotating_output == True` it will just unlink the symlink forever.

gprofiler/main.py
```diff
@@ -127,6 +143,17 @@ def _generate_output_files(
...
+            # point last_flamegraph.html at the new file; and possibly, delete the previous one.
+            last_flamegraph = os.path.join(self._output_dir, "last_flamegraph.html")
+            prev_flamegraph = Path(last_flamegraph).resolve()
+            atomically_symlink(os.path.basename(flamegraph_path), last_flamegraph)
+            if self._rotating_output:
+                try:
+                    prev_flamegraph.unlink()
+                except FileNotFoundError:
+                    pass
+
```
leonid-granulate, line 156: This code is duplicated, consider to create a function

gprofiler/main.py
```diff
@@ -103,16 +107,28 @@ def _generate_output_files(
...
     ) -> None:
         start_ts = get_iso8061_format_time(local_start_time)
         end_ts = get_iso8061_format_time(local_end_time)
         base_filename = os.path.join(self._output_dir, "profile_{}".format(end_ts))
+
         collapsed_path = base_filename + ".col"
         Path(collapsed_path).write_text(collapsed_data)
+
+        # point last_profile.col at the new file; and possibly, delete the previous one.
+        last_profile = os.path.join(self._output_dir, "last_profile.col")
```
leonid-granulate, line 119: It always return the same value, can be initialized in `__init__`?

leonid-granulate (changes requested):

gprofiler/main.py
Jongy: Since `last_profile.col` is a symlink, `Path(last_profile).resolve()` is never `last_profile.name` itself, but whatever the symlinks points to (which is not the symlinks itself).

Jongy: Or, I didn't understand what you mean here?

gprofiler/main.py
Jongy: Hmm yeah but it's really only private to this function so I see no reason to store it as an attribute (it's not that this "calculation" costs time)

Jongy: Hmm yeah but it's really only private to this function so I see no reason to store it as an attribute (it's not that this "calculation" costs time)

gprofiler/main.py
Jongy: Yeah I thought about it but haven't done before I pushed. Will do

gprofiler/main.py
Jongy: Okay I see. On the first run the symlink is missing then `resolve()` returns the same file. Fixed

leonid-granulate: Exactly

leonid-granulate (dismissed):

leonid-granulate (dismissed):

leonid-granulate (approved):

PR #54, opened 2021-05-01
java: Use run_in_ns() instead of nsenter for JDK version check

Jongy:
## Description
Use new `run_in_ns()` helper instead of using `nsenter` to implement the JDK version checking.

Based on https://github.com/Granulate/gprofiler/pull/52 for some of the changes. Will merge after it.

## Motivation and Context
Drop dependency on `nsenter`.

## How Has This Been Tested?
Checked that we correctly get the JDK version for the `java` image (Debian 8 based) and `java:alpine` (Alpine based.. duh :)

## Types of changes
## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,9 +60,17 @@ jobs:
       fail-fast: false
       matrix:
         # add ubuntu 14.04 and alpine, when they will be ready
-        containers: [ 'ubuntu:16.04', 'ubuntu:18.04',
-                      'ubuntu:20.04', 'ubuntu:20.10', 'centos:7', 'centos:8', 'debian:8',
-                      'debian:9', 'debian:10' ]
+        containers: [
+                      'ubuntu:16.04',
+                      'ubuntu:18.04',
+                      'ubuntu:20.04',
+                      'ubuntu:20.10',
+                      'centos:7',
+                      'centos:8',
+                      'debian:8',
+                      'debian:9',
+                      'debian:10',
+                    ]
     runs-on: ubuntu-20.04
     steps:
       - name: Set up Python 3.8
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,8 +59,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # add alpine, when it's ready
         containers: [
+                      'alpine',
                       'ubuntu:14.04',
                       'ubuntu:16.04',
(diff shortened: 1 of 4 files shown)


gprofiler/utils.py
```diff
@@ -261,6 +266,45 @@ def get_run_mode() -> str:
         return "local_python"
 
 
+def run_in_ns(nstypes: List[str], callaback: Callable[[], None], target_pid: int = 1) -> None:
```
d3dave, line 269: I see you draw your inspiration from Gwen Stefani 😄 
```suggestion
def run_in_ns(nstypes: List[str], callback: Callable[[], None], target_pid: int = 1) -> None:
```

Jongy: Never listened to Hollaback Girl

Jongy: Will do now

Jongy: Oh okay, ofc I have

Jongy:
Heh, CI failed because we use `git clone --depth=1` but @d3dave has pushed new commits to https://github.com/Granulate/bcc 's master branch. We probably need to `git clone --depth=100` or so haha :)

Jongy:
Build failure is fixed in https://github.com/Granulate/gprofiler/pull/56

Jongy:
Relevant for Ubuntu 14.04 where there's no `nsenter` by default :/

gprofiler/utils.py
Jongy: Done

Jongy:
I can't recall why tests fail on `alpine`, so I'll try it again and note the reason in comment this time.

Also, enabled testing on `ubuntu:14.04` because it should work now.

.github/workflows/build.yml
```diff
@@ -59,10 +59,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # add ubuntu 14.04 and alpine, when they will be ready
-        containers: [ 'ubuntu:16.04', 'ubuntu:18.04',
-                      'ubuntu:20.04', 'ubuntu:20.10', 'centos:7', 'centos:8', 'debian:8',
-                      'debian:9', 'debian:10' ]
+        containers: [
```
guybortnikov, line 62: Please change to a normal YAML array.

Jongy: Mhm

guybortnikov (approved):

guybortnikov (approved):

PR #55, opened 2021-05-02
Allow disabling runtime-specific profilers

Jongy:
## Description
Allow disabling the runtime-specific profilers via `--no-java` and `--no-python`.

Since there are only 2 runtime-specific profilers now, I didn't invest too much in the "NoopProfiler" infrastructure. Later on we can replace those sets of funny `if`s with a proper class registry for profilers (profilers can be `@register("java")` and then we can decide if we use the registered class or `NoopProfiler` at 1 spot)

## Motivation and Context
Give more control to the user. It might be useful to disable runtime-specific profilers in case either of them cause issues / compatibility problems. If need the results only from part of them, disabling the others (especially as we grow to add more profilers) might be helpful in such cases.

## How Has This Been Tested?
Tested running locally with the 4 combinations of `--no-java`, `--no-python` and verified that Java/Python profilers are not initialized when not requested.

## Types of changes

## Checklist:


Jongy:
Build failure is fixed in https://github.com/Granulate/gprofiler/pull/56

gprofiler/java.py
```diff
@@ -95,7 +84,7 @@ def get_async_profiler_stop_cmd(
             f"stop,file={output_path},{self.OUTPUT_FORMAT},{self.FORMAT_PARAMS},log={log_path}",
         ]
 
-    def run_async_profiler(self, cmd: str, log_path_host: str):
+    def _run_async_profiler(self, cmd: str, log_path_host: str):
```
leonid-granulate, line 87:
```suggestion
    def _run_async_profiler(self, cmd: Union[str, List[str]], log_path_host: str):
```

leonid-granulate (changes requested):
I saw your comment about an `if` statements and replacement by decorators, maybe it worth to create an issue.

gprofiler/java.py
Jongy: Just `List[str]` will do

leonid-granulate (approved):

Jongy:
> I saw your comment about an `if` statements and replacement by decorators, maybe it worth to create an issue.

Will do

PR #77, opened 2021-05-18
Support merging the result of fp and dwarf perfs for accuracy

alonbentamar:
## Description
In order to increase the accuracy and reliability of perf results, 2 global perfs will run in parallel - one with FP (frame pointers) and one with Dwarf. The result with the highest average of frames per stack will be chosen for the final output, per process.
## How Has This Been Tested?
I used the new `--perf-mode` parameter to test all of the different modes - FP, dwarf and "smart" and compared the results.
## Screenshots:

FP:  fp flamegraph

Dwarf: dwarf flamegraph

Smart: smart flamegraph

## Types of changes

## Checklist:


README.md
```diff
@@ -49,6 +49,12 @@ The default duration is *60 seconds*, and the default interval matches it. So gP
 
 * `--no-java`, `--no-python`: Disable the runtime-specific profilers of Java and/or Python, accordingly.
 
+* `--perf-mode`: Controls the global perf strategy. Must be one of the following options:
```
Jongy, line 52: There's existing reference in the footnotes to frame pointer, please fix it (refer to the new `--perf-mode` parameter)

gprofiler/perf.py
```diff
@@ -55,29 +55,23 @@ def _run_perf(self, filename_base: str, dwarf=False):
...
-                    [resource_path("perf")] + buildid_args + ["script", "-F", "+pid", "-i", record_file.name], stdout=f
-                )
-            return parsed_path
+            perf_script_result = run_process(
+                [resource_path("perf")] + buildid_args + ["script", "-F", "+pid", "-i", record_file.name]
+            )
+            return perf_script_result.stdout.decode('utf8')
 
-    def snapshot(self) -> Iterable[Mapping[str, str]]:
+    def snapshot(self) -> Dict[int, List[Dict[str, str]]]:
```
Jongy, line 63: Funny that mypy hasn't whined on this

gprofiler/perf.py
```diff
@@ -42,8 +43,7 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.stop()
 
-    def _run_perf(self, filename_base: str, dwarf=False):
-        parsed_path = os.path.join(self._storage_dir, f"{filename_base}.parsed")
+    def _run_perf(self, dwarf=False):
```
Jongy, line 46: Add typing :)

gprofiler/perf.py
```diff
@@ -55,29 +55,23 @@ def _run_perf(self, filename_base: str, dwarf=False):
...
             )
-            with open(parsed_path, "w") as f:
-                run_process(
-                    [resource_path("perf")] + buildid_args + ["script", "-F", "+pid", "-i", record_file.name], stdout=f
-                )
-            return parsed_path
+            perf_script_result = run_process(
+                [resource_path("perf")] + buildid_args + ["script", "-F", "+pid", "-i", record_file.name]
+            )
+            return perf_script_result.stdout.decode('utf8')
```
Jongy, line 61: In that case we're limited to the file buffer size, I think (since you're not reading along the way, only when the process is done). Default size is 65k which is too low for common use-cases. Keep it as a file.

gprofiler/perf.py
```diff
@@ -55,29 +55,23 @@ def _run_perf(self, filename_base: str, dwarf=False):
...
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+            fp_future = executor.submit(self._run_perf, False) if self._fp_perf else None
+            dwarf_future = executor.submit(self._run_perf, True) if self._dwarf_perf else None
+        fp_perf = fp_future.result() if fp_future is not None else None
+        dwarf_perf = dwarf_future.result() if dwarf_future is not None else None
 
-        # TODO: run dwarf in parallel, after supporting it in merge.py
-        # Alternatively: fix golang dwarf problems and the run just dwarf
-
-        # if get_cpu_count() > 32:
```
Jongy, line 77: Place some limit on the dwarf frequency - it's much heavier than FP, if someone runs gProfiler with e.g freq 1k~ and they have lots of cores, the effect will be substantial.

gprofiler/perf.py
```diff
@@ -55,29 +55,23 @@ def _run_perf(self, filename_base: str, dwarf=False):
...
-        record_path = self._run_perf("global")
-        logger.info("Finished running global perf")
-        return parse_perf_script(Path(record_path).read_text())
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+            fp_future = executor.submit(self._run_perf, False) if self._fp_perf else None
+            dwarf_future = executor.submit(self._run_perf, True) if self._dwarf_perf else None
+        fp_perf = fp_future.result() if fp_future is not None else None
+        dwarf_perf = dwarf_future.result() if dwarf_future is not None else None
 
-        # TODO: run dwarf in parallel, after supporting it in merge.py
```
Jongy, line 74: Should also add support to changing dwarf stack size - this is the number of bytes `perf` collects per event. The default size is `8192`. It is passed like `--call-graph dwarf,{stack_size}`. Please add it as `--perf-dwarf-stack-size`.

Jongy:
The screenshots look good :) I still have to go over the core logic.

gprofiler/perf.py
DanielShaulov: We are using `process.communicate()`, it reads along the way

gprofiler/perf.py
DanielShaulov: The upper limit is 64k, btw

Jongy: `-8` to be precise :) `65528` IIRC

DanielShaulov:
The comparison photo is not the best, since it only shows the golang process and for that we know that `fp` is better so it just chooses fp.
Let's run it alongside something that we know need dwarf, like umm.. @Jongy - you have something easy to run?

Jongy:
> Let's run it alongside something that we know need dwarf, like umm.. @Jongy - you have something easy to run?

CPython, and run gProfiler with `--no-python` so we see native stacks and not py-spy/pyperf

Jongy:
@alonbentamar , the examples you've given now are still "dwarf only" or "fp only". Can you show 2 processes w/ smart - 1 that'll use fp (Go) and 1 that'll use dwarf (CPython)

README.md
```diff
@@ -49,6 +49,12 @@ The default duration is *60 seconds*, and the default interval matches it. So gP
...
+  * `fp` - Use Frame Pointers for the call graph
+  * `dwarf` - Use Dwarf for the call graph (adds the `--call-graph dwarf` argument to the `perf` command)
```
Jongy, line 54:
```suggestion
  * `dwarf` - Use DWARF for the call graph (adds the `--call-graph dwarf` argument to the `perf` command)
```

styling-wise

gprofiler/main.py
```diff
@@ -326,6 +329,24 @@ def parse_cmd_args():
         help="Do not invoke runtime-specific profilers for Python processes",
     )
 
+    parser.add_argument(
+        "--perf-mode",
+        dest="perf_mode",
+        default="smart",
+        choices=["fp", "dwarf", "smart"],
+        help="Run perf with either FP (Frame Pointers), Dwarf, or run both and intelligently merge them "
```
Jongy, line 337:
```suggestion
        help="Run perf with either FP (Frame Pointers), DWARF, or run both and intelligently merge them "
```

gprofiler/merge.py
```diff
@@ -82,7 +82,85 @@ def collapse_stack(stack: str, comm: str) -> str:
     return ";".join(funcs)
 
 
-def parse_perf_script(script: str):
+def merge_global_perfs(
+    raw_fp_perf: Optional[str], raw_dwarf_perf: Optional[str]
+) -> Tuple[DefaultDict[int, MutableMapping[str, int]], Dict[int, str]]:
+    # The FP perf is used here as the "main" perf, to which the Dwarf perf is scaled.
```
Jongy, line 88:
```suggestion
    # The FP perf is used here as the "main" perf, to which the DWARF perf is scaled.
```

Jongy, line 88: I think that this comment belongs ahead?

gprofiler/merge.py
```diff
@@ -82,7 +82,85 @@ def collapse_stack(stack: str, comm: str) -> str:
...
+    merged_pid_to_name = dwarf_pid_to_name
+
+    if raw_fp_perf is None:
+        return dwarf_perf, merged_pid_to_name
+    elif raw_dwarf_perf is None:
+        return fp_perf, merged_pid_to_name
+
+    total_fp_samples = sum([sum(stacks.values()) for stacks in fp_perf.values()])
+    total_dwarf_samples = sum([sum(stacks.values()) for stacks in dwarf_perf.values()])
+    fp_to_dwarf_total_sample_count_ratio = total_fp_samples / total_dwarf_samples
```
Jongy, line 102: Shorter name pls

gprofiler/merge.py
```diff
@@ -82,7 +82,85 @@ def collapse_stack(stack: str, comm: str) -> str:
     return ";".join(funcs)
 
 
-def parse_perf_script(script: str):
+def merge_global_perfs(
+    raw_fp_perf: Optional[str], raw_dwarf_perf: Optional[str]
+) -> Tuple[DefaultDict[int, MutableMapping[str, int]], Dict[int, str]]:
+    # The FP perf is used here as the "main" perf, to which the Dwarf perf is scaled.
+    merged_pid_to_stacks_counters: DefaultDict[int, MutableMapping[str, int]] = defaultdict(Counter)
+    fp_perf, fp_pid_to_name = parse_perf_script(raw_fp_perf)
```
Jongy, line 90: `fp_perf` and `dwarf_perf` are dicts of pid to a dict of stack to count? Can add a comment explaining on `parse_perf_script` maybe. It's becoming complex

gprofiler/merge.py
Jongy, line 102: Can you please log this number? `logger.debug`. It should remain roughly 1. If not, it can generate weird results. I wanna monitor it over the logs we have in our systems.

gprofiler/merge.py
```diff
@@ -82,7 +82,85 @@ def collapse_stack(stack: str, comm: str) -> str:
...
+    add_highest_avg_depth_stacks_per_process(
+        dwarf_perf, fp_perf, fp_to_dwarf_total_sample_count_ratio, merged_pid_to_stacks_counters
+    )
+    add_missing_dwarf_stacks(dwarf_perf, fp_to_dwarf_total_sample_count_ratio, merged_pid_to_stacks_counters)
+
+    return merged_pid_to_stacks_counters, merged_pid_to_name
+
+
+def add_highest_avg_depth_stacks_per_process(
+    dwarf_perf, fp_perf, fp_to_dwarf_total_sample_count_ratio, merged_pid_to_stacks_counters
```
Jongy, line 113: Typing everywhere please

gprofiler/merge.py
```diff
@@ -82,7 +82,85 @@ def collapse_stack(stack: str, comm: str) -> str:
...
+):
+    for pid, fp_collapsed_stacks_counters in fp_perf.items():
+        if pid not in dwarf_perf:
```
Jongy, line 116: What about the other way (only in `dwarf` and not in `fp`)

Jongy: I'm actually not sure this is the behavior we want. Since both `perf`s run concurrently and generate their samples at different intervals, we shouldn't inject samples from one into another.
Let's imagine a simple case - 1 core, 2 apps running at 100%. `fp` and `dwarf` were started at different intervals so there's a given offset between their interrupt (say, 30 ms). The `fp` might "fall" onto stacks from app 1, and the `dwarf` on stacks from app 2. If you merge in this case - then we're taking double the stacks for no good reason.

@DanielShaulov I recall we've already discussed it but I don't remember what was the outcome :sweat_smile:

Jongy: Whatever we pick - like suggested in https://github.com/Granulate/gprofiler/pull/77/files#r636259102, please add a `logger.debug` call giving the numbers:
* total collected fp
* total collected dwarf
* total merged

gprofiler/merge.py
Jongy: Since many of these are long, you can define type helpers, e.g

```
PerProcessStackCounts = Dict[..., Dict[....]]
```

gprofiler/merge.py
```diff
@@ -82,7 +82,85 @@ def collapse_stack(stack: str, comm: str) -> str:
...
+            merged_pid_to_stacks_counters[pid] = fp_collapsed_stacks_counters
+            continue
+
+        fp_frame_count_average = get_average_frame_count(fp_collapsed_stacks_counters.keys())
+        dwarf_collapsed_stacks_counters = dwarf_perf[pid]
```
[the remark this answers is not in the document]
Jongy: (typing would tell me easily :) )

gprofiler/merge.py
DanielShaulov: What you are saying is that we have 2 apps running at 100%, and the `fp` graph would only get samples from one of them?
Regardless of dwarf samples, this would mean current gprofiler would show a very skewed result (all samples from app1 vs 50-50).
If you mean that it might align perfectly with context switches, then we do need to set our frequency to something not divisible by 10, it is a common suggestion (like our default of 3999 in gserver, so maybe just 9 here? Or increase to 19?).
We should be able to trust the `fp` graph to be a good representation of how much CPU each process takes.

In any case - I didn't review the code yet to see if it fits the algorithm, but what it should do is treat the `fp` run as the "master" flamegraph, and assume all process proportions in it are "correct", then everything else is scaled to it - so a per-process perf results are scaled to the amount of samples in the `fp` graph for that process.
Dwarf is no different, we should look at the amount of samples of that process in the `dwarf` and the `fp` graph and scale to fit the `fp`. If it doesn't exist in the `fp` - it should not be in the end result at all. If it exists only in the `fp`, we would just the `fp` result for it. (Exactly the same logic is true for java and python currently)

gprofiler/merge.py
```diff
@@ -92,33 +170,37 @@ def parse_perf_script(script: str):
...
-                process_names[pid] = parsed["comm"]
-            elif parsed["stack"] is not None:
-                new_samples[collapse_stack(parsed["stack"], parsed["comm"])] += 1
-        except Exception:
-            logger.exception(f"Error processing sample: {parsed}")
+    for pid, stacks_counters in system_perf_pid_to_stacks_counter.items():
+        if pid in process_perfs:
+            per_process_samples[pid] += sum(stacks_counters.values())
+            continue
+        new_samples += stacks_counters  # type: ignore
```
Jongy, line 195: Why `# type: ignore`?

gprofiler/merge.py
Jongy: > If you mean that it might align perfectly with context switches, then we do need to set our frequency to something not divisible by 10.

My example was "catastrophic" to demonstrate the issue at its full extent, but I assume that even less catastrophic "flows" can yield skewed results.
I don't mean the extreme case that we happen to sample at the context switch times. Changing to 11hz/9hz/19hz is a good idea nonetheless (I've got no idea why I'd set it to 10 in the first place...) - but it won't mitigate what I have in mind.

> it should not be in the end result at all.

Yes, that's what I meant in https://github.com/Granulate/gprofiler/pull/77/files#r636267222 and that's the behavior now. So all good in that regard...

gprofiler/merge.py
```diff
@@ -82,7 +82,85 @@ def collapse_stack(stack: str, comm: str) -> str:
...
+        if fp_frame_count_average > dwarf_frame_count_average:
+            merged_pid_to_stacks_counters[pid] = fp_collapsed_stacks_counters
+        else:
+            dwarf_collapsed_stacks_counters = scale_dwarf_samples_count(
+                dwarf_collapsed_stacks_counters, fp_to_dwarf_total_sample_count_ratio
+            )
+            merged_pid_to_stacks_counters[pid] = dwarf_collapsed_stacks_counters
+
+
+def scale_dwarf_samples_count(
```
Jongy, line 132: I wonder if we need it at all. The number should be very close to 1. If this function actually gets to make any changes in the sample count for a process, that's weird for itself...

gprofiler/merge.py
alonbentamar: Added typing names

gprofiler/merge.py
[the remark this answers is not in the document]
alonbentamar: Added typing, but in any case I'm checking if the `dwarf_perf` has this key (`if pid not in dwarf_perf`) beforehand

gprofiler/merge.py
alonbentamar: Kept it for now, tell me if you want me to remove it

alonbentamar:
Pushed fixes for your comments, I'm now updating the screenshots

gprofiler/merge.py
Jongy: k

gprofiler/merge.py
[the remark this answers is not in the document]
Jongy: Mhm

gprofiler/main.py
```diff
@@ -389,6 +410,11 @@ def parse_cmd_args():
     if not args.upload_results and not args.output_dir:
         parser.error("Must pass at least one output method (--upload-results / --output-dir)")
 
+    if args.dwarf_stack_size > 65528:
+        parser.error("--perf-dwarf-stack-size maximum size is 65528")
+
+    if args.perf_mode in ("dwarf", "smart") and args.frequency > 100:
+        parser.error("--profiling-frequency|-f can't be larger than 100 when using --perf-mode smart|dwarf")
     return args
```
Jongy, line 418: Missing one blank before this

gprofiler/merge.py
```diff
@@ -82,7 +87,86 @@ def collapse_stack(stack: str, comm: str) -> str:
     return ";".join(funcs)
 
 
-def parse_perf_script(script: str):
+def merge_global_perfs(
+    raw_fp_perf: Optional[str], raw_dwarf_perf: Optional[str]
+) -> Tuple[ProcessToStackSampleCounters, ProcessIdToNameMapping]:
+    merged_pid_to_stacks_counters: ProcessToStackSampleCounters = defaultdict(Counter)
```
Jongy, line 93: Can be moved below, to where you first use it

gprofiler/perf.py
```diff
@@ -42,42 +46,39 @@ def __enter__(self):
...
             run_process(
                 [resource_path("perf")] + buildid_args + ["record"] + args + ["--", "sleep", str(self._duration)],
                 stop_event=self._stop_event,
             )
-            with open(parsed_path, "w") as f:
-                run_process(
-                    [resource_path("perf")] + buildid_args + ["script", "-F", "+pid", "-i", record_file.name], stdout=f
-                )
-            return parsed_path
+            perf_script_result = run_process(
```
Jongy, line 61: Pass `suppress_log=True` here

Jongy: (the output is very long, and we'll see the stdout/stderr on error anyway)

gprofiler/perf.py
```diff
@@ -42,42 +46,39 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.stop()
 
-    def _run_perf(self, filename_base: str, dwarf=False):
-        parsed_path = os.path.join(self._storage_dir, f"{filename_base}.parsed")
+    def _run_perf(self, dwarf: bool = False) -> str:
```
Jongy, line 50: Remove blank

gprofiler/perf.py
```diff
@@ -42,42 +46,39 @@ def __enter__(self):
...
+        if not self._dwarf_perf:
+            return merge_global_perfs(self._run_perf(dwarf=False), None)
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+            fp_future = executor.submit(self._run_perf, False)
+            dwarf_future = executor.submit(self._run_perf, True)
+        fp_perf = fp_future.result()
+        dwarf_perf = dwarf_future.result()
 
-        # TODO: run dwarf in parallel, after supporting it in merge.py
-        # Alternatively: fix golang dwarf problems and the run just dwarf
```
Jongy, line 75: Keep this TODO here somewhere - to remind why we're running both. The only currently issue I know of, is Go where DWARF doesn't seem to work.

gprofiler/merge.py
```diff
@@ -92,33 +176,39 @@ def parse_perf_script(script: str):
...
                 raise Exception("Failed to match sample")
-            yield match.groupdict()
+            sample_dict = match.groupdict()
+
+            pid = int(sample_dict["pid"])
+            process_name = sample_dict["comm"]
+            stack = sample_dict["stack"]
+            if stack is not None:
+                pid_to_collapsed_stacks_counters[pid][collapse_stack(stack, process_name)] += 1
+            pid_to_name.setdefault(pid, process_name)
```
Jongy, line 186: Why `setdefault`?

gprofiler/merge.py
```diff
@@ -92,33 +176,39 @@ def parse_perf_script(script: str):
...
+) -> str:
     per_process_samples: MutableMapping[int, int] = Counter()
-    new_samples: MutableMapping[str, int] = Counter()
-    process_names = {}
-    for parsed in perf_all:
-        try:
-            pid = int(parsed["pid"])
-            if pid in process_perfs:
-                per_process_samples[pid] += 1
-                process_names[pid] = parsed["comm"]
```
Jongy, line 109: I just realized there was a bug here (and it's also relevant for the new code): the `comm` may be thread-based, and until https://github.com/Granulate/gprofiler/pull/69 we always saw `comm` of threads. Now we should be seeing the `comm` of the process, but in older kernels, we'll still see the threads' one. So this code will set the `comm` of the last thread in the `perf script` output, for *all* other threads in that process...

Jongy: Creating an issue for it

Jongy:
Will continue the CR today

gprofiler/perf.py
alonbentamar: Not sure what you meant here, but added a comment so check it out

gprofiler/merge.py
```diff
@@ -82,7 +87,86 @@ def collapse_stack(stack: str, comm: str) -> str:
...
+            dwarf_collapsed_stacks_counters = scale_dwarf_samples_count(
+                dwarf_collapsed_stacks_counters, fp_to_dwarf_sample_ratio
+            )
+            merged_pid_to_stacks_counters[pid] = dwarf_collapsed_stacks_counters
+
+
+def scale_dwarf_samples_count(
+    dwarf_collapsed_stacks_counters: StackToSampleCount, fp_to_dwarf_sample_ratio: float
+) -> StackToSampleCount:
+    if fp_to_dwarf_sample_ratio == 1:
```
Jongy, line 146: I doubt it'll ever be 1 lol. But okay, can keep

alonbentamar: It should be always 1, because we use the same frequency and duration

gprofiler/merge.py
```diff
@@ -82,7 +87,86 @@ def collapse_stack(stack: str, comm: str) -> str:
...
+        return dwarf_collapsed_stacks_counters
+    # scale the dwarf stacks to the FP stacks to avoid skewing the results
+    for stack, sample_count in dwarf_collapsed_stacks_counters.items():
+        new_count = sample_count * fp_to_dwarf_sample_ratio
+        if new_count < 1 and random.random() > new_count:
```
Jongy, line 151: 1. Explain the rational behind this randomness
2. Any reason we apply this logic only for the `< 1` case and not for all cases? If the `new_count` is `1.13` then I want 1 sample at 13% and 2 samples at 87%, no? Although, if the `new_count` is `1.13` then something is way off anyway lol

gprofiler/merge.py
Jongy: "Should" is nice. Have you tried it? I never see `1` as the ratio.
This is, for example, from 5 seconds:
```
[2021-05-30 13:46:05,875] DEBUG: gprofiler.merge: Total FP samples: 416; Total DWARF samples: 441; FP to DWARF ratio: 0.9433106575963719; Total merged samples: 472
```
and this is from 60 seconds
```
[2021-05-30 13:47:30,700] DEBUG: gprofiler.merge: Total FP samples: 4817; Total DWARF samples: 4785; FP to DWARF ratio: 1.006687565308255; Total merged samples: 4833
```

gprofiler/perf.py
Jongy: I mean that in the longer term I prefer that we solve the core issue (DWARF producing bad samples in certain cases) instead of running both. So the comment "fix golang ..." should remain.

gprofiler/merge.py
alonbentamar: That's odd, I always see 1.0 - 
`DEBUG: gprofiler.merge: Total FP samples: 220; Total DWARF samples: 220; FP to DWARF ratio: 1.0; Total merged samples: 221`

Jongy: Baremetal vs virtual, I guess

gprofiler/merge.py
```diff
@@ -92,33 +176,39 @@ def parse_perf_script(script: str):
             match = SAMPLE_REGEX.match(sample)
             if match is None:
                 raise Exception("Failed to match sample")
-            yield match.groupdict()
+            sample_dict = match.groupdict()
+
+            pid = int(sample_dict["pid"])
+            process_name = sample_dict["comm"]
```
Jongy, line 182:
```suggestion
            comm = sample_dict["comm"]
```
Keep the logical naming please

gprofiler/merge.py
```diff
@@ -92,33 +176,39 @@ def parse_perf_script(script: str):
...
+            stack = sample_dict["stack"]
+            if stack is not None:
```
Jongy, line 184: I'm not sure in which cases we get an empty stack. But the sample should still be accounted for the PID, otherwise we'll see skewed results. I think you should insert an empty stack in that case. I know it's from the previous state of the code, but let's fix it along the way.

CC @DanielShaulov

gprofiler/merge.py
```diff
@@ -92,33 +176,39 @@ def parse_perf_script(script: str):
...
-            elif parsed["stack"] is not None:
-                new_samples[collapse_stack(parsed["stack"], parsed["comm"])] += 1
-        except Exception:
-            logger.exception(f"Error processing sample: {parsed}")
+    new_samples: StackToSampleCount = Counter()
+    for pid, stacks_counters in system_perf_pid_to_stacks_counter.items():
+        if pid in process_perfs:
+            per_process_samples[pid] += sum(stacks_counters.values())
+            continue
+        new_samples += stacks_counters
```
Jongy, line 203: I think it's nicer with an `else` here, to signify that these are 2 exclusive cases (we either account it for a specific PID, or we take the `perf` result)

gprofiler/merge.py
```diff
@@ -92,33 +176,39 @@ def parse_perf_script(script: str):
...
 
     for pid, perf_all_count in per_process_samples.items():
         process_stacks = process_perfs[pid]
         process_perf_count = sum(process_stacks.values())
         if process_perf_count > 0:
             ratio = perf_all_count / process_perf_count
             for stack, count in process_stacks.items():
-                full_stack = ";".join([process_names[pid], stack])
+                full_stack = ";".join([pid_to_name[pid], stack])
```
Jongy, line 211:
```suggestion
                full_stack = ";".join([pid_to_comm[pid], stack])
```
etc

gprofiler/perf.py
```diff
@@ -42,42 +46,41 @@ def __enter__(self):
...
-
-        # TODO: run dwarf in parallel, after supporting it in merge.py
-        # Alternatively: fix golang dwarf problems and the run just dwarf
+        if not self._fp_perf:
+            return merge_global_perfs(None, self._run_perf(dwarf=True))
+        if not self._dwarf_perf:
+            return merge_global_perfs(self._run_perf(dwarf=False), None)
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+            # We are running 2 perfs in parallel - one with DWARF and one with FP, and then we merge their results.
+            # This improves the results from software that is compiled without frame pointers, like some Go software.
```
Jongy, line 78: It's actually the other way around - Go has FP but doesn't work with DWARF for whatever reason which we believe is a bug in DWARF unwinding in `perf` or one of the libraries it's based on.

Jongy (changes requested):
Done reviewing. Let's talk about benching it

gprofiler/merge.py
alonbentamar: That would only change it to a different misrepresentation - a few stacks that originate from different stacks but don't have for some reason the stack string would all be aggregated to the same value

alonbentamar:
Need to fix conflicts

gprofiler/perf.py
Jongy: Your existing comment doesn't explain why we don't use just DWARF - the note on "like some native software" is correct, but explain why just DWARF is not enough.

gprofiler/merge.py
Jongy: ?

gprofiler/merge.py
Jongy: I didn't understand your comment lol.

My point here: sometimes, `perf` yields empty stacks for samples - without even 1 frame. I don't know why it happens. But the sample did happen - that process *did* execute at that time, so it should still be accounted....

Anyway, since this is here from before of your PR, I guess we can tackle it later.

Jongy:
2 small comments left. Please fix the conflicts & report benchmark results - the numbers I care about are how much did it worsen the effect of gProfiler, compared to when running with just FP.

gprofiler/merge.py
alonbentamar: To avoid checking if the PID already exists and setting it if it doesn't. It does all of that in one atomic operation.

alonbentamar:
@Jongy re-CR please?

Jongy:
> @Jongy re-CR please?

Yes.

Made the default `fp` until we bench?

alonbentamar:
> > @Jongy re-CR please?
> 
> Yes.
> 
> Made the default `fp` until we bench?

Done

Jongy:
I'll CR and we merge it, then add your manual tests as automatic in another PR.

Blackbox tests: run 2 apps - 1 python `while: 1` and 1 Go, same.
* Run the gProfiler container with `--perf-mode fp` and ensure there are proper Go stacks and no Python stacks (also run with `--no-python` so we get native CPython)
* Run the gProfiler container with `--perf-mode dwarf` and ensure the opposite.
* Run the gProfiler container with `--perf-mode auto` and ensure both look good.

Whitebox tests:
* on `merge_perfs`
  * scaling of runtime profilers based on global perf samples for that PID
  * scaling of fp vs dwarf
  * attaching container names per PID + container PID
  * check that the container names cache is refreshed every run

gprofiler/merge.py
```diff
@@ -96,33 +185,35 @@ def parse_perf_script(script: str):
...
                 raise Exception("Failed to match sample")
-            yield match.groupdict()
+            sample_dict = match.groupdict()
+
+            pid = int(sample_dict["pid"])
+            comm = sample_dict["comm"]
+            stack = sample_dict["stack"]
+            if stack is not None:
+                pid_to_collapsed_stacks_counters[pid][collapse_stack(comm, stack)] += 1
+            pid_to_comm.setdefault(pid, comm)
```
Jongy, line 195: Why `setdefault`?

gprofiler/perf.py
```diff
@@ -41,42 +46,43 @@ def __enter__(self):
...
+        dwarf_perf = dwarf_future.result()
 
-        # TODO: run dwarf in parallel, after supporting it in merge.py
-        # Alternatively: fix golang dwarf problems and the run just dwarf
-
-        # if get_cpu_count() > 32:
-        #     dwarf_frequency = 99
-        # else:
-        #     dwarf_frequency = 999
+        logger.info("Finished running global perf")
```
Jongy, line 86: This message is not shown anymore if we're not in `auto` mode.

How about you restructure this function so there's only 1 call to `merge_global_perfs` with possible `None` placeholders?

Jongy (changes requested):
2 final final final comments and let's merge.

gprofiler/merge.py
alonbentamar: Pretty sure I already answered a similar one - it's to set the comm only if the PID doesn't exist yet in the `pid_to_comm` dict in a single atomic operation

Jongy: Hmm haven't seen that previous one then. Okay

Jongy: I'm not sure I understand tho. You never actually write to this dict? Only with `setdefault` and with `update`

alonbentamar: The `setdefault` does the writing. I could replace this code with:
```python
if pid not in pid_to_comm:
    pid_to_comm[pid] = comm
```

Jongy: Yes, but you only write to it in this function. I thought `setdefault` overrides, but apparently it doesn't:
```
In [1]: x = {}                                                                                                                                                                                             

In [2]: x.setdefault("y", 1)                                                                                                                                                                               
Out[2]: 1

In [3]: x["y"]                                                                                                                                                                                             
Out[3]: 1

In [4]: x.setdefault("y", 2)                                                                                                                                                                               
Out[4]: 1

In [5]: x["y"]                                                                                                                                                                                             
Out[5]: 1
```

Alright then. Previous code used to take the last `comm` and now you're taking the first `comm`. What's the reason?

Jongy (approved):
1 small question left, can answer & update the branch & merge.

gprofiler/merge.py
alonbentamar: It's arbitrary either way.

Jongy: Hmm. Yup. Well, we will need to improve it somehow. It's problematic

Jongy: Yields weird stacks

PR #78, opened 2021-05-18
Fix linting so it would ignore external packages and use the venv if necessary

alonbentamar:
## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
## Checklist:

- [ ] My code follows the code style of this project.


lint.sh
```diff
@@ -1,7 +1,11 @@
...
 set -e
 
+if [ -f venv/bin/activate ]; then
+  source venv/bin/activate
+fi
+
 isort --settings-file .isort.cfg .
 flake8 --config .flake8 .
-black --line-length 120 --skip-string-normalization --check .
+black --line-length 120 --skip-string-normalization .
```
Jongy, line 10: Does black exit with nonzero in this case as well? Because otherwise, it will noe pass the CI

alonbentamar: It reformats the code so I don't think it ever exits with an error

Jongy: So if I push misformatted code, the CI will now pass... lol

I suggest we can check if we're running in the CI - then don't pass `--check`. Otherwise, do pass

alonbentamar: But you won't push misformatted code, because this line formats it for you

alonbentamar: Oh I see what you mean. I wanted to have a git hook automatically format the files, I didn't know the CI uses this as well. What do you think of separating them? (identical scripts but the hook removed the `--check` from the black argument)

Jongy (approved):
LGTM. Make sure that locally it indeed runs w/o `--check`.

alonbentamar:
> LGTM. Make sure that locally it indeed runs w/o `--check`.

Yeah, I checked it

PR #87, opened 2021-05-25
Initial PHP support using phpspy

guybortnikov:
## Description
Add initial support for PHP processes profiling using phpspy project (Granulate's fork) in continuous mode (pgrep mode in phpspy)

## Motivation and Context
Like Java and Python, we would like to support PHP runtime profiling.
At this point only PHP workloads running on top of [php-fpm](https://www.php.net/manual/en/install.fpm.php), in the future adding support to PHP running inside apache workers [`mod_php`](https://cwiki.apache.org/confluence/display/HTTPD/PHP) and other options.

### Notes:
* Based on Granulate's fork release  [Granulate phpspy v0.6.0-g3](https://github.com/Granulate/phpspy/releases/tag/v0.6.0-g3)

## How Has This Been Tested?
* Automatic tests have been written for both container and executable packaging methods. 
* Granulate phpspy fork was tested manually.

## Screenshots (if appropriate):
Drupal (running inside a container on the same host as the gProfiler): 
image

## Types of changes

## Checklist:

## Credits:
* [phpspy](https://github.com/adsr/phpspy) by [Adam Saponara](https://github.com/adsr). See [our fork](https://github.com/Granulate/phpspy).

### DIFF
--- a/gprofiler/php.py
+++ b/gprofiler/php.py
@@ -1,0 +1,234 @@
+import fcntl
+import glob
+import logging
+import os
+import re
+import signal
+from collections import Counter, defaultdict
+from functools import lru_cache
+from pathlib import Path
+from subprocess import Popen
+from threading import Event
+from typing import List, Mapping, MutableMapping, Optional, Pattern
+
+from gprofiler.exceptions import StopEventSetException
+from gprofiler.profiler_base import ProfilerBase
+from gprofiler.utils import limit_frequency, resource_path, start_process, wait_event
+
+logger = logging.getLogger(__name__)
+
+
+class PHPSpyProfiler(ProfilerBase):
+    PHPSPY_RESOURCE = "php/phpspy"
+    # Currently tracing only php-fpm, TODO: support mod_php in apache.
+    DEFAULT_PROCESS_FILTER = "php-fpm"
+    dump_signal = signal.SIGUSR2
+    poll_timeout = 10  # seconds
+    MAX_FREQUENCY = 999
+    MIN_DURATION = 3  # seconds, phpspy is running commands at bootstrap and it takes some time.
+    BUFFER_SIZE = 16384
+    NUM_WORKERS = 16  # Num of workers following unique PHP processes (one worker per _process)
+
+    # regex required for parsing phpspy output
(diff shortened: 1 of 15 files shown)


gprofiler/php.py
```diff
@@ -0,0 +1,234 @@
...
+        php_process_filter: str = DEFAULT_PROCESS_FILTER,
+    ):
+        self._frequency = limit_frequency(self.MAX_FREQUENCY, frequency, "phpspy", logger)
+
+        if duration < self.MIN_DURATION:
+            logger.warning(
+                f"Minimum duration for phpspy is {self.MIN_DURATION} (given {duration}), "
+                "raise duration in order to profile php processes"
+            )
+            self._enabled = False
```
Jongy, line 52: You fallthrough to `self._enabled = True`

Jongy: Why did you take this approach (of `_enabled`) instead of raising? I think it complicates things (for example, you can drop a `return` here, but then the `stop()` method which tries to access `self.process` will fail

gprofiler/php.py
```diff
@@ -0,0 +1,234 @@
...
+
+        self._duration = max(duration, self.MIN_DURATION)
+        self._stop_event = stop_event or Event()
+        self._storage_dir = storage_dir
+        logger.info(f"Initializing PHP profiler (frequency: {self._frequency}hz, duration: {self._duration}s)")
+        self.process: Optional[Popen] = None
```
Jongy, line 58: Can be `_process` and `_output_path` since they are private

gprofiler/utils.py
```diff
@@ -420,3 +420,14 @@ def reset_umask() -> None:
     Resets our umask back to a sane value.
     """
     os.umask(0o022)
+
+
+def limit_frequency(limit: int, requested: int, msg_header: str, runtime_logger: logging.Logger):
+    if requested > limit:
+        runtime_logger.warning(
+            f"{msg_header}: Requested frequency ({requested}) is higher than the limit {limit} "
```
Jongy, line 428:
```suggestion
            f"{msg_header}: Requested frequency ({requested}) is higher than the limit {limit}, "
```

Jongy (approved):

PR #88, opened 2021-05-25
Ensure gProfiler is run in init PID NS

Jongy:
## Description
Otherwise, perf produces garbage results, and the runtime profilers can't run because
they don't see any processes.

## Motivation and Context
This is a common mistake that's happened more than once to users. It causes the profiler to produce confusing results. Better catch the error early and notify.

## How Has This Been Tested?
I ran gProfiler as a Docker container without `--pid=host` and got the expected message.

I checked some old CentOS 6 (kernel 2.6.32) machine and verified my assumption of `kthreadd` holds there as well.

## Types of changes

## Checklist:


guybortnikov (approved):

leonid-granulate (approved):

Jongy:
Still flaky @guybortnikov :/

PR #89, opened 2021-05-25
Allow running without the system perf

Jongy:
## Description
Adds `--no-perf` to be able to run gProfiler without invoking `perf`.

TODOs:
* [x] Update README
* [x] Catch `perf record` errors, suggesting the user to try `--no-perf`
* [ ] ~~Ensure all profilers run with the same frequency to avoid incorrect scaling in the `--no-perf` case.~~
* [x] Add `comm` or another "first frame", because now we don't have it.

## Motivation and Context
`perf` features may be unavailable even when we're "root", for example due to lack of certain capabilities. Now, the runtime profilers might work because we can still `ptrace` target processes, request the JVM to load async-profier, etc.

For starters, this PR aims at the first case (lack of capabilities, but still being root, because we still test for it in `verify_preconditions`).

## How Has This Been Tested?
Tested without `--no-perf` - results from runtime profilers are still merged as expected.

Tested with `--no-perf` - we get the expected runtime profilers only concatenation.

## Types of changes

## Checklist:


gprofiler/main.py
```diff
@@ -245,7 +240,13 @@ def _snapshot(self):
                 logger.exception(f"{future.name} profiling failed")
 
         local_end_time = local_start_time + datetime.timedelta(seconds=(time.monotonic() - monotonic_start_time))
-        merged_result = merge.merge_perfs(system_future.result(), process_perfs)
+
+        system_result = system_future.result()
+        if self._runtimes["system"]:
+            merged_result = merge.merge_perfs(system_result, process_perfs)
+        else:
+            assert system_result == {}  # should be empty!
```
guybortnikov, line 248: Add assert message

Jongy:
Will add after #77 .

guybortnikov (changes requested):
Looks good. single small comment

Jongy:
Sanity works. Will work on the TODOs listed.

gprofiler/main.py
guybortnikov: meh

guybortnikov (dismissed):

michelhe:
@Jongy merge?

Jongy:
> @Jongy merge?

Still has some open TODOs :) Will fix them soon, and then we'll merge it.

Jongy:
I have finished all TODOs. Tested running gProfiler w/o `--privileged`, it fails and logs the error line suggesting you to run `--perf-mode none` (which doesn't work w/o `--privileged`, but that's not the point :)

Note: I've given up on the task:

>  Ensure all profilers run with the same frequency to avoid incorrect scaling in the --no-perf case.

because the differences between runtime profilers are just too much now: rbspy & phpspy don't check if the thread is active before collecting its stacks; py-spy collects with `--gil` so it won't see >100% CPU; ...

These ^^ are all points that should be improved in any case, but my point is that given this state *now*, I think that adding the frequencies verification won't help me with trusting the relative results (and in any case, if you're not using `perf`, you probably wish to focus certain processes, so the ratios between processes of different runtimes don't matter much).

Jongy:
Ready now @guybortnikov . I did make many reorganizations in `merge.py`, tested stuff manually but I think it might be time to add automatic tests for these parts..

tests/test_sanity.py
```diff
@@ -23,64 +23,76 @@
...
     tmp_path: Path,
     application_pid: int,
-    assert_collapsed: Callable[[Optional[Mapping[str, int]]], None],
+    assert_collapsed,
 ) -> None:
     with JavaProfiler(1000, 1, Event(), str(tmp_path)) as profiler:
-        process_collapsed = profiler.snapshot()
-        assert_collapsed(process_collapsed.get(application_pid))
+        process_collapsed = profiler.snapshot().get(application_pid)
+        assert_collapsed(process_collapsed, True)
```
guybortnikov, line 30:
```suggestion
        assert_collapsed(process_collapsed, check_comm=True)
```
please be explicit here since `assert_collapsed` is partial without full signature..
In all occurences

guybortnikov (changes requested):
looks good, a small change in tests

tests/test_sanity.py
Jongy: k

guybortnikov (approved):

PR #98, opened 2021-06-02
Get hostname from host UTS NS and log it

Jongy:
## Description
Need to get the hostname in init UTS NS, otherwise we might be seeing the container / Pod name.

## Motivation and Context
Get the correct hostname in containerized applications. Currently we instruct people to run gProfiler with `--network=host` for no good reason. I will remove it. Also, our k8s YAML doesn't have `hostNetwork: true`.

## How Has This Been Tested?
I added a print for `socket.gethostname()` w/o escaping the UTS NS. It prints the container hostname. This code prints the real hostname.

## Checklist:
- [x] Remove `--network=host` from README & in the backend
- [x] Update other callers of `socket.gethostname()` to use a cached value I get from the host UTS NS.

### DIFF
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -466,7 +466,5 @@ def limit_frequency(limit: int, requested: int, msg_header: str, runtime_logger:
 
 
 def get_hostname() -> str:
-    if hostname is None:
-        raise Exception("hostname not initialized!")
-
+    assert hostname is not None, "hostname not initialized!"
     return hostname
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -461,3 +461,10 @@ def limit_frequency(limit: int, requested: int, msg_header: str, runtime_logger:
         return limit
 
     return requested
+
+
+def get_hostname() -> str:
+    if hostname is None:
+        raise Exception("hostname not initialized!")
+
+    return hostname
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -332,6 +332,10 @@ def run_in_ns(nstypes: List[str], callback: Callable[[], None], target_pid: int 
 
 
 def log_system_info():
+    # initialized first
+    global hostname
+    hostname = "unknown"
+
     uname = platform.uname()
     logger.info(f"gProfiler Python version: {sys.version}")
     logger.info(f"gProfiler run mode: {get_run_mode()}")
@@ -342,8 +346,6 @@ def log_system_info():
 
     distribution = "unknown"
     libc_version = "unknown"
-    global hostname
-    hostname = "unknown"
 
     # move to host mount NS for distro & ldd.
     # now, distro will read the files on host.
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -296,7 +296,7 @@ def run_in_ns(nstypes: List[str], callback: Callable[[], None], target_pid: int 
     Runs a callback in a new thread, switching to a set of the namespaces of a target process before
     doing so.
 
-    Needed initially for swithcing mount namespaces, because we can't setns(CLONE_NEWNS) in a multithreaded
+    Needed initially for switching mount namespaces, because we can't setns(CLONE_NEWNS) in a multithreaded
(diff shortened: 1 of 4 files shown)


alonbentamar:
How could this solve the hostname that is sent to the backend?

Jongy:
> How could this solve the hostname that is sent to the backend?

The 5th task is "replace other calls of `socket.gethostname()` with the cached value from host UTS ..."

gprofiler/utils.py
```diff
@@ -338,19 +344,36 @@ def log_system_info():
...
+        except Exception:
+            logger.exception("Failed to get distribution")
+
+        try:
+            libc_version = get_libc_version()
+        except Exception:
+            logger.exception("Failed to get libc version")
 
-    run_in_ns(["mnt"], get_distro_and_libc)
-    assert len(results) == 2, f"only {len(results)} results, expected 2"
```
alonbentamar, line 350: We previously raised an error when we didn't get all of the expected values, but now you are only logging them and using `unknown` as a replacement value. Is this the new desired behavior?

gprofiler/utils.py
```diff
@@ -440,3 +463,10 @@ def limit_frequency(limit: int, requested: int, msg_header: str, runtime_logger:
         return limit
 
     return requested
+
+
+def get_hostname() -> str:
+    if hostname is None:
+        raise Exception("hostname not initialized!")
+
+    return hostname
```
alonbentamar, line 472: Do we want to return `unknown` when it fails?

gprofiler/utils.py
```diff
@@ -330,6 +332,10 @@ def _switch_and_run():
 
 
 def log_system_info():
```
alonbentamar, line 334: The function name doesn't make sense anymore, it does more then just logging

alonbentamar (changes requested):

gprofiler/utils.py
Jongy: Yep. This function is known to fail in certain cases (for example, there's no `ldd` executable/script in Google Co-OS so that part always fails). The different data items really are unrelated so I prefer that each can fail separately, raise its exception separately and use a placeholder for "error".

gprofiler/utils.py
Jongy: Yeah I was just lazy not splitting it off. Will do

gprofiler/utils.py
Jongy: Nope, I raise intentionally because it shouldn't be called before we try to initialize the hostname. That's actually an `assert`.

gprofiler/utils.py
alonbentamar: OK, so do you think it's fine to display a hostname filter with `unknown`? It might be better to handle that specific case

gprofiler/utils.py
alonbentamar: I meant that after you initialize it already, the hostname fetching might get an error (which you log). In that case, the hostname would be `unknown`.

alonbentamar (changes requested):
Replied to your comments

gprofiler/utils.py
Jongy: That's the same discussion as https://github.com/Granulate/gprofiler/pull/98#discussion_r646115821, no?

gprofiler/utils.py
Jongy: Oh, in the backend we should do it, yes. Was planning to after this PR.

What we can instead do is use a more specific "error" hostname, like `<unknown>` (which is less likely to happen in real life)

Jongy: Changed to `<unknown>`

gprofiler/utils.py
alonbentamar: yeah, pretty much

alonbentamar (approved):

alonbentamar:
Handle the `<unknown>` hostname in the backend before merging

Jongy:
> Handle the `<unknown>` hostname in the backend before merging

TBH I'm not sure there's any specific handling to do. Grouping the unknowns under `<unknown>` does make sense.

Jongy:
Also tested with `--uts=host`, works as well.

PR #99, opened 2021-06-03
Include parameters in POST query string for requests logging

Jongy:
## Description
Based on https://github.com/Granulate/gprofiler/pull/98.

This PR lets gProfiler send a bunch of useful parameters in the query string of its requests. This is done merely for logging purposes - we don't extract these values in our backend; our ALB is able to extract parameters from the query string, and we can use these for proper logging & monitoring of sent requests.

## Motivation and Context
Help us debug & track requests to the backend, by logging useful pieces of data by our ALB.

## How Has This Been Tested?
I ran this version and was able to view all data correctly in the ALB logs.

## Screenshots
Athena query sample:
Screenshot from 2021-06-04 01-03-03

## Checklist:

### DIFF
--- a/gprofiler/client.py
+++ b/gprofiler/client.py
@@ -56,7 +56,12 @@ class APIClient:
         return "{}/{}/{}".format(self._host.rstrip("/"), self.BASE_PATH, version)
 
     def _get_query_params(self) -> List[Tuple[str, str]]:
-        return [("key", self._key), ("service", self._service), ("hostname", self._hostname)]
+        return [
+            ("key", self._key),
+            ("service", self._service),
+            ("hostname", self._hostname),
+            ("timestamp", get_iso8601_format_time(datetime.datetime.utcnow())),
+        ]
 
     def _send_request(
         self,
--- a/gprofiler/client.py
+++ b/gprofiler/client.py
@@ -37,15 +37,16 @@ class APIClient:
         self._host: str = host
         self._upload_timeout = upload_timeout
         self._version: str = version
+        self._key = key
         self._service = service
         self._hostname = hostname
 
-        self._init_session(key, service)
+        self._init_session()
         logger.info(f"The connection to the server was successfully established (service {service!r})")
 
-    def _init_session(self, key: str, service: str):
+    def _init_session(self):
         self._session: Session = requests.Session()
-        self._session.headers.update({"GPROFILER-API-KEY": key, "GPROFILER-SERVICE-NAME": service})
+        self._session.headers.update({"GPROFILER-API-KEY": self._key, "GPROFILER-SERVICE-NAME": self._service})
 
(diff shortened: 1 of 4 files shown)


gprofiler/client.py
```diff
@@ -73,11 +83,13 @@ def _send_request(
...
 
+        opts["params"] = self._get_query_params() + [(k, v) for k, v in params.items()]
+
         resp = self._session.request(method, "{}/{}".format(self.get_base_url(api_version), path), **opts)
         if 400 <= resp.status_code < 500:
             try:
-                data = resp.json()
-                raise APIError(data["message"], data)
+                resp_data = resp.json()
+                raise APIError(resp_data["message"], resp_data)
```
guybortnikov, line 92: If you already touched this section I suggest  changing `resp_data` -> `response_data`

guybortnikov, line 92: And the case that message is a valid JSON but without a "message" key is not handled properly, either except `KeyError` or get with default

guybortnikov (changes requested):
few small comments

alonbentamar:
Change the base branch so we'll only see the relevant changes

Jongy:
> Change the base branch so we'll only see the relevant changes

Yup, after the merge.

gprofiler/client.py
```diff
@@ -52,19 +55,26 @@ def get_base_url(self, api_version: str = None) -> str:
         version = api_version if api_version is not None else self._version
         return "{}/{}/{}".format(self._host.rstrip("/"), self.BASE_PATH, version)
 
+    def _get_query_params(self) -> List[Tuple[str, str]]:
+        return [("key", self._key), ("service", self._service), ("hostname", self._hostname)]
```
alonbentamar, line 59: Perhaps add the current date? I think it will be useful, especially if we somehow have a request delay

alonbentamar (changes requested):

gprofiler/client.py
Jongy: k

Jongy: Replacing with `.get()`

gprofiler/client.py
Jongy: Mhm good call

Jongy:
Here's how it looks now:
```
https://profiler.granulate.io:443/api/v2/profiles?key=...&service=1.0.8-rc&hostname=jong-XPS-15-7590&timestamp=2021-06-06T12%3A59%3A55&samples=298
```

timestamp gets URL encoded, I thought of replacing `:` with `-` so it looks better in Athena, but I think that we'll query via some Pythonic API anyway.

alonbentamar (approved):

guybortnikov (approved):

PR #100, opened 2021-06-05
java: Robustness improvements, round 2

Jongy:
## Description
This PR improves the robustness of the Java profiler in several aspects:
1. We now handle processes where AP was already loaded & started - for example, of gProfiler is brutally stopped while profiling a Java process, AP remains running and the next time gProfiler runs, we can't start AP again. We now identify that case, stop AP and re-start it.
2. We no longer load AP from multiple paths - there is a single path `/tmp/gprofiler_tmp/async-profiler/libasyncProfiler.so` (per namespace, ofc) where AP resides.
3. We no longer store multiple copies of AP on-disk - previously we'd store 1 copy per process, now we store 1 copy per namespace.
4. ... to be added

It also improves the code by adding the `AsyncProfiledProcess` helper class, so we avoid long methods with large amount of parameters.

## Motivation and Context
Improve robustness of the Java profiler.

## How Has This Been Tested?
* Ran gProfiler, killed it (`kill -9`), ran it again and verified that it restarts AP.
* Stressed it on >100 Java processes to ensure that the AP copying logic works (injected random sleeps in it)

## Checklist:
  - [ ] AP already started in a container
  - [ ] AP already started on host
  - [ ] stress test with >100 java processes?

### DIFF
--- a/gprofiler/java.py
+++ b/gprofiler/java.py
@@ -75,7 +75,7 @@ class JavaProfiler(ProfilerBase):
 
     def _get_async_profiler_stop_cmd(
         self, pid: int, output_path: str, jattach_path: str, async_profiler_lib_path: str, log_path: str
-    ):
+    ) -> List[str]:
         return [
             jattach_path,
             str(pid),
@@ -85,12 +85,40 @@ class JavaProfiler(ProfilerBase):
             f"stop,file={output_path},{self.OUTPUT_FORMAT},{self.FORMAT_PARAMS},log={log_path}",
         ]
 
-    def _run_async_profiler(self, cmd: List[str], log_path_host: str):
+    def _run_async_profiler(self, cmd: List[str], log_path_host: str) -> None:
         try:
             run_process(cmd)
         except CalledProcessError:
             if os.path.exists(log_path_host):
                 logger.warning(f"async-profiler log: {Path(log_path_host).read_text()}")
+            raise
+
+    def _start_async_profiler(
+        self,
+        process: Process,
+        log_path_host: str,
+        log_path_process: str,
+        output_path_process: str,
+        libasyncprofiler_path_process: str,
+    ) -> None:
+        profiler_event = "itimer" if self._use_itimer else "cpu"
+
+        try:
+            self._run_async_profiler(
+                self._get_async_profiler_start_cmd(
+                    process.pid,
+                    profiler_event,
+                    self._interval,
+                    output_path_process,
+                    resource_path("java/jattach"),
+                    libasyncprofiler_path_process,
+                    log_path_process,
+                ),
+                log_path_host,
+            )
+        except CalledProcessError:
+            is_loaded = f" {libasyncprofiler_path_process}" in Path(f"/proc/{process.pid}/maps").read_text()
+            logger.warning(f"async-profiler DSO was{'' if is_loaded else ' not'} loaded into {process.pid}")
             raise
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


gprofiler/java.py
```diff
@@ -143,99 +287,45 @@ def _profile_process(self, process: Process) -> Optional[Mapping[str, int]]:
...
+            return self._profile_ap_process(ap_proc)
+
+    def _profile_ap_process(self, ap_proc: AsyncProfiledProcess) -> Optional[Mapping[str, int]]:
+        started = ap_proc.start_async_profiler(self._interval)
+        if not started:
+            logger.info(f"Found async-profiler already started on {ap_proc.process.pid}, trying to stop it...")
+            # stop, and try to start again. this might happen if AP & gProfiler go out of sync: for example,
+            # gProfiler being stopped brutally, while AP keeps running. If gProfiler is later started again, it will
+            # try to start AP again...
+            ap_proc.stop_async_profiler(with_output=False)
```
Jongy, line 300: Explain why stopping and not just using the `resume` action

guybortnikov (approved):

tests/conftest.py
```diff
@@ -118,18 +117,6 @@ def gprofiler_docker_image(docker_client: DockerClient) -> Iterable[Image]:
     yield docker_client.images.get("gprofiler")
 
 
-@fixture(scope="session")
-def gprofiler_docker_image_resources(gprofiler_docker_image: Image):
```
guybortnikov, line 122: hmm, If you removed this fixture, and now you rely that resources will be available on disk, execute the `copy_resource_from_image.sh` in the `test.sh` script

gprofiler/java.py
```diff
@@ -33,66 +34,214 @@
...
+                assert os.path.exists(self._libap_path_host)
+            else:
+                raise
+
+    def _recreate_log(self) -> None:
+        touch_path(self._log_path_host, self.OUTPUTS_MODE)
+
+    def _check_disk_requirements(self) -> None:
+        free_disk = psutil.disk_usage(self._ap_dir_host).free
+        if free_disk < 250 * 1024:
```
guybortnikov, line 149: Please explain in a comment the 250kb (I guess it's the ap itself + log + outputs) estimation

gprofiler/java.py
```diff
@@ -33,66 +34,214 @@
...
+            raise Exception(f"Not enough free disk space: {free_disk}kb (path: {self._output_path_host}")
```
guybortnikov, line 150:
```suggestion
            raise Exception(f"Not enough free disk space: {free_disk}kb (required...) (path: {self._output_path_host})")
```

guybortnikov (changes requested):
.

tests/conftest.py
Jongy: I don't want it to run implicitly because the user might need to supply the image name. It's now explained in README for local installation, so I think it's alright

gprofiler/java.py
Jongy: Yeah something like that. Just a limit to avoid killing the disk

gprofiler/java.py
Jongy: Improved

Jongy:
Not clear, passes on my box w/ Python 3.9.5

Jongy:
Welp, now the tests do pass

tests/conftest.py
guybortnikov: I see, let's do something else, let's check in the `test.sh` script that the resources are available prior to running the pytest test suite and if not prints a message and exits

gprofiler/java.py
guybortnikov: > Please explain in a comment


?

guybortnikov (changes requested):
looks very good. 
single unresolved comment

tests/conftest.py
Jongy: kk

gprofiler/java.py
Jongy: explained?

guybortnikov (dismissed):

Jongy:
Uhh last commit now breaks the executable tests

guybortnikov (approved):

Jongy:
Unrelated error

PR #101, opened 2021-06-05
Log command line of execution

Jongy:
## Description
Log the commandline gProfiler was executed with.

## Motivation and Context
Helps the reader of the logs understand the exact settings gProfiler is run with.

## How Has This Been Tested?
Tested locally (both container & standalone executable) and verified the output appears as I expect it to.

## Screenshots
```
[2021-06-05 22:04:20,441] INFO: gprofiler: Running gprofiler (version 1.1.0): -o /tmp/output --token <....> --service-name blabla -d 60 -v -c --no-python --no-php
```

## Checklist:

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -504,7 +504,7 @@ def main():
     reset_umask()
 
     try:
-        logger.info(f"Running gprofiler (version {__version__})...")
+        logger.info(f"Running gprofiler (version {__version__}): {' '.join(sys.argv[1:])}")
         try:
             log_system_info()
         except Exception:
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -504,7 +504,7 @@ def main():
     reset_umask()
 
     try:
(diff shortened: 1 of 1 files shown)


gprofiler/main.py
```diff
@@ -504,7 +504,7 @@ def main():
     reset_umask()
 
     try:
-        logger.info(f"Running gprofiler (version {__version__})...")
+        logger.info(f"Running gprofiler (version {__version__}): {' '.join(sys.argv[1:])}")
```
guybortnikov, line 507: wrap with quotes,
```suggestion
        logger.info(f"Running gprofiler (version {__version__}): '{' '.join(sys.argv[1:])}'")
```
maybe even add a "command line" prefix, so it will be very explicit (that this is the command line)

guybortnikov (changes requested):
small comment

gprofiler/main.py
Jongy: k

guybortnikov (approved):

PR #102, opened 2021-06-06
Don't install unneeded packages when building perf

d3dave:
## Description
The build installs `texlive-*-doc` packages which amount to 500MB+ download. For example, `texlive-latex-extra-doc`:
```
Get:359 http://archive.ubuntu.com/ubuntu xenial/main amd64 texlive-latex-extra-doc all 2015.20160320-1 [338 MB]
```
On a slow/metered connection this is a considerable barrier to building the image. We don't need these packages.

## How Has This Been Tested?
Build succeeds, perf still works. I don't think perf requires any of the recommended packages besides what's already specified.

### DIFF
--- a/scripts/perf_env.sh
+++ b/scripts/perf_env.sh
@@ -5,12 +5,18 @@
 #
 set -euo pipefail
 
-apt-get update && apt-get install -y \
-    build-essential \
+apt-get update
+apt-get install -y --no-install-recommends \
+    gcc \
+    g++ \
+    make \
     git \
     curl \
+    ca-certificates \
+    lbzip2 \
+    unzip \
+    python3 \
     autoconf \
-    asciidoc \
     libssl-dev \
     zlib1g-dev \
     libaudit-dev \


d3dave:
Here's the original list of additional recommended packages (excluding ones we already require):
```
patch
less
rsync
<ssh-client>
dblatex
docbook-utils
libxml2-utils
xmlto
libssl-doc
bzip2-doc
```

Jongy:
Very nice!

To make sure `perf` really hasn't changed, you can run `perf test` and verify the outputs match. That's what I did when I worked on https://github.com/Granulate/gprofiler/pull/69.

d3dave:
> Very nice!
> 
> To make sure `perf` really hasn't changed, you can run `perf test` and verify the outputs match. That's what I did when I worked on #69.

Output is identical.

Jongy:
kk

Jongy (approved):

PR #103, opened 2021-06-07
Use perfs output switching capability

d3dave:
## Description
Currently for each snapshot we execute perf on `sleep {duration}`. This PR replaces this with running perf without stopping and using the output switching capability to dump the recorded events on each snapshot.

## Motivation and Context
This should avoid enabling/disabling of perf events every `duration` step for each core.

## How Has This Been Tested?
Ran locally on Ubuntu 18 VM. Output files look ok.

## Checklist:

### DIFF
--- a/gprofiler/perf.py
+++ b/gprofiler/perf.py
@@ -4,15 +4,23 @@
 #
 import logging
 import os
+import signal
 from pathlib import Path
-from tempfile import NamedTemporaryFile
 from threading import Event
-from typing import Iterable, Mapping
+from typing import Iterable, Mapping, Optional
 
 import psutil
 
+from gprofiler.exceptions import CalledProcessError
 from gprofiler.merge import parse_perf_script
-from gprofiler.utils import TEMPORARY_STORAGE_PATH, resource_path, run_process
+from gprofiler.utils import (
+    TEMPORARY_STORAGE_PATH,
+    poll_process,
+    resource_path,
+    run_process,
+    start_process,
+    wait_for_file,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


d3dave:
I combined some misc changes in the same commit. I can split them if you prefer.

Jongy:
> I combined some misc changes in the same commit. I can split them if you prefer.

I don't mind, I can CR as a whole.

But please rebase before I start

gprofiler/utils.py
```diff
@@ -117,6 +121,14 @@ def wait_event(timeout: float, stop_event: Event, condition: Callable[[], bool])
             raise TimeoutError()
 
 
+def wait_for_file(pattern: Union[str, Path], timeout: float, stop_event: Event) -> Path:
+    if isinstance(pattern, Path):
+        pattern = str(pattern)
+    files = wait_event(timeout, stop_event, lambda: glob.glob(pattern))
```
alonbentamar, line 127: You should not do the `glob` with no delay, that's a ton of IO calls

alonbentamar: `os.path.exists` should be more efficient, since you know the exact file path for some functions

gprofiler/perf.py
```diff
@@ -80,3 +60,76 @@ def snapshot(self) -> Iterable[Mapping[str, str]]:
...
+        except TimeoutError:
+            process.kill()
+            logger.error(f"perf failed to start. stdout {process.stdout.read()!r} stderr {process.stderr.read()!r}")
+            raise
+        else:
+            self.process = process
+
+    def _dump(self) -> Path:
+        assert self.process is not None, "profiling not started!"
+        self.process.send_signal(signal.SIGUSR2)
```
alonbentamar, line 81: Document what this does (and maybe add a link to the relevant perf docs)

alonbentamar (changes requested):

gprofiler/utils.py
d3dave: First, there is a delay in `wait_event` of 0.1s (via `Event.wait`). Second, `glob` actually does `os.path.exists` when the pattern does not contain any wildcards (`Lib/glob.py`):
```
...
    if not has_magic(pathname):
        if basename:
            if os.path.lexists(pathname):
                yield pathname
...
```

Jongy: Since there's a sleep, it doesn't matter at all. But if there wasn't any sleep, it wouldn't matter if we use `glob.glob` or `os.path.exists` because both would reach 100% CPU. It doesn't matter if you do 1 IO call or 20 IO calls if you do them repeatedly (there's shouldn't be any disk access here, it's just the CPU we're using)

Jongy:
Superseded by #113 .

PR #105, opened 2021-06-07
Bump gProfiler version to 1.1.1

Jongy:
:)


d3dave:
We shouldn't deploy new PyPerf yet because we haven't fixed yet the issue with the `bpf_probe_read_user_str` return codes.

scripts/pyperf_build.sh
```diff
@@ -5,7 +5,7 @@
 #
 set -e
 
-git clone --depth 1 -b v1.0.5 https://github.com/Granulate/bcc.git && cd bcc && git reset --hard a2d0ae75ce59813eb571e0019f6640d2ee539bce
+git clone --depth 1 -b v1.0.7 https://github.com/Granulate/bcc.git && cd bcc && git reset --hard 3ffa8982fdbfce183da81af634dc39f743e65d59
```
d3dave, line 8: We shouldn't deploy new PyPerf yet because we haven't fixed yet the issue with the bpf_probe_read_user_str return codes.

d3dave (dismissed):

scripts/pyperf_build.sh
Jongy: But it's not new? Which issue are you referring to, exactly.

d3dave: It is new. Since https://github.com/Granulate/bcc/commit/af26c253f6421720efdb32a7230eae4032b14cf4 which is not in v1.0.5 - the currently deployed version.

Jongy: Given an error post https://github.com/Granulate/bcc/commit/af26c253f6421720efdb32a7230eae4032b14cf4, how would it appear before that patch?

d3dave: A valid frame might appear as an error.

d3dave: Woops, I meant the issue with module-level no-args functions (`get_classname`).

d3dave: Actually it seems like it would be ok. A bit fickle however.

Jongy: Besides of all issues that we *don't* have a solution for, all issues are solved

Jongy: #ציטוטים

guybortnikov (approved):

PR #108, opened 2021-06-10
Extract ProfilerInterface from ProfilerBase, and put more logic in the latter

Jongy:
## Description
Merge reusable logic into `ProfilerBase`.

## Motivation and Context
Improving the codebase.

## How Has This Been Tested?
CI

## Checklist:
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have updated the relevant documentation.
- [ ] I have added tests for new logic.

PR #109, opened 2021-06-10
python: Add --python-mode option

Jongy:
~~Based on https://github.com/Granulate/gprofiler/pull/108 and https://github.com/Granulate/gprofiler/pull/89, review only after.~~

## Description
Adds the `--python-mode` option:
* `auto` - try PyPerf, fallback to py-spy
* `pyperf` - PyPerf, and no fallback
* `pyspy` - py-spy w/o trying PyPerf
* `none` - like `--no-python`.

## Motivation and Context
Give tighter control over which profilers are run, exactly. It's possible to want to use PyPerf with no fallback, or to avoid trying PyPerf to begin with.

## How Has This Been Tested?
* Running with `auto` - PyPerf is selected, if I "hide" kernel headers then py-spy is selected
* Running with `auto` and killing PyPerf - `Python eBPF profiler failed (exit code: -15), falling back to py-spy`
* Running with `pyperf` and no kernel headers - no Python profiler is used
* Running with `pyperf` and PyPerf is killed - `Python eBPF profiler failed (exit code: -15)` and no pyspy next round.
* Running with `pyspy` just runs py-spy w/o PyPerf initialization
* Running with `none` runs no Python profiler at all.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,13 +47,32 @@ The default profiling frequency is *11 hertz*. Using higher frequency will lead 
 
 The default duration is *60 seconds*, and the default interval matches it. So gProfiler runs the profiling sessions back-to-back - the next session starts as soon as the previous session is done.
 
-* `--no-java`, `--no-python`, `--no-php`, `--no-ruby`: Disable the runtime-specific profilers.
+### Java profiling options
+
+* `--no-java`: Disable profilers for Java.
+
+### Python profiling options
+* `--no-python`: Alias of `--python-mode none`.
+* `--python-mode`: Controls which profiler is used for Python.
+    * `auto` - (default) try with PyPerf, fall back to py-spy.
+    * `pyperf` - Use PyPerf with no py-spy fallback.
+    * `pyspy` - Use py-spy.
+    * `none` - Disable profilers for Python.
+
+### PHP profiling options
+* `--no-php`: Disable profilers for PHP.
+* `--php-proc-filter`: Process filter (`pgrep`) to select PHP processes for profiling (this is phpspy's `-P` option)
+
+### Ruby profiling options
+* `--no-ruby`: Disable profilers for Ruby.
+
+### System profiling options
 
 * `--perf-mode`: Controls the global perf strategy. Must be one of the following options:
-  * `fp` - Use Frame Pointers for the call graph
-  * `dwarf` - Use DWARF for the call graph (adds the `--call-graph dwarf` argument to the `perf` command)
-  * `smart` - Run both `fp` and `dwarf`, then choose the result with the highest average of stack frames count, per process.
-  * `none` - Avoids running `perf` at all. See [perf-less mode](#perf-less-mode).
+    * `fp` - Use Frame Pointers for the call graph
+    * `dwarf` - Use DWARF for the call graph (adds the `--call-graph dwarf` argument to the `perf` command)
+    * `smart` - Run both `fp` and `dwarf`, then choose the result with the highest average of stack frames count, per process.
+    * `none` - Avoids running `perf` at all. See [perf-less mode](#perf-less-mode).
 
 ### Continuous mode
 gProfiler can be run in a continuous mode, profiling periodically, using the `--continuous`/`-c` flag.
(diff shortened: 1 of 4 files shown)


michelhe:
Use `PythonProfiler` in `test_sanity.py`, the ebpf profiler is broken

Jongy:
> Use `PythonProfiler` in `test_sanity.py`, the ebpf profiler is broken

Nope, it should remain `PythonEbpfProfiler` explicitly. The failure is just flakiness :( we happened to fall only on other syscalls and not on `getdents`. Which is really bad luck considering the frequency and duration...

Jongy:
Ready for review

gprofiler/python.py
```diff
@@ -118,6 +116,12 @@ def snapshot(self) -> ProcessToStackSampleCounters:
         return results
 
 
+class PythonEbpfError(CalledProcessError):
+    """ "
```
Jongy, line 120:
```suggestion
    """
```

gprofiler/python.py
```diff
@@ -280,31 +276,67 @@ def stop(self):
...
 
+    def __init__(
+        self,
+        frequency: int,
+        duration: int,
+        stop_event: Event,
+        storage_dir: str,
+        python_mode: str,
+    ):
+        # not calling my c'tor - we're just a proxy class
```
Jongy, line 293: Is it relevant? I use `ProfilerInterface` here, not `ProfilerBase`

michelhe: Doesn't really matter but if someone adds a constructor later..

README.md
```diff
@@ -47,13 +47,32 @@ The default profiling frequency is *11 hertz*. Using higher frequency will lead
...
+
+* `--no-java`: Disable profilers for Java.
+
+### Python profiling options
+* `--no-python`: Alias of `--python-mode none`.
+* `--python-mode`: Controls which profiler is used for Python.
+    * `auto` - (default) try with PyPerf, fall back to py-spy.
+    * `pyperf` - Use PyPerf with no py-spy fallback.
+    * `pyspy` - Use py-spy.
+    * `none` - Disable profilers for Python.
```
michelhe, line 60: worth to include explanation of pyspy vs PyPerf

gprofiler/main.py
```diff
@@ -392,20 +390,31 @@ def parse_cmd_args():
         "--rotating-output", action="store_true", default=False, help="Keep only the last profile result"
     )
 
-    parser.add_argument(
+    java_options = parser.add_argument_group("Java")
+    java_options.add_argument(
```
michelhe, line 394: do the same for php & ruby

gprofiler/python.py
```diff
@@ -245,18 +249,10 @@ def _dump(self) -> Path:
...
-            finally:
-                # in any case, we should stop PyPerf.
-                logger.warning("PyPerf dead/not responding, killing it")
-                process = self.process  # save it
-                self._terminate()
-                self._pyperf_error(process)
+            logger.warning("PyPerf dead/not responding, killing it")
+            process = self.process  # save it
+            self._terminate()
+            self._pyperf_error(process)
```
michelhe, line 255: Did you remove the try-finally for a reason ?

gprofiler/python.py
Jongy: Will change to the interface

gprofiler/main.py
Jongy: mhm

gprofiler/python.py
Jongy: Yes - the swap logic is now done in `PythonProfiler.snapshot`, so the `try` part actually happens later, and the `finally` part can happen immediately.

README.md
Jongy: k added a bit

gprofiler/python.py
Jongy: Hmm I already use the interface. So comment is irrelevant.

michelhe (approved):
LGTM, Waiting for CI

michelhe (approved):

PR #112, opened 2021-06-17
Warn on leftover unaccounted samples

d3dave:
## Description
Occasionally we may get samples for runtime processes from other profilers and not have any samples from `perf`. Currently these samples are just dropped and ignored (because we have no "global" measure to scale their counts). This change will warn when this happens. These samples are more probable for mostly inactive processes, so I don't expect their absence to affect results too much.

## How Has This Been Tested?
I ran it for about an hour on an idle Ubuntu VM and occasionally we get the message:
```
[16:37:36] Unaccounted samples for runtime pids 5254!
```
In this case the pid is gprofiler itself.

## Checklist:

### DIFF
--- a/gprofiler/merge.py
+++ b/gprofiler/merge.py
@@ -215,7 +215,7 @@ def merge_perfs(
                 new_samples[f"{container_name};{stack}"] += count
 
     for pid, perf_all_count in per_process_samples.items():
-        process_stacks = process_perfs[pid]
+        process_stacks = process_perfs.pop(pid)
         process_perf_count = sum(process_stacks.values())
         if process_perf_count > 0:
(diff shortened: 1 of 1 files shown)


DanielShaulov (approved):

Jongy (changes requested):
As you mentioned yourself - this thing "just happens" from time to time. On a system with many "runtime" processes running, this is even more likely - so why add a warning that pops on a regular basis? What're you trying to see with it?

What I think we care about here is if we happen to lose a large amount of samples. "Large", of course, is defined by the frequency and duration - so we can create a threshold based on them. Additionally, I think you should display the actual count of unaccounted samples, next to each PID.

PR #113, opened 2021-06-20
Run perf continuously

Jongy:
## Description
This PR changes `SystemProfiler` (which runs `perf`) to run in true continuous mode (like PyPerf & phpspy, and more to come, hopefully).
Continuous mode produces more accurate results because we don't miss anything - we're always sampling. Also, avoiding the re-starts of `perf` should yield better performance.

Supersedes https://github.com/Granulate/gprofiler/pull/103.
~~Based on https://github.com/Granulate/gprofiler/pull/89.~~

## Motivation and Context
Performance & accuracy.

## How Has This Been Tested?
* Basic sanity works.
* I ran it for 30 mins, then ran master for 30 mins, and compared the results - they were very much the same (on the same machine with similar load)

I will decide on more elaborate tests & preferably add them to our test suite.

## Checklist:
- [x] Test performance effect - ~~does it really help in benchmarks & do we see less `perf` in the flamegraph now :)~~ I ran `master` and this branch on the same machine, and compared the numbers of `--log-cpu-usage` & how much of `perf` I see in the graph. Both went down.
- [x] Complete all TODOs in code
- [x] Time to get rid of the `--profiling-interval` option? With `perf` now being continuous, there's no sense in allowing `duration != interval`.


Jongy:
> Time to get rid of the --profiling-interval option? With perf now being continuous, there's no sense in allowing duration != interval.

We have decided to get rid of it, in favor of true continuity.

tests/test_java.py
```diff
@@ -46,8 +46,17 @@ def test_java_async_profiler_stopped(
...
+    # run Java only (just so initialization is faster w/o others) for 1000 seconds
+    args = [
+        "-v",
+        "-d",
+        "1000",
+        "-o",
+        inner_output_directory,
+        "--no-php",
+        "--no-python",
+        "--perf-mode=none",
```
michelhe, line 58: Add --no-ruby?

gprofiler/perf.py
```diff
@@ -2,77 +2,139 @@
...
 from gprofiler.merge import ProcessToStackSampleCounters, merge_global_perfs
 from gprofiler.profiler_base import ProfilerBase
-from gprofiler.utils import TEMPORARY_STORAGE_PATH, resource_path, run_process
+from gprofiler.utils import resource_path, run_process, start_process, wait_event, wait_for_file_by_prefix
 
 logger = get_logger_adapter(__name__)
 
-PERF_BUILDID_DIR = os.path.join(TEMPORARY_STORAGE_PATH, "perf-buildids")
+
+class PerfProcess:
```
michelhe, line 21: Looks like your can derive from ProfilerBase here and get away with much of the code

gprofiler/perf.py
```diff
@@ -2,77 +2,139 @@
...
+            logger.error(f"perf failed to start. stdout {process.stdout.read()!r} stderr {process.stderr.read()!r}")
+            raise
+        else:
+            self._process = process
+            logger.info(f"Started perf ({self._type} mode)")
+
+    def stop(self) -> None:
+        if self._process is not None:
+            self._process.terminate()  # okay to call even if process is already dead
+            self._process.wait()
```
michelhe, line 63: timeout?

michelhe:
Changes LGTM

tests/test_java.py
Jongy: Can. I added `--perf-mode=none` specifically because perf was heavy and caused issues. Can disable ruby as well

gprofiler/perf.py
Jongy: Not really. It's not the same API. For example, its result isn't `ProcessToStackSampleCounters` but plain `perf script`.

gprofiler/perf.py
Jongy: `terminate` sends `SIGTERM`, which kills `perf`.

But I agree timeouts are nice nonetheless

Jongy: But we'll add timeouts somewhen else (we have plenty of other missing timeouts - for example, when running py-spy, I happened to see it hang). So in another PR. There's an issue open (or I will open one)

Jongy:
I'll update the README

michelhe (approved):

PR #115, opened 2021-06-20
Allow disabling the init PID NS check

Jongy:
## Description
Skipping the check allows running gProfiler in a container, if a user wants to do it for whatever reason.

## Motivation and Context
Allow running gProfiler in a container.

## How Has This Been Tested?
Without passing this flag, we get:
```
Please run me in the init PID namespace! In Docker, make sure you pass '--pid=host'. In Kubernetes, add 'hostPID: true' in the Pod spec.
You can disable this check with --disable-pidns-check
```

If passing this flag (and `--perf-mode none`) - gProfiler works :)

## Checklist:

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -393,6 +393,14 @@ def parse_cmd_args():
     )
 
     parser.add_argument(
+        "--disable-pidns-check",
+        action="store_false",
+        default=True,
+        dest="pid_ns_check",
+        help="Disable host PID NS check on startup",
+    )
+
+    parser.add_argument(
         "--no-java",
         dest="java",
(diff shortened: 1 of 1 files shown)


gprofiler/main.py
```diff
@@ -392,6 +392,14 @@ def parse_cmd_args():
         "--rotating-output", action="store_true", default=False, help="Keep only the last profile result"
     )
 
+    parser.add_argument(
+        "--disable-pidns-check",
+        action="store_false",
+        default=True,
```
michelhe, line 398: Do we want `default=True` here ?

guybortnikov (approved):

PR #120, opened 2021-06-24
Add assert failure messages where missing

Jongy:
## Description
Makes it possible to get more information about what happened by reading the exception
message.
I didn't add it for obvious cases (e.g: assert x is not None).
Also didn't add it for tests code, because pytest adds its introspection.

## Motivation and Context
See description ^^

## How Has This Been Tested?
CI

## Checklist:


Jongy:
In the longer term we should avoid using `assert`s since they allow such cases to happen. Exceptions with no information are very hard to handle (when all I have is the logs)

guybortnikov (approved):

PR #126, opened 2021-06-27
Preliminary NodeJS support

Jongy:
## Description
Small PoC that modifies our `perf` execution to add `perf inject --jit` - making it automatically handle processes that support `perf`'s "JIT" protocol, such as `node --perf-prof`.

This is less nice because it requires modifying code / configuration, unlike other profilers which require no changes at all (and besides async-profiler, truly non-intrusive to the app).

But... it just works :)

TODOs:
* [x] Add better explanation in README
* [ ] Add coloring for Node frames / builtins (perhaps color all frames that come from `/tmp/jitted-*.so`?)
* [x] Clean up jitted leftovers? After `perf inject --jit` or when gProfiler goes down?
* [x] Run `perf inject --jit` only if needed (to avoid overhead); or prove its overhead when there are no `jitdump` map entries is null
* [ ] Think about making this automatic (i.e, enabling `--perf-prof` in runtime)
* [x] ~~Possibly `-k 1` is not supported on all systems; in which case, don't use it and don't run `perf inject --jit`. Test it on some 2.6 kernel~~ now it's not automatically selected, so I don't mind
* [x] Add sanity tests

## Motivation and Context
NodeJS profiling support :)

## How Has This Been Tested?
Local sanity on `docker run -it --rm node:14 --perf-prof --interpreted-frames-native-stack` which runs `while (1) { }`

## Screenshots
Screenshot from 2021-06-28 01-15-52

## Checklist:


Jongy:
Since (for now) this ain't automatic and requires the user to pass `--perf-prof` in their app, I decided to also require the user to pass `--nodejs-mode perf` in order to enable the `perf inject --jit` part (avoiding it when not needed, for performance reasons)

Jongy:
@michelhe , *now* it will pass

Jongy:
Merged https://github.com/Granulate/gprofiler/pull/113 into here

michelhe:
Tests still failing

Jongy:
Kill me why, but doesn't work when Node runs on the host. Well, the `inject --jit` part does work:
```
node;__libc_start_main;node::Start;node::NodeMainInstance::Run;node::LoadEnvironment;node::StartExecution;node::StartExecution;node::ExecuteBootstrapper;v8::Function::Call;v8::internal::Execution::Call;v8::internal::(anonymous namespace)::Invoke;Builtin:JSEntry;Builtin:JSEntryTrampoline;InterpretedFunction: internal/main/run_main_module.js:1;InterpretedFunction:executeUserEntryPoint internal/modules/run_main.js:69;InterpretedFunction:Module._load internal/modules/cjs/loader.js:725;InterpretedFunction:Module.load internal/modules/cjs/loader.js:938;InterpretedFunction:Module._extensions..js internal/modules/cjs/loader.js:1094;InterpretedFunction:Module._compile internal/modules/cjs/loader.js:1039;[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];[/tmp/perf-91190.map];asm_sysvec_hyperv_stimer0_[k];sysvec_hyperv_stimer0_[k];irq_exit_rcu_[k];do_softirq_own_stack_[k];asm_call_sysvec_on_stack_[k];__softirqentry_text_start_[k]
``` 

but our fibonacci functions remains an unknown entry.

Possibly a difference in versions between the Node installed on the host vs the one we run in container...

Jongy:
> Possibly a difference in versions between the Node installed on the host vs the one we run in container...

Pfft, seems like this was the problem. I'm not even sure which version did it have installed. I'll test other LTS versions (12.x, 14.x) and post here; possibly will also add them as tests

Jongy:
> > Possibly a difference in versions between the Node installed on the host vs the one we run in container...
> 
> Pfft, seems like this was the problem. I'm not even sure which version did it have installed. I'll test other LTS versions (12.x, 14.x) and post here; possibly will also add them as tests

Ran the tests Dockerfile with `node:12` and `node:14`. They both exhibit the same results as https://github.com/Granulate/gprofiler/pull/126#issuecomment-882128624. I don't understand Node's "jitdump policy" (why does my fibonacci has no symbol in that case ....?) but that's out of scope for the PR, I guess.

@michelhe can review

gprofiler/main.py
```diff
@@ -551,7 +563,10 @@ def parse_cmd_args():
         parser.error("--perf-dwarf-stack-size maximum size is 65528")
 
     if args.perf_mode in ("dwarf", "smart") and args.frequency > 100:
-        parser.error("--profiling-frequency|-f can't be larger than 100 when using --perf-mode smart|dwarf")
+        parser.error("--profiling-frequency|-f can't be larger than 100 when using --perf-mode 'smart' or 'dwarf'")
```
michelhe, line 566: Hmm first time I notice this.
I don't think its a good idea to deny users the option to sample at a higher frequency. We could instead warn about disk space implications.

gprofiler/perf.py
```diff
@@ -18,21 +20,35 @@
 logger = get_logger_adapter(__name__)
 
 
+@functools.lru_cache(maxsize=1)
+def perf_path() -> str:
+    return resource_path("perf")
```
michelhe, line 25: why not do the lru_cache on `resource_path` itself?

gprofiler/perf.py
```diff
@@ -18,21 +20,35 @@
...
+        inject_jit: bool,
+        extra_args: List[str],
+    ):
         self._frequency = frequency
         self._stop_event = stop_event
         self._output_path = output_path
-        self._extra_args = extra_args
         self._type = "dwarf" if is_dwarf else "fp"
+        self._inject_jit = inject_jit
+        self._extra_args = extra_args + (["-k", "1"] if self._inject_jit else [])
```
michelhe, line 46: why is it needed for inject-jit ?

gprofiler/perf.py
Jongy: Smart one

gprofiler/perf.py
```diff
@@ -76,8 +92,16 @@ def wait_and_script(self) -> str:
         assert self._process is not None
         logger.debug(f"perf stderr: {self._process.stderr.read1(4096)}")
 
+        if self._inject_jit:
+            inject_data = Path(f"{str(perf_data)}.inject")
+            run_process(
+                [perf_path(), "inject", "--jit", "-o", str(inject_data), "-i", str(perf_data)],
```
michelhe, line 98: I guess we can't pass same file for `-i` and `-o` right ?

gprofiler/perf.py
Jongy: It needs to correlate timestamps in the `jitdump` file with those from `perf.data`. So both need to use the same clock

gprofiler/main.py
Jongy: It's not about disk space but about CPU usage. Anyway, that limit is indeed less relevant, we don't want to deny a conscious user from doing whatever they want. Will remove in another PR

tests/conftest.py
```diff
@@ -82,6 +88,12 @@ def gprofiler_exe(request, tmp_path: Path) -> Path:
     return tmp_path / "gprofiler"
 
 
+def _print_process_output(popen: subprocess.Popen) -> None:
```
michelhe, line 91: if the process is hanging, this will hang as well.
add a timeout

gprofiler/perf.py
Jongy: Nope, it reads as a stream

tests/conftest.py
```diff
@@ -93,20 +105,31 @@ def lower_privs():
...
+
+        # ensure, again, that it still alive (if it exited prematurely it might provide bad data for the tests)
+        try:
+            popen.wait(0)
+        except subprocess.TimeoutExpired:
+            pass
+        else:
+            _print_process_output(popen)
+            raise Exception(f"Command {command_line} exited unexpectedly during the test with {popen.returncode}")
+
```
michelhe, line 130: There are no tests that actively kill the process?

tests/conftest.py
Jongy: It's called only after kill/after it has exited

tests/test_sanity.py
```diff
@@ -67,6 +68,19 @@ def test_rbspy(
...
 
+@pytest.mark.parametrize("runtime", ["nodejs"])
+def test_nodejs(
+    tmp_path: Path,
+    application_pid: int,
+    assert_collapsed,
+    gprofiler_docker_image: Image,
+    runtime: str,
+) -> None:
+    with SystemProfiler(1000, 3, Event(), str(tmp_path), "fp", True, 0) as profiler:
```
michelhe, line 79: I genuinely don't like the invocation here, too many const arguments. kwargs will look better

michelhe (commented):
LGTM, notify me when ready for merge

tests/test_sanity.py
Jongy: k

tests/conftest.py
Jongy: Nope, why would we (it's the process being profiled, and these tests check just the profiling of it)

gprofiler/utils.py
```diff
@@ -46,6 +46,7 @@
 hostname: Optional[str] = None
 
 
+@lru_cache(maxsize=None)
```
michelhe, line 49: use `functools.cache` if you are already passing None.

image

Jongy: Too new (missing in e.g 3.7)

michelhe (dismissed):

michelhe (approved):

PR #127, opened 2021-06-28
Feature/metrics

alonbentamar:
Added metrics collection for new insights on the Performance Studio

## Description
I've added metrics collection, with the following data:
* Cloud Metadata - The instance type, region, and more
* System Metadata - CPU core count, total RAM, UNIX distro, kernel version and more.
* Average Metrics over collection period - average CPU and RAM usage 

## Motivation and Context

It will allow the performance studio to display relevant insights

## How Has This Been Tested?
I've logged the added data in the `profiles` endpoint in a custom performance studio version, and validated each value. 
For example, this is the data received in the `profile` upload JSON (without the profile itself):
```python
{'cpu_avg': 6.4,
 'end_time': '2021-07-02T10:46:11',
 'hostname': 'ip-10-0-0-9',
 'mem_avg': 2.0,
 'spawn_time': '2021-07-02T10:31:01',
 'start_time': '2021-07-02T10:46:06'}
```
This is the metadata received in the profile file content, with some values I changed manually (I've changed it so the metadata is added to the first line as a comment in the profile file) -
```json
{
  "containers": [
    "gprofiler"
  ],
  "container_names_enabled": true,
  "metadata": {
    "cloud_provider": "aws",
    "agent_version": "1.1.6",
    "spawn_time": "2021-07-20T13:58:06",
    "python_version": "3.8.5 (default, May 27 2021, 13:30:53) \n[GCC 9.3.0]",
    "run_mode": "container",
    "deployment_type": "containers",
    "kernel_release": "5.4.0-1029-aws",
    "kernel_version": "#30-Ubuntu SMP Tue Oct 20 10:06:38 UTC 2020",
    "system_name": "Linux",
    "processors": 16,
    "memory_capacity_mb": 127363,
    "hostname": "ip-10-0-0-9",
    "os_name": "Ubuntu",
    "os_release": "20.04",
    "os_codename": "focal",
    "libc_type": "glibc",
    "libc_version": "2.31-0ubuntu9.1",
    "hardware_type": "x86_64",
    "pid": 397028,
    "mac_address": "0D:FF:D6:81:1B:0B",
    "private_ip": "10.0.0.9",
    "spawn_uptime_ms": 709741.060411254,
    "cloud_info_wrapped": {
      "region": "us-east-1",
      "zone": "us-east-1a",
      "instance_type": "r5.4xlarge",
      "life_cycle": "on-demand",
      "account_id": "123123",
      "image_id": "ami-asdasdasd",
      "instance_id": "i-basdbasd"
    },
    "run_arguments": {
      "config": null,
      "frequency": 11,
      "duration": 60,
      "output_dir": null,
      "flamegraph": true,
      "rotating_output": false,
      "log_cpu_usage": false,
      "java": true,
      "python_mode": "auto",
      "php": true,
      "php_process_filter": "php-fpm",
      "ruby": true,
      "perf_mode": "fp",
      "dwarf_stack_size": 8192,
      "upload_results": true,
      "server_host": "http://host.docker.internal:8080",
      "server_upload_timeout": 120,
      "server_token": "<TOKEN>",
      "service_name": "my-test",
      "verbose": false,
      "log_file": "/var/log/gprofiler/gprofiler.log",
      "log_rotate_max_size": 5242880,
      "log_rotate_backup_count": 1,
      "log_to_server": true,
      "disable_container_names": false,
      "continuous": true,
      "continuous_profiling_interval": 60,
      "pid_ns_check": true,
      "collect_metrics": true,
      "collect_metadata": true
    },
    "current_time": "2021-07-20T13:59:12"
  }
}
```

## Checklist:


gprofiler/cloud_metadata.py
```diff
@@ -0,0 +1,161 @@
...
+    return AwsInstanceMetadata("aws", region, zone, instance_type, life_cycle, account_id, image_id, instance_id)
+
+
+def get_gcp_metadata() -> Optional[GcpInstanceMetadata]:
+    response = send_request(
+        "http://metadata.google.internal/computeMetadata/v1/instance/?recursive=true",
+        headers={"Metadata-Flavor": "Google"},
+    )
+    if response is None:
+        return response
```
sobolron, line 77:
```suggestion
        return None
```
for consistency

gprofiler/cloud_metadata.py
```diff
@@ -0,0 +1,161 @@
+import logging
+from dataclasses import dataclass
+from typing import Dict, List, Optional, Union
+
+import requests
+from requests import Response
+
+from gprofiler.exceptions import BadResponseCode
+
+AWS_TIMEOUT = 5
```
sobolron, line 10: Not used only for AWS.

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,81 @@
...
+import psutil
+
+DEFAULT_POLLING_INTERVAL_SECONDS = 5
+
+
+class SystemMetricsMonitor:
+    def __init__(self, max_memory_poll_age_seconds: int, polling_rate_seconds: int = DEFAULT_POLLING_INTERVAL_SECONDS):
+        self._cpu_count = psutil.cpu_count() or 1
+        self._mem_percentages: Deque[float] = deque(maxlen=round(max_memory_poll_age_seconds / polling_rate_seconds))
+        self._last_cpu_poll_time: Optional[float] = None
```
sobolron, line 16: Does this ever change? Not really sure I understand it's usage.

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,81 @@
...
+        self._last_cpu_utilization_percentages: Optional[Tuple[float, float]] = None
+        self._stop_event = Event()
+        self._thread = Thread(target=self._continuously_polling_memory, args=(polling_rate_seconds,))
+
+        self.get_cpu_utilization()  # Call this once to set the necessary data
+
+    def start(self):
+        self._stop_event.clear()
+        self._thread.start()
```
sobolron, line 25: Thread is not reusable after stopping it... so Event doesn't have to be cleared, or thread can be created in here.

sobolron (changes requested):

gprofiler/system_metrics.py
alonbentamar: Yes, every time `get_cpu_utilization()` is called:
```python
current_time = self._last_cpu_poll_time = time.monotonic() * self._cpu_count
```

gprofiler/system_metrics.py
alonbentamar: Made it reusable

gprofiler/system_metrics.py
```diff
@@ -11,22 +11,27 @@
...
     def start(self):
+        assert self._thread is None, "SystemMetricsMonitor is already running"
         self._stop_event.clear()
+        self._thread = Thread(target=self._continuously_polling_memory, args=(self._polling_rate_seconds,))
         self._thread.start()
 
     def stop(self):
         self._stop_event.set()
+        assert self._thread is not None, "SystemMetricsMonitor was not running"
         self._thread.join()
```
sobolron, line 33: Maybe add timeout to the join in order to not block everything? (raise if timed out)

sobolron (changes requested):

gprofiler/cloud_metadata.py
```diff
@@ -0,0 +1,161 @@
...
+
+def send_request(url: str, headers: Dict[str, str] = None) -> Optional[Response]:
+    response = requests.get(url, headers=headers or {}, timeout=AWS_TIMEOUT)
+    if not response.ok:
+        raise BadResponseCode(response.status_code)
+    return response
+
+
+def get_cloud_instance_metadata() -> Optional[Dict[str, Union[str, bool, int]]]:
+    cloud_metadata_fetchers = [get_aws_metadata, get_gcp_metadata, get_azure_metadata]
```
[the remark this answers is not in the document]
alonbentamar: As far as I'm aware, this is the only way that works consistently on all of the cloud providers

gprofiler/system_metrics.py
```diff
@@ -30,7 +31,7 @@ def start(self):
     def stop(self):
         self._stop_event.set()
         assert self._thread is not None, "SystemMetricsMonitor was not running"
-        self._thread.join()
+        self._thread.join(STOP_TIMEOUT_SECONDS)
```
sobolron, line 34: Please raise if timeout happened. I don't want to leak threads with this.
image

sobolron (changes requested):

gprofiler/client.py
```diff
@@ -78,7 +78,11 @@ def _send_request(
             opts["headers"]["Content-type"] = "application/json"
             buffer = BytesIO()
             with gzip.open(buffer, mode="wt", encoding="utf-8") as gzip_file:
-                json.dump(data, gzip_file, ensure_ascii=False)  # type: ignore
+                try:
+                    json.dump(data, gzip_file, ensure_ascii=False)  # type: ignore
+                except TypeError:
+                    logger.exception("Given data is not a valid JSON!", bad_json=str(data))
```
Jongy, line 84: Why do you need this? How can it happen with our structured data? And logging `str(dta)` is probably too much (requests can be in the order of MBs, don't log them all)

gprofiler/client.py
```diff
@@ -122,8 +133,19 @@ def submit_profile(
...
+                "cpu_avg": cpu_avg,
+                "mem_avg": mem_avg,
+                "spawn_time": get_iso8601_format_time_from_epoch_time(spawn_time),
             },
             timeout=self._upload_timeout,
             api_version="v2",
             params={"samples": str(total_samples), "version": __version__},
         )
+
+    def submit_metadata(self, metadata: Dict):
```
Jongy, line 145:
```suggestion
    def submit_metadata(self, metadata: Dict[str, str]):
```
?

gprofiler/client.py
```diff
@@ -122,8 +133,19 @@ def submit_profile(
...
+        return self.post(
+            "agent-metadata",
+            metadata,
+            timeout=self._upload_timeout,
```
Jongy, line 149: I'd expect `post()` to pass `timeout=self._upload_timeout` itself. The purpose of that API is to hide common request details. Fix it?

gprofiler/cloud_metadata.py
```diff
@@ -0,0 +1,161 @@
+from dataclasses import dataclass
```
Jongy, line 1: Add file header (copy from another one)

gprofiler/cloud_metadata.py
```diff
@@ -0,0 +1,161 @@
...
+logger = get_logger_adapter(__name__)
+
+
+@dataclass
+class InstanceMetadataBase:
+    provider: str
+
+
+@dataclass()
+class AwsInstanceMetadata(InstanceMetadataBase):
```
Jongy, line 21: Document these classes. Where did you get their structure from (web link to relevant AWS/GCP/whatever page)

gprofiler/cloud_metadata.py
```diff
@@ -0,0 +1,161 @@
...
+
+logger = get_logger_adapter(__name__)
+
+
+@dataclass
+class InstanceMetadataBase:
+    provider: str
+
+
+@dataclass()
```
Jongy, line 20:
```suggestion
@dataclass
```

gprofiler/cloud_metadata.py
Jongy, line 21: You can add here `provider: str = "aws"` to have it for all instances for this class. (just need to add it last)

gprofiler/cloud_metadata.py
```diff
@@ -0,0 +1,161 @@
...
+    resource_id: str
+    instance_id: str
+    name: str
+    image_info: Optional[Dict[str, str]]
+
+
+def get_aws_metadata() -> Optional[AwsInstanceMetadata]:
+    metadata_response = send_request("http://169.254.169.254/latest/dynamic/instance-identity/document")
+    life_cycle_response = send_request("http://169.254.169.254/latest/meta-data/instance-life-cycle")
+    if life_cycle_response is None or metadata_response is None:
```
Jongy, line 58: Why either? If any succeeds, isn't it AWS?

Jongy: Or both must always succeed anyway

gprofiler/cloud_metadata.py
```diff
@@ -0,0 +1,161 @@
...
+        return None
+    instance_metadata = metadata_response.json()
+    region = instance_metadata["region"]
+    zone = instance_metadata["availabilityZone"]
+    instance_type = instance_metadata["instanceType"]
+    account_id = instance_metadata["accountId"]
+    image_id = instance_metadata["imageId"]
+    instance_id = instance_metadata["instanceId"]
+    life_cycle = life_cycle_response.text
+    return AwsInstanceMetadata("aws", region, zone, instance_type, life_cycle, account_id, image_id, instance_id)
```
Jongy, line 68: Can you make it shorter somehow? For example, you can use the exact field names of AWS (also makes it more readable), then you can just `return AwsInstance(**instance_metadata)`

gprofiler/cloud_metadata.py
```diff
@@ -0,0 +1,161 @@
...
+    raised_exceptions: List[Exception] = []
+    for fetcher in cloud_metadata_fetchers:
+        try:
+            response = fetcher()
+            if response is not None:
+                return response.__dict__
+        except Exception as exception:
+            raised_exceptions.append(exception)
+    formatted_exceptions = ', '.join([repr(exception) for exception in raised_exceptions])
+    logger.debug(
```
Jongy, line 158: Make sure you don't display any error if it's really not AWS

gprofiler/main.py
```diff
@@ -532,6 +565,14 @@ def parse_cmd_args():
         help="Disable host PID NS check on startup",
     )
 
+    parser.add_argument(
+        "--disable-metrics-collection",
+        action="store_false",
+        default=True,
+        dest="collect_metrics",
+        help="Disable sending metrics and cloud metadata to the performance studio",
```
Jongy, line 573:
```suggestion
        help="Disable sending metrics and cloud metadata to the Performance Studio",
```

gprofiler/main.py
```diff
@@ -662,13 +703,15 @@ def main():
...
             runtimes,
             client,
+            args.collect_metrics,
             state,
             not args.disable_container_names,
             remote_logs_handler,
             args.php_process_filter,
         )
         logger.info("gProfiler initialized and ready to start profiling")
-
```
Jongy, line 671: Bring it back

gprofiler/main.py
```diff
@@ -230,8 +239,23 @@ def _strip_container_data(collapsed_data):
             lines.append(line[line.find(';') + 1 :])
         return '\n'.join(lines)
 
+    def send_metadata(self):
```
Jongy, line 242: `GProfiler` grows too large and this logic really doesn't need to be here. Move to another class that also has access to the `APIClient` object? (and it can be a field of `GProfiler`, or altogether a different class / function that's just invoked once with the `APIClient` as gProfiler boots)

Jongy, line 242: This may raise - wherever you end up putting it, make sure it doesn't stop gProfiler from running

Jongy, line 242: This blocks. It runs after profiling has started. This logic should run in a different thread - don't slow down profiling / harm its accuracy by any way.

gprofiler/main.py
```diff
@@ -230,8 +239,23 @@ def _strip_container_data(collapsed_data):
...
+            "start_time": round(self._spawn_time),
+        }
+        metadata_dict.update(system_metadata.get_dict())
+        if cloud_metadata is not None:
+            metadata_dict["cloud_info_wrapped"] = cloud_metadata
+        self._client.submit_metadata(metadata_dict)
+
     def start(self):
         self._stop_event.clear()
+        if self._system_metrics_monitor is not None:
```
Jongy, line 257: To avoid all the `is None` checks, I think a nicer impl is to create a dummy class (like `NoopProfiler`) that returns `None`  when you call your methods

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+from gprofiler.exceptions import ThreadStopTimeoutError
+
+DEFAULT_POLLING_INTERVAL_SECONDS = 5
+STOP_TIMEOUT_SECONDS = 30
+
+
+class SystemMetricsMonitor:
+    def __init__(self, max_memory_poll_age_seconds: int, polling_rate_seconds: int = DEFAULT_POLLING_INTERVAL_SECONDS):
+        self._polling_rate_seconds = polling_rate_seconds
+        self._cpu_count = psutil.cpu_count() or 1
```
Jongy, line 18: `or 1`?

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+        self._stop_event.clear()
+        self._thread = Thread(target=self._continuously_polling_memory, args=(self._polling_rate_seconds,))
+        self._thread.start()
+
+    def stop(self):
+        self._stop_event.set()
+        assert self._thread is not None, "SystemMetricsMonitor was not running"
+        self._thread.join(STOP_TIMEOUT_SECONDS)
+        if self._thread.is_alive():
+            raise ThreadStopTimeoutError("The SystemMetricsMonitor internal thread could not stop because of a timeout")
```
Jongy, line 38:
```suggestion
            raise ThreadStopTimeoutError("Timed out while waiting for ... to stop")
```
better explains it

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _continuously_polling_memory(self, polling_rate_seconds: int):
+        while not self._stop_event.is_set():
+            start_time = time.monotonic()
+            current_ram_percent = psutil.virtual_memory().percent
+            self._mem_percentages.append(current_ram_percent)
+            elapsed = time.monotonic() - start_time
+            self._sleep(polling_rate_seconds - elapsed)
+
+    def _sleep(self, amount_seconds: float):
```
Jongy, line 49: Hmm? Why not `self._stop_event.wai()`?

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+            time.sleep(sleep_interval / 100)
+            if self._stop_event.is_set():
+                return
+
+    def get_average_memory_utilization(self) -> Optional[float]:
+        current_length = len(self._mem_percentages)
+        if current_length == 0:
+            return None
+        all_percentages: List[float] = []
+        # Avoid a race condition by only fetching the current X values. More values can be added in the meantime by
```
Jongy, line 61: What race? Why is it racy? It's okay, just pop all values until none are left

Jongy: This can also just be a list and you can swap it with an empty one before calculating the mean. Will simplify the code

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def get_cpu_utilization(self) -> float:
+        """
+        Returns the CPU utilization percentage since the last time this method was called.
+        Based on the psutil.cpu_percent method.
+        """
+        last_user, last_system = self._last_cpu_utilization_percentages or (None, None)
+        current_cpu_times = psutil.cpu_times()
+        current_user, current_system = current_cpu_times.user, current_cpu_times.system
+        self._last_cpu_utilization_percentages = (current_user, current_system)
```
Jongy, line 75: Don't call it `percentages` since it's not percentages

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+        last_time = self._last_cpu_poll_time
+        current_time = self._last_cpu_poll_time = time.monotonic() * self._cpu_count
+
+        if None in (last_user, last_system):
+            return 0.0
+        assert last_time is not None
+        delta_cpu = (current_user - last_user) + (current_system - last_system)
```
Jongy, line 82: You're calculating here what `psutil.cpu_percent` already does. Use it?

gprofiler/client.py
leonid-granulate: This endpoint is not in use, we get the metadata with each profiler request

gprofiler/client.py
alonbentamar: It's useful mainly for development to have a better and more indicative error message, I don't expect it to happen in production.
I've added a cutoff

gprofiler/client.py
alonbentamar: Removed in #134 , no point in fixing

gprofiler/client.py
alonbentamar: Removed in #134

gprofiler/cloud_metadata.py
alonbentamar: Dataclasses completely failed on me on Python 3.6 in mypy, so I gave up on them in #134

gprofiler/cloud_metadata.py
alonbentamar: Removed in #134

gprofiler/cloud_metadata.py
alonbentamar: Removed in #134

gprofiler/cloud_metadata.py
alonbentamar: I don't want to send a partial response. If one of them fails, it will also fail on the other cloud types and therefore it will log the exceptions

gprofiler/cloud_metadata.py
alonbentamar: It won't work on all of the cloud providers anyway, and also somewhat changed in #134

gprofiler/cloud_metadata.py
alonbentamar: Only on `verbose` mode

gprofiler/main.py
alonbentamar: Removed in #134

alonbentamar: Removed in #134

alonbentamar: Removed in #134

gprofiler/system_metrics.py
alonbentamar: Taken from `psutil. cpu_percent`, I suppose it can fail and return 0 or `None`

gprofiler/system_metrics.py
alonbentamar: Oh my God

gprofiler/system_metrics.py
alonbentamar: I didn't want to use an interval, as it would block and would make me open a new thread. Don't really remember why I didn't use it without an interval, as it should keep the "last values" itself, but pretty sure I had a good reason

gprofiler/main.py
```diff
@@ -546,6 +577,30 @@ def parse_cmd_args():
         help="Disable host PID NS check on startup",
     )
 
+    parser.add_argument(
+        "--disable-metrics-collection",
+        action="store_false",
+        default=True,
+        dest="collect_metrics",
+        help="Disable sending system metrics to the performance studio",
```
michelhe, line 585: performance studio -> Performance Studio

gprofiler/main.py
```diff
@@ -546,6 +577,30 @@ def parse_cmd_args():
...
+        dest="collect_metadata",
+        help="Disable sending system and cloud metadata to the Performance Studio",
+    )
+
+    parser.add_argument(
+        "--disable-argument-uploading",
+        action="store_false",
+        default=True,
+        dest="send_args",
+        help="Disable sending the gProfiler arguments to the performance studio",
```
michelhe, line 601: here as well, performance studio -> Performance Studio

gprofiler/main.py
michelhe: image

README.md
```diff
@@ -81,6 +81,14 @@ Profiling using eBPF incurs lower overhead & provides kernel stacks. This (curre
 **By default, gProfiler sends logs to Granulate Performance Studio** (when using `--upload-results`/`-u` flag)
 This behavior can be disabled by passing `--dont-send-logs` or the setting environment variable `GPROFILER_DONT_SEND_LOGS=1`.
 
+### Metrics and metadata collection
```
Jongy, line 84: Also mention that if `-u` is not given, nothing is uploaded (and ensure it's actually so :sweat_smile: )

gprofiler/client.py
```diff
@@ -78,7 +79,14 @@ def _send_request(
             opts["headers"]["Content-type"] = "application/json"
             buffer = BytesIO()
             with gzip.open(buffer, mode="wt", encoding="utf-8") as gzip_file:
-                json.dump(data, gzip_file, ensure_ascii=False)  # type: ignore
+                try:
+                    json.dump(data, gzip_file, ensure_ascii=False)  # type: ignore
+                except TypeError:
+                    bad_json = str(data)
```
Jongy, line 85: I don't understand how it can happen. We are passing structured data

gprofiler/client.py
```diff
@@ -113,7 +121,14 @@ def get_health(self):
...
 
     def submit_profile(
-        self, start_time: datetime.datetime, end_time: datetime.datetime, profile: str, total_samples: int
+        self,
+        start_time: datetime.datetime,
+        end_time: datetime.datetime,
+        profile: str,
+        total_samples: int,
+        spawn_time: float,
+        cpu_avg: Optional[float],
```
Jongy, line 130: Encapsulate in some "metrics" object (so later we can add more fields with less frtiction)

gprofiler/main.py
```diff
@@ -98,8 +100,11 @@ def __init__(
         python_mode: str,
         runtimes: Dict[str, bool],
         client: APIClient,
+        collect_metrics: bool,
+        collect_metadata: bool,
         state: State,
         cpu_usage_logger: CpuUsageLogger,
+        run_args: Dict,
```
Jongy, line 107:
```suggestion
        run_args: Dict[str, str],
```
?

gprofiler/main.py
```diff
@@ -289,26 +306,40 @@ def _snapshot(self):
...
                 process_profiles,
                 self._docker_client,
+                metadata,
             )
 
         if self._output_dir:
             self._generate_output_files(merged_result, local_start_time, local_end_time)
 
         if self._client:
+            cpu_avg = self._system_metrics_monitor.get_cpu_utilization()
```
Jongy, line 331: Like I commented elsewhere - instead of 2 calls here, 1 call `get_metrics()` and pass it onward. That "metric" can be a dict you wanna send back.

README.md
```diff
@@ -81,6 +81,14 @@ Profiling using eBPF incurs lower overhead & provides kernel stacks. This (curre
...
+By default, the gProfiler agent sends system metrics (CPU and RAM usage) and metadata to the Performance Studio.
+The metadata includes system metadata like the kernel version and CPU count, and cloud metadata like the type of the instance you are running on.
+You can disable metrics and metadata by using the following parameters:
+* Use `--disable-metrics-collection` to disable metrics collection
```
Jongy, line 88: Any reason you split it into 3 types?

Now we have:
* `-u` for upload in general
* logs
* metrics
* metadata
* arguments

kinda confusing.

Also, some are already included by others (e.g, args are logged so "logs" imply "arguments").

Let's narrow it? I think that simply arguments can be merged into "metadata".
"metrics" can remain separate since it really doesn't contain (and shouldn't contain) anything identifying - just raw numbers.

gprofiler/merge.py
```diff
@@ -218,13 +219,14 @@ def _make_profile_metadata(docker_client: Optional[DockerClient]) -> str:
...
-            'container_names_enabled': enabled,
-        }
-    )
+    profile_metadata = {
+        'containers': container_names,
+        'container_names_enabled': enabled,
+    }
+    profile_metadata["metadata"] = metadata
+    formatted_profile_metadata = "# " + json.dumps(profile_metadata)
+    formatted_profile_metadata = formatted_profile_metadata.replace('\n', ' ')
```
Jongy, line 228: Why this?

If removing, you can step back to:
```
    return "# " + json.dumps(
        {
            'containers': container_names,
            'hostname': get_hostname(),
            'container_names_enabled': enabled,
            'metadata': metadata,
        }
    )
```

gprofiler/metadata/cloud_metadata.py
```diff
@@ -0,0 +1,199 @@
...
+        self.resource_group_name = resource_group_name
+        self.resource_id = resource_id
+        self.instance_id = instance_id
+        self.name = name
+        self.image_info = image_info
+
+
+def get_aws_metadata() -> Optional[AwsInstanceMetadata]:
+    # Documentation: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
+    metadata_response = send_request("http://169.254.169.254/latest/dynamic/instance-identity/document")
```
Jongy, line 92: Shorter names, please make this function easier for the eyes. You can avoid extracting entries from the dict and just include them in the call: `AwsInstanceMetadata("aws", instance_metadata["region"], ...)`. Also, `instance_metadata` could be shorter. ...

gprofiler/metadata/cloud_metadata.py
```diff
@@ -0,0 +1,199 @@
...
+    availability_zone = instance_metadata["zone"]
+    instance_type = instance_metadata["machineType"]
+    preempted = instance_metadata["preempted"] == "TRUE"
+    preemptible = instance_metadata["scheduling"]["preemptible"] == "TRUE"
+    instance_id = str(instance_metadata["id"])
+    image_id = instance_metadata["image"]
+    name = instance_metadata["name"]
+    return GcpInstanceMetadata(
+        provider="gcp",
+        zone=availability_zone,
```
Jongy, line 125: Consistency? Here you use kwargs, in the others you do not

gprofiler/metadata/cloud_metadata.py
Jongy: Please do so in the other providers as well.

gprofiler/metadata/metadata_collector.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import Dict, Optional
+
+from gprofiler import __version__
+from gprofiler.metadata.cloud_metadata import get_static_cloud_instance_metadata
+from gprofiler.metadata.metadata_type import Metadata
+from gprofiler.metadata.system_metadata import get_static_system_info
+
+
+def get_static_metadata(spawn_time: Optional[float], run_args: Dict = None) -> Metadata:
+    if spawn_time is None:
```
Jongy, line 12: ? why can it be `None`? Isn't it the spawn time of gProfiler?

gprofiler/metadata/metadata_type.py
```diff
@@ -0,0 +1,3 @@
+from typing import Dict, Union
+
+Metadata = Dict[str, Union[str, int, bool, Dict]]
```
Jongy, line 3: Explaining comment about the fields?

gprofiler/system_metrics.py
Jongy: I prefer 0 in this case (or None. Anything else we can mark as an error value)

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,240 @@
...
+        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        try:
+            s.connect(("8.8.8.8", 53))
+            local_ips.append(s.getsockname()[0])
+        finally:
+            s.close()
+    return local_ips[0] if local_ips else "unknown"
+
+
+def get_mac_address() -> str:
```
Jongy, line 86: Remind me again why not `getmac`? :sweat_smile: 

Anyway, if keeping it - add an explaining comment about why not using `getmac` and/or why implemented this way (and what is "this way", what's the logic this function is employing with the loopback etc)

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,240 @@
...
+import distro  # type: ignore
+import psutil
+
+from gprofiler.log import get_logger_adapter
+
+logger = get_logger_adapter(__name__)
+hostname: Optional[str] = None
+
+IS_64BIT = sys.maxsize > 2 ** 32
+IFNAMSIZ = 16
```
Jongy, line 22: All these consts that are used only in `get_mac_address` can move into it

gprofiler/client.py
alonbentamar: As I said, it shouldn't happen in production, and only while in development. It's useful to have when adding new features.

gprofiler/merge.py
alonbentamar: Some of the metadata fields (like the Python version) contain `\n` so I'm removing them to make sure we won't have issues parsing them later

gprofiler/metadata/cloud_metadata.py
alonbentamar: Shortened the entire file, now using dataclasses as well

gprofiler/metadata/metadata_collector.py
alonbentamar: It could be used either way, if this method is called at spawn time

gprofiler/metadata/system_metadata.py
alonbentamar: It's so the MAC is the always the same one as the one fetched by the gAgent. If we used `getmac` here, a different interface could be potentially chosen. Will leave a comment

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,219 @@
...
+    pid: int
+    mac_address: str
+    private_ip: str
+    spawn_uptime_ms: float
+
+
+def get_static_system_info() -> SystemInfo:
+    hostname, distribution, libc_tuple, mac_address, local_ip = _initialize_system_info()
+    clock_boottime_type = getattr(time, "CLOCK_BOOTTIME", None)
+    assert clock_boottime_type is not None, "The gProfiler only supports Unix with a kernel version of >=2.6.39"
```
Jongy, line 136: We support CentOS 6 (I tested on 2.6.32).

So don't remove that support - instead, use another clock or so. (although specifically in the CentOS case, I'm fairly certain they have backported `CLOCK_BOOTTIME`)

Jongy:
@alonbentamar can backmerge?

README.md
```diff
@@ -81,6 +81,14 @@ Profiling using eBPF incurs lower overhead & provides kernel stacks. This (curre
 **By default, gProfiler sends logs to Granulate Performance Studio** (when using `--upload-results`/`-u` flag)
 This behavior can be disabled by passing `--dont-send-logs` or the setting environment variable `GPROFILER_DONT_SEND_LOGS=1`.
 
+### Metrics and metadata collection
+By default, the gProfiler agent sends system metrics (CPU and RAM usage) and metadata to the Performance Studio.
+The metadata includes system metadata like the kernel version and CPU count, and cloud metadata like the type of the instance you are running on.
+If the `--upload-results`/`-u` is not enabled, then the metrics and metadata collection is disabled.
```
Jongy, line 87: Instead of negative negative, explain it positive positive: metrics & metadata are sent only iff `-u`.

dev-requirements.txt
```diff
@@ -4,3 +4,4 @@ black==21.5b2
 mypy==0.901
 isort==5.8.0
 types-requests==0.1.9
+types-dataclasses==0.1.5;
```
Jongy, line 7: y `;`?

gprofiler/system_metrics.py
Jongy: Added

gprofiler/client.py
Jongy: k. since it confused me twice already, add a comment? Usually when you catch exceptions, it's because you expect them to happen. If you don't expect, comment that

gprofiler/cloud_metadata.py
Jongy: k

gprofiler/cloud_metadata.py
Jongy: k

gprofiler/cloud_metadata.py
Jongy: Still annoying when running in debug mode locally (for devs). You can differentiate between a real error and "endpoint doesn't exist"

gprofiler/client.py
Jongy: As commented in the other one - explain in comment that this is not expected as a real comment and it's only for devs

gprofiler/merge.py
Jongy: What issues? `json.dumps` escapes `\n`s. `json.loads` later unescapes them

gprofiler/metadata/metadata_collector.py
Jongy: There's exactly one call site:
```
        self._spawn_time = time.time()
        if collect_metadata and self._client is not None:
            self._static_metadata = get_static_metadata(spawn_time=self._spawn_time, run_args=run_args)
```

gprofiler/metadata/system_metadata.py
Jongy: Let's talk about it IRL

gprofiler/client.py
```diff
@@ -78,7 +82,14 @@ def _send_request(
             opts["headers"]["Content-type"] = "application/json"
             buffer = BytesIO()
             with gzip.open(buffer, mode="wt", encoding="utf-8") as gzip_file:
-                json.dump(data, gzip_file, ensure_ascii=False)  # type: ignore
+                try:
+                    json.dump(data, gzip_file, ensure_ascii=False)  # type: ignore
+                except TypeError:
+                    bad_json = str(data)
+                    if len(bad_json) > MAX_BAD_JSON_STRING_LENGTH:
```
Jongy, line 89: Don't need the `if`
```
In [1]: "aaaa"[:100]                                                                                                                                                                                       
Out[1]: 'aaaa'
```

Jongy: Anyway if used only for developers you don't need the cutoff

gprofiler/client.py
```diff
@@ -113,7 +124,13 @@ def get_health(self):
...
 
     def submit_profile(
-        self, start_time: datetime.datetime, end_time: datetime.datetime, profile: str, total_samples: int
+        self,
+        start_time: datetime.datetime,
+        end_time: datetime.datetime,
+        profile: str,
+        total_samples: int,
+        spawn_time: float,
+        metrics: 'Metrics',
```
Jongy, line 133:
```suggestion
        metrics: Metrics,
```
and add `from __future__ import annotations`

gprofiler/main.py
```diff
@@ -289,26 +306,40 @@ def _snapshot(self):
         except Exception:
             logger.error("Running perf failed; consider running gProfiler with '--perf-mode none' to avoid using perf")
             raise
-
+        metadata = (
+            get_current_metadata(self._static_metadata)
```
Jongy, line 310: Why sending the static metadata every time?

gprofiler/main.py
```diff
@@ -113,7 +118,13 @@ def __init__(
         self._client = client
         self._state = state
         self._remote_logs_handler = remote_logs_handler
+        self._collect_metrics = collect_metrics
+        self._collect_metadata = collect_metadata
         self._stop_event = Event()
+        self._static_metadata: Optional[Metadata] = None
```
Jongy, line 124: Don't need `Optional` here (since we're setting to `None`)

gprofiler/main.py
```diff
@@ -289,26 +306,40 @@ def _snapshot(self):
...
+            get_current_metadata(self._static_metadata)
+            if self._collect_metadata and self._client
+            else {"hostname": get_hostname()}
+        )
         if self._runtimes["system"]:
             merged_result, total_samples = merge.merge_profiles(
                 system_result,
                 process_profiles,
                 self._docker_client,
+                metadata,
```
Jongy, line 319: Why do we encode it with the profiles? This should be a different endpoint, what reason is there to mix them?

The existing (in master) `_make_profile_metadata` is a hack; it contains the container names because they really are required together with the profile itself, to do the mixing in the server.

All this metadata you added here - is not relevant (as far as I'm concerned?) so let's avoid mixing.

gprofiler/merge.py
```diff
@@ -2,6 +2,7 @@
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
+
```
Jongy, line 5: Remove

gprofiler/metadata/cloud_metadata.py
```diff
@@ -0,0 +1,154 @@
...
+        name=instance["name"],
+        image_info=image_info,
+    )
+
+
+def send_request(url: str, headers: Dict[str, str] = None) -> Optional[Response]:
+    response = requests.get(url, headers=headers or {}, timeout=METADATA_REQUEST_TIMEOUT)
+    if not response.ok:
+        raise BadResponseCode(response.status_code)
+    return response
```
Jongy, line 137: When are you returning `None` exactly? At which situation does `requests` return a timeout

gprofiler/metadata/metadata_collector.py
```diff
@@ -0,0 +1,35 @@
...
+    cloud_metadata = get_static_cloud_instance_metadata()
+
+    metadata_dict: Metadata = {
+        "cloud_provider": cloud_metadata.pop("provider") if cloud_metadata is not None else "unknown",
+        "agent_version": __version__,
+        "spawn_time": formatted_spawn_time,
+    }
+    metadata_dict.update(static_system_metadata.__dict__)
+    if cloud_metadata is not None:
+        metadata_dict["cloud_info_wrapped"] = cloud_metadata
```
Jongy, line 25: Let's talk about this one IRL

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+
+def get_libc_version() -> Tuple[str, str]:
+    # platform.libc_ver fails for musl, sadly (produces empty results).
+    # so we'll run "ldd --version" and extract the version string from it.
+    # not passing "encoding"/"text" - this runs in a different mount namespace, and Python fails to
+    # load the files it needs for those encodings (getting LookupError: unknown encoding: ascii)
+    def decode_libc_version(version: bytes) -> str:
+        return version.decode("utf-8", errors="replace")
+
+    # Avoid circular imports
```
Jongy, line 31: Oof :(

You can move `log_system_info`, it makes mroe sense now, anyway.

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+        return "glibc", decode_libc_version(m.group(1))
+    # musl
+    m = re.search(br"musl libc.*?\nVersion (.*?)\n", ldd_version, re.M)
+    if m is not None:
+        return "musl", decode_libc_version(m.group(1))
+
+    return "unknown", decode_libc_version(ldd_version)
+
+
+def get_run_mode_and_deployment_type() -> Tuple[str, str]:
```
Jongy, line 53: I think it's nicer to keep these 2 logics separate - `get_run_mode` as one, and another:
```
def get_deployment_type(run_mode: str):
   if run_mode == k8s ..
   elif ...
....
```

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+        return "k8s", "k8s"
+    elif os.getenv("GPROFILER_IN_CONTAINER") is not None:  # set by our Dockerfile
+        return "container", "containers"
+    elif os.getenv("STATICX_BUNDLE_DIR") is not None:  # set by staticx
+        return "standalone_executable", "instances"
+    else:
+        return "local_python", "instances"
+
+
+def get_local_ip() -> str:
```
Jongy, line 64: Document it

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+    try:
+        local_ips = [ip for ip in socket.gethostbyname_ex(socket.gethostname())[2] if not ip.startswith("127.")]
+    except Exception:
```
Jongy, line 67: What are you expecting to catch here? Maybe catch a specific type?

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+    pid: int
+    mac_address: str
+    private_ip: str
+    spawn_uptime_ms: float
+
+
+def get_static_system_info() -> SystemInfo:
+    hostname, distribution, libc_tuple, mac_address, local_ip = _initialize_system_info()
+    clock = getattr(time, "CLOCK_BOOTTIME", time.CLOCK_MONOTONIC)
+    spawn_uptime_ms = time.clock_gettime(clock)
```
Jongy, line 136: You generate multiple different "spawn time" values. Let's avoid that, it's gonna be confusing. Generate it at one place and make sure it's used by all consumers.

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+    libc_type, libc_version = libc_tuple
+    os_name, os_release, os_codename = distribution
+    uname = platform.uname()
+    cpu_count = os.cpu_count()
+    cpu_count = cpu_count if cpu_count is not None else 0
```
Jongy, line 141: Can just use `cpu_count = os.cpu_count() or 0`

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+        hostname=hostname,
+        os_name=os_name,
+        os_release=os_release,
+        os_codename=os_codename,
+        libc_type=libc_type,
+        libc_version=libc_version,
+        hardware_type=uname.machine,
+        pid=os.getpid(),
+        mac_address=mac_address,
+        private_ip=local_ip,
```
Jongy, line 161: Consistent naming

Jongy: I think the better name is `local_ip`. "private" or not depends on the server

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+            mac_address = get_mac_address()
+        except Exception:
+            logger.exception("Failed to get MAC address")
+
+        try:
+            local_ip = get_local_ip()
+        except Exception:
+            logger.exception("Failed to get the local IP")
+
+    # Avoid circular imports
```
Jongy, line 213: Let's fix this 1 as well

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,128 @@
...
+
+import psutil
+
+from gprofiler.exceptions import ThreadStopTimeoutError
+
+DEFAULT_POLLING_INTERVAL_SECONDS = 5
+STOP_TIMEOUT_SECONDS = 30
+
+
+class Metrics:
```
Jongy, line 15: Document the fields of this struct

Also - it can be a dataclass, no?

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class SystemMetricsMonitor(SystemMetricsMonitorBase):
+    def __init__(self, polling_rate_seconds: int = DEFAULT_POLLING_INTERVAL_SECONDS):
+        self._polling_rate_seconds = polling_rate_seconds
+        self._cpu_count = psutil.cpu_count() or 0
+        self._mem_percentages: List[float] = []
+        self._last_cpu_poll_time: Optional[float] = None
+        self._last_cpu_times: Optional[Tuple[float, float]] = None
+        self._stop_event = Event()
```
Jongy, line 52: Can't use the same stop event used globally in the profiler?

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def start(self):
+        assert self._thread is None, "SystemMetricsMonitor is already running"
+        self._stop_event.clear()
+        self._thread = Thread(target=self._continuously_polling_memory, args=(self._polling_rate_seconds,))
+        self._thread.start()
+
+    def stop(self):
+        self._stop_event.set()
+        assert self._thread is not None, "SystemMetricsMonitor was not running"
```
Jongy, line 66:
```suggestion
        assert self._thread is not None, "SystemMetricsMonitor is not running"
```

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,128 @@
...
+        self._stop_event = Event()
+        self._thread = None
+        self._lock = RLock()
+
+        self._get_cpu_utilization()  # Call this once to set the necessary data
+
+    def start(self):
+        assert self._thread is None, "SystemMetricsMonitor is already running"
+        self._stop_event.clear()
+        self._thread = Thread(target=self._continuously_polling_memory, args=(self._polling_rate_seconds,))
```
Jongy, line 61:
```suggestion
        self._thread = Thread(target=self._continuously_poll_memory, args=(self._polling_rate_seconds,))
```

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,128 @@
...
+        self._stop_event.set()
+        assert self._thread is not None, "SystemMetricsMonitor was not running"
+        self._thread.join(STOP_TIMEOUT_SECONDS)
+        if self._thread.is_alive():
+            raise ThreadStopTimeoutError("Timed out while waiting for the SystemMetricsMonitor internal thread to stop")
+        self._thread = None
+
+    def _continuously_polling_memory(self, polling_rate_seconds: int):
+        while not self._stop_event.is_set():
+            start_time = time.monotonic()
```
Jongy, line 74: You don't need the `elapsed` calculation here. This scheme is used in other parts of the profiler where the operation can actually take time. Here it can't

Jongy: (it == `psutil.virtual_memory()`)

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,128 @@
...
+            current_ram_percent = psutil.virtual_memory().percent
+            self._mem_percentages.append(current_ram_percent)
+            elapsed = time.monotonic() - start_time
+            self._stop_event.wait(timeout=polling_rate_seconds - elapsed)
+
+    def _get_average_memory_utilization(self) -> Optional[float]:
+        with self._lock:
+            # Make sure there's only one thread that takes out the values
```
Jongy, line 82: This function is called only in one site? It can't be called from multiple threads.

IMO just declare it "not thread safe" and remove the lock

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,128 @@
...
+        assert self._thread is not None, "SystemMetricsMonitor was not running"
+        self._thread.join(STOP_TIMEOUT_SECONDS)
+        if self._thread.is_alive():
+            raise ThreadStopTimeoutError("Timed out while waiting for the SystemMetricsMonitor internal thread to stop")
+        self._thread = None
+
+    def _continuously_polling_memory(self, polling_rate_seconds: int):
+        while not self._stop_event.is_set():
+            start_time = time.monotonic()
+            current_ram_percent = psutil.virtual_memory().percent
```
Jongy, line 75: Hmm, this seems to be "(total - available) / total" which is what we want

gprofiler/system_metrics.py
Jongy: Well I don't see why not to, lol. You can use it to maintain the value since last call and basically avoid this entire function

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,128 @@
...
+        with self._lock:
+            # Make sure there's only one thread that takes out the values
+            current_length = len(self._mem_percentages)
+            if current_length == 0:
+                return None
+            average_memory = statistics.mean(self._mem_percentages[:current_length])
+            self._mem_percentages[:current_length] = []
+            return average_memory
+
+    def _get_cpu_utilization(self) -> Optional[float]:
```
Jongy, line 90: I have some more ideas for CPU-related fields that we can easily add in this PR, talk w/ me IRL

gprofiler/system_metrics.py
Jongy: Meanwhile I'm not CRing this logic here until we decide what to do with it

Jongy (changes requested):
Done going over all

gprofiler/client.py
alonbentamar: Added a comment

gprofiler/system_metrics.py
alonbentamar: It's perfectly fine for multiple threads to use this function (have multiple "consumer" threads). Is the comment confusing?

gprofiler/system_metrics.py
alonbentamar: I don't understand the comment. If it's what we want, is there anything I need change?

gprofiler/system_metrics.py
alonbentamar: It's still technically better than not having the `elapsed`, even if it takes a couple of MS. I see no reason to worsen the code

gprofiler/metadata/system_metadata.py
alonbentamar: I tend to agree, but it would require a backend change (and I'm pretty sure it's also `private_ip` in our DB so pretty hard to change). I'll change to `private_ip` for consistency, and I'll see if it's hard to change.
@leonid-granulate What do you say? Is it hard to rename `private_ip`->`local_ip`?

gprofiler/client.py
alonbentamar: We officially support Python 3.6, which doesn't have this feature

dev-requirements.txt
alonbentamar: Whoops, fixed. Will only install on Python < 3.6

gprofiler/metadata/system_metadata.py
leonid-granulate: It's not hard to rename, the reason we choose that names is for consistency with gAgent metadata and DB schema which is the same for both.

alonbentamar: @Jongy Your thoughts?

Jongy: > "private" or not depends on the server

That's what I think, lol.

Anyway, for consistency, ok, keep it.

gprofiler/system_metrics.py
Jongy: No, just commented it because the last time I read it I wasn't sure if it's correct

gprofiler/system_metrics.py
Jongy: It's just weird to add locking to a function that's called from a single thread. Can explain that it's thinking about the future. Can also move that comment one line up (usually comments refer to the next line)

alonbentamar: Added a comment

gprofiler/system_metrics.py
alonbentamar: Replaced with `psutil.cpu_percent`.

gprofiler/metadata/metadata_collector.py
alonbentamar: @leonid-granulate I'm changing to `cloud_info` instead of `cloud_info_wrapped`

gprofiler/metadata/system_metadata.py
alonbentamar: Discussed, will keep

gprofiler/system_metrics.py
alonbentamar: removed

gprofiler/system_metrics.py
alonbentamar: Discussed, might add in the future

gprofiler/cloud_metadata.py
alonbentamar: Improved the log and ignoring 404s

gprofiler/main.py
alonbentamar: We have both versions in the code :(

gprofiler/main.py
alonbentamar: Talked about this IRL (we can add an ID later to send the metadata once on boot for a new endpoint and get and ID which the agent will send on each profile upload)

gprofiler/client.py
Jongy: Oof

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,106 @@
...
+        self._mem_percentages: List[float] = []
+        self._stop_event = stop_event
+        self._thread = None
+        self._lock = RLock()
+
+        self._get_cpu_utilization()  # Call this once to set the necessary data
+
+    def start(self):
+        assert self._thread is None, "SystemMetricsMonitor is already running"
+        self._stop_event.clear()
```
Jongy, line 59: You shouldn't call `clear` or `set` on this event - you're not controlling it, you're merely using it. And you won't need to - by the time `stop` is called, the event will be already set. You can add `assert self._stop_event.is_set()`?

gprofiler/cloud_metadata.py
Jongy: Cool. Now it looks good.

gprofiler/merge.py
```diff
@@ -218,13 +218,13 @@ def _make_profile_metadata(docker_client: Optional[DockerClient], add_container_
...
-            'hostname': get_hostname(),
-            'container_names_enabled': enabled,
-        }
-    )
+    profile_metadata = {
+        'containers': container_names,
+        'container_names_enabled': enabled,
+    }
+    profile_metadata["metadata"] = metadata
+    formatted_profile_metadata = "# " + json.dumps(profile_metadata)
```
Jongy, line 226: Just `return` this? Why the extra variable

gprofiler/main.py
Jongy: k

gprofiler/metadata/system_metadata.py
Jongy: k

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,106 @@
...
+        self._thread.join(STOP_TIMEOUT_SECONDS)
+        if self._thread.is_alive():
+            raise ThreadStopTimeoutError("Timed out while waiting for the SystemMetricsMonitor internal thread to stop")
+        self._thread = None
+
+    def _continuously_poll_memory(self, polling_rate_seconds: int):
+        while not self._stop_event.is_set():
+            current_ram_percent = psutil.virtual_memory().percent
+            self._mem_percentages.append(current_ram_percent)
+            self._stop_event.wait(timeout=polling_rate_seconds)
```
Jongy, line 75: You need to exit the thread if it returns `True` (or `False`, IDK, whatever it returns if the event was set)

Jongy: Well the `while` condition is enough I guess

gprofiler/system_metrics.py
Jongy: k

gprofiler/metadata/system_metadata.py
Jongy: @Jongy to split it up

gprofiler/main.py
Jongy: k. will need to maintain compatibility later.

gprofiler/main.py
Jongy: Explained in https://github.com/Granulate/gprofiler/pull/127#discussion_r676918552

Jongy: We'll improve later...

gprofiler/metadata/system_metadata.py
Jongy: I decided to skip splitting, it's easier to reuse the same struct objects etc, instead of creating new raw array objects for each call.

Jongy: Added some comments instead

Jongy (changes requested):
.

Jongy:
@alonbentamar fix linting?

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,108 @@
...
+    def start(self):
+        assert self._thread is None, "SystemMetricsMonitor is already running"
+        assert (
+            not self._stop_event.is_set()
+        ), "Stop condition is already set (perhaps the gProfiler was already stopped?)"
+        self._thread = Thread(target=self._continuously_poll_memory, args=(self._polling_rate_seconds,))
+        self._thread.start()
+
+    def stop(self):
+        self._stop_event.set()
```
Jongy, line 66: Remove this 1

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,108 @@
...
+        self._thread = None
+        self._lock = RLock()
+
+        self._get_cpu_utilization()  # Call this once to set the necessary data
+
+    def start(self):
+        assert self._thread is None, "SystemMetricsMonitor is already running"
+        assert (
+            not self._stop_event.is_set()
+        ), "Stop condition is already set (perhaps the gProfiler was already stopped?)"
```
Jongy, line 61: Forgot to comment this elsewhere - don't use "the gProfiler". gProfiler is a name. It doesn't need "the"

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,236 @@
...
+    else:
+        return "local_python"
+
+
+def get_deployment_type(run_mode: str) -> str:
+    return RUN_MODE_TO_DEPLOYMENT_TYPE.get(run_mode, "unknown")
+
+
+def get_private_ip() -> str:
+    # Fetches the local IP. Attempts to get it locally. If it fails, it will attempt to fetch it by connecting to
```
Jongy, line 73: Either case it's locally. The "connect" to 8.8.8.8 is UDP - this just tells Linux to figure out which local endpoint is going to be used to communicate with `8.8.8.8:53`; no packet is sent.

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,236 @@
...
+    hardware_type: str
+    pid: int
+    mac_address: str
+    private_ip: str
+    spawn_uptime_ms: float
+
+
+def get_static_system_info() -> SystemInfo:
+    hostname, distribution, libc_tuple, mac_address, private_ip = _initialize_system_info()
+    clock = getattr(time, "CLOCK_BOOTTIME", time.CLOCK_MONOTONIC)
```
Jongy, line 156: Not sure that's enough. This tests if `CLOCK_BOOTTIME` exists on `time`; from CPython:
```
#ifdef CLOCK_BOOTTIME
        if (PyModule_AddIntMacro(module, CLOCK_BOOTTIME) < 0) {
            return -1;
        }
#endif
```
we're using Python built with one libc, and run it on older kernels.

Basically, this should run on CentOS 6 which is the oldest we have. This may fail and then we want `MONOTONIC`. Let's try running on that CentOS tomorrow (it's a good idea to test this PR on that old system anyway, since it makes many OS-related changes)

gprofiler/metadata/system_metadata.py
```diff
@@ -0,0 +1,236 @@
...
+        except Exception:
+            logger.exception("Failed to get libc version")
+
+        try:
+            hostname = socket.gethostname()
+        except Exception:
+            logger.exception("Failed to get hostname")
+
+        try:
+
```
Jongy, line 224:
```suggestion
```

gprofiler/system_metrics.py
```diff
@@ -0,0 +1,108 @@
...
+from dataclasses import dataclass
+from threading import Event, RLock, Thread
+from typing import List, Optional
+
+import psutil
+
+from gprofiler.exceptions import ThreadStopTimeoutError
+
+DEFAULT_POLLING_INTERVAL_SECONDS = 5
+STOP_TIMEOUT_SECONDS = 30
```
Jongy, line 12: Too high. This blocks after the user has requested us to stop. Not that I think this will ever trigger, but anyway, bring it lower (2 seconds are more than enough)

Jongy (changes requested):
Final few comments. Let's also run this tomorrow together with prints to review the collected data (on the Centos 6 and on another standard machine)

gprofiler/metadata/system_metadata.py
Jongy:
```
In [4]: time.clock_gettime(time.CLOCK_BOOTTIME + 5)                                                                                                                                                        
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-4-d8bd7da901c2> in <module>
----> 1 time.clock_gettime(time.CLOCK_BOOTTIME + 5)

OSError: [Errno 22] Invalid argument
```

let's catch that error and try `MONOTONIC`?

Jongy:
@alonbentamar and I have tested the metrics on a CentOS 6 & CentOS 7 machines; both look good.

@alonbentamar , please do the last comment https://github.com/Granulate/gprofiler/pull/127#discussion_r681316723 then disable these new features so we can merge now w/o waiting for backend support. (to avoid conflicts, ...)

gprofiler/metadata/system_metadata.py
```diff
@@ -70,21 +70,16 @@ def get_deployment_type(run_mode: str) -> str:
 
 
 def get_private_ip() -> str:
```
Jongy, line 72: `get_local_ip`

Jongy (approved):

PR #128, opened 2021-06-29
python, ruby: Remove collapsed files after reading

Jongy:
## Description
We do not delete the output files of py-spy and rbspy. This PR fixes that.

The issue was introduced in 1efe4a7f87d322beaa1708cb56a5913c18e316e9 (version 1.0.0) - we ceased from deleting the core temporary folder each iteration, instead it survives for the lifetime of the `GProfiler` object.
It was worsened by 0ab54a089f9c31d267b29e4a1b33060c68ff7e58 which added random a prefix in each output file - so now files of the same target PID don't override each other; so storage keeps on growing!

## Motivation and Context
Bugfix.

## How Has This Been Tested?
1. Verified manually that files are now gone.
2. Added logic to the tests to ensure it from now on.

## Checklist:

### DIFF
--- a/gprofiler/merge.py
+++ b/gprofiler/merge.py
@@ -7,6 +7,7 @@ import math
 import random
 import re
 from collections import Counter, defaultdict
+from pathlib import Path
 from typing import Iterable, Optional, Tuple
 
 from gprofiler.docker_client import DockerClient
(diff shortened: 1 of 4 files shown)


Jongy:
Tests fail because the intermediate storage dir (e.g: `tmpygfl6_1a`) also exists.

Also, I realized that the test isn't good: after going down, it makes sense if gProfiler removes its files. What's interesting to check if *before* going down & after a bunch of iterations. Will add that test instead.

sobolron (approved):

sobolron:
Code is good.
Test doesn't work because we have another temp dir.

Jongy:
We will add these tests later, first I want to solve this issue

guybortnikov (approved):

PR #131, opened 2021-06-30
Create ProcessProfilerBase and treat NoSuchProcess as DEBUG

Jongy:
## Description
Started with logging NoSuchProcess exceptions from profilers as DEBUG, not ERROR.

Then created `ProcessProfilerBase` to reduce code dup, it was time...

## Motivation and Context
Reduce the error logs & reduce code dup.

## How Has This Been Tested?
CI

## Checklist:

### DIFF
--- a/gprofiler/java.py
+++ b/gprofiler/java.py
@@ -322,7 +322,7 @@ class JavaProfiler(ProfilerBase):
         if ap_proc.process.is_running():
             ap_proc.stop_async_profiler(True)
         else:
-            logger.info(f"Profiled process {ap_proc.process.pid} exited before stopping async-profiler")
+            logger.debug(f"Profiled process {ap_proc.process.pid} exited before stopping async-profiler")
             # no output in this case :/
             return None
 
@@ -358,6 +358,8 @@ class JavaProfiler(ProfilerBase):
                         results[futures[future]] = result
                 except StopEventSetException:
                     raise
+                except psutil.NoSuchProcess:
+                    logger.debug(f"Process process went down during profiling {futures[future]}", exc_info=True)
                 except Exception:
                     logger.exception(f"Failed to profile Java process {futures[future]}")
 
--- a/gprofiler/java.py
+++ b/gprofiler/java.py
@@ -2,14 +2,13 @@
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
-import concurrent.futures
 import errno
 import functools
 import os
 import shutil
 from pathlib import Path
 from threading import Event
-from typing import Dict, List, Mapping, Optional
+from typing import List, Optional
 
 import psutil
 from psutil import Process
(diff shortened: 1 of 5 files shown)


gprofiler/java.py
```diff
@@ -358,6 +358,8 @@ def snapshot(self) -> ProcessToStackSampleCounters:
                         results[futures[future]] = result
                 except StopEventSetException:
                     raise
+                except psutil.NoSuchProcess:
```
Jongy, line 361: Maybe in this PR it's time to merge all those 3 `snapshot` impls?

gprofiler/python.py
```diff
@@ -110,6 +110,8 @@ def snapshot(self) -> ProcessToStackSampleCounters:
                     results[futures[future]] = future.result()
                 except StopEventSetException:
                     raise
+                except NoSuchProcess:
+                    logger.debug(f"Process process went down during profiling {futures[future]}", exc_info=True)
```
alonbentamar, line 114: Shouldn't it be a warning? What if the profiler crashed? We can't even check the output

alonbentamar (changes requested):

gprofiler/java.py
alonbentamar: Removing code duplication is always welcome

guybortnikov: Yes please share, too much duplication

gprofiler/python.py
Jongy: `NoSuchProcess` is most likely for interaction with the target process. It's much more likely that the target process exited (because processes exit, it happens).

Jongy: We don't use `psutil` on our profilers anyway

gprofiler/java.py
Jongy: yeah... done. we shoulda done it a while ago

gprofiler/python.py
alonbentamar: > most likely

So it is possible that we could miss errors that we would like to log otherwise? Let's improve the reliability as much as possible if we can

Jongy: Note - I'm not missing any logs, just changed specific ones to `DEBUG` so they don't spam our backend with `ERROR`s that are, in 99.99% of the cases, just processes going down. In case there's a real problem with one of the profilers that causes processes to go down (e.g with async-profiler) - we will always have the full logs.

alonbentamar: Yeah I know, but I'm talking about that error that would happen in 0.01% of the time - maybe check for a certain REGEX on the output to make sure the error is okay?

Jongy: I don't know how will it look like when it's a real error. These logs will be able to help when debugging once we *know* there was a real problem.

They can also tell us there's a problem if it happens "every time" but that's more complex to catch. Therefore I think it's best kept at DEBUG

gprofiler/profiler_base.py
```diff
@@ -75,3 +79,46 @@ class NoopProfiler(ProfilerInterface):
...
+                futures[executor.submit(self._profile_process, process)] = process.pid
+
+            results = {}
+            for future in concurrent.futures.as_completed(futures):
+                try:
+                    result = future.result()
+                    if result is not None:
+                        results[futures[future]] = result
+                except StopEventSetException:
+                    raise
```
alonbentamar, line 115: Consider raising this after calling `.result()` on all of the threads to make sure you are not losing exceptions

guybortnikov (approved):

gprofiler/profiler_base.py
Jongy: That's true - we are aware of the issue, will be solved separately (there are other cases as well)

Jongy:
@alonbentamar can you approve please? Or dismiss your review.

alonbentamar (approved):

PR #133, opened 2021-06-30
ci: Reject non-matching version between git tag and code

Jongy:
## Description
Latest release 1.1.5 was done w/o bumping `__version__` (my :dumb:), adding this to avoid such cases in the future.

## Motivation and Context
Avoid making human errors when releasing a version.

## How Has This Been Tested?
Tested this snippet locally on my shell. Fails if the tag doesn't match.

## Checklist:


guybortnikov (approved):
nice, good idea

PR #140, opened 2021-07-14
Feature/auto profilers arguments

alonbentamar:
Automatically register and add the arguments for the profilers

## Description
Refactored the code, as adding each new profiler bloated the `main.py` module.

## Motivation and Context
Code was getting messy and bloated

## Checklist:


Jongy:
I'm not sure this abstraction is useful, if the PR adding it ends up adding lines and not removing :(

gprofiler/profilers/perf.py
```diff
@@ -12,14 +12,36 @@
...
 from gprofiler.merge import ProcessToStackSampleCounters, merge_global_perfs
-from gprofiler.profiler_base import ProfilerBase
+from gprofiler.profilers.profiler_base import ProfilerBase
+from gprofiler.profilers.registry import ProfilerArgument, register_profiler
 from gprofiler.utils import TEMPORARY_STORAGE_PATH, resource_path, run_process
 
 logger = get_logger_adapter(__name__)
 
 PERF_BUILDID_DIR = os.path.join(TEMPORARY_STORAGE_PATH, "perf-buildids")
+PERF_MODE_ARGUMENT_HELP = (
```
Jongy, line 22: What's the use of placing it here? This "const" is meaningful next to the other related arguments (e.g `possible_modes`). Just embed it inside `register_profiler(...)`

gprofiler/main.py
```diff
@@ -557,6 +490,28 @@ def parse_cmd_args():
...
+            default=profiler_config.default_mode,
+            help=profiler_config.profiler_mode_help,
+            choices=profiler_config.possible_modes,
+        )
+        profiler_argument_group.add_argument(
+            f"--no-{profiler_name.lower()}",
+            dest=profiler_name.lower(),
+            action="store_false",
+            default=True,
+            help=f"Deprecated. Use '--{profiler_name.lower()}-mode none' instead",
```
Jongy, line 509: You're actually adding new options by this (we didn't have `--no-perf`, for example). Can just remove the "deprecated" note if you are maintaining it here

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,77 @@
...
+    profiler_name: str,
+    possible_modes: Optional[List] = None,
+    default_mode="auto",
+    profiler_mode_argument_help: Optional[str] = None,
+    profiler_arguments: Optional[List[ProfilerArgument]] = None,
+):
+    if profiler_mode_argument_help is None:
+        profiler_mode_argument_help = (
+            f"Choose the mode for profiling {profiler_name} processes. 'auto'"
+            f" to automatically profile them, or 'none' to disable {profiler_name} profiling"
```
Jongy, line 60: That's not the meaning of "auto". We had "auto" only in Python, and that means "select the profiler automatically"

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,77 @@
...
+        profilers_config[profiler_name] = ProfilerConfig(
+            profiler_mode_argument_help, possible_modes, default_mode, profiler_arguments
+        )
+        return profiler_class
+
+    return profiler_decorator
+
+
+def get_profilers_registry():
+    global profilers_config
```
Jongy, line 76: Needn't this line

gprofiler/profilers/registry.py
alonbentamar: Changing to `enabled` and `disabled`

gprofiler/profilers/perf.py
alonbentamar: Moved it because it's a big string, I'll move it back

gprofiler/profilers/registry.py
Jongy: So what's the options are, now? e.g in Python?

gprofiler/profilers/perf.py
Jongy: I don't think there's use in moving away big strings. This number ain't a "meaningful" const like e.g `PERF_BUILDID_DIR` one line above it. It's just a string and it's easier to read the string right next to where it's used.

README.md
```diff
@@ -49,32 +49,32 @@ The default duration is *60 seconds*, and the default interval matches it. So gP
...
-* `--no-java`: Disable profilers for Java.
+* `--no-java` or `--java-mode disabled`: Disable profilers for Java.
 
 ### Python profiling options
 * `--no-python`: Alias of `--python-mode none`.
 * `--python-mode`: Controls which profiler is used for Python.
     * `auto` - (default) try with PyPerf (eBPF), fall back to py-spy.
     * `pyperf` - Use PyPerf with no py-spy fallback.
     * `pyspy` - Use py-spy.
-    * `none` - Disable profilers for Python.
```
Jongy, line 60: This breaks compatibility, and we have users that already use `none`.

I suggest: remove `none` from here, but maintain compatibility in code

gprofiler/main.py
```diff
@@ -557,6 +490,28 @@ def parse_cmd_args():
...
+            choices=profiler_config.possible_modes,
+        )
+        profiler_argument_group.add_argument(
+            f"--no-{profiler_name.lower()}",
+            dest=profiler_name.lower(),
+            action="store_false",
+            default=True,
+            help=f"Disable the profiling of {profiler_name} processes",
+        )
+        for profiler_argument in profiler_config.profiler_arguments:
```
Jongy, line 511: I think it's nicer to avoid creating the `profiler_arguments` abstraction and simply allow each registered profiler provide its own `add_my_arguments` method, which you can call. It will get the `parser` object (or actually, the `profiler_argument_group`)

gprofiler/main.py
```diff
@@ -645,11 +600,11 @@ def main():
             remote_logs_handler.init_api_client(client)
 
         runtimes = {
-            "system": args.perf_mode != "none",
-            "java": args.java,
-            "python": args.python_mode != "none",
-            "php": args.php,
-            "ruby": args.ruby,
+            "system": args.perf_mode not in ["none", "disabled"],
```
Jongy, line 625: Instead of implementing this logic for each case - add a helper method

gprofiler/main.py
```diff
@@ -645,11 +600,11 @@ def main():
...
+            "java": args.java and args.java_mode == "enabled",
```
Jongy, line 626: Why do you need to check 2 fields? What sense does it make to have 2 fields o.O can make `--no-java` write `disabled` to `java_mode`.

gprofiler/main.py
```diff
@@ -557,6 +490,28 @@ def parse_cmd_args():
     return args
 
 
+def _add_profilers_arguments(parser):
+    profilers_registry = get_profilers_registry()
+    for profiler_name, profiler_config in profilers_registry.items():
```
Jongy, line 509: Please, shorter names. All names here are very long. `profiler_name` can be `name`. `profilers_registry` can be `registry`. `profiler_argument_group` can be just `arg_group`. I think it's obvious enough from context.

gprofiler/profilers/python.py
```diff
@@ -29,6 +30,12 @@
 
 logger = get_logger_adapter(__name__)
 
+PYTHON_MODE_ARGUMENT_HELP = (
```
Jongy, line 33: Push into the `register_profiler` call

gprofiler/profilers/python.py
```diff
@@ -255,6 +262,12 @@ def stop(self):
             logger.info("Finished profiling Python processes with PyPerf")
 
 
+@register_profiler(
+    "Python",
+    possible_modes=["auto", "pyperf", "pyspy", "disabed", "none"],
```
Jongy, line 267: Comment on `none` that's it legacy.

Or actually - just add the legacy automatically in the shared code?

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,76 @@
+from typing import Any, Dict, List, Optional, Type
+
+
+class ProfilerArgument:
```
Jongy, line 4: I see no need in this abstraction

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,76 @@
...
+                argument_dict[argument_name] = argument_value
+        return argument_dict
+
+
+class ProfilerConfig:
+    def __init__(
+        self,
+        profiler_mode_help: str,
+        possible_modes: List[str] = None,
+        default_mode: str = "auto",
```
Jongy, line 38: Any reason for the default (of this & the others?)

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,76 @@
...
+    if profiler_mode_argument_help is None:
+        profiler_mode_argument_help = (
+            f"Choose the mode for profiling {profiler_name} processes. 'enabled'"
+            f" to automatically profile them, or 'disabled' to disable {profiler_name} profiling"
+        )
+    if possible_modes is None:
+        possible_modes = ["enabled", "disabled"]
+
+    def profiler_decorator(profiler_class):
+        global profilers_config
```
Jongy, line 66: Don't need `global` here either :) You need it only if assigning into the name (not if assigning via subscript, ...)

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,76 @@
...
+    profiler_name: str,
+    possible_modes: Optional[List] = None,
+    default_mode="enabled",
+    profiler_mode_argument_help: Optional[str] = None,
+    profiler_arguments: Optional[List[ProfilerArgument]] = None,
+):
+    if profiler_mode_argument_help is None:
+        profiler_mode_argument_help = (
+            f"Choose the mode for profiling {profiler_name} processes. 'enabled'"
+            f" to automatically profile them, or 'disabled' to disable {profiler_name} profiling"
```
Jongy, line 60: "to profile them with the default method" is a better description than "automatic".

Also, where do you join in the descriptions specific to the profiler? like `perf`'s smart/fp/...

Jongy: So this one is used only as default?

gprofiler/profilers/perf.py
```diff
@@ -12,14 +12,34 @@
...
+@register_profiler(
+    "Perf",
+    possible_modes=["fp", "dwarf", "smart", "disabled", "none"],
+    default_mode="fp",
+    profiler_mode_argument_help="Run perf with either FP (Frame Pointers), DWARF, or run both and intelligently merge"
+    " them by choosing the best result per process. If 'none' is chosen, do not invoke"
+    " 'perf' at all. The output, in that case, is the concatenation of the results from all"
+    " of the runtime profilers.",
+    profiler_arguments=[
+        ProfilerArgument(
```
Jongy, line 33: I admit this does look nice, but find it weird that you need to have this `ProfilerArgument` class which is merely arguments to an `add_argument` call. Any other way to do it more nicely (either providing a function here to do the `add_argument` calls, or ... need to think)

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+    if possible_modes is None:
+        possible_modes = ["enabled", "disabled"]
+
+    def profiler_decorator(profiler_class):
+        global profilers_config
+        profilers_config[profiler_name] = ProfilerConfig(
```
Jongy, line 67: `assert profiler_name not in profilers_config`

gprofiler/profilers/registry.py
alonbentamar: It's not an abstraction, but rather an improvement than the alternative of a dict

gprofiler/profilers/registry.py
alonbentamar: Fixed and changed to "enabled"

gprofiler/profilers/registry.py
alonbentamar: Yeah, it's the defaut

gprofiler/profilers/perf.py
alonbentamar: Yeah, I agree. I wanted to shorten the huge register line, but in this case it only hurt readablity

gprofiler/profilers/registry.py
alonbentamar: `auto`, `disabled`, `pyperf`, `pyspy`, `none` (deprecated)

README.md
alonbentamar: Yep, supporting `none` behind the scenes

gprofiler/main.py
alonbentamar: Meh, I don't like that at all. I'm trying to make creating a new profiler / adding arguments the easiest process as possible, your solution is too low-level imho

gprofiler/profilers/perf.py
sobolron: I actually like this. It's basically a dataclass that wraps params for `add_argument`. Sounds like something you'd want when wrapping the use of `add_argument` with a list of objects to call with.

sobolron:
> I'm not sure this abstraction is useful, if the PR adding it ends up adding lines and not removing :(

Usually I agree with this. And less code is better. But most added lines are caused by the `registry` which I think it's a file you won't need to normally read or change, and that the benefit of having the profilers arguments in profilers file is really nice.

gprofiler/main.py
```diff
@@ -365,6 +366,7 @@ def parse_cmd_args():
         add_config_file_help=True,
         add_env_var_help=False,
         default_config_files=["/etc/gprofiler/config.ini"],
+        conflict_handler='resolve',
```
Jongy, line 369: Why? I think the default of `'error'` makes more sense

gprofiler/profilers/python.py
```diff
@@ -255,6 +256,14 @@ def stop(self):
             logger.info("Finished profiling Python processes with PyPerf")
 
 
+@register_profiler(
+    "Python",
+    possible_modes=["auto", "pyperf", "pyspy", "disabed"],
+    default_mode="auto",
+    profiler_mode_argument_help="Select the Python profiling mode: auto (try PyPerf, resort to py-spy if it fails), "
+    "pyspy (always use py-spy), pyperf (always use PyPerf, and avoid py-spy even if it fails)"
+    " or none (no runtime profilers for Python).",
```
Jongy, line 265:
```suggestion
    " or disabled (no runtime profilers for Python).",
```

gprofiler/profilers/registry.py
Jongy: It is an abstraction/wrapper over the arguments of `add_argument`. I agree that it looks better than a plain dict. But we don't need a class for it. A `namedtuple` will suffice, no?

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,79 @@
...
+    profiler_arguments: Optional[List[ProfilerArgument]] = None,
+):
+    if profiler_mode_argument_help is None:
+        profiler_mode_argument_help = (
+            f"Choose the mode for profiling {profiler_name} processes. 'enabled'"
+            f" to profile them with the default method, or 'disabled' to disable {profiler_name} profiling"
+        )
+    if possible_modes is None:
+        possible_modes = ["enabled", "disabled"]
+    elif "none" not in possible_modes:
```
Jongy, line 64: Remove it from all existing ones, and remove this check

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,79 @@
...
+    profiler_mode_argument_help: Optional[str] = None,
+    profiler_arguments: Optional[List[ProfilerArgument]] = None,
+):
+    if profiler_mode_argument_help is None:
+        profiler_mode_argument_help = (
+            f"Choose the mode for profiling {profiler_name} processes. 'enabled'"
+            f" to profile them with the default method, or 'disabled' to disable {profiler_name} profiling"
+        )
+    if possible_modes is None:
+        possible_modes = ["enabled", "disabled"]
```
Jongy, line 63: I don't like it that not all have `enabled`. It's confusing. Perhaps remove `enabled` altogether and provide the name of the profiler? So:
* `perf` - `fp`, `dwarf`, `smart`, `disabled`/`none`
* `python` - `auto`, `pyspy`, `pyperf`, `disabled`/`none`
* `java` - `ap`, `disabled`/`none`
....

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,79 @@
...
+def register_profiler(
+    profiler_name: str,
+    possible_modes: Optional[List] = None,
+    default_mode="enabled",
+    profiler_mode_argument_help: Optional[str] = None,
+    profiler_arguments: Optional[List[ProfilerArgument]] = None,
+):
+    if profiler_mode_argument_help is None:
+        profiler_mode_argument_help = (
+            f"Choose the mode for profiling {profiler_name} processes. 'enabled'"
```
Jongy, line 59: You may add it but `possible_modes` isn't `None` so you won't add `enabled`. Also, the other way around can happen. So like in https://github.com/Granulate/gprofiler/pull/140/files#r670691470 I suggest we remove it

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,79 @@
...
+            f" to profile them with the default method, or 'disabled' to disable {profiler_name} profiling"
```
Jongy, line 60:
```suggestion
            f" to profile them with the default method, or 'disabled' to disable {profiler_name}-specific profiling"
```
Emphasize - we will still profile these processes (with perf, ...) but not with a runtime-specific profiler.

gprofiler/profilers/perf.py
Jongy: `namedtuple` will suffice IMO

Jongy: Agree that this way looks better than passing a function object here (which gets called and adds the new arguments)

gprofiler/main.py
```diff
@@ -571,6 +505,30 @@ def parse_cmd_args():
...
+            help=config.profiler_mode_help,
+            choices=config.possible_modes,
+        )
+        arg_group.add_argument(
+            f"--no-{name.lower()}",
+            action="store_const",
+            const="disabled",
+            dest=mode_var,
+            default=True,
+            help=f"Disable the profiling of {name} processes",
```
Jongy, line 526:
```suggestion
            help=f"Disable the runtime-profiling of {name} processes",
```

Jongy, line 526: Do note that `perf`/system needs a different message here

gprofiler/main.py
Jongy: Agree. Currently way looks better, just make the `ProfilerArgument` class simpler & this function was already improved for readability with shorter names

alonbentamar: Simplified the `get_dict` method, really not sure how to shorten the names as they are the same as the kwargs for the `add_argument` method (and they are already very short)

gprofiler/profilers/registry.py
alonbentamar: No, because I still need to get the dict without `None` values, and pretty much only a class can do that

gprofiler/profilers/perf.py
alonbentamar: Wrote in another comment why `namedtuple` is not enough

gprofiler/main.py
alonbentamar: It's so the `--no-<PROFILER>` arguments will work with the same dest arg as `--<PROFILER>-mode`

gprofiler/profilers/registry.py
Jongy: K. Can explain in comment then

gprofiler/profilers/perf.py
Jongy: k

gprofiler/main.py
Jongy: I didn't get it?

gprofiler/main.py
Jongy: By shorted names I referred mostly to `profiler_argument`, `profiler_config` etc

gprofiler/profilers/registry.py
```diff
@@ -53,22 +51,26 @@ def register_profiler(
...
 ):
+    # NOTE: Currently, the default_mode value is used as the default "enabled" mode in the default help string.
     if profiler_mode_argument_help is None:
         profiler_mode_argument_help = (
-            f"Choose the mode for profiling {profiler_name} processes. 'enabled'"
-            f" to profile them with the default method, or 'disabled' to disable {profiler_name} profiling"
+            f"Choose the mode for profiling {profiler_name} processes. '{default_mode}'"
+            f" to profile them with the default method, or 'disabled' to disable {profiler_name}-specific profiling"
         )
     if possible_modes is None:
```
Jongy, line 62: Any reason to support this case? No profiler should use it

gprofiler/profilers/perf.py
Jongy: resolving

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,81 @@
...
+        self.disablement_help = disablement_help
+
+
+profilers_config: Dict[str, ProfilerConfig] = {}
+
+
+def register_profiler(
+    profiler_name: str,
+    possible_modes: Optional[List] = None,
+    default_mode="enabled",
```
Jongy, line 51:
```suggestion
    default_mode: str,
```

no need for default + add typing

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,81 @@
...
+        self.profiler_args: List[ProfilerArgument] = arguments if arguments is not None else []
+        self.disablement_help = disablement_help
+
+
+profilers_config: Dict[str, ProfilerConfig] = {}
+
+
+def register_profiler(
+    profiler_name: str,
+    possible_modes: Optional[List] = None,
```
Jongy, line 50:
```suggestion
    possible_modes: List[str],
```

must have values

gprofiler/profilers/registry.py
```diff
@@ -0,0 +1,81 @@
...
+    default_mode="enabled",
+    profiler_mode_argument_help: Optional[str] = None,
+    profiler_arguments: Optional[List[ProfilerArgument]] = None,
+    disablement_help: Optional[str] = None,
+):
+    # NOTE: Currently, the default_mode value is used as the default "enabled" mode in the default help string.
```
Jongy, line 56: No need

Jongy:
@alonbentamar , I left a few comments. Looks good nonetheless. Please clean it up and we can merge tomorrow.

Also, if possible please take a look at https://github.com/Granulate/gprofiler/pull/126 and see how it's gonna work after your branch (adding NodeJS-specific options that actually go onto the `SystemProfiler` class)

gprofiler/main.py
alonbentamar: There's a conflict, as I add two arguments (the `--no-X` and `--X-mode`) with the same dest arg (`X_mode`). If I didn't add the conflict handler here it wouldn't work

gprofiler/profilers/registry.py
alonbentamar: Are you sure we don't ever want to enable profilers having just `enabled` and `disabled` options?

alonbentamar:
> @alonbentamar , I left a few comments. Looks good nonetheless. Please clean it up and we can merge tomorrow.
> 
> Also, if possible please take a look at #126 and see how it's gonna work after your branch (adding NodeJS-specific options that actually go onto the `SystemProfiler` class)

I think the node.js parameter would be a custom perf parameter instead of its own thing

gprofiler/profilers/registry.py
Jongy: Nah, let's be explicit and consistent and always call optimizers by their name

gprofiler/main.py
Jongy: But we already have it this way - for example, from `nodejs` branch:
```
    python_options = parser.add_argument_group("Python")
    python_options.add_argument(
        "--python-mode",
        dest="python_mode",
        default="auto",
        choices=["auto", "pyspy", "pyperf", "none"],
        help="Select the Python profiling mode: auto (try PyPerf, resort to py-spy if it fails), pyspy (always use"
        " py-spy), pyperf (always use PyPerf, and avoid py-spy even if it fails) or none (no runtime-specific profilers"
        " for Python).",
    )
    python_options.add_argument(
        "--no-python",
        dest="python_mode",
        action="store_const",
        const="none",
        help="Do not invoke runtime-specific profilers for Python processes (legacy option for '--python-mode none')",
    )
```

both have `dest="python_mode"`.

I think the conflict handler refers to conflicts in argument *names*.

Jongy: See https://docs.python.org/3/library/optparse.html#conflicts-between-options

Jongy (changes requested):
Small comments left + fix tests

Jongy (approved):
LGTM; just fix tests

alonbentamar:
@Jongy review #142 next and we'll merge both?

Jongy (approved):

PR #141, opened 2021-07-14
Log CPU usage

Jongy:
## Description
Add `--log-cpu-usage` which prints the CPU usage (as measured by cgroups) of each profiling session, and also when gProfiler exits.

## Motivation and Context
To allow us to measure optimizations & check for regressions.

## How Has This Been Tested?
Ran it locally and verified the numbers make sense. Also ran `yes > /dev/null` alongside in the container, and this pumps the reported value.

## Checklist:

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -32,6 +32,7 @@ from gprofiler.ruby import RbSpyProfiler
 from gprofiler.state import State, init_state
 from gprofiler.utils import (
     TEMPORARY_STORAGE_PATH,
+    CpuUsageLogger,
     TemporaryDirectoryWithMode,
     atomically_symlink,
     get_hostname,
@@ -97,6 +98,7 @@ class GProfiler:
         runtimes: Dict[str, bool],
         client: APIClient,
         state: State,
+        cpu_usage_logger: CpuUsageLogger,
         include_container_names=True,
         remote_logs_handler: Optional[RemoteLogsHandler] = None,
         php_process_filter: str = PHPSpyProfiler.DEFAULT_PROCESS_FILTER,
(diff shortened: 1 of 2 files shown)


gprofiler/main.py
```diff
@@ -575,6 +589,12 @@ def verify_preconditions(args):
         print("Could not acquire gProfiler's lock. Is it already running?", file=sys.stderr)
         sys.exit(1)
 
+    if args.log_cpu_usage and get_run_mode() not in ("k8s", "container"):
+        # TODO: we *can* move into another cpuacct cgroup, to let this work also when run as a standalone
+        # executable.
+        print("--log-cpu-usage is available only when run as a container!")
```
guybortnikov, line 595: Hmm, maybe exit and fail the execution of the gProfiler?

guybortnikov (changes requested):
small comment

gprofiler/main.py
```diff
@@ -604,6 +624,8 @@ def main():
 
     setup_signals()
     reset_umask()
+    # assume we run in the root cgroup (when containerized, that's our view)
```
guybortnikov, line 627: can't we check it, the get_mode relies on arguments/environment variables, which the user might miss.

gprofiler/main.py
Jongy: Next line is `sys.exit(1)` lol

gprofiler/main.py
Jongy: Both `GPROFILER_IN_K8S` and `GPROFILER_IN_CONTAINER` are set by us, not by the user

guybortnikov: I know, but no better way to actually test it for sure?

Jongy: I agree that it's nicer if it is fully foolproof. What I think we'll add later on is that TODO I'd written before - moving ourselves into a sub-cgroup of `cpuacct` so it also works with executables. It will solve this issue as well

guybortnikov: ok

guybortnikov (approved):

PR #142, opened 2021-07-15
Feature/refactor profilers usage

alonbentamar:
Remove a lot of boilerplate code by using a factory class to initiate the profilers

## Description
Dynamically handle profilers instead of creating and using each one explicitly. 

## Motivation and Context
Code was bloated and maintainability suffered because of it

## How Has This Been Tested?
Ran the gProfiler with different arguments (`--no-python`, `--python-mode disabled`, `--no-perf` and more) and made sure that all of the appropriate profilers were used and with the correct arguments
## Checklist:


gprofiler/main.py
```diff
@@ -122,52 +100,22 @@ def __init__(
...
         if include_container_names:
             self._docker_client: Optional[DockerClient] = DockerClient()
         else:
             self._docker_client = None
         self._cpu_usage_logger = cpu_usage_logger
 
+    @property
+    def all_profilers(self) -> Iterable[ProfilerInterface]:
+        for process_profiler in self.process_profilers:
+            yield process_profiler
```
Jongy, line 116:
```suggestion
        yield from process_profiler
```

gprofiler/main.py
```diff
@@ -291,7 +222,7 @@ def _snapshot(self):
             logger.error("Running perf failed; consider running gProfiler with '--perf-mode none' to avoid using perf")
             raise
 
-        if self._runtimes["system"]:
+        if isinstance(self.system_profiler, SystemProfiler):
```
Jongy, line 225: Uhh this check is kinda weird. You're trying to check if it's the `NoopProfiler`, right? Maybe add a method `is_noop()` or alternatively, check the opposite condition (`isinstance(.., NoopProfiler)`) which makes more sense when reading.

gprofiler/main.py
```diff
@@ -567,7 +498,7 @@ def setup_signals() -> None:
 
 def main():
     args = parse_cmd_args()
-    verify_preconditions(args)
+    # verify_preconditions(args)
```
Jongy, line 501: Revert

gprofiler/merge.py
```diff
@@ -11,8 +11,8 @@
 from typing import Iterable, Optional, Tuple
 
 from gprofiler.docker_client import DockerClient
+from gprofiler.gprofiler_types import ProcessToStackSampleCounters, StackToSampleCount
```
Jongy, line 14: Why this rename?

gprofiler/profilers/__init__.py
```diff
@@ -0,0 +1,7 @@
+# flake8: noqa
+# TODO: Make sure to import any new process filters to load it
```
Jongy, line 2: "process filters"? "profilers"?

gprofiler/profilers/__init__.py
```diff
@@ -0,0 +1,7 @@
+# flake8: noqa
```
Jongy, line 1: ?

gprofiler/profilers/__init__.py
Jongy: Why is it a `TODO`?

gprofiler/profilers/factory.py
```diff
@@ -0,0 +1,38 @@
...
+
+from gprofiler.profilers.perf import SystemProfiler
+from gprofiler.profilers.registry import get_profilers_registry
+
+logger = get_logger_adapter(__name__)
+
+
+def get_profilers(
+    user_args: 'UserArgs', **profiler_init_kwargs: Any
+) -> Tuple[Union['SystemProfiler', 'NoopProfiler'], List['ProcessProfilerBase']]:
```
Jongy, line 18:
```suggestion
) -> Tuple[Union[SystemProfiler, NoopProfiler], List[ProcessProfilerBase]]:
```
might need to add `from __future__ import annotations`?

gprofiler/profilers/factory.py
```diff
@@ -0,0 +1,38 @@
...
+    profilers_registry = get_profilers_registry()
+    process_profilers_instances: List['ProcessProfilerBase'] = []
+    system_profiler: Union['SystemProfiler', 'NoopProfiler'] = NoopProfiler()
+    for profiler_name, profiler_config in profilers_registry.items():
+        profiler_mode = user_args.get(f"{profiler_name.lower()}_mode")
+        if profiler_mode in ("none", "disabled") or not user_args.get(profiler_name.lower(), True):
```
Jongy, line 24: Use your same helper method as used in `main.py` from that comment I gave earlier

gprofiler/profilers/factory.py
```diff
@@ -0,0 +1,38 @@
...
+            continue
+        try:
+            profiler_instance = profiler_config.profiler_class(
+                **profiler_init_kwargs, **user_args, profiler_mode=profiler_mode
+            )
+        except Exception:
+            logger.exception(f"Couldn't create {profiler_name} profiler, continuing without this runtime profiler")
```
Jongy, line 31: This message doesn't fit when it's `perf`

gprofiler/profilers/java.py
```diff
@@ -245,8 +245,8 @@ class JavaProfiler(ProcessProfilerBase):
     JDK_EXCLUSIONS = ["OpenJ9", "Zing"]
     SKIP_VERSION_CHECK_BINARIES = ["jsvc"]
 
-    def __init__(self, frequency: int, duration: int, stop_event: Event, storage_dir: str):
-        super().__init__(frequency, duration, stop_event, storage_dir)
+    def __init__(self, frequency: int, duration: int, stop_event: Event, storage_dir: str, **profiler_kwargs):
```
Jongy, line 248: What are these for?

Jongy (changes requested):
.

Jongy:
@alonbentamar waiting for your update here

gprofiler/merge.py
alonbentamar: `types` messes up a lot of imports, it prevented me from debugging the code

gprofiler/profilers/__init__.py
alonbentamar: > Why is it a TODO?
To remind anyone that adds a new profiler

gprofiler/profilers/__init__.py
alonbentamar: Without it:
```
./gprofiler/profilers/__init__.py:2:1: F401 'gprofiler.profilers.java.JavaProfiler' imported but unused
./gprofiler/profilers/__init__.py:3:1: F401 'gprofiler.profilers.perf.SystemProfiler' imported but unused
./gprofiler/profilers/__init__.py:4:1: F401 'gprofiler.profilers.php.PHPSpyProfiler' imported but unused
./gprofiler/profilers/__init__.py:5:1: F401 'gprofiler.profilers.python.PythonProfiler' imported but unused
./gprofiler/profilers/__init__.py:6:1: F401 'gprofiler.profilers.ruby.RbSpyProfiler' imported but unused
```

gprofiler/profilers/factory.py
alonbentamar: This future doesn't exist in Python 3.6 I believe :/

gprofiler/profilers/factory.py
alonbentamar, line 31: @Jongy Please note that if the perf initialization fails, we will continue in a "non-perf" mode. Want me to change the behavior and crash instead?

gprofiler/profilers/java.py
alonbentamar: So I can send ALL of the possible args to ALL of the profilers, and each profiler defines the kwargs it is actually interested in. A pretty clean way to ensure dependency injection for a factory.

gprofiler/profilers/factory.py
alonbentamar: The one that check if a profiler is Noop?

gprofiler/merge.py
Jongy: .

gprofiler/profilers/__init__.py
Jongy: A "TODO" marks a task that can be done *now* (and just waits for someone to pick it up?) or alternatively as a reminder for when a specific issue is done. Here it's just a note for the future, unrelated to any specific issue/task. So no need to mark as TODO

gprofiler/profilers/__init__.py
Jongy: `__all__ = ....`

gprofiler/profilers/factory.py
Jongy: Blehhh

gprofiler/profilers/factory.py
Jongy: Yes. That was the behvaior (e.g, in 1.1.3 it crashes). We (or I :) kinda broke it in 1.1.7, now it fails with some `TimeoutError` and doesn't print the error from `perf`, which is pretty indicative (`perf_event_open(..., PERF_FLAG_FD_CLOEXEC) failed with unexpected error 1 (Operation not permitted)`).

Please fix your code to crash as it used to. And also fix gProfiler to print the proper error message.

gprofiler/profilers/java.py
Jongy: Uhh. I don't like this implicitness. It is *very* error prone (e.g non-matching strings preventing the real argument from being passed).

The `**` unpacking part is okay (since Python checks you pretty good - denies passing the same arg twice, denies passing an arg that's not accepted - as long as you don't allow extra unchecked `**`)

It reminded me this post: https://rhettinger.wordpress.com/2011/05/26/super-considered-super/ which I remember reading once (and I encourage you to!). It shows how `**kwargs` can be used in inheritance chains, while the last entry in the chain calls `super().__init__(**kwargs)` and in this case `super().__init__ == object.__init__` which expects 0 arguments - so it makes sure all arguments were really unpacked and accepted by someone along the chain. It's legit to use this way, but not without verifying.

We can discuss other ways to achieve the implicit factory you built here; otherwise let's cancel it

gprofiler/profilers/factory.py
Jongy: Seems irrelevant now since I can't find anything matching anymore. Perhaps I meant the Noop one

gprofiler/main.py
```diff
@@ -124,54 +99,24 @@ def __init__(
...
-        )
-        self.ruby_profiler = create_profiler_or_noop(
-            self._runtimes,
-            lambda: RbSpyProfiler(self._frequency, self._duration, self._stop_event, self._temp_storage_dir.name),
-            "ruby",
+        self.system_profiler, self.process_profilers = get_profilers(
+            user_args,
+            storage_dir=self._temp_storage_dir.name,
+            stop_event=self._stop_event,
+            inject_jit=user_args["nodejs_mode"] == "perf",
```
Jongy, line 106: Kinda weird that specifically `inject_jit` deserves to be special here?

Jongy (changes requested):
Done reviewing all

gprofiler/profilers/factory.py
alonbentamar: Will run without `--privileged`

gprofiler/main.py
```diff
@@ -523,13 +455,13 @@ def parse_cmd_args():
...
 
-    if args.dwarf_stack_size > 65528:
+    if args.perf_dwarf_stack_size > 65528:
         parser.error("--perf-dwarf-stack-size maximum size is 65528")
 
     if args.perf_mode in ("dwarf", "smart") and args.frequency > 100:
         parser.error("--profiling-frequency|-f can't be larger than 100 when using --perf-mode 'smart' or 'dwarf'")
 
-    if args.nodejs_mode == "perf" and args.perf_mode not in ("fp", "smart"):
+    if args.perf_nodejs_mode == "perf" and args.perf_mode not in ("fp", "smart"):
```
Jongy, line 464: It is kinda weird, though, because `nodejs_mode` may contain values other than `perf`, in which case it might be totally unrelated to `perf`.

The correct way, maybe, is to define `nodejs_mode` and add an argument `perf_inject` which is a boolean set to `True` if `nodejs_mode == "perf"`.

gprofiler/profilers/factory.py
```diff
@@ -0,0 +1,45 @@
...
+            continue
+        try:
+            profiler_kwargs = profiler_init_kwargs.copy()
+            for key, value in user_args.items():
+                if key.startswith(lower_profiler_name) or key in COMMON_PROFILER_ARGUMENT_NAMES:
+                    profiler_kwargs[key] = value
+            profiler_instance = profiler_config.profiler_class(**profiler_kwargs)
+        except Exception:
+            if profiler_config.profiler_class == SystemProfiler:
+                raise SystemProfilerInitFailure("Could not create the system profiler")
```
Jongy, line 37: Why swallowing the error here and removing all information attached to the exception?

Do you have any reason for `SystemProfilerInitFailure`? If not, I suggest you remove it and just re-raise the exception. If you have, please attach the original exception values in it

gprofiler/profilers/factory.py
```diff
@@ -0,0 +1,45 @@
...
+        if profiler_mode in ("none", "disabled"):
+            continue
+        try:
+            profiler_kwargs = profiler_init_kwargs.copy()
+            for key, value in user_args.items():
+                if key.startswith(lower_profiler_name) or key in COMMON_PROFILER_ARGUMENT_NAMES:
+                    profiler_kwargs[key] = value
+            profiler_instance = profiler_config.profiler_class(**profiler_kwargs)
+        except Exception:
+            if profiler_config.profiler_class == SystemProfiler:
```
Jongy, line 36:
```suggestion
            if profiler_config.profiler_class is SystemProfiler:
```

gprofiler/profilers/factory.py
```diff
@@ -0,0 +1,45 @@
...
+) -> Tuple[Union['SystemProfiler', 'NoopProfiler'], List['ProcessProfilerBase']]:
+    profilers_registry = get_profilers_registry()
+    process_profilers_instances: List['ProcessProfilerBase'] = []
+    system_profiler: Union['SystemProfiler', 'NoopProfiler'] = NoopProfiler()
+    for profiler_name, profiler_config in profilers_registry.items():
+        lower_profiler_name = profiler_name.lower()
+        profiler_mode = user_args.get(f"{lower_profiler_name}_mode")
+        if profiler_mode in ("none", "disabled"):
+            continue
+        try:
```
Jongy, line 29: This `try` should be only on the call to `profiler_class`. Errors in this loop should not be ignored, they are a hard error in your code

gprofiler/profilers/java.py
```diff
@@ -244,7 +244,8 @@ def read_output(self) -> Optional[str]:
 class JavaProfiler(ProcessProfilerBase):
     JDK_EXCLUSIONS = ["OpenJ9", "Zing"]
 
-    def __init__(self, frequency: int, duration: int, stop_event: Event, storage_dir: str):
+    def __init__(self, frequency: int, duration: int, stop_event: Event, storage_dir: str, java_mode: str):
+        self._java_mode = java_mode  # Not currently in use
```
Jongy, line 248: Add `assert` on its value to be what you expect.

And whys saving it on the instance?

gprofiler/profilers/php.py
```diff
@@ -59,12 +59,14 @@ def __init__(
...
         storage_dir: str,
-        php_process_filter: str = DEFAULT_PROCESS_FILTER,
+        php_process_filter: str,
+        php_mode: str,
     ):
         super().__init__(frequency, duration, stop_event, storage_dir)
         self._process: Optional[Popen] = None
         self._output_path = Path(self._storage_dir) / f"phpspy.{random_prefix()}.col"
         self._process_filter = php_process_filter
+        self._php_mode = php_mode  # Not currently in use
```
Jongy, line 69: https://github.com/Granulate/gprofiler/pull/142/files#r681752321 likewise

gprofiler/profilers/ruby.py
```diff
@@ -24,6 +25,10 @@ class RbSpyProfiler(ProcessProfilerBase):
     RESOURCE_PATH = "ruby/rbspy"
     MAX_FREQUENCY = 100
 
+    def __init__(self, frequency: int, duration: int, stop_event: Optional[Event], storage_dir: str, ruby_mode: str):
+        super().__init__(frequency, duration, stop_event, storage_dir)
+        self._ruby_mode = ruby_mode  # Not currently in use
```
Jongy, line 30: https://github.com/Granulate/gprofiler/pull/142/files#r681752321 likewise

Jongy (changes requested):
Reviewed all again

gprofiler/profilers/factory.py
Jongy: Tried?

alonbentamar: It didn't work as it failed on another area, so I manually added a `raise Exception()` to the `SystemProfiler.__init__` method. It stopped the agent from booting.

gprofiler/main.py
alonbentamar: The only possible values are either `perf` or `disabled`. And in any case, I think the name is not horrible because behind the scenes the value is only relevant for the `perf` profiler.

gprofiler/profilers/factory.py
alonbentamar: 1. The exception is _not_ swallowed - the traceback is attached to the new exception and will be logged. 
2. I think it cleans the code, doing `except SystemProfilerInitFailure` to add a new specific log (`System profiler initialization has failed`) is way cleaner than catching all of the exceptions and logging that _something_ failed while attempting to get the profilers. I would not be able to have a precise log otherwise.

gprofiler/profilers/factory.py
Jongy: good

gprofiler/main.py
Jongy: good

gprofiler/profilers/factory.py
Jongy: Right. No `from None` so it will be attached.

Anyway, I ran gProfiler such that `perf` will fail, and I get:
```
2021-08-05 11:00:36,946] ERROR: gprofiler: Running perf failed; consider running gProfiler with '--perf-mode disabled' to avoid using perf
  File "/app/gprofiler/main.py", line 240, in _snapshot
    system_result = system_future.result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 444, in result
    return self.__get_result()
    raise self._exception
  File "/app/gprofiler/profilers/perf.py", line 196, in snapshot
    self._perf_fp.wait_and_script() if self._perf_fp is not None else None,
  File "/app/gprofiler/profilers/perf.py", line 87, in wait_and_script
    perf_data = wait_for_file_by_prefix(f"{self._output_path}.", self._dump_timeout_s, self._stop_event)
  File "/app/gprofiler/utils.py", line 132, in wait_for_file_by_prefix
    wait_event(timeout, stop_event, lambda: len(glob.glob(glob_pattern)) > 0)
  File "/app/gprofiler/utils.py", line 119, in wait_event
    raise TimeoutError()
TimeoutError
[2021-08-05 11:00:36,953] INFO: gprofiler: Stopping ...
[2021-08-05 11:00:36,954] INFO: gprofiler.profilers.perf: Stopped perf (fp mode)
[2021-08-05 11:00:36,954] ERROR: gprofiler: Unexpected error occurred
  File "/app/gprofiler/main.py", line 667, in main
    gprofiler.run_single()
  File "/app/gprofiler/main.py", line 311, in run_single
    self._snapshot()
  File "/app/gprofiler/main.py", line 240, in _snapshot
    system_result = system_future.result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 444, in result
    return self.__get_result()
    raise self._exception
  File "/app/gprofiler/profilers/perf.py", line 196, in snapshot
    self._perf_fp.wait_and_script() if self._perf_fp is not None else None,
  File "/app/gprofiler/profilers/perf.py", line 87, in wait_and_script
    perf_data = wait_for_file_by_prefix(f"{self._output_path}.", self._dump_timeout_s, self._stop_event)
  File "/app/gprofiler/utils.py", line 132, in wait_for_file_by_prefix
    wait_event(timeout, stop_event, lambda: len(glob.glob(glob_pattern)) > 0)
  File "/app/gprofiler/utils.py", line 119, in wait_event
    raise TimeoutError()
TimeoutError
```

which makes sense, since `perf` is only executed on` star` - there's no real reason for `SystemProfiler.__init__` to fail, it doesn't call anything, You can retain the exception tho..

gprofiler/profilers/ruby.py
```diff
@@ -27,7 +27,7 @@ class RbSpyProfiler(ProcessProfilerBase):
 
     def __init__(self, frequency: int, duration: int, stop_event: Optional[Event], storage_dir: str, ruby_mode: str):
         super().__init__(frequency, duration, stop_event, storage_dir)
-        self._ruby_mode = ruby_mode  # Not currently in use
+        assert ruby_mode == "rubyspy", "Ruby profiler should not be initialized, wrong ruby_mode value given"
```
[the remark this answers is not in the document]
Jongy: Ahh done. I went commit by commit

Jongy (approved):

PR #144, opened 2021-07-19
Add an option to use the legacy profile upload API

alonbentamar:
Because the legacy V1 profile upload method is still useful in some scenarios, I've added the option to use a new gProfiler with the legacy endpoint. 

## Description
I've added the new `--legacy-profile-api-version` option, which currently supports the `V1` value. If this option is enabled, then the gProfiler will not get the container names of stacks and each line in the profile content will not begin with the `;` character. In addition, the `client` will use the `V1` API endpoint when uploading profiles.

## How Has This Been Tested?
Ran the gProfiler with and without the new option. I've made sure that the correct API endpoint is used, and the profile content is as expected.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,6 @@ gProfiler can be run in a continuous mode, profiling periodically, using the `--
 Note that when using `--continuous` with `--output-dir`, a new file will be created during *each* sampling interval.
 Aggregations are only available when uploading to the Granulate Performance Studio.
 
-
 ## Running as a Docker container
 Run the following to have gProfiler running continuously, uploading to Granulate Performance Studio:
 ```bash
--- a/README.md
+++ b/README.md
@@ -85,8 +85,6 @@ gProfiler can be run in a continuous mode, profiling periodically, using the `--
 Note that when using `--continuous` with `--output-dir`, a new file will be created during *each* sampling interval.
 Aggregations are only available when uploading to the Granulate Performance Studio.
 
-### Legacy profile upload mode
-In some cases, using the old V1 profile upload endpoint is useful. If that's the case, you may use the `--legacy-profile-api-version v1` flag to enable the legacy mode.
 
 ## Running as a Docker container
 Run the following to have gProfiler running continuously, uploading to Granulate Performance Studio:
--- a/README.md
+++ b/README.md
@@ -85,6 +85,9 @@ gProfiler can be run in a continuous mode, profiling periodically, using the `--
 Note that when using `--continuous` with `--output-dir`, a new file will be created during *each* sampling interval.
 Aggregations are only available when uploading to the Granulate Performance Studio.
 
+### Legacy profile upload mode
+In some cases, using the old V1 profile upload endpoint is useful. If that's the case, you may use the `--legacy-profile-api-version v1` flag to enable the legacy mode.
+
 ## Running as a Docker container
 Run the following to have gProfiler running continuously, uploading to Granulate Performance Studio:
 ```bash
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -100,7 +100,7 @@ class GProfiler:
         state: State,
         cpu_usage_logger: CpuUsageLogger,
         include_container_names=True,
-        legacy_profile_api_version: Optional[str] = None,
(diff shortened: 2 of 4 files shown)


gprofiler/main.py
```diff
@@ -528,6 +534,15 @@ def parse_cmd_args():
         "--continuous", "-c", action="store_true", dest="continuous", help="Run in continuous mode"
     )
 
+    parser.add_argument(
+        "--legacy-profile-api-version",
```
sobolron, line 538: This is one long argument name... Can you make it shorter?

sobolron (changes requested):

README.md
```diff
@@ -85,6 +85,9 @@ gProfiler can be run in a continuous mode, profiling periodically, using the `--
 Note that when using `--continuous` with `--output-dir`, a new file will be created during *each* sampling interval.
 Aggregations are only available when uploading to the Granulate Performance Studio.
 
+### Legacy profile upload mode
```
Jongy, line 88: This is currently a private option for our use so I don't think it deems documentation here

gprofiler/merge.py
```diff
@@ -245,19 +246,20 @@ def concatenate_profiles(
...
-    lines.insert(0, _make_profile_metadata(docker_client))
+    lines.insert(0, _make_profile_metadata(docker_client, add_container_names))
     return "\n".join(lines), total_samples
 
 
 def merge_profiles(
     perf_pid_to_stacks_counter: ProcessToStackSampleCounters,
     process_profiles: ProcessToStackSampleCounters,
     docker_client: Optional[DockerClient],
+    add_container_names: bool,
```
Jongy, line 262: Perhaps the API can be reworked in a way that `merge_profiles` doesn't have to receive `docker_client` & `add_container_names` as arguments?

gprofiler/merge.py
```diff
@@ -209,8 +209,8 @@ def _parse_perf_script(script: Optional[str]) -> ProcessToStackSampleCounters:
     return pid_to_collapsed_stacks_counters
 
 
-def _make_profile_metadata(docker_client: Optional[DockerClient]) -> str:
-    if docker_client is not None:
+def _make_profile_metadata(docker_client: Optional[DockerClient], add_container_names: bool) -> str:
+    if docker_client is not None and add_container_names:
```
Jongy, line 213:
```suggestion
    if add_container_names docker_client is not None:
```

for consistency with the other check u do in this file

gprofiler/main.py
```diff
@@ -160,7 +162,7 @@ def __init__(
             lambda: RbSpyProfiler(self._frequency, self._duration, self._stop_event, self._temp_storage_dir.name),
             "ruby",
         )
-        if include_container_names:
+        if include_container_names and legacy_profile_api_version is None:
```
Jongy, line 165:
```suggestion
        if include_container_names and legacy_profile_api_version != "v1":
```

like the other check in this file?

gprofiler/merge.py
alonbentamar: The problem is that it's two different situations - 
If the `docker_client` is None I still want to add a `;` at the start of each line, but not if the `add_container_names` is false. I don't think it's too bad sending these two different args

Jongy: Aha, because in `v2` we have to send it in any way. Oh well. This code will have to be restructured some when

gprofiler/main.py
Jongy: Solving

README.md
```diff
@@ -85,6 +85,7 @@ gProfiler can be run in a continuous mode, profiling periodically, using the `--
 Note that when using `--continuous` with `--output-dir`, a new file will be created during *each* sampling interval.
 Aggregations are only available when uploading to the Granulate Performance Studio.
 
+
```
Jongy, line 88: Remove

Jongy (approved):

gprofiler/main.py
sobolron: Still long if you ask me. lol. But OK :)

sobolron (approved):

PR #154, opened 2021-08-01
python: pyperf: Automatically fix up numerous output-files situations

Jongy:
## Description
We used to bail out on this case (that `assert`).
If it ever happens, for whatever reason, it will continue happening forever for that invocation of gProfiler, because we don't take care of it and just keep bailing out on every iteration.

This has 2 parts:
* Always remove the output file, even if reading / decoding of it fails.
* If numerous output files are found, remove all and use the last.

I saw it triggered by a `UnicodeDecodeError` inside `read_text`. This PR, of course, does not solve that issue, but only attempts to fix up profiling in case that issue was only a one-timer.

## Motivation and Context
Automatically fix up the described error case, without user intervention/restart.

## How Has This Been Tested?
T.B.A

## Checklist:

### DIFF
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -148,6 +148,7 @@ def wait_for_file_by_prefix(prefix: str, timeout: float, stop_event: Event) -> P
             f"One output file expected, but found {len(output_files)}."
             f" Removing all and using the last one. {output_files}"
         )
+        # timestamp format guarantees alphabetical order == chronological order.
         output_files.sort()
         for f in output_files[:-1]:
             os.unlink(f)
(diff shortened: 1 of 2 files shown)


gprofiler/utils.py
```diff
@@ -137,7 +137,22 @@ def wait_for_file_by_prefix(prefix: str, timeout: float, stop_event: Event) -> P
...
+        # * the profiler generating those files is erroneous
+        # * the profiler received many signals (and it generated files based on signals)
+        # * errors in gProfiler led to previous output fails remain not removed
+        # in any case, we remove all old files, and assume the last one (after sorting by timestamp)
+        # is the one we want.
+        logger.warning(
+            f"One output file expected, but found {len(output_files)}."
+            f" Removing all and using the last one. {output_files}"
+        )
+        output_files.sort()
```
michelhe, line 151: Add comment explaining that the timestamp format in the filenames guarantee alphabetical sort == chronological sort

michelhe (changes requested):
.

gprofiler/utils.py
Jongy: okie

michelhe (approved):

PR #156, opened 2021-08-03
utils: Log a nicer message if "ldd" is not found

Jongy:
Based on #127


gprofiler/metadata/system_metadata.py
```diff
@@ -36,9 +36,12 @@ def get_libc_version() -> Tuple[str, str]:
...
 
-    ldd_version = run_process(
-        ["ldd", "--version"], stdout=subprocess.PIPE, stderr=subprocess.STDOUT, suppress_log=True, check=False
-    ).stdout
+    try:
+        ldd_version = run_process(
+            ["ldd", "--version"], stdout=subprocess.PIPE, stderr=subprocess.STDOUT, suppress_log=True, check=False
+        ).stdout
+    except FileNotFoundError:
+        ldd_version = "ldd not found"
```
alonbentamar, line 44:
```suggestion
        ldd_version = b"ldd not found"
```
It's decoded later

alonbentamar (changes requested):

gprofiler/metadata/system_metadata.py
Jongy: hmm right

Jongy: Stupid mypy

alonbentamar (approved):

PR #157, opened 2021-08-05
Differentiate between "lock taken" and "error while trying to take the lock"

Jongy:
## Description
Errors while trying to take the lock are common (well, now it's the first "privileged" thing we're trying to do).

Running as a container w/o `--privileged` currently yields:
```
Exception in thread Thread-1:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/app/gprofiler/utils.py", line 362, in _switch_and_run
    if not is_same_ns(target_pid, nstype):
  File "/app/gprofiler/utils.py", line 291, in is_same_ns
    return os.stat(f"/proc/self/ns/{nstype}").st_ino == os.stat(f"/proc/{pid}/ns/{nstype}").st_ino
PermissionError: [Errno 13] Permission denied: '/proc/1/ns/net'
Could not acquire gProfiler's lock. Is it already running?
```

The exception is not very clear, and the message "is it already running" doesn't help.

After this PR:
```
Exception in thread Thread-1:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/app/gprofiler/utils.py", line 355, in _switch_and_run
    if not is_same_ns(target_pid, nstype):
  File "/app/gprofiler/utils.py", line 317, in is_same_ns
    return os.stat(f"/proc/self/ns/{nstype}").st_ino == os.stat(f"/proc/{pid}/ns/{nstype}").st_ino
PermissionError: [Errno 13] Permission denied: '/proc/1/ns/net'
Could not acquire gProfiler's lock due to an error. Are you running gProfiler in privileged mode?
```

Better.

## Motivation and Context
Better UX on errors

## How Has This Been Tested?
Tested w/o `--privileged` and verified the exception is as I showed above.
Tested w/ `--privileged` and ran 2 profilers - I get the old `Could not acquire gProfiler's lock. Is it already running?` correctly.

## Checklist:

### DIFF
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -406,7 +406,10 @@ def grab_gprofiler_mutex() -> bool:
     res = run_in_ns(["net"], _take_lock)
     if res is None:
         # exception in run_in_ns
-        print("Could not acquire gProfiler's lock due to an error. Are you running gProfiler in privileged mode?")
+        print(
+            "Could not acquire gProfiler's lock due to an error. Are you running gProfiler in privileged mode?",
+            file=sys.stderr,
+        )
         return False
     elif res[0]:
         assert res[1] is not None
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -42,7 +42,7 @@ logger = get_logger_adapter(__name__)
 
 TEMPORARY_STORAGE_PATH = "/tmp/gprofiler_tmp"
 
-gprofiler_mutex: Optional[socket.socket]
+gprofiler_mutex: Optional[socket.socket] = None
 
 
 @lru_cache(maxsize=None)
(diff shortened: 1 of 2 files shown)


michelhe:
@Jongy  ❤️ rust now ?

Also, tests fail

gprofiler/utils.py
```diff
@@ -385,7 +386,7 @@ def grab_gprofiler_mutex() -> bool:
     """
     GPROFILER_LOCK = "\x00gprofiler_lock"
 
-    def _take_lock() -> Optional[socket.socket]:
+    def _take_lock() -> Tuple[bool, Optional[socket.socket]]:  # like Rust's Result :(
```
michelhe, line 389: You mean, `Result<Option<T>>`?

gprofiler/utils.py
```diff
@@ -395,20 +396,28 @@ def _take_lock() -> Optional[socket.socket]:
...
 
-    s = run_in_ns(["net"], _take_lock)
-
-    global gprofiler_mutex
-    # hold the reference so lock remains taken
-    gprofiler_mutex = s
+    res = run_in_ns(["net"], _take_lock)
+    if res is None:
+        # exception in run_in_ns
+        print("Could not acquire gProfiler's lock due to an error. Are you running gProfiler in privileged mode?")
```
michelhe, line 409: stderror?

gprofiler/utils.py
```diff
@@ -395,20 +396,28 @@ def _take_lock() -> Optional[socket.socket]:
...
+    elif res[0]:
+        assert res[1] is not None
 
-    return s is not None
+        global gprofiler_mutex
+        # hold the reference so lock remains taken
+        gprofiler_mutex = res[1]
+        return True
+    else:
+        print("Could not acquire gProfiler's lock. Is it already running?", file=sys.stderr)
```
michelhe, line 419: here too

michelhe: We can be more elaborate here and write which PID holds the lock.

Jongy: Already `sys.stderr`

gprofiler/utils.py
Jongy: Mhm

gprofiler/utils.py
Jongy: > We can be more elaborate here and write which PID holds the lock.

Hmm, it's kinda complicated to run that. I'll drop in better comment

michelhe (approved):
Wait for tests 😈

PR #163, opened 2021-08-18
bugfix: gProfiler excessive restarts when run in a container and failing to grab the mutex

tomersa:
Change exit code to 0 when failing to grab the mutex. This help gprofiler container avoid
excessive restarts when running using K8s or docker

## Description
Exiting with exit code 0 when failing to grab the mutex. 

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ Aggregations are only available when uploading to the Granulate Performance Stud
 Run the following to have gProfiler running continuously, uploading to Granulate Performance Studio:
 ```bash
 docker pull granulate/gprofiler:latest
-docker run --name gprofiler -d --restart=always \
+docker run --name gprofiler -d --restart=on-failure:10 \
     --pid=host --userns=host --privileged \
     -v /lib/modules:/lib/modules:ro -v /usr/src:/usr/src:ro \
     -v /var/run/docker.sock:/var/run/docker.sock \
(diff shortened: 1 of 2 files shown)


Jongy:
Also change the documentation to use `--restart=on-failure`?

tomersa:
I've also made a maximum of 10 retries by default FYI

Jongy:
> I've also made a maximum of 10 retries by default FYI

Cool. I wasn't aware of this feature. Can you please check from which Docker version is it available? (we refrained from adding `--pull=always` because it's too new and it caused `docker run` to fail in many servers; hence it shouldn't be in the "standard" command)

tomersa:
> > I've also made a maximum of 10 retries by default FYI
> 
> Cool. I wasn't aware of this feature. Can you please check from which Docker version is it available? (we refrained from adding `--pull=always` because it's too new and it caused `docker run` to fail in many servers; hence it shouldn't be in the "standard" command)

Couldn't find it. I guess it's old enough to be used here.

Jongy (approved):

tomersa:
@Jongy I resolved conflict in README please reapprove.

Jongy (approved):

tomersa:
@Jongy You know what are this problems in the CI?
GProfiler CI / test (3.6) (pull_request) Failing after 41s — test (3.6)

It fails on async_profiler_env.sh

Jongy:
> @Jongy You know what are this problems in the CI?
> GProfiler CI / test (3.6) (pull_request) Failing after 41s — test (3.6)
> 
> It fails on async_profiler_env.sh

Nope. Will check in the next PR

PR #164, opened 2021-08-18
Use "granulate-gprofiler" as container name

Jongy:
Better than just "gprofiler".


alonbentamar (approved):
Make sure to update the `performance studio` as well

PR #168, opened 2021-08-22
java: Allow skipping the version check

Jongy:
## Description
We've seen it fail, for whatever reasons (e.g invoking via /proc/pid/exe doesn't work).
Such safety checks should be override-able so we can skip them if we know it's okay, but
they fail us.

## How Has This Been Tested?
Ran with `--java-no-version-check` and saw that this line:
```
[2021-08-22 22:59:40,095] DEBUG: gprofiler.utils: Running command: (/proc/40/exe -version)
```
is gone.

## Checklist:

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -258,8 +258,8 @@ class AsyncProfiledProcess:
         ),
         ProfilerArgument(
             "--java-no-version-check",
-            dest="java_no_version_check",
-            action="store_true",
+            dest="java_version_check",
+            action="store_false",
             help="Skip the JDK version check (that is done before invoking async-profiler)",
         ),
     ],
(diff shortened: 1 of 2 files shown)


gprofiler/profilers/java.py
```diff
@@ -256,6 +256,12 @@ def read_output(self) -> Optional[str]:
             " The added buildid+offset can be resolved & symbolicated in the Performance Studio."
             " This is useful if debug symbols are unavailable for the relevant DSOs (libjvm, libc, ...).",
         ),
+        ProfilerArgument(
+            "--java-no-version-check",
+            dest="java_no_version_check",
+            action="store_true",
+            help="Skip the JDK version check (that is done before invoking async-profiler)",
+        ),
```
alonbentamar, line 264: We usually do `store_false` for these arguments (so the dest would be `java_version_check` and the default would be `True`) so the condition later on would be `if java_version_check` which is a bit more readable imho

alonbentamar (changes requested):

gprofiler/profilers/java.py
Jongy: Fair

Jongy:
Tested again as described in the description

alonbentamar (approved):

PR #169, opened 2021-08-24
Simplify async-profiler's build & COPY statements

Jongy:
Nothing major, just simplify the build script & remove the complex COPY & `tar -xf` lines. Copy only the files we need from the build.

### DIFF
--- a/scripts/perf_env.sh
+++ b/scripts/perf_env.sh
@@ -41,7 +41,7 @@ cd /tmp
 curl -L ftp://sourceware.org/pub/elfutils/0.179/elfutils-0.179.tar.bz2 -o elfutils-0.179.tar.bz2
 tar -xf elfutils-0.179.tar.bz2
 pushd elfutils-0.179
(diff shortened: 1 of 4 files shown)


scripts/perf_env.sh
```diff
@@ -41,7 +41,7 @@ cd /tmp
 curl -L ftp://sourceware.org/pub/elfutils/0.179/elfutils-0.179.tar.bz2 -o elfutils-0.179.tar.bz2
 tar -xf elfutils-0.179.tar.bz2
 pushd elfutils-0.179
-./configure --disable-debuginfod --prefix=/usr && make && make install
+./configure --disable-debuginfod --prefix=/usr && make -j 8 && make install
```
michelhe, line 44: are you sure github runners can handle `-j8` ?

Jongy: Obv they can, otherwise it wouldn't have passed

michelhe (approved):

PR #170, opened 2021-08-24
Initial Aarch64 support

Jongy:
## Description
This PR adds support for building gProfiler's container as Aarch64.

Only part of the features are supported & enabled:
* `perf` is built
* async-profiler is built
* gProfiler is built as a *container*

This is all done via `scripts/build_aarch64_container.sh`.

The following are *not* supported:
* py-spy, PyPerf, rbspy, phpspy and burn - they are not built & their respective profilers are disabled (in the case of burn, local flamegraphs are not generated - only collapsed files are written).
* `pyi.Dockerfile` is not supported - gProfiler can not be built as an executable.

async-profiler was chosen to be supported first; `perf` is obvious.
Machines based on Aarch64 are modern and usually (if not always) have containers support, hence containers are supported first. We can add exe support later.

I have tried hard to use the same Dockerfile & to retain it a simple Dockerfile - i.e, avoided falling back to generating it via some script. So, x86_64 can still be built in a standard fashion - `docker build .` The Aarch64 related consts are embedded in `scripts/build_aarch64_container.sh` which contains the relevant `docker buildx build` command.

What might be improved here is the set of hacks I used to maintain a single Dockerfile. CR fellows, tell me if you think this one is too complex and I'll try to simplify things. Otherwise I suppose it's not too bad and I actually improved a bunch of things along the way ;)

This doesn't run as part of the CI at the moment (`docker buildx build` is very slow over qemu). I'll check if we can get some Aarch64 machine for the CI so it could run this build & tests regularly.

## Motivation and Context
Allows profiling on Linux Aarch64 :)

## How Has This Been Tested?
Tested `perf` and async-profiler with some demo apps on this machine:
```
centos@xxx:~$ uname -a
Linux xxx 4.18.0-305.10.2.el8_4.aarch64 #1 SMP Tue Jul 20 17:47:41 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux
```

* [ ] Run gProfiler's test suite on Aarch64

## Checklist:
- [x] I have read the **CONTRIBUTING** document.
- [x] I have updated the relevant documentation.
- [ ] I have added tests for new logic.

PR #176, opened 2021-09-08
Aarch64 as an executable

Jongy:
## Description
Implement executable build for Aarch64. Continues from https://github.com/Granulate/gprofiler/pull/170.

This also adds proper build scripts + documentation for building locally.

## Motivation and Context
Support running gProfiler on Aarch64 with no containers.

## How Has This Been Tested?
I can run the resulting `build/aarch64/gprofiler` on an Aarch64, and it currently crashes in the staticx bootloader. Very weird stuff, see [here](https://github.com/JonathonReinhart/staticx/issues/181#issuecomment-915616425). We might be able to disable staticx for now, though, because Aarch64 machines are new libc & we currently don't compile PyPerf which IIRC brings most of the dynamic dependencies.  Running gProfiler in `gdb` doesn't exhibit the crash, and it works fine.

I'll fix the staticx issues, meanwhile this can be CRed as most of the changes are unrelated to staticx itself.

## Checklist:
- [x] Fix / disable staticx


README.md
```diff
@@ -191,6 +187,19 @@ It is possible to run gProfiler without using `perf` - this is useful where `per
...
+
+* x86_64: `./scripts/build_x86_64_container.sh -t gprofiler` will create a local image `gprofiler`.
+* Aarch64: `./scripts/build_x86_64_container.sh`, you will need to set up buildx for building cross architecture before, if you're building on x86_64.
+
+## Executable
+
+* x86_64: `./scripts/build_x86_64_executable.sh` will build the executable into `build/x86_64/gprofiler`.
+* Aarch64: `./scripts/build_aarch64_executable.sh` will build the executable into `build/aarch64/gprofiler`. As with the Aarch64 container build - this can be used to cross-compile on x86_64, you just need to set up buildx for that, see notes above.
+
+
```
Jongy, line 202:
```suggestion
```

README.md
```diff
@@ -191,6 +187,18 @@ It is possible to run gProfiler without using `perf` - this is useful where `per
 
 In this mode, gProfiler uses runtime-specific profilers only, and their results are concatenated (instead of scaled into the results collected by `perf`). This means that, although the results from different profilers are viewed on the same graph, they are not necessarily of the same scale: so you can compare the samples count of Java to Java, but not Java to Python.
 
+# Building
+
+## Container
+
+* x86_64: `./scripts/build_x86_64_container.sh -t gprofiler` will create a local image `gprofiler`.
+* Aarch64: `./scripts/build_x86_64_container.sh`, you will need to set up buildx for building cross architecture before, if you're building on x86_64.
```
alonbentamar, line 195: Script name seems off if it's used for all archs

pyi.Dockerfile
```diff
@@ -31,37 +49,33 @@ COPY scripts/perf_build.sh .
...
+FROM centos${GPROFILER_BUILDER} AS build-stage
 
 # bcc part
 # TODO: copied from the main Dockerfile... but modified a lot. we'd want to share it some day.
 
-RUN yum install -y \
-    git \
+RUN yum install -y git
+
+RUN if [ $(uname -m) = "aarch64" ]; then exit 0; fi; yum install -y \
```
alonbentamar, line 78: Doesn't the ARM build need any of these packages?

.github/workflows/ci.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       # the tests need the gprofiler image built (from Dockerfile). I run it separately here, because "docker build" prints the build logs
       # more nicely. the tests will then be able to use the built image.
     - name: Build gProfiler image
-      run: DOCKER_BUILDKIT=1 docker build -t gprofiler .
+      run: ./scripts/build_x86_64_container.sh -t gprofiler
```
alonbentamar, line 64: Perhaps also build the ARM one in this workflow?

alonbentamar (changes requested):

README.md
Jongy: Uhh it's a typo, should be `aarch64`

.github/workflows/ci.yml
Jongy: ARM is emulated and takes lots of time (container build takes >3hrs). Not worth the time

pyi.Dockerfile
Jongy: Nope, they're needed for PyPerf which we currently don't build for ARM

michelhe:
This PR breaks the `build-exectuable` job in CI

.github/workflows/ci.yml
michelhe: I think in the future it would be easier to install a cross-compiler for aarch64

README.md
michelhe: @Jongy  fix typo

.github/workflows/ci.yml
Jongy: It will be easier for the processor to run it, but I doubt it'll be easier for us to write it (it's not just cross compiler, it's also getting `pip` to download the correct packages, and getting Rust, and fixing many different Makefiles with their different packages, ...)

pyi.Dockerfile
alonbentamar: So perhaps add a comment at the top of the file explaining why it's not installed

README.md
```diff
@@ -116,7 +116,9 @@ For profiling with eBPF, kernel headers must be accessible from within the conta
 The command above mounts both of these directories.
 
 ## Running as an executable
-Run the following to have gprofiler running continuously, uploading to Granulate Performance Studio:
+First, check if gProfiler is already running - run `pgrep gprofiler`. You should see 3 PIDs (staticx bootloader, PyInstaller and gProfiler's Python itself). If no processes match, then gProfiler is not running.
+
```
alonbentamar: The doc here doesn't seem to follow the "happy flow", and is overall confusing. Since this is the doc for running the gProfiler, most of the time for the first time, I would replace `You should see 3 PIDs` with `The command should show no output. If you see 3 PIDs, it means the gProfiler is already running, and you could stop it by running...`

alonbentamar (changes requested):

README.md
Jongy: K will improve. I might change the "numerous" PIDs part altogether, because I'm considering to skip staticx for Aarch64 since it's messed up (see https://github.com/JonathonReinhart/staticx/issues/181#issuecomment-915616425, and now I'm getting:
```
 => ERROR [build-stage 43/43] RUN if [ $(uname -m) != "aarch64" ]; then source scl_source enable devtoolset-8 llvm-toolset-7; fi && libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gpro  0.9s
------
 > [build-stage 43/43] RUN if [ $(uname -m) != "aarch64" ]; then source scl_source enable devtoolset-8 llvm-toolset-7; fi && libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gprofiler dist/gprofiler:
#80 0.803 ldd failed: ldd: exited with unknown exit code (139)
------
error: failed to solve: executor failed running [/bin/sh -c if [ $(uname -m) != "aarch64" ]; then source scl_source enable devtoolset-8 llvm-toolset-7; fi && libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gprofiler dist/gprofiler]: exit code: 1
```
(`ldd` gets SIGSEGV...))

Jongy: I decided to skip staticx

Jongy:
I decided to skip the staticx for Aarch64 - we don't need it for any of the executables IIRC, only for PyPerf, and we don't build PyPerf for Aarch64 now, so :shrug:

gprofiler/utils.py
```diff
@@ -524,3 +524,15 @@ def read_perf_event_mlock_kb() -> int:
...
 def write_perf_event_mlock_kb(value: int) -> None:
     Path(PERF_EVENT_MLOCK_KB).write_text(str(value))
+
+
+def is_pyinstaller() -> bool:
+    """
+    Are we running in PyInstaller?
+    """
+    # https://pyinstaller.readthedocs.io/en/stable/runtime-information.html#run-time-information
+    return getattr(sys, 'frozen', False) and hasattr(sys, '_MEIPASS')
```
alonbentamar, line 534: This file is getting way too big. We need to refactor it (not necessarily in this PR though). A lot of these "utils" functions would be better as classes

gprofiler/metadata/system_metadata.py
```diff
@@ -64,7 +64,7 @@ def get_run_mode() -> str:
         return "k8s"
     elif os.getenv("GPROFILER_IN_CONTAINER") is not None:  # set by our Dockerfile
         return "container"
-    elif os.getenv("STATICX_BUNDLE_DIR") is not None:  # set by staticx
+    elif is_pyinstaller():
```
alonbentamar, line 67: PyInstaller is always used right? There's no standalone build without it?

Jongy: Yup. And staticx is only x86_64 now

gprofiler/utils.py
Jongy: This `utils.py` file might be hideous but how are classes related here? lol (specifically asking about this 1?)

Jongy:
I ran the aarch64 exe build once (see https://github.com/Granulate/gprofiler/pull/176/commits/308861d0c33c9f345e87d8afff0ced9fde4b0cb5), downloaded the binary and tested it. Works like a charm. This PR is ready.

PR #178, opened 2021-09-09
Get process comm from /proc/PID/status instead of /proc/PID/comm

liorgorb:
## Description
Get process comm from `/proc/PID/status` instead of `/proc/PID/comm`, as it only exists since kernel 2.6.33

## Motivation and Context
gProfiler fails in kernel versions prior to 2.6.33.

## Checklist:

### DIFF
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -332,8 +332,7 @@ def assert_program_installed(program: str):
 
 def run_in_ns(nstypes: List[str], callback: Callable[[], T], target_pid: int = 1) -> Optional[T]:
     """
-    Runs a ca
-    llback in a new thread, switching to a set of the namespaces of a target process before
+    Runs a callback in a new thread, switching to a set of the namespaces of a target process before
     doing so.
 
     Needed initially for switching mount namespaces, because we can't setns(CLONE_NEWNS) in a multithreaded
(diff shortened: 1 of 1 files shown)


gprofiler/utils.py
```diff
@@ -344,36 +345,7 @@ def run_in_ns(nstypes: List[str], callback: Callable[[], T], target_pid: int = 1
     By default, run stuff in init NS. You can pass 'target_pid' to run in the namespace of that process.
     """
 
-    # make sure "mnt" is last, once we change it our /proc is gone
```
Jongy, line 347: Remove this patch :)

gprofiler/utils.py
```diff
@@ -332,7 +332,8 @@ def assert_program_installed(program: str):
 
 def run_in_ns(nstypes: List[str], callback: Callable[[], T], target_pid: int = 1) -> Optional[T]:
     """
-    Runs a callback in a new thread, switching to a set of the namespaces of a target process before
+    Runs a ca
```
Jongy, line 335: And this

gprofiler/utils.py
```diff
@@ -487,7 +459,7 @@ def random_prefix() -> str:
 
 
 def process_comm(process: Process) -> str:
-    comm = Path(f"/proc/{process.pid}/comm").read_text()
-    # the kernel always adds \n
-    assert comm.endswith('\n'), f"unexpected comm: {comm!r}"
-    return comm[:-1]
+    status = Path(f"/proc/{process.pid}/status").read_text()
+    name_line = status.splitlines()[0]
+    assert name_line.startswith("Name:")
```
Jongy, line 464:
```suggestion
    assert name_line.startswith("Name:\t")
```

gprofiler/utils.py
```diff
@@ -487,7 +459,7 @@ def random_prefix() -> str:
...
+    return name_line.split("\t")[1]
```
Jongy, line 465:
```suggestion
    return name_line.split("\t", 1)[1]
```

Jongy, line 465: I checked and both use `proc_task_name` so I guess it should produce matching results with `/proc/pid/comm`?

Jongy (changes requested):
.

gprofiler/utils.py
liorgorb: Yeah, should be fine.

Jongy:
Verified that the outputs are consistent between the 2:
1. For PID 1 (`systemd` in my case)
2. For a process started as `./aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` - both versions of this function return `'a' * 15`.

Jongy (approved):

PR #184, opened 2021-10-01
Limit the amount of locally-buffered log records before being sent

Jongy:
## Description
Limit the amount of locally-buffered log records before being sent.

## Motivation and Context
Avoid memory explosion if logs aren't being sent, for whatever reason.

## How Has This Been Tested?
Locally - using this:
```
diff --git a/gprofiler/log.py b/gprofiler/log.py
index dd2291c..8a17148 100644
--- a/gprofiler/log.py
+++ b/gprofiler/log.py
@@ -107,7 +107,7 @@ class RemoteLogsHandler(logging.Handler):
     APIClient, while logs are still accumulated from the beginning.
     """
 
-    MAX_BUFFERED_RECORDS = 100 * 1000  # max number of records to buffer locally
+    MAX_BUFFERED_RECORDS = 100 * 1  # max number of records to buffer locally
 
     def __init__(self, path: str = "logs", api_client: Optional['APIClient'] = None) -> None:
         super().__init__(logging.DEBUG)
@@ -178,6 +178,8 @@ class RemoteLogsHandler(logging.Handler):
         # Snapshot the current num logs because logs list might be extended meanwhile.
         logs_count = len(self._logs)
         try:
+            print(logs_count)
+            raise 1
             self._api_client.post(self._path, data=self._logs[:logs_count], api_version='v1')
         except (APIError, RequestException):
             self._logger.exception("Failed sending logs to server")
(END)
```
I verified the printed number doesn't exceed 100.

## Checklist:

### DIFF
--- a/gprofiler/log.py
+++ b/gprofiler/log.py
@@ -107,6 +107,8 @@ class RemoteLogsHandler(logging.Handler):
     APIClient, while logs are still accumulated from the beginning.
     """
 
+    MAX_BUFFERED_RECORDS = 100 * 1000  # max number of records to buffer locally
+
     def __init__(self, path: str = "logs", api_client: Optional['APIClient'] = None) -> None:
         super().__init__(logging.DEBUG)
         self._api_client = api_client
@@ -125,6 +127,8 @@ class RemoteLogsHandler(logging.Handler):
             return
 
         self._logs.append(self._make_dict_record(record))
+        # trim logs to last N entries
+        self._logs[: -self.MAX_BUFFERED_RECORDS] = []
 
     def _make_dict_record(self, record: LogRecord):
         formatted_timestamp = datetime.datetime.utcfromtimestamp(record.created).isoformat()


gprofiler/log.py
```diff
@@ -125,6 +127,8 @@ def emit(self, record: LogRecord) -> None:
             return
 
         self._logs.append(self._make_dict_record(record))
+        # trim logs to last N entries
+        self._logs[: -self.MAX_BUFFERED_RECORDS] = []
```
alonbentamar, line 131: We should be aware that a trimming happened, at least once. Perhaps add a single log that the logs were trimmed, set a bool flag in the logger that the trimming message was logged, and reset the flag on a successful logs upload?

alonbentamar (changes requested):
Small change

gprofiler/log.py
Jongy: Good call. I'll add such log. Logically, as long as trimming has happened, there'll always be one such log in the buffer (because once one gets trimmed, a new one is added).

gprofiler/log.py
```diff
@@ -125,6 +127,12 @@ def emit(self, record: LogRecord) -> None:
             return
 
         self._logs.append(self._make_dict_record(record))
+        # truncate logs to last N entries
+        if len(self._logs) > self.MAX_BUFFERED_RECORDS:
+            self._logger.warn(
+                f"Truncating log buffer as the maximum number of records ({self.MAX_BUFFERED_RECORDS}) has been reached"
+            )
```
alonbentamar, line 134:
```suggestion
            self._logger.warning(
                f"Truncating log buffer as the maximum number of records ({self.MAX_BUFFERED_RECORDS}) has been reached"
            )
```
`warn` is deprecated

gprofiler/log.py
alonbentamar: I don't think your logic works, seems to me like it would remove the new "truncated" log

Jongy:
Still sucks now:
```
[2021-10-03 17:29:16,183] WARNING: gprofiler.RemoteLogsHandler: Truncating log buffer as the maximum number of records (100) has been reached (log_buffer_check=False)
[2021-10-03 17:29:16,185] DEBUG: gprofiler.utils: Running command: (grep -lP '^.+/libjvm\.so$' /proc/*/maps)
[2021-10-03 17:29:16,185] WARNING: gprofiler.RemoteLogsHandler: Truncating log buffer as the maximum number of records (100) has been reached (log_buffer_check=False)
[2021-10-03 17:29:16,351] DEBUG: gprofiler.utils: (["grep -lP '^.+/libjvm\\.so$' /proc/*/maps"]) exit code: 1
[2021-10-03 17:29:16,352] WARNING: gprofiler.RemoteLogsHandler: Truncating log buffer as the maximum number of records (100) has been reached (log_buffer_check=False)
[2021-10-03 17:29:18,286] DEBUG: gprofiler.profilers.perf: perf stderr: b'[ perf record: dump data: Woken up 4 times ]\n[ perf record: Dump /tmp/gprofiler_tmp/tmpjpgmos30/perf.fp.2021100317291821 ]\n'
[2021-10-03 17:29:18,286] WARNING: gprofiler.RemoteLogsHandler: Truncating log buffer as the maximum number of records (100) has been reached (log_buffer_check=False)
[2021-10-03 17:29:18,286] DEBUG: gprofiler.utils: Running command: (/app/gprofiler/resources/perf script -F +pid -i /tmp/gprofiler_tmp/tmpjpgmos30/perf.fp.2021100317291821)
[2021-10-03 17:29:18,287] WARNING: gprofiler.RemoteLogsHandler: Truncating log buffer as the maximum number of records (100) has been reached (log_buffer_check=False)
```
getting complex. Will think of simplification

alonbentamar:
> getting complex. Will think of simplification

I think my bool flag solution is decent

Jongy:
> > getting complex. Will think of simplification
> 
> I think my bool flag solution is decent

So I'll set a flag and in `try_send_log_to_server`, if that flag is set, I'll add that log message. Though then it won't be in its correct spot (as in, right after the truncation. oh well, also with the current code it's not haha)

Jongy:
Okie, now it'll write the "truncation" log message on each iteration and not on every log message. Acceptable

gprofiler/log.py
```diff
@@ -174,6 +182,12 @@ def try_send_log_to_server(self):
         # Snapshot the current num logs because logs list might be extended meanwhile.
         logs_count = len(self._logs)
         try:
+            if self._truncated:
+                self._logger.warning(
+                    f"Log buffer truncation has occured as the maximum number of records ({self.MAX_BUFFERED_RECORDS}) "
+                    "was reached",
+                )
+                self._truncated = False
```
alonbentamar, line 190: append the log to the `_logs` list only if `self._truncated == False` in `RemoteLogsHandler.emit`, and set `self._truncated = False` only after you successfully send the logs home. Currently, you will add tons of `Log buffer truncation has occured` messages, which override older messages which would be more interesting.

gprofiler/log.py
```diff
@@ -174,6 +182,12 @@ def try_send_log_to_server(self):
         # Snapshot the current num logs because logs list might be extended meanwhile.
         logs_count = len(self._logs)
         try:
+            if self._truncated:
+                self._logger.warning(
+                    f"Log buffer truncation has occured as the maximum number of records ({self.MAX_BUFFERED_RECORDS}) "
```
alonbentamar, line 187:
```suggestion
                    f"Log buffer truncation has occurred as the maximum number of records ({self.MAX_BUFFERED_RECORDS}) "
```

PR #185, opened 2021-10-01
Amend the executable run command for background execution

Jongy:


README.md
```diff
@@ -119,13 +119,19 @@ The command above mounts both of these directories.
...
-sudo ./gprofiler -cu --token <token> --service-name <service> [options]
+sudo sh -c "setsid ./gprofiler -cu --token <token> --service-name <service> [options] > /dev/null 2>&1 &"
 ```
 
+You can verify that it has started correctly with `pgrep gprofiler`. If that doesn't find any process, try running without `> /dev/null 2>&1 &` so you can inspect the output, and look for errors.
+
+For non-daemon mode runes, you can remove the `setsid` and `> /dev/null 2>&1 &` parts.
+
+The logs can then be viewed in their default location (`/var/log/gprofiler`).
+
```
alonbentamar, line 134: I think you should add the `pgrep` command to the commands list above, I can assure you some people will not read this part and will only run the commands. Adding the `pgrep` above will help mitigate that

alonbentamar (changes requested):

README.md
Jongy: K will reorder (tho my intention was not as a pre-check but merely as a way to know if it's running - since it starts as a daemon, you have no feedback whether it has even started or not)

README.md
```diff
@@ -119,13 +119,21 @@ The command above mounts both of these directories.
...
 
-Run the following to have gprofiler running continuously, uploading to Granulate Performance Studio:
+First, check if gProfiler is already running - run `pgrep gprofiler`. You should see 3 PIDs (staticx bootloader, PyInstaller and gProfiler's Python itself). If no processes match, then gProfiler is not running.
+
+Run the following to have gprofiler running continuously, in the background, uploading to Granulate Performance Studio:
 ```bash
 wget https://github.com/Granulate/gprofiler/releases/latest/download/gprofiler
 sudo chmod +x gprofiler
-sudo ./gprofiler -cu --token <token> --service-name <service> [options]
+sudo sh -c "setsid ./gprofiler -cu --token <token> --service-name <service> [options] > /dev/null 2>&1 &"
```
alonbentamar, line 128:
```suggestion
sudo sh -c "setsid ./gprofiler -cu --token <token> --service-name <service> [options] > /dev/null 2>&1 &"
pgrep gprofiler  # Make sure gProfiler is running
```
I actually meant that the `pgrep` command to be here as well, as some people (like me sometimes) only read the commands and skip the docs

Jongy: If so, we should add a `sleep` (as it takes a while for staticx & PyInstaller to start, if both commands are pasted together it's not enough time for it to start)

alonbentamar: Sounds good to me

alonbentamar (approved):

PR #186, opened 2021-10-03
Feature/local metrics output

alonbentamar:
Save the metrics and metadata to the local profile files

## Checklist:

### DIFF
--- a/gprofiler/merge.py
+++ b/gprofiler/merge.py
@@ -14,6 +14,7 @@ from gprofiler.docker_client import DockerClient
 from gprofiler.gprofiler_types import ProcessToStackSampleCounters, StackToSampleCount
 from gprofiler.log import get_logger_adapter
 from gprofiler.metadata.metadata_type import Metadata
+from gprofiler.system_metrics import Metrics
 
 logger = get_logger_adapter(__name__)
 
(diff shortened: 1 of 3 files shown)


gprofiler/merge.py
```diff
@@ -14,6 +14,7 @@
 from gprofiler.gprofiler_types import ProcessToStackSampleCounters, StackToSampleCount
 from gprofiler.log import get_logger_adapter
 from gprofiler.metadata.metadata_type import Metadata
+from gprofiler.system_metrics import Metrics
```
Jongy, line 17: Oof, I don't like it that we have to change this file at all. The "metadata" and "metrics" is unrelated, logically, to this file.

@alonbentamar think if you have any way to change the API such that adding new metadata/metrics/whatever doesn't involve touching this file.

Jongy (changes requested):
1 small, if you can't find a good way to do that, let's merge

gprofiler/merge.py
alonbentamar: Meh, this file is the one that outputs the final stack collapse, I don't think it's bad that it knows about the metadata and metrics as they are part of the file

Jongy: It's not "bad" but it's not useful either, these are different parts of the code. `merge.py` is about the merging of stackcollapse files, not appending random data as the header in files. Can we maybe extract that logic away from this file? It's something that I'd wanted to do a while back.

alonbentamar: It's pretty messy right now, can't think of a nice way to do it without a pretty big refactor :/

Jongy: We'll need to do that one soon

Jongy (approved):

PR #187, opened 2021-10-03
Disable PHP profiling by default

Jongy:
## Description
Don't run `phpspy` by default. It seems to leak memory in some cases. Let's enable it only when needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ For each profiling session (each profiling duration), gProfiler produces outputs
 Profiling using eBPF incurs lower overhead & provides kernel stacks. This (currently) requires kernel headers to be installed.
 
 ### PHP profiling options
+* `--php-mode phpspy`: Enable PHP profiling with phpspy.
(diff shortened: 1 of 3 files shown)


alonbentamar:
I worry about users that use PHP, and after upgrading to the latest version they will stop seeing the PHP stacks and probably won't understand why. Can you think of a solution?

Jongy:
> I worry about users that use PHP, and after upgrading to the latest version they will stop seeing the PHP stacks and probably won't understand why. Can you think of a solution?

phpspy is not mature enough IMO. I encountered this memory leak on one deployment of gProfiler. I haven't pinpointed / debugged it yet. If that's the case, I think we shouldn't run phpspy by default. As long as the documentation explains it, I think that it's alrigt.

alonbentamar (approved):
Fair enough, perhaps alert product so they will decide whether or not to talk with certain clients about it

PR #188, opened 2021-10-03
Include profiler name in the "Profiling process" log message

Jongy:
Makes it easier to read the logs.


michelhe:
Can add a `_PROFILER_NAME` to each profiler class, and create a logger adapter for each class that consists of this name? this way the profiler name will be included in all the log messages it generates

Jongy:
> Can add a `_PROFILER_NAME` to each profiler class, and create a logger adapter for each class that consists of this name? this way the profiler name will be included in all the log messages it generates

Actually this is already included as the `logger_name` field, since each file has its own `logger` object. This does make it a bit more readable.

michelhe (approved):
ok

PR #189, opened 2021-10-07
Update AWS Fargate docs

Jongy:
Update the docs based on some usage experience :)

This involves:
1. Giving the full `curl` command.
2. Overriding `entrypoint` (otherwise, if the user's Dockerfile uses it, our "bash" won't run)
3. Don't redirect stdout & stderr by default, as they're often needed for debugging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ So in order to deploy gProfiler, we need to modify a container definition to inc
 
     `--perf-mode none` is required because our container will not have permissions to run system-wide `perf`, so gProfiler will profile only runtime processes. See [perf-less mode](#perf-less-mode) for more information.
 
-    gProfiler and its installation process will send the outputs to your container's stdout & stderr. After verifying that everything works, you can append `> /tmp/gprofiler_log 2>&1` to the gProfiler command parenthesis (in this example, before the `& python ...`).
+    gProfiler and its installation process will send the outputs to your container's stdout & stderr. After verifying that everything works, you can append `> /dev/null 2>&1` to the gProfiler command parenthesis (in this example, before the `& python ...`) to prevent it from spamming your container logs.
 
     This requires your image to have `wget` installed - you can make sure `wget` is installed, or substitute the `wget` command with `curl -SL https://github.com/Granulate/gprofiler/releases/latest/download/gprofiler --output /tmp/gprofiler`, or any other HTTP-downloader you wish.
 2. Add `linuxParameters` to the container definition (this goes directly in your entry in `containerDefinitinos`):


README.md
```diff
@@ -183,9 +183,11 @@ At the time of this writing, Fargate does not support `DAEMON` tasks (see [this]
 Furthermore, Fargate does not allow using `"pidMode": "host"` in the task definition (see documentation of `pidMode` [here](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskDefinition.html)). Host PID is required for gProfiler to be able to profile processes running in other containers (in case of Fargate, other containers under the same `containerDefinition`).
 
 So in order to deploy gProfiler, we need to modify a container definition to include running gProfiler alongside the actual application. This can be done with the following steps:
-1. Modify the `command` parameter of your entry in the `containerDefinitions` array. The new command should include downloading of gProfiler & executing it in the background.
+1. Modify the `command` & `entrypoint` parameters of your entry in the `containerDefinitions` array. The new command should include downloading of gProfiler & executing it in the background, and `entrypoint` will be `bash`.
```
Ernyoke, line 186: I think we should specify that the `entrypoint` should be an array as well, like `"entryPoint": ["/bin/bash"]`. Otherwise the ECS JSON validation will pass, but the task creation will raise an error without any exact description what the actual issue is.

README.md
```diff
@@ -183,15 +183,19 @@ At the time of this writing, Fargate does not support `DAEMON` tasks (see [this]
...
+
+    Additionally, we will set `entryPoint` to `["/bin/bash"]`. If you had used `entryPoint` prior to incorporating gProfiler, make sure to use it in the new `command`.
 
     `--disable-pidns-check` is required because, well, we won't run in init PID NS :)
 
     `--perf-mode none` is required because our container will not have permissions to run system-wide `perf`, so gProfiler will profile only runtime processes. See [perf-less mode](#perf-less-mode) for more information.
 
-   This requires your image to have `wget` installed - you can make sure `wget` is installed, or substitute it with `curl` or any other HTTP-downloader you wish.
+    gProfiler and its installation process will send the outputs to your container's stdout & stderr. After verifying that everything works, you can append `> /tmp/gprofiler_log 2>&1` to the gProfiler command parenthesis (in this example, before the `& python ...`).
```
alonbentamar, line 196: 1. If it's Fargate, why would you need a log file? I don't think you could ever access it. Might as well send it to `/dev/null`
2. You could explain how to setup Cloudwatch logs here

Jongy: 1. I suppose, yes
2. I could but I don't know how myself haha. Currently we don't need - if possible, it's best to just send the logs to the backend. If we ever end setting this up, we'll add it here

alonbentamar (approved):

PR #192, opened 2021-10-19
Add a guide on Dataproc installation

alonbentamar:
## Checklist:

### DIFF
--- a/deploy/dataproc/gprofiler_initialization_action.sh
+++ b/deploy/dataproc/gprofiler_initialization_action.sh
@@ -19,7 +19,7 @@ if [ "$ENABLE_STDOUT" != "1" ]; then
   OUTPUT_REDIRECTION="> /dev/null 2>&1"
 fi
 
-wget --no-verbose https://github.com/Granulate/gprofiler/releases/latest/download/gprofiler_$(uname -m)
+wget --no-verbose https://github.com/Granulate/gprofiler/releases/latest/download/gprofiler_$(uname -m) -O gprofiler
 sudo chmod +x gprofiler
 sudo sh -c "setsid ./gprofiler -cu --token '$GPROFILER_TOKEN' --service-name '$GPROFILER_SERVICE' $OUTPUT_REDIRECTION &"
 echo "gProfiler installed successfully."


README.md
```diff
@@ -265,6 +265,28 @@ setuptools.setup(
 )
 ```
 
+## Running on Google Dataproc
+To run gProfiler on your cluster, you will need to add an [initialization action](https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/init-actions)
+that will install agent on all of your workers when the cluster is created.
```
Jongy, line 270:
```suggestion
that will install the agent on all of your workers when the cluster is created.
```

README.md
```diff
@@ -265,6 +265,28 @@ setuptools.setup(
...
+gsutil cp gprofiler_initialization_action.sh gs://<YOUR BUCKET>
+```
+If you don't have a Google Storage bucket, make sure you create one ([documentation](https://cloud.google.com/storage/docs/creating-buckets)).
+Then, create your Dataproc cluster with the `--initialization-actions` flag -
+```shell
+TOKEN=<TOKEN> \
+SERVICE=<SERVICE NAME> \
+gcloud dataproc clusters create <CLUSTER NAME> \
+--initialization-actions gs://<YOUR BUCKET>/gprofiler_initialization_action.sh \
+--metadata gprofiler-token=$TOKEN,gprofiler-service=$SERVICE --region us-central1
```
Jongy, line 283: Doube quote the expanded values of token and service here

deploy/dataproc/gprofiler_initialization_action.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
Jongy, line 1: Add license

deploy/dataproc/gprofiler_initialization_action.sh
```diff
@@ -0,0 +1,13 @@
...
+
+GPROFILER_TOKEN=$(/usr/share/google/get_metadata_value attributes/gprofiler-token)
+readonly GPROFILER_TOKEN
+
+GPROFILER_SERVICE=$(/usr/share/google/get_metadata_value attributes/gprofiler-service)
+readonly GPROFILER_SERVICE
+
+wget https://github.com/Granulate/gprofiler/releases/latest/download/gprofiler
+sudo chmod +x gprofiler
+sudo sh -c "setsid ./gprofiler -cu --token '$GPROFILER_TOKEN' --service-name '$GPROFILER_SERVICE' > /dev/null 2>&1 &"
```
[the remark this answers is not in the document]
Jongy: I also think that, learning from the Fargate case, we should not redirect to /dev/null by default, as it hinders debugging which is often needed. I suggest you remove it and instead mention in a comment that it can be added if desired to avoid spamming (like we did in the Fargate section)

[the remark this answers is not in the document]
alonbentamar: What do I need to fix in the redirection?

PR #201, opened 2021-11-09
Bugfix/python kernel headers

glipner:
Fix a bug that was encountered on Amazon Linux 2018.3, which resulted in `gprofiler` being unable to properly locate the kernel headers, thus unable to use PyPerf.

Specifically, the bug manifests itself when:
1. running in a container
2. on the host, `/lib` is not a symlink
3. on the host, `/lib/modules/$(uname -r)/build` is a relative symlink (i.e. `../../../usr/src/...`)

Inside the `gprofiler` container `/lib` is linked to `/usr/lib`, which means that the relative link from the host becomes broken.
The solution is to fix the symlink inside the container, by:
1. mounting `/lib/modules/$(uname -r)` as an overlay inside the container
2. converting the relative link to an absolute one
3. removing and recreating the link

### DIFF
--- a/gprofiler/profilers/python.py
+++ b/gprofiler/profilers/python.py
@@ -155,6 +155,26 @@ class PythonEbpfProfiler(ProfilerBase):
 
     @staticmethod
     def _fix_kernel_headers_symlink():
+        """Fix a bug preventing PyPerf from being used even though kernel headers were present on the host.
+
+
+        Fix a bug that was encountered on Amazon Linux 2018.3,
+        which resulted in gprofiler being unable to properly locate the kernel headers, thus unable to use PyPerf.
+
+        Specifically, the bug manifests itself when:
+
+        + Running in a container.
+        + On the host, /lib is not a symlink.
+        + On the host, /lib/modules/$(uname -r)/build is a relative symlink (i.e. ../../../usr/src/...).
+
+        Inside the gprofiler container /lib is linked to /usr/lib, which means that the relative link from the host
+        becomes broken.
+        The solution is to fix the symlink inside the container, by:
+
+        + Mounting /lib/modules/$(uname -r) as an overlay inside the container.
+        + Converting the relative link to an absolute one.
+        + Removing and recreating the link.
+        """
         try:
             if get_run_mode() not in ("container", "k8s"):
                 return
--- a/gprofiler/profilers/python.py
+++ b/gprofiler/profilers/python.py
@@ -4,7 +4,9 @@
 #
 import glob
 import os
+import re
 import signal
+import subprocess
 from pathlib import Path
 from subprocess import Popen, TimeoutExpired
 from threading import Event
@@ -16,7 +18,7 @@ from gprofiler.exceptions import CalledProcessError, ProcessStoppedException, St
 from gprofiler.gprofiler_types import ProcessToStackSampleCounters, StackToSampleCount, nonnegative_integer
 from gprofiler.log import get_logger_adapter
 from gprofiler.merge import parse_many_collapsed, parse_one_collapsed_file
-from gprofiler.metadata.system_metadata import get_arch
+from gprofiler.metadata.system_metadata import get_arch, get_run_mode
 from gprofiler.profilers.profiler_base import ProcessProfilerBase, ProfilerBase, ProfilerInterface
 from gprofiler.profilers.registry import ProfilerArgument, register_profiler
 from gprofiler.utils import (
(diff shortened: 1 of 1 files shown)


gprofiler/profilers/python.py
```diff
@@ -151,6 +153,49 @@ def __init__(
...
         self.user_stacks_pages = user_stacks_pages
 
+    @staticmethod
+    def _fix_kernel_headers_symlink():
+        try:
+            if get_run_mode() not in ("container", "k8s"):
+                return
+
+            try:
+                uname = subprocess.check_output(["uname", "-r"]).decode('latin-1').strip()
```
Jongy, line 163: `latin-1`? o.O

gprofiler/profilers/python.py
```diff
@@ -151,6 +153,49 @@ def __init__(
         self.output_path = Path(self._storage_dir) / f"pyperf.{random_prefix()}.col"
         self.user_stacks_pages = user_stacks_pages
 
+    @staticmethod
+    def _fix_kernel_headers_symlink():
```
Jongy, line 157: This function should be documented (I suppose, with the comment in the PR's description)

gprofiler/profilers/python.py
```diff
@@ -151,6 +153,49 @@ def __init__(
...
+            build_path = f"/lib/modules/{uname}/build"
+            if not os.path.islink(build_path):
+                return
+
+            if os.path.lexists(build_path + "/"):
+                return
+
+            lib_modules_uname_path = f"/lib/modules/{uname}"
+
+            os.makedirs('/dev/shm/modules_mount/upper', 555)
```
Jongy, line 176: `/dev/shm` is used as a tmpfs here?

gprofiler/profilers/python.py
glipner: "Latin-1, also known as ISO-8859-1, is a similar encoding. Unicode code points 0–255 are identical to the Latin-1 values, so converting to this encoding simply requires converting code points to byte values; if a code point larger than 255 is encountered, the string can’t be encoded into Latin-1."

https://docs.python.org/3.4/howto/unicode.html

gprofiler/profilers/python.py
```diff
@@ -151,6 +153,49 @@ def __init__(
...
+                    '-o',
+                    f'upperdir=/dev/shm/modules_mount/upper,lowerdir={lib_modules_uname_path}'
+                    ',workdir=/dev/shm/modules_mount/workdir',
+                    lib_modules_uname_path,
+                ]
+            )
+
+            os.chmod(lib_modules_uname_path, 0o755)
+            link_target = os.readlink(build_path)
+            new_target = re.sub(r'(\.\./)+usr/src', '/usr/src', link_target, count=1)
```
Jongy, line 193: Don't we want to count the number of `../` and make sure they are what we expect (4, I suppose)? Or is there an hidden assumption here that the `/usr/src` part will always be the root?

Jongy (changes requested):

gprofiler/profilers/python.py
glipner: Yes, for technical reasons, it didn't work on `/tmp` because that one isn't

gprofiler/profilers/python.py
glipner: Assumption.

If the kernel headers actually exist in `/usr/src/`, we can point there just fine.
Otherwise, we probably won't have access from the container any way.

gprofiler/profilers/python.py
glipner: Added in the latest commit

gprofiler/profilers/python.py
Jongy: Yeah it should be on a tmpfs. `/tmp` is not a tmpfs on all distros and I suppose `/dev/shm` is. Could you drop a comment on that? it's not obvious, took me a moment to figure out why you do that

gprofiler/profilers/python.py
Jongy: Okay. Can comment the assumption

Jongy (changes requested):
.

gprofiler/profilers/python.py
glipner: Done

gprofiler/profilers/python.py
glipner: done

Jongy (approved):

PR #205, opened 2021-11-13
lint: Enable black's string normalization

Jongy:
## Description
@d3dave , it was disabled by you in #47 . I don't recall why. Now, it's become quite annoying for me to manually fix strings :sweat_smile: 

Personally I prefer `"` to `'` but I really don't mind configuring either (I suppose `black` supports it?). Just let's keep it a standard and let the linter do its job.

@alonbentamar tagging you as well because I know you're up for `'`s.

### DIFF
--- a/deploy/dataflow/setup.py
+++ b/deploy/dataflow/setup.py
@@ -48,7 +48,7 @@ class ProfilerInstallationCommands(setuptools.Command):
         p = subprocess.Popen(
             command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True
         )
-        stdout_data, _ = p.communicate(input=b'\n')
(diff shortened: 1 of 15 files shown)


d3dave:
I prefer the single quotes. Saves having to press shift.

Jongy:
> I prefer the single quotes. Saves having to press shift.

Sigh, I was wrong - they don't seem to support it and they don't plan to: https://github.com/psf/black/issues/118#issuecomment-393298377

Yet I believe that consistency is more important than having to press shift :) So I'm still up with merging this.

alonbentamar:
> @alonbentamar tagging you as well because I know you're up for 's.
@Jongy I'm all for `"`

alonbentamar (approved):

d3dave (approved):

PR #208, opened 2021-11-14
Use proc_events on Java processes

YishaiZinkin:
## Description
Use `proc_events` from `granulate-utils` to track async-profiled processes and log when they exit with non-zero code.

## Checklist:


gprofiler/proc_events.py
```diff
@@ -0,0 +1,129 @@
...
+
+# struct exit_proc_event {
+#         __kernel_pid_t process_pid;
+#         __kernel_pid_t process_tgid;
+#         __u32 exit_code, exit_signal;
+# } exit;
+_exit_proc_event = struct.Struct("=4I")
+
+
+def _proc_events_listener():
```
Jongy, line 70: I'd wrap it in a class with proper lifetime management (`__enter__`, `__exit__` etc)

Jongy: Thread control, ...

Jongy (changes requested):
Looks good. I'll do a full CR once we restructure it ,let's make it an object instead of a set of functions, I think that's more managable.

Jongy:
@YishaiZinkin  - let's move this code to https://github.com/Granulate/utils so it can be used by other components. I'll explain on Sunday.

gprofiler/proc_events.py
YishaiZinkin: What're the advantages you see over the current approach?
The way I see it: you'll never want more than a single thread to listen to connector messages, and you would never want it to die. Hence there is no meaning for creating or holding an object to represent this thread, and there is no reason to use OOP when you don't need an object.
In this case exporting just a bunch of `register_.*_callback` functions is a much clearer and simpler API.

YishaiZinkin: Summary of my verbal discussion with @Jongy: We might want to restart the thread during execution, hence `__enter__` and `__exit__` functionality will be useful.

gprofiler/proc_events.py
```diff
@@ -9,121 +9,134 @@
...
+        try:
+            self._socket.bind((os.getpid(), self._CN_IDX_PROC))
+        except socket.error:
+            logger.exception("")
 
+        self._register_to_connector_events(self._socket)
 
-_listener_thread = threading.Thread(target=_proc_events_listener, daemon=True)
+        while True:
+            (readable, _, _) = select([self._socket], [], [])
```
d3dave, line 105: Use `selectors.DefaultSelector` instead of using `select`.

d3dave (dismissed):

YishaiZinkin:
I suggest we finish the PR here just to concentrate the comments in one place. I'll move the module to `utils` once we're finished.

gprofiler/proc_events.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def __init__(self):
+        self._socket = socket.socket(socket.AF_NETLINK, socket.SOCK_DGRAM, self._NETLINK_CONNECTOR)
+        self._exit_callbacks = []
+        self._should_stop = False
+
+        super().__init__(target=self._proc_events_listener, name="Process Events Listener", daemon=True)
+
+    def _register_to_connector_events(self, socket):
+        """Notify the kernel that we're listening to events on the connector"""
```
Jongy, line 88:
```suggestion
        """Notify the kernel that we're listening for events on the connector"""
```

I think? (also function name)

gprofiler/proc_events.py
```diff
@@ -0,0 +1,159 @@
...
+        cn_proc_op = struct.Struct("=I").pack(self._PROC_CN_MCAST_LISTEN)
+        cn_msg = self._cn_msg.pack(self._CN_IDX_PROC, self._CN_VAL_PROC, 0, 0, len(cn_proc_op), 0) + cn_proc_op
+        nl_msg = self._nlmsghdr.pack(self._nlmsghdr.size + len(cn_msg), self._NLMSG_DONE, 0, 0, os.getpid()) + cn_msg
```
Jongy, line 91: Which man page describes the magic that goes on here? :sweat_smile: can you drop a comment?

gprofiler/proc_events.py
```diff
@@ -0,0 +1,159 @@
...
+
+        socket.send(nl_msg)
+
+    def _proc_events_listener(self):
+        """Runs forever and calls registered callbacks on process events"""
+        try:
+            self._socket.bind((os.getpid(), self._CN_IDX_PROC))
+        except socket.error:
+            logger.exception("")
```
Jongy, line 100: What do we do on this case?

gprofiler/proc_events.py
```diff
@@ -0,0 +1,159 @@
...
+                            self._exit_proc_event.unpack(
+                                data[
+                                    self._base_proc_event.size : self._base_proc_event.size + self._exit_proc_event.size
+                                ]
+                            ),
+                        )
+                    )
+
+                    for callback in self._exit_callbacks:
+                        callback(event_data["pid"], event_data["exit_code"])
```
Jongy, line 146: If there's `pid` and `tgid` - then `tgid` is the "user-space" PID and `pid` is the TID.

In any case - I suggest the API should pass them all? We also want the `exit_signal` because for Java we want to check the `exit_signal`

gprofiler/proc_events.py
YishaiZinkin: There is man 7 netlink, but it doesn't says much. AFAIK if you want to use this mechanism you'll have to read the headers.

gprofiler/profilers/java.py
```diff
@@ -54,6 +55,23 @@ def get_ap_log(self) -> str:
...
+class AsyncProfiledProcessMonitor:
+    def __init__(self):
+        self._attached_processes = []
+        self._proc_events_listener = ProcEventsListener()
+        self._proc_events_listener.start()
+        self._proc_events_listener.register_exit_callback(self._proc_exit_callback)
+
+    def _proc_exit_callback(self, pid, exit_code):
+        if pid in self._attached_processes:
+            if exit_code != 0:
```
Jongy, line 123: We want to check the exit_signal here (JVM dies with SIGABRT I think, or SIGKILL for OOM, or SIGTERM, ...)

Jongy (changes requested):
Very clean API :+1:

Some small comments.

gprofiler/proc_events.py
```diff
@@ -0,0 +1,159 @@
...
+            self._socket.bind((os.getpid(), self._CN_IDX_PROC))
+        except socket.error:
+            logger.exception("")
+
+        self._register_to_connector_events(self._socket)
+        selector = selectors.DefaultSelector()
+        selector.register(self._socket, selectors.EVENT_READ)
+
+        while not self._should_stop:
+            events = selector.select()
```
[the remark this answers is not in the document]
d3dave: classic EBADF case, hopefully before the fd is reopened again

gprofiler/proc_events.py
YishaiZinkin: Here we're just logging the exception. I wasn't sure whether to let it pass through or just log it. Not that we're aiming to make this code generic it's better to `raise` with descriptive error message.
The main error that may happen is insufficient permissions.

gprofiler/proc_events.py
YishaiZinkin: This `pid` is user-space pid. As we discussed yesterday, I'll also parse here the signal out of the exit code.

gprofiler/proc_events.py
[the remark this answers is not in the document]
YishaiZinkin: You mean between the `bind` and the `select`?

gprofiler/proc_events.py
YishaiZinkin: Now we just `raise`

gprofiler/proc_events.py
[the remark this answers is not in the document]
Jongy: No, between `close` and this call. 

I also meant - how does `selector.select()` behave if it's currently blocking and you are closing the socket in another thread?

gprofiler/proc_events.py
```diff
@@ -96,8 +92,8 @@ def _proc_events_listener(self):
         """Runs forever and calls registered callbacks on process events"""
         try:
             self._socket.bind((os.getpid(), self._CN_IDX_PROC))
-        except socket.error:
-            logger.exception("")
+        except PermissionError as e:
+            raise PermissionError("You don't have permissions to bind to the process events connector") from e
```
Jongy, line 96: Don't need the extra layer IMO. Just let the original error pass

YishaiZinkin: The original error isn't very clear:
```
  File "<ipython-input-1-1a014f10004c>", line 98, in _proc_events_listener
    self._socket.bind((os.getpid(), self._CN_IDX_PROC))
PermissionError: [Errno 1] Operation not permitted
```
The description I added is nicer IMO.

Jongy: Both are permission errors and the stacktrace originates from `proc_events_listener`. Personally I dislike wrapping exceptions and prefer to view the original one, especially when it's not a-framework-inside-a-framework case (which might try to hide the original exception because it's too deep).

Anyway, since you `raise ... from e` I suppose it's fine, because the original info is kept. Up to you to decide

gprofiler/proc_events.py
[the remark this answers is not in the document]
d3dave: `select(2)`:
```
   Multithreaded applications
       If  a  file  descriptor  being  monitored by select() is closed in another thread, the
       result is unspecified.  On some UNIX systems, select() unblocks and returns,  with  an
       indication  that  the file descriptor is ready (a subsequent I/O operation will likely
       fail with an error, unless another the  file  descriptor  reopened  between  the  time
       select()  returned  and  the  I/O operations was performed).  On Linux (and some other
       systems), closing the file descriptor in another thread has no effect on select().  In
       summary, any application that relies on a particular behavior in this scenario must be
       considered buggy.
```

gprofiler/proc_events.py
YishaiZinkin: > Anyway, since you raise ... from e I suppose it's fine

Yup, that's the idea

gprofiler/proc_events.py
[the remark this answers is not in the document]
YishaiZinkin: Hmm I wrote the code this way based on the behavior I saw in my Linux machine.
Anyway I fixed it by registering some `pipe` to the `selector` as well, just to make it return on will.

Jongy:
The centos:6 error was fixed in https://github.com/Granulate/gprofiler/pull/221, so u can ignore

gprofiler/proc_events.py
```diff
@@ -0,0 +1,179 @@
...
+    # } exit;
+    _exit_proc_event = struct.Struct("=4I")
+
+    def __init__(self):
+        self._socket = socket.socket(socket.AF_NETLINK, socket.SOCK_DGRAM, self._NETLINK_CONNECTOR)
+        self._exit_callbacks = []
+        self._should_stop = False
+
+        self._selector = selectors.DefaultSelector()
+        # Create a pip so we can make select() return
```
Jongy, line 88:
```suggestion
        # Create a pipe so we can make select() return
```

gprofiler/proc_events.py
```diff
@@ -0,0 +1,179 @@
...
+        r, w = os.pipe()
+        self._select_breaker = os.fdopen(w, "w")
```
Jongy, line 90: Just making sure - Python's `fdopen` does not `dup` or anything, right? So `r, w` are not leaked

gprofiler/proc_events.py
```diff
@@ -0,0 +1,179 @@
...
+        # Write to make select() return
+        self._select_breaker.write("\0")
+
+    @_raise_if_not_running
+    def register_exit_callback(self, callback):
+        self._exit_callbacks.append(callback)
+
+
+_proc_events_listener = _ProcEventsListener()
+_proc_events_listener.start()
```
Jongy, line 171: I suggest we start it upon the first registration. Starting threads & logic upon `import` is not cool. Make it a global but start only upon first registration (I suggest that like all other mechanisms we introduce - we should add a flag to allow disabling it; so when disabled, it should really do nothing)

gprofiler/profilers/java.py
```diff
@@ -649,6 +668,7 @@ def _profile_process(self, process: Process) -> Optional[StackToSampleCount]:
         with AsyncProfiledProcess(
             process, self._storage_dir, self._buildids, self._mode, self._safemode, self._java_safemode
         ) as ap_proc:
+            self._process_monitor.register_process(process.pid)
```
Jongy, line 671: You never remove PIDs from here. @d3dave added some mechanism in his PR: https://github.com/Granulate/gprofiler/pull/219/files#diff-77067d3a05ebeab8e9e2620faefa95849baa469195b8c07aa6283cc160e00df3R98

Well, I guess we can go with your code and whomever merges first can use the other's code. Just make sure your API stands with his.

gprofiler/profilers/java.py
Jongy: Talked IRL - it's complicated. @YishaiZinkin , I suggest you spend more time looking into after *after* moving the existing mechanism to the utils library. Currently, this won't work for the SIGKILL case you've shown me (because the main thread seemed to exit with `0`, and the main thread is the PID we're looking at...)

gprofiler/profilers/java.py
```diff
@@ -110,6 +111,23 @@ def get_ap_log(self) -> str:
...
+        self._attached_processes = []
+        proc_events.register_exit_callback(self._proc_exit_callback)
+
+    def _proc_exit_callback(self, tid, pid, exit_code):
+        if pid in self._attached_processes:
+            if exit_code != 0:
+                logger.warning(
+                    f"Async-profiled Java process [{pid}] (TID: [{tid}]) exited with error code: {exit_code}"
+                )
+            self._attached_processes.remove(pid)
```
Jongy, line 125: This is not correct. A thread may exit and you should not remove the attached process upon it. I'm not sure of the exact semantics, but to my understandings only certain signals lead to killing the entire process.

I suggest that we focus only on the `tid == pid` case (== main thread) and figure out why we get `0` upon SIGKILL.. But first, let's move the proc_events code to utils please.

gprofiler/proc_events.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def register_exit_callback(callback):
```
Jongy, line 174: Also expose "unregister"?

Jongy, line 174: Types. Everywhere please

Jongy (changes requested):
.

gprofiler/proc_events.py
YishaiZinkin: If it does it would've been a bug.
On the other hand, I never close the pipe files, which I should.

gprofiler/profilers/java.py
YishaiZinkin: PIDs are removed in the exit callback

gprofiler/profilers/java.py
YishaiZinkin: Maybe the correct behavior should be to check whether `is_process_running` whenever the callback is called (it's invoked for each thread) and remove it only if it's not?

gprofiler/profilers/java.py
Jongy: Yeah I commented that thinking that the removal in the exit callback is incorrect :sweat_smile:

gprofiler/profilers/java.py
Jongy: Decided to check on TID == PID

gprofiler/profilers/java.py
Jongy: The SIGKILL thing was a faux. And we decided to go with `!= 0` to begin with.

Jongy (approved):

Jongy (approved):

Jongy:
@YishaiZinkin can you merge from master? The other PR got merged

gprofiler/profilers/java.py
```diff
@@ -114,6 +115,23 @@ def get_ap_log(self) -> str:
         return self._ap_log
 
 
+class AsyncProfiledProcessMonitor:
+    def __init__(self):
+        self._attached_processes: List[int] = []
```
Jongy, line 120: You're now tracking processes both here and in `JavaProfiler._profiled_processes` :(

Jongy:
Getting this error:
```
Exception in thread Process Events Listener:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.8/dist-packages/granulate_utils/linux/proc_events.py", line 165, in _proc_events_listener
    self._register_for_connector_events(self._socket)
  File "/usr/local/lib/python3.8/dist-packages/granulate_utils/linux/proc_events.py", line 110, in _register_for_connector_events
    socket.send(nl_msg)
ConnectionRefusedError: [Errno 111] Connection refused
```

Jongy:
> Getting this error:
> 
> ```
> Exception in thread Process Events Listener:
> Traceback (most recent call last):
>   File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
>     self.run()
>   File "/usr/lib/python3.8/threading.py", line 870, in run
>     self._target(*self._args, **self._kwargs)
>   File "/usr/local/lib/python3.8/dist-packages/granulate_utils/linux/proc_events.py", line 165, in _proc_events_listener
>     self._register_for_connector_events(self._socket)
>   File "/usr/local/lib/python3.8/dist-packages/granulate_utils/linux/proc_events.py", line 110, in _register_for_connector_events
>     socket.send(nl_msg)
> ConnectionRefusedError: [Errno 111] Connection refused
> ```

We're failing tests upon ERRORs in the log, but since this happens in another thread and it's not logged, it doesn't fail the tests :/

Jongy:
I pushed an update that merges this logic with #219 , as suggested in https://github.com/Granulate/gprofiler/pull/219#discussion_r757057637.

Jongy:
> Getting this error:
> 
> ```
> Exception in thread Process Events Listener:
> Traceback (most recent call last):
>   File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
>     self.run()
>   File "/usr/lib/python3.8/threading.py", line 870, in run
>     self._target(*self._args, **self._kwargs)
>   File "/usr/local/lib/python3.8/dist-packages/granulate_utils/linux/proc_events.py", line 165, in _proc_events_listener
>     self._register_for_connector_events(self._socket)
>   File "/usr/local/lib/python3.8/dist-packages/granulate_utils/linux/proc_events.py", line 110, in _register_for_connector_events
>     socket.send(nl_msg)
> ConnectionRefusedError: [Errno 111] Connection refused
> ```

It seems like this happens since we're running in a network namespace. `forkstat` fails on a privileged container but passes on if `--net=host`.

Jongy (approved):

PR #209, opened 2021-11-14
Add --controller-pid

Jongy:
## Description
Lets gProfiler know it was invoked by a specific process; and should that process go down for whatever
reason, gProfiler should go down as well.

## Motivation and Context
Automatically stop gProfiler if it was started by another component which has gone down.

## How Has This Been Tested?
* PID not running - gProfiler doesn't start
* PID running then stops - gProfiler stops
* PID running then zombie - gProfiler stops

## Checklist:

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -512,7 +512,7 @@ def parse_cmd_args():
         "--controller-pid",
         default=None,
         type=int,
-        help="PID of the process who invoked gProfiler; if given and that process exits, gProfiler will exit"
+        help="PID of the process that invoked gProfiler; if given and that process exits, gProfiler will exit"
         " as well",
     )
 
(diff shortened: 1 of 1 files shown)


gprofiler/main.py
```diff
@@ -499,6 +508,14 @@ def parse_cmd_args():
         help="Disable sending system and cloud metadata to the Performance Studio",
     )
 
+    parser.add_argument(
+        "--controller-pid",
+        default=None,
+        type=int,
+        help="PID of the process who invoked gProfiler; if given and that process exits, gProfiler will exit"
```
YoniKF, line 515:
```suggestion
        help="PID of the process that invoked gProfiler; if given and that process exits, gProfiler will exit"
```

YoniKF (dismissed):

YoniKF (approved):

PR #212, opened 2021-11-15
Java: Check for Hotspot error files after process termination

d3dave:
## Description
If java process terminates due to an error, detect it and report a log message. In addition, extract siginfo, native frames, vm info, and container info from the hotspot log and report them in our log.

## Motivation and Context
Debugging async-profiler misbehaviors.

## How Has This Been Tested?
1. Changed async-profiler to force a crash.
2. Started OpenJDK example process.
3. Ran gprofiler built from this branch.
4. Ran crashing async-profiler.

### Log output
```
[15:40:25] Found Hotspot error log at /proc/1/root/home/ubuntu/jbb/21-11-16_153905/hs_err_pid32629.log
[15:40:25] Pid 32629 Hotspot VM info: OpenJDK 64-Bit Server VM (25.292-b10) for linux-amd64 JRE (1.8.0_292-8u292-b10-0ubuntu1~18.04-b10), built on Apr 21 2021 11:46:09 by "buildd" with gcc 7.5.0
[15:40:25] Pid 32629 Hotspot siginfo: si_signo: 11 (SIGSEGV), si_code: 1 (SEGV_MAPERR), si_addr: 0x0000000000000004
[15:40:25] Pid 32629 Hotspot native frames:
C  [libc.so.6+0x18e4e1]
C  [libasyncProfiler.so+0x1bb4e]  Profiler::dump(std::ostream&, Arguments&)+0xce
C  [libasyncProfiler.so+0x1bcae]  Profiler::runInternal(Arguments&, std::ostream&)+0x9e
C  [libasyncProfiler.so+0x1c242]  Profiler::run(Arguments&)+0x212
C  [libasyncProfiler.so+0x48d81]  Agent_OnAttach+0x1e1
V  [libjvm.so+0x7ea65b]
V  [libjvm.so+0x2f5e62]
V  [libjvm.so+0xb08d2f]
V  [libjvm.so+0xb0a0fa]
V  [libjvm.so+0x990552]
C  [libpthread.so.0+0x76db]  start_thread+0xdb
[15:40:25] Pid 32629 Hotspot container info:
container_type: cgroupv1
cpu_cpuset_cpus: 0-3
cpu_memory_nodes: 0
active_processor_count: 4
cpu_quota: -1
cpu_period: 100000
cpu_shares: -1
memory_limit_in_bytes: -1
memory_and_swap_limit_in_bytes: -2
memory_soft_limit_in_bytes: -1
memory_usage_in_bytes: 5046411264
[...]

PR #213, opened 2021-11-16
Added java version parsing and enhanced version checks

IlayRosenberg:
## Description

Enhanced the current java version check mechanism to parse the output of `java -version` and compare it against known "good" versions

## How Has This Been Tested?

Added new tests

## Checklist:

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -6,9 +6,12 @@ import errno
 import functools
 import os
 import shutil
+from dataclasses import dataclass
 from pathlib import Path
 from threading import Event
 from typing import List, Optional
+from packaging.version import Version
+import re
 
 import psutil
 from psutil import Process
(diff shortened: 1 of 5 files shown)


gprofiler/profilers/java.py
```diff
@@ -5,12 +5,15 @@
 import errno
 import functools
 import os
+import re
 import shutil
+from dataclasses import dataclass
```
Jongy, line 10: Does it work with Python 3.6? (the standalone executable uses it)

gprofiler/profilers/java.py
```diff
@@ -318,8 +321,76 @@ def read_output(self) -> Optional[str]:
...
+    #   openjdk version "1.8.0_265"
+    #   OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_265-b01)
+    #   OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.265-b01, mixed mode)
+    # We are taking the version from the first line, and the build number and vm name from the last line
+
+    lines = version_string.splitlines()
+    # version is always in quotes
+    _, version_str, _ = lines[0].split('"')
+    build_str = lines[2].split("(build ")[1]
+    assert "," in build_str, "Didn't find comma in build information"
```
Jongy, line 346:
```suggestion
    assert "," in build_str, f"Didn't find comma in build information: {build_str!r}"
```

gprofiler/profilers/java.py
```diff
@@ -318,8 +321,76 @@ def read_output(self) -> Optional[str]:
...
+    if version_str.endswith("-internal"):
+        # Not sure what this means, ignore
+        version_str = version_str[: -len("-internal")]
+
+    version_list = version_str.split(".")
+    if version_list[0] == "1":
+        # For java 8 and prior, versioning looks like
+        # 1.<major>.0_<minor>-b<build_number>
+        # For example 1.8.0_242-b12 means 8.242 with build number 12
+        assert len(version_list) == 3, "Unexpected number of elements for old-style java version"
```
Jongy, line 359:
```suggestion
        assert len(version_list) == 3, "Unexpected number of elements for old-style java version: {version_list!r}"
```

Jongy: Generally in all `assert`s, give the bad object in the message

gprofiler/profilers/java.py
```diff
@@ -318,8 +321,76 @@ def read_output(self) -> Optional[str]:
...
     ],
 )
+@dataclass
+class JvmVersion:
+    version: Version
+    build: int
+    name: str
+
+
+JAVA11_BUILD_REGEX = re.compile("\\d+")
```
Jongy, line 331:
```suggestion
JAVA11_BUILD_REGEX = re.compile(r"\d+")
```

also since it's very small I suggest you just inline it? Would make the code more readable. Don't need to cache compilation, Python does it anyway :)

IlayRosenberg: >Don't need to cache compilation, Python does it anyway :)

Cool, didn't know that

gprofiler/profilers/java.py
```diff
@@ -342,12 +413,36 @@ def __init__(
...
+        for exclusion in self.JDK_EXCLUSIONS:
+            if exclusion in java_version_cmd_output:
+                logger.error(f"Spotted unsupported keyword {exclusion} in java version, not profiling")
+                return False
+
+        try:
+            jvm_version = parse_jvm_version(java_version_cmd_output)
+            logger.info(f"Checking support for java version {jvm_version}")
+        except Exception as e:
+            logger.error(f"Failed to parse java -version output {java_version_cmd_output}: {e}")
```
Jongy, line 433: Need to `return` after that? Because `jvm_version` is not defined

gprofiler/profilers/java.py
```diff
@@ -342,12 +413,36 @@ def __init__(
...
+
+        if jvm_version.version.major not in self.MINIMAL_SUPPORTED_VERSIONS:
+            logger.error(f"Unsupported java version {jvm_version.version}")
+            return False
+        min_version, min_build = self.MINIMAL_SUPPORTED_VERSIONS[jvm_version.version.major]
```
Jongy, line 438: Handle the possible `KeyError` here?

gprofiler/profilers/java.py
IlayRosenberg: No it doesn't, I'll remove it

gprofiler/profilers/java.py
IlayRosenberg: I already do (in the preceding `if`)

Jongy: Oh, cool

gprofiler/profilers/java.py
```diff
@@ -333,21 +405,51 @@ def __init__(
...
+            if exclusion in java_version_cmd_output:
+                logger.error(f"Spotted unsupported keyword {exclusion} in java version, not profiling")
+                return False
+
+        try:
+            jvm_version = parse_jvm_version(java_version_cmd_output)
+            logger.info(f"Checking support for java version {jvm_version}")
+        except Exception as e:
+            logger.error(f"Failed to parse java -version output {java_version_cmd_output}: {e}")
+            return False
```
Jongy, line 440: This part (and generally - failing the `JavaProfiler` if the JDK version is too old) - should only be enabled if `java_safemode` is set.

Jongy (changes requested):
1 last

gprofiler/profilers/java.py
```diff
@@ -392,15 +492,25 @@ def _run_java_version() -> None:
 
     def _profile_process(self, process: Process) -> Optional[StackToSampleCount]:
         logger.info(f"Profiling process {process.pid} with async-profiler")
+        process_basename = os.path.basename(process.exe())
+        if self._java_safemode and process_basename != "java":
+            logger.error(f"Non-java basenamed process {process.pid} ({process.exe()}), skipping...")
+            return None
 
         # Get Java version
+        java_version_output = self._get_java_version(process)
```
Jongy, line 501: You shouldn't call it if `basename != "java"`, but it will happen in this flow

gprofiler/profilers/java.py
```diff
@@ -392,15 +492,25 @@ def _run_java_version() -> None:
...
-        if self._version_check and os.path.basename(process.exe()) == "java":
-            if not self._is_jdk_version_supported(self._get_java_version(process)):
-                logger.warning(f"Process {process.pid} running unsupported Java version, skipping...")
+        if self._version_check and process_basename == "java":
+            if not self._is_jvm_type_supported(java_version_output):
+                logger.error(f"Process {process.pid} running unsupported JVM, skipping...")
                 return None
 
+        if self._java_safemode and not self._is_jvm_version_supported(java_version_output):
+            logger.error(f"Process {process.pid} running unsupported Java version, skipping...")
```
Jongy, line 511:
```suggestion
            logger.error(f"Process {process.pid} running unsupported Java version, skipping... (disable --java-safemode to profile it anyway")
```
or something like that

Jongy (changes requested):
.

Jongy (dismissed):
LGTM. I haven't tried it locally yet, please provide logs of how it looks when the version check fails? i.e by increasing the minimum version.

Jongy:
We can add simple automatic tests that patch the min version and check for the matching error in the log, similarly to https://github.com/Granulate/gprofiler/pull/212/commits/a15735d93232c6a6225a5ae8890e473114923897

Jongy (dismissed):

Jongy (dismissed):

Jongy (approved):

Jongy (approved):

PR #218, opened 2021-11-21
java: not profiling processes that were injected with another async-profiler

IlayRosenberg:
## Description

When using `--java-safemode`, java processes that have async-profiler injected to them _not_ from gprofiler will not be profiled

## How Has This Been Tested?

Added a pytest test

## Checklist:

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -162,7 +162,6 @@ class AsyncProfiledProcess:
             f"async-profiler-{get_ap_version()}",
             "musl" if self._is_musl() else "glibc",
         )
-        print(self._ap_dir)
         self._ap_dir_host = resolve_proc_root_links(self._process_root, self._ap_dir)
 
         self._libap_path_host = os.path.join(self._ap_dir_host, "libasyncProfiler.so")
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -22,9 +22,9 @@ from gprofiler.merge import parse_one_collapsed
 from gprofiler.profilers.profiler_base import ProcessProfilerBase
 from gprofiler.profilers.registry import ProfilerArgument, register_profiler
 from gprofiler.utils import (
-    TEMPORARY_STORAGE_PATH,
     get_mnt_ns_ancestor,
     get_process_nspid,
+    get_temporary_storage_path,
     is_process_running,
     pgrep_maps,
     process_comm,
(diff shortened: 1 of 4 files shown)


gprofiler/profilers/java.py
```diff
@@ -284,6 +288,13 @@ def _check_disk_requirements(self) -> None:
                 f" required (on path: {self._output_path_host!r}"
             )
 
+    def _check_async_profiler_not_loaded(self) -> None:
+        if not self._java_safemode:
+            return
+        for mmap in self.process.memory_maps():
+            if "libasyncProfiler.so" in mmap.path and mmap.path != self._libap_path_process:
```
Jongy, line 295: Over different gProfiler versions, async-profiler version may change (and we include the path in the filename to have the DSO re-loaded). I suggest you check for gProfiler's prefix? like `/tmp/gproifler_tmp/...`?

tests/test_java.py
```diff
@@ -221,3 +226,37 @@ def sap_and_crash(self, *args, **kwargs):
...
+            shutil.copy(ap_proc._libap_path_process, copied_libap_path)
+            os.chmod(copied_libap_path, 0o777)
+            cmd = [
+                jattach_path(),
+                str(process.pid),
+                "load",
+                remove_prefix(copied_libap_path, ap_proc._process_root),
+                "true",
+            ]
+            run_process(cmd)
```
Jongy, line 260: Make sure that AP was loaded from your path after this command?

tests/test_java.py
```diff
@@ -228,7 +225,23 @@ def sap_and_crash(self, *args, **kwargs):
     assert "memory_usage_in_bytes:" in caplog.text
 
 
-def test_already_loaded_ap_profiling_failure(tmp_path, caplog, application_pid) -> None:
+def test_already_loaded_ap_profiling_failure(tmp_path, monkeypatch, caplog, application_pid) -> None:
+    with monkeypatch.context() as m:
+        m.setattr("gprofiler.profilers.java.get_temporary_storage_path", lambda: "/tmp/fake_gprofiler_tmp")
+        with JavaProfiler(
+            1,
```
Jongy, line 232: Can keep the args in some dict and expand it twice so you don't copy them all over

gprofiler/profilers/java.py
```diff
@@ -284,6 +289,13 @@ def _check_disk_requirements(self) -> None:
                 f" required (on path: {self._output_path_host!r}"
             )
 
+    def _check_async_profiler_not_loaded(self) -> None:
+        if not self._java_safemode:
+            return
+        for mmap in self.process.memory_maps():
+            if "libasyncProfiler.so" in mmap.path and not mmap.path.startswith(get_temporary_storage_path()):
+                raise Exception(f"Non-Gprofiler Async-profiler is already loaded to the target process: {mmap.path!r}")
```
Jongy, line 297:
```suggestion
                raise Exception(f"Non-gProfiler async-profiler is already loaded to the target process: {mmap.path!r}")
```

Jongy: Also note "disable --java-safemode to bypass this"

tests/test_java.py
IlayRosenberg: Tried but then type checking failed

Jongy (approved):
LGTM :(

PR #220, opened 2021-11-22
java: Disabling java profiling on hotspot error file detection

IlayRosenberg:
## Description

When a hotspot error file is detected for a profiled process, disable current and future profiling of all java processes

## How Has This Been Tested?

Added pytest tests

## Checklist:

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -538,10 +538,10 @@ class JavaProfiler(ProcessProfilerBase):
         if self._java_safemode:
             logger.debug("Java safemode enabled")
 
-    def _disable_profiling(self):
-        logger.error("Java profiling has been disabled, avoiding from profiling any java process")
-        JavaProfiler._should_profile = False
-        self._stop_event.set()
+    @classmethod
+    def _disable_profiling(cls):
+        logger.error("Java profiling has been disabled, will avoid profiling any new java process")
+        cls._should_profile = False
 
     def _is_jvm_type_supported(self, java_version_cmd_output: str) -> bool:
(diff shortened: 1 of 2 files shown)


gprofiler/profilers/java.py
```diff
@@ -537,6 +538,11 @@ def __init__(
         if self._java_safemode:
             logger.debug("Java safemode enabled")
 
+    def _disable_profiling(self):
+        logger.error("Java profiling has been disabled, avoiding from profiling any java process")
+        JavaProfiler._should_profile = False
+        self._stop_event.set()
```
Jongy, line 544: Don't need to stop gProfiler itself

gprofiler/profilers/java.py
```diff
@@ -642,6 +648,10 @@ def _is_profiling_supported(self, process: Process) -> bool:
         return True
 
     def _profile_process(self, process: Process) -> Optional[StackToSampleCount]:
+        if not JavaProfiler._should_profile:
```
Jongy, line 651: Can instead fix `_select_processes_to_profile`... to return an empty list.

Jongy, line 651: You can also access it via `self` (if attr is not found on `self`, it looks up the attr on the class)

tests/test_java.py
```diff
@@ -214,6 +214,8 @@ def sap_and_crash(self, *args, **kwargs):
         return result
 
     monkeypatch.setattr(AsyncProfiledProcess, "start_async_profiler", sap_and_crash)
+    # To make sure it is reverted to True (the original value) after the test
```
Jongy, line 217: lol

Jongy (changes requested):
small

Jongy (approved):

PR #221, opened 2021-11-23
gProfiler 1.2.8

Jongy:
Waiting for:
* https://github.com/Granulate/gprofiler/pull/208
* https://github.com/Granulate/gprofiler/pull/219

PR #223, opened 2021-11-25
tests: Skip PyPerf test on centos:6

Jongy:
## Description
It fails with the following error:

	[2021-11-23 11:10:14,538] DEBUG: gprofiler.profilers.python: eBPF profiler error: [Errno 2] No such file or directory: 'mount': 'mount'

Cherry-picked from https://github.com/Granulate/gprofiler/pull/221.

### DIFF
--- a/tests/test_executable.py
+++ b/tests/test_executable.py
@@ -32,6 +32,10 @@ def test_from_executable(
     _ = application_pid  # Fixture only used for running the application.
 
     if exec_container_image is not None:
+        if "centos:6" in exec_container_image.tags and any("pyperf" in flag for flag in profiler_flags):
+            # don't run PyPerf on the centos:6 image, it fails. And in any case PyPerf can't run on centos:6.
+            pytest.skip("PyPerf test on centos:6")
+
         gprofiler_inner_dir = Path("/app")
         inner_output_dir = Path("/app/output")
         cwd = Path(os.getenv("GITHUB_WORKSPACE", os.getcwd()))


tests/test_executable.py
```diff
@@ -32,6 +32,10 @@ def test_from_executable(
     _ = application_pid  # Fixture only used for running the application.
 
     if exec_container_image is not None:
+        if "centos:6" in exec_container_image.tags and any("pyperf" in flag for flag in profiler_flags):
+            # don't run PyPerf on the centos:6 image, it fails. And in any case PyPerf can't run on centos:6.
```
YishaiZinkin, line 36: Do we know why it fails in this case?

Jongy: Yeah, noted in the commit msg & pr desc

YishaiZinkin: You mean the "no such file" error? I saw that, but why does it happen?

YishaiZinkin: I feel like the condition you added lacks some additional explanations

Jongy: Because there's no `mount` in the container. But I don't care because we don't need to support PyPerf on CentOS 6 since it's a very old kernel.

YishaiZinkin: Hmm kk

Jongy: I can add the explanation for the `mount` but there's no real reason behind it. Had I installed `mount`, it'd work

YishaiZinkin (approved):

tests/test_executable.py
Jongy: Explained in the commit message of the merge

PR #224, opened 2021-11-25
Replace the vertical bar in method descriptors with a semicolon

liorgorb:
## Description
AP replaces semicolons in method descriptors with a vertical bar, so that we can later split frames by the semicolon character. This patch turns the vertical bar back into a semicolon.

## Checklist:

### DIFF
--- a/gprofiler/merge.py
+++ b/gprofiler/merge.py
@@ -45,6 +45,7 @@ def parse_one_collapsed(collapsed: str, add_comm: Optional[str] = None) -> Stack
             continue
         try:
             stack, _, count = line.rpartition(" ")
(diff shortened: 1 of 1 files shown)


gprofiler/merge.py
```diff
@@ -45,6 +45,7 @@ def parse_one_collapsed(collapsed: str, add_comm: Optional[str] = None) -> Stack
             continue
         try:
             stack, _, count = line.rpartition(" ")
+            stack = stack.replace("|", ";")
```
Jongy, line 48: The backend then splits again by `;`. As long as we remain with the basic stackcollapse format, we can only do this translation in the backend.

PR #225, opened 2021-11-26
java: Cleanup & fixes for the next release

Jongy:
* Put log params in extras for easier parsing
* Merge PIDs tracking
* Make sure that kernel messages & proc_events are checked for PIDs (i.e don't remove from the profiled PIDs list immediately)
* Start proc_events in init net NS because it wouldn't work otherwise.

After https://github.com/Granulate/gprofiler/pull/226

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -545,6 +545,7 @@ class JavaProfiler(ProcessProfilerBase):
         self._profiled_pids: Set[int] = set()
         self._pids_to_remove: Set[int] = set()
         self._kernel_messages_provider = DefaultMessagesProvider()
+        self._enabled_proc_events = False
 
     @classmethod
     def _disable_profiling(cls):
@@ -769,10 +770,18 @@ class JavaProfiler(ProcessProfilerBase):
             self._saved_mlock = read_perf_event_mlock_kb()
             write_perf_event_mlock_kb(self._new_perf_event_mlock_kb)
 
-        # needs to run in init net NS
-        run_in_ns(["net"], lambda: proc_events.register_exit_callback(self._proc_exit_callback), 1)
+        try:
+            # needs to run in init net NS
+            run_in_ns(["net"], lambda: proc_events.register_exit_callback(self._proc_exit_callback), 1)
+        except Exception:
+            logger.warning("Failed to enable proc_events listener for exited Java processes", exc_info=True)
+        else:
+            self._enabled_proc_events = True
 
     def stop(self) -> None:
+        if self._enabled_proc_events:
+            proc_events.unregister_exit_callback(self._proc_exit_callback)
+            self._enabled_proc_events = False
         if self._saved_mlock is not None:
             write_perf_event_mlock_kb(self._saved_mlock)
         super().stop()
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -543,6 +543,7 @@ class JavaProfiler(ProcessProfilerBase):
         if self._java_safemode:
             logger.debug("Java safemode enabled")
         self._profiled_pids: Set[int] = set()
+        self._pids_to_remove: Set[int] = set()
         self._kernel_messages_provider = DefaultMessagesProvider()
 
     @classmethod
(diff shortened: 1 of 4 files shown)


gprofiler/profilers/java.py
```diff
@@ -814,11 +805,11 @@ def _handle_new_kernel_messages(self):
...
-        finally:
-            self._prune_profiled_processes()
 
     def snapshot(self):
         try:
             return super().snapshot()
         finally:
             self._handle_new_kernel_messages()
+            self._profiled_pids -= self._pids_to_remove
+            self._pids_to_remove = set()
```
d3dave, line 815:
```suggestion
            self._pids_to_remove.clear()
```

d3dave (changes requested):

gprofiler/profilers/java.py
Jongy: Mhm makes sense

Jongy:
I want to merge https://github.com/Granulate/gprofiler/pull/226 first, and I'll update `run_in_ns` to re-raise the exception in the calling thread, so it gets logged properly and stopa execution (would have caught this bug fixed here).

I also need to test low-privileges execution (specifically non-init PID, how does proc_events react; and /dev/kmsg in non-privileged cases)

gprofiler/profilers/java.py
```diff
@@ -784,28 +769,34 @@ def start(self) -> None:
             self._saved_mlock = read_perf_event_mlock_kb()
             write_perf_event_mlock_kb(self._new_perf_event_mlock_kb)
 
+        # needs to run in init net NS
+        run_in_ns(["net"], lambda: proc_events.register_exit_callback(self._proc_exit_callback), 1)
```
YishaiZinkin, line 773: Did you figure out why?

Jongy: I suspect it's because the kernel listens only on the init net. I don't know netlink good enough to be able to prove/disprove that

Jongy: Planning to check further tho

YishaiZinkin, line 773: As we discussed verbally, we should wrap it in a `try...except` or at least leave a `TODO` as a reminder

Jongy:
```
	/* 
	 * Events are reported with respect to the initial pid
	 * and user namespaces so ignore requestors from
	 * other namespaces.
	 */
	if ((current_user_ns() != &init_user_ns) ||
	    (task_active_pid_ns(current) != &init_pid_ns))
		return;

	/* Can only change if privileged. */
	if (!__netlink_ns_capable(nsp, &init_user_ns, CAP_NET_ADMIN)) {
		err = EPERM;
		goto out;
	}
```

Pfft, it silently ignores requests from non init PID / init user namespaces.

YishaiZinkin: Lol well played. I wonder why they're silent about it

Jongy: Still doesn't explain the net NS thing (which results in ECONNREFUSED). I'll continue checking

Jongy:
```
static int cn_init(void)
{
	struct cn_dev *dev = &cdev;
	struct netlink_kernel_cfg cfg = {
		.groups	= CN_NETLINK_USERS + 0xf,
		.input	= cn_rx_skb,
	};

	dev->nls = netlink_kernel_create(&init_net, NETLINK_CONNECTOR, &cfg);
	if (!dev->nls)
		return -EIO;
```

I'll add a comment

Jongy: Added

Jongy: Done

Jongy:
I checked now in an unprivileged container: `./gprofiler -v -d 5 -o xxx --disable-pidns-check --no-perf` fails due to proc_events starting but it gets an exception in its thread, and then upon stop I get `RuntimeError: Process Events Listener isn't running`. Will be solved after we merge #226 which makes `proc_events` raise in the calling thread if it fails to start.

Jongy:
> I checked now in an unprivileged container: `./gprofiler -v -d 5 -o xxx --disable-pidns-check --no-perf` fails due to proc_events starting but it gets an exception in its thread, and then upon stop I get `RuntimeError: Process Events Listener isn't running`. Will be solved after we merge #226 which makes `proc_events` raise in the calling thread if it fails to start.
> 
> Will check again after we merge #226

Checked now - profiling with low privileges work - I get a warning for the failure in `proc_events,` and a warning for failing to open `/dev/kmsg` (which is not necessarily permissions-related, but as with everything, should be gracefully handled as well)

d3dave (approved):

d3dave (approved):

PR #226, opened 2021-11-26
Move namespace utils to granulate_utils and make granulate_utils a submodule

Jongy:
* Move namespace utils to granulate_utils so they can be reused by other components besides gProfiler.
* Make granulate_utils a submodule
* Upgrade version of granulate_utils (after https://github.com/Granulate/granulate-utils/pull/8)


Jongy:
I'm not sure yet if these errors are real or just flaky :/

d3dave (approved):

PR #233, opened 2021-11-30
java: Filter signal codes only in proc_events callback (from all non-zero exit codes)

Jongy:
## Description
Log only upon signal exits - we don't care about the rest. We care specifically about SIGKILL, SIGTERM which are common sources of trouble.

* SIGKILL can't be caught, so it results in exit code `0x0009` which makes `WIFSIGNALED` return true and the signal code is 9.
* I checked SIGSEGV, SIGILL. They are caught by the JVM and cause generation of hs_err files; then the JVM exits with SIGABRT (`VMError::report_and_die` calls `os::die` which calls `abort`). So the signal code is 6.
* SIGTERM is the last question. The JVM handles it (seen in `strace -f`) and eventually does `exit_group(143)`. I have yet to find where it happens in JVM sources.

I have found this gem, tho:
```
    } else if (WIFSIGNALED(status)) {
       // The child exited because of a signal
       // The best value to return is 0x80 + signal number,
       // because that is what all Unix shells do, and because
       // it allows callers to distinguish between process exit and
       // process death by signal.
       return 0x80 + WTERMSIG(status);
    } else {
```

(in `os::fork_and_exec`, it's not the relevant code ofc, but it is a lead)

## Motivation and Context
Reduce noise by this log message. We only cared about signals in the first place, but delayed the work of actually figuring out how signals look like in the JVM.

## How Has This Been Tested?
Tested signaling a JVM with SIGKILL/SIGSEGV/SIGILL/SIGTERM

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -749,8 +749,7 @@ class JavaProfiler(ProcessProfilerBase):
                 # not a signal - don't report it.
                 return
 
-            if exit_code != 0:
-                logger.warning("async-profiled Java process exited with signal", pid=tid, signal=signo)
+            logger.warning("async-profiled Java process exited with signal", pid=tid, signal=signo)
 
     def _handle_kernel_messages(self, messages):
         for message in messages:
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -8,6 +8,7 @@ import json
 import os
 import re
 import shutil
+import signal
 from pathlib import Path
 from threading import Event
 from typing import List, Optional, Set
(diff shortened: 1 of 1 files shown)


Jongy:
I traced the `exit_group`:
```
VM Thread 199659 [007] 54013.438812: syscalls:sys_enter_exit_group: error_code: 0x0000008f
            7fb17d07d2c6 __GI__exit+0x36 (inlined)
            7fb17cfe0b41 __run_exit_handlers+0x211 (/lib/x86_64-linux-gnu/libc-2.31.so)
            7fb17cfe0bdf __GI_exit+0x1f (inlined)
            7fb17c7e7ed2 vm_direct_exit+0x22 (/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server/libjvm.so)
            7fb17cc8caaa VM_Operation::evaluate+0x4a (/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server/libjvm.so)
            7fb17cc8a6f7 VMThread::evaluate_operation+0x117 (/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server/libjvm.so)
            7fb17cc8ad33 VMThread::loop+0x4f3 (/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server/libjvm.so)
            7fb17cc8b06c VMThread::run+0xbc (/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server/libjvm.so)
            7fb17cacc78f java_start+0x14f (/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server/libjvm.so)
            7fb17cf5b608 start_thread+0xd8 (/lib/x86_64-linux-gnu/libpthread-2.31.so)
            7fb17d0b9292 __GI___clone+0x42 (inlined)
```

This is SIGTERMing the Java. It seems to happen from the Java side. Anyway this assures the point for me - Java exits with 143 deliberately.

YishaiZinkin (approved):

gprofiler/profilers/java.py
```diff
@@ -737,10 +738,20 @@ def _proc_exit_callback(self, tid: int, pid: int, exit_code: int):
...
+            if os.WIFSIGNALED(exit_code):
+                signo = os.WTERMSIG(exit_code)
+            elif exit_code == 0x8F00:
+                # java exits with 143 upon SIGTERM
+                signo = signal.SIGTERM.value
+            else:
+                # not a signal - don't report it.
+                return
+
+            if exit_code != 0:
```
Jongy, line 752: Hmm can be removed I guess

YishaiZinkin (approved):

PR #238, opened 2021-12-02
Add timeouts to run_process() invocations

Jongy:
## Description
Basically - make sure gProfiler can't hang due to external processes it invokes.

## Motivation and Context
Stability

## How Has This Been Tested?
CI


gprofiler/profilers/java.py
```diff
@@ -305,9 +312,10 @@ def _get_stop_cmd(self, with_output: bool) -> List[str]:
         ap_params.append(f"log={self._log_path_process}")
         return self._get_base_cmd() + [",".join(ap_params)]
 
-    def _run_async_profiler(self, cmd: List[str]) -> None:
+    def _run_async_profiler(self, cmd: List[str], timeout: int) -> None:
         try:
-            run_process(cmd)
+            # kill jattach with SIGTERM if it hangs. it will go down
+            run_process(cmd, stop_event=self._stop_event, timeout=timeout, kill_signal=signal.SIGTERM)
```
adi-benz, line 318: There's logs cleanup code in the `except CalledProcessError` clause, do we want this as well?

adi-benz, line 318: What happens if this really gets killed by a timeout / stop event? Should we call `stop_async_profiler()`?

I see `JattachException` doesn't do that even if it detects async profiler is loaded...

gprofiler/profilers/java.py
```diff
@@ -128,15 +128,22 @@ class AsyncProfiledProcess:
     OUTPUT_FORMAT = "collapsed"
     OUTPUTS_MODE = 0o622  # readable by root, writable by all
 
+    # timeouts in seconds
+    _FDTRANSFER_TIMEOUT = 3
+    _START_AP_TIMEOUT = 3
```
adi-benz, line 133: Why so low? jattach has a 6 seconds timeout on its [own](https://github.com/jvm-profiling-tools/async-profiler/blob/4a77d68bcb2e998d5a13f9ca80ab11803c3c12ec/src/jattach/jattach_hotspot.c#L69), and this timeout is for extreme cases anyway, so why not specify really long timeouts? Like 15 seconds? It will reduce the chance for false-negatives.

(same for `_STOP_AP_TIMEOUT`)

adi-benz (changes requested):

gprofiler/profilers/java.py
```diff
@@ -325,7 +333,12 @@ def _run_fdtransfer(self) -> None:
         """
         Start fdtransfer; it will fork & exit once ready, so we can continue with jattach.
         """
-        run_process([fdtransfer_path(), str(self.process.pid)], communicate=False)
+        run_process(
```
adi-benz, line 336: OT: When running gprofiler as an executable, do we make sure not to leave processes like these running?

Jongy: Generally, gProfiler waits for all executed profilers (e.g py-spy) before it exits. Errors while waiting & bugs in the code are what this PR addresses (and possibly other processes like this one, which were invoked with no timeout at all).

gprofiler/profilers/python.py
```diff
@@ -196,6 +196,7 @@ class PythonEbpfProfiler(ProfilerBase):
     dump_signal = signal.SIGUSR2
     dump_timeout = 5  # seconds
     poll_timeout = 10  # seconds
+    get_offsets_timeout = 5  # seconds
```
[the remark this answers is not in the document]
Jongy: Done for all

gprofiler/profilers/java.py
Jongy: Selected arbitrarily, will increase

gprofiler/profilers/java.py
Jongy: You mean, the code checking for `_log_path_host`? No reason not to have it (at this point, we don't know if the blocker is jattach itself, or something in AP).

Jongy: Or you meant - we want that code to run for timeouts? Yes, we do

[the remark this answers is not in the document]
Jongy: Hmm, I suppose that what we actually want is to have the timeout case still raise `CalledProcessError` in that case. Will check if we can make this change wide, or only for relevant sites (with an argument)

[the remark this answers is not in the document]
Jongy: Only used here:
```
            except TimeoutExpired:
                logger.error(f"Profiling with py-spy timed out on process {process.pid}")
```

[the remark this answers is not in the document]
Jongy: I think that the case you have suggested is irrelevant - if jattach timed out, there's no much sense in checking the logs for the "ap already loaded" case. I do want the stderr/stdout to get logged, tho, so will fix the code according to that.

Jongy: Re-starting AP is handled in `_profile_ap_process`. An event of timeout is something unknown and I don't think we should run any additional operations on the target Java process

Jongy: After a bit of thinking: not needed - see https://github.com/Granulate/gprofiler/pull/238#discussion_r790306959

adi-benz: The error handling for `CalledProcessError` handles AP log as well, when raising `JattachException` the `__str__` also includes AP's log contents, I think it will be useful to have AP's log in the case of timeout as well

adi-benz (changes requested):

gprofiler/profilers/java.py
```diff
@@ -128,16 +129,23 @@ class AsyncProfiledProcess:
     OUTPUT_FORMAT = "collapsed"
     OUTPUTS_MODE = 0o622  # readable by root, writable by all
 
+    # timeouts in seconds
+    _FDTRANSFER_TIMEOUT = 10
+    _START_AP_TIMEOUT = 10  # higher than jattach's timeout
+    _STOP_AP_TIMEOUT = 10  # longer as it does more work
```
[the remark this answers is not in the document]
Jongy: Now indeed, tho `stop` really does more work

gprofiler/profilers/java.py
Jongy: Hmm. K, I'll make the conversion: I'll extend `CalledProcessError` into `CalledProcessTimeout` which will contain the same fields + the timeout.

In case of `ProcessStoppedException` - we don't want anything.

adi-benz: Sounds good, but `CalledProcessTimeoutError ` (Error suffix?)

Jongy: Okie

Jongy: Tested with this diff:
```
diff --git a/gprofiler/profilers/python.py b/gprofiler/profilers/python.py
index 6ca2c2c..db15c53 100644
--- a/gprofiler/profilers/python.py
+++ b/gprofiler/profilers/python.py
@@ -85,7 +85,7 @@ def _add_versions_to_stacks(
 class PySpyProfiler(ProcessProfilerBase):
     MAX_FREQUENCY = 50
     _BLACKLISTED_PYTHON_PROCS = ["unattended-upgrades", "networkd-dispatcher", "supervisord", "tuned"]
-    _EXTRA_TIMEOUT = 10  # give py-spy some seconds to run (added to the duration)
+    _EXTRA_TIMEOUT = -20  # give py-spy some seconds to run (added to the duration)
 
     def __init__(
         self,
```

we now get the timed out log, plus the full stderr stdout etc:
```
[2022-01-24 10:22:34,414] ERROR: gprofiler.profilers.python: Profiling with py-spy timed out on process 1750271
[2022-01-24 10:22:34,415] ERROR: gprofiler.profilers.profiler_base: PySpyProfiler: failed to profile process 1750271
  File "/app/gprofiler/profilers/profiler_base.py", line 111, in snapshot
    result = future.result()
    return self.__get_result()
    raise self._exception
  File "/app/gprofiler/profilers/python.py", line 133, in _profile_process
    run_process(
  File "/app/gprofiler/utils.py", line 258, in run_process
    raise reraise_exc
gprofiler.exceptions.CalledProcessTimeoutError: Timed out after 5 seconds
Command '['/app/gprofiler/resources/python/py-spy', 'record', '-r', '50', '-d', '25', '--nonblocking', '--format', 'raw', '-F', '--gil', '--output', '/tmp/gprofiler_tmp/tmp0hw95d08/pyspy.kiUkeeSWwTKiqubZ.1750271.col', '-p', '1750271', '--full-filenames']' died with <Signals.SIGKILL: 9>.
stdout: b'py-spy> Sampling process 50 times a second for 25 seconds. Press Control-C to exit.\n\n'
stderr: b''
```

adi-benz (approved):

adi-benz:
Tests aren't passing

Jongy:
Fixed

adi-benz (approved):

PR #240, opened 2021-12-02
Python: Get modules package info

YishaiZinkin:
## Description
Add package name and version to Python frames.

For example, this symbol:
`scapy.utils.PcapReader.read_packet (/usr/lib/python3.6/site-packages/scapy/utils.py:1347)_[p]`
Will be now presented like this:
`scapy.utils.PcapReader.read_packet (/usr/lib/python3.6/site-packages/scapy/utils.py:1347 (scapy-2.4.5))_[p]`

## Motivation and Context
When looking at a flamegraph, the file name and line number of a module isn't enough to find the exact corresponding function because the package version and sometimes the package name is unknown.

## How Has This Been Tested?

Added a basic test to the automation.

## Checklist:


Jongy (changes requested):
Ran a simple test locally, seems to work alright :) but the real test will be to see for how many libraries it manages to find versions, and I'm too lazy to add entries to the `modules_list` so I'll wait until you incorporate it into gProfiler to test that.

Not a full CR yet

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+        else:
+            info = info[:-1]
+        entry = entry[1:]
+    return str(pathlib.Path(*info, *entry))
+
+
+def _files_from_record(dist: pkg_resources.Distribution) -> Optional[Iterator[str]]:
+    try:
+        text = dist.get_metadata("RECORD")
+    except:
```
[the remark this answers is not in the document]
YishaiZinkin: There are several unrelated exceptions that can be caught here. Since the operation of this module isn't crucial for gProfiler an error here is always non-critical and should only fail the operation of figuring the version, regardless of the exception type.
Maybe I should log unpredicted exceptions though.

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+    pkg_resources used in this function is a part of the setuptools package).
+    This function gathers info for all the packages in the (dist|site)-packages
+    directory in each path, including a list of all the files associated with
+    each package. This list is searched for the given module path.
+    """
+    result = dict.fromkeys(modules_paths)
+    path_to_dist = {}
+
+    for path in modules_paths:
+        if not path.startswith("/"):
```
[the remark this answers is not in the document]
YishaiZinkin: Yes

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+# A little monkey patch to prevent pkg_resources from converting "/proc/{pid}/root/" to "/"
+pkg_resources._normalize_cached = lambda path: path
+
+
+__all__ = ["get_versions"]
+
+
+def _convert_to_proc_root_path(path: str, pid: int) -> str:
+    assert path.startswith("/")
+    return os.path.join(f"/proc/{pid}/root", path)
```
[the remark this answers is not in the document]
YishaiZinkin: Which operations might do this resolving? E.g. does it happen on `read`?

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+        return None
+
+    idx = file_path.rfind("-packages/")
+    if idx == -1 or (not file_path[:idx].endswith("site") and not file_path[:idx].endswit("dist")):
+        return None
+
+    return file_path[:idx] + "-packages/"
+
+
+def _get_metadata(dist: pkg_resources.Distribution) -> email.message.Message:
```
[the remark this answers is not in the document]
YishaiZinkin: Just a dict-like object that metadata files can be parsed into. Actually I could equivalently cast the result to `dict` before returning it, it would've been clearer.

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+__all__ = ["get_versions"]
+
+
+def _convert_to_proc_root_path(path: str, pid: int) -> str:
+    assert path.startswith("/")
+    return os.path.join(f"/proc/{pid}/root", path)
+
+
+def _get_packages_dir(file_path: str) -> Optional[str]:
+    if not file_path.startswith("/"):
```
guybortnikov, line 28: why? `co_filename` can be relative.

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+        return None
+
+    idx = file_path.rfind("-packages/")
+    if idx == -1 or (not file_path[:idx].endswith("site") and not file_path[:idx].endswit("dist")):
```
guybortnikov, line 32:
```suggestion
    if idx == -1 or (not file_path[:idx].endswith("site") and not file_path[:idx].endswith("dist")):
```

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+def _convert_to_proc_root_path(path: str, pid: int) -> str:
+    assert path.startswith("/")
+    return os.path.join(f"/proc/{pid}/root", path)
+
+
+def _get_packages_dir(file_path: str) -> Optional[str]:
+    if not file_path.startswith("/"):
+        return None
+
+    idx = file_path.rfind("-packages/")
```
guybortnikov, line 31: Maybe just `rfind` for `site-packages`/`dist-packages`, it will be a lot more simple and readable

gprofiler/py_package_version.py
Jongy: Not sure that's true, do you have any example?

Anyway I think what @YishaiZinkin tests here is to avoid stuff like `<string>` etc. Could be commented.

gprofiler/py_package_version.py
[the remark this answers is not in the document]
Jongy: `SystemExit` and possibly others, I'm not sure. What you can do is `except Exception` which should not catch those like keyboard interrupt. See https://docs.python.org/3/library/exceptions.html#exception-hierarchy.

For better code readability, I'd suggest:
* At each function / possibly-failing-operation,you catch the specific exceptions it can raise (and place the `try except` on the smallest possible code area)
* The module entrypoint itself should be wrapped with `except Exception` in no way it interferes with gProfiler

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+        # does, fall back to the known canonical name.
+        return metadata.get("Name", dist.project_name)
+    return None
+
+
+def get_versions(modules_paths: List[str], pid: int):
+    """Return a dict with module_path: (package_name, version). If couldn't
+    determine the version the value is None.
+
+    modules_paths must be absoulte. pid is required to access the path via
```
guybortnikov, line 125:
```suggestion
    modules_paths must be absolute. pid is required to access the path via
```

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+
+    if metadata is None:
+        return {}
+
+    feed_parser = email.parser.FeedParser()
+    feed_parser.feed(metadata)
+    return feed_parser.close()
+
+
+def _convert_legacy_entry(entry: Tuple[str, ...], info: Tuple[str, ...]) -> str:
```
guybortnikov, line 56: Please specify from which function from pip this was takes/inspired

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+        if not info or info[-1] == "..":
+            info += ("..",)
+        else:
+            info = info[:-1]
+        entry = entry[1:]
+    return str(pathlib.Path(*info, *entry))
+
+
+def _files_from_record(dist: pkg_resources.Distribution) -> Optional[Iterator[str]]:
+    try:
```
guybortnikov, line 84: Same as https://github.com/Granulate/gprofiler/pull/240/files#r762542684

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+        text = dist.get_metadata("RECORD")
+    except:
+        return None
+    # This extra Path-str cast normalizes entries.
+    return (str(pathlib.Path(row[0])) for row in csv.reader(text.splitlines()))
+
+
+def _files_from_legacy(dist: pkg_resources.Distribution) -> Optional[Iterator[str]]:
+    try:
```
guybortnikov, line 93: Same as https://github.com/Granulate/gprofiler/pull/240/files#r762542684

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+    # TODO: Test
+    metadata = _get_metadata(dist)
+    if metadata:
+        # The metadata should NEVER be missing the Name: key, but if it somehow
+        # does, fall back to the known canonical name.
+        return metadata.get("Name", dist.project_name)
+    return None
+
+
+def get_versions(modules_paths: List[str], pid: int):
```
guybortnikov, line 121:
```suggestion
def get_packages_versions(modules_paths: List[str], pid: int):
```

gprofiler/py_package_version.py
YishaiZinkin: Generally this module converts paths to `/proc/{pid}/root/` paths, so paths here should be absolute. I think it's pretty clear from the code, and described in the docstring of the main function.

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
...
+                files_iter = _files_from_record(dist) or _files_from_legacy(dist)
+                if files_iter is not None:
+                    path_to_dist.update(dict.fromkeys((os.path.join(packages_path, file) for file in files_iter), dist))
+
+        dist = path_to_dist.get(_convert_to_proc_root_path(path, pid))
+        if dist is not None:
+            name = _get_package_name(dist)
+            if name is not None:
+                result[path] = (_get_package_name(dist), dist.version)
+
```
guybortnikov, line 159: Maybe just store package name, version tuple in `path_to_dist` dict values, because now you're parsing the metadata file on each call to `_get_package_name` which does IO that can be reduced easily

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,160 @@
+"""Package name and version for Python files.
```
guybortnikov, line 1: Consider moving this module to `gprofiler/metadata` directory

gprofiler/py_package_version.py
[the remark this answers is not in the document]
guybortnikov: Good idea with the dict, this `email.message.Message` isn't known to anyone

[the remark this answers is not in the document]
guybortnikov: Is this function is taken from pip/setuptools? If so please specify

guybortnikov (changes requested):
.

gprofiler/py_package_version.py
```diff
@@ -0,0 +1,190 @@
...
+            # interfere with gProfiler
+            try:
+                for dist in pkg_resources.find_distributions(packages_path):
+                    files_iter = _files_from_record(dist) or _files_from_legacy(dist)
+                    if files_iter is not None:
+                        path_to_dist.update(
+                            dict.fromkeys((os.path.join(packages_path, file) for file in files_iter), dist)
+                        )
+            except Exception:
+                pass
```
YishaiZinkin, line 180: I think we should log here, but I'm afraid the log will be spammed heavily in case the same error occur for every module.
@Jongy @guybortnikov WDYT?

gprofiler/py_package_version.py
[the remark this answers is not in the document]
YishaiZinkin: Based on a function from there, I'll add documentation for all of them.

gprofiler/py_package_version.py
guybortnikov: I do think we should log here, but we should catch more specific exceptions + don't try more than 1 time on a failed unit (failed unit may be for example a "dist", so if we already failed on a certain "dist" we shouldn't try again)

gprofiler/py_package_version.py
YishaiZinkin: Used `rpartition` instead

gprofiler/py_package_version.py
YishaiZinkin: I'm not sure it's true. I might use `_get_package_name` twice for the same dist, but on the other hand I perform it only for the dists of the relevant modules as opposed to all the many dists in the directory.

gprofiler/py_package_version.py
YishaiZinkin: Any specific exceptions are caught before this `try`, so what we're interested in here are actually the exceptions that aren't expected.
With this implementation units are different between processes, so the log we'll be spammed nonetheless. I suggest we add a log here once we implement the cache.

tests/containers/python/Dockerfile
```diff
@@ -1,6 +1,8 @@
-FROM python:3.6-slim
+FROM python:3.6.15-slim
 
 WORKDIR /app
 ADD lister.py /app
+# Install some package so we can test that its info appears in the collapsed
+RUN pip install setuptools==59.4.0
```
guybortnikov, line 6: I think it'll be better to install something that will be less confusing, maybe some parsing library or something that doesn't do IO/networking

YishaiZinkin: How is this confusing?

tests/containers/python/lister.py
```diff
@@ -16,7 +18,15 @@ def burner():
         pass
 
 
+def getter():
```
guybortnikov, line 21: If you stay with this function name
```suggestion
def platform_infinite_getter():
```

tests/containers/python/Dockerfile
guybortnikov: because we're testing packages info gathering on package that interact with packages.

gprofiler/metadata/py_package_version.py
```diff
@@ -0,0 +1,258 @@
...
+    except ValueError:  # info is not relative to root.
+        return paths
+    if not info_rel.parts:  # info *is* root.
+        return paths
+    return (_convert_legacy_entry(pathlib.Path(p).parts, info_rel.parts) for p in paths)
+
+
+def _get_package_name(dist: pkg_resources.Distribution) -> Optional[str]:
+    """Based on pip._internal.metadata.base.BaseDistribution.raw_name"""
+    # TODO: Test
```
guybortnikov, line 125: ?

tests/containers/python/lister.py
YishaiZinkin: I chose a name that fits the other silly names in the file (`burner` for a function that does nothing and `lister` for a function that calls `os.listdir`)

gprofiler/metadata/py_package_version.py
YishaiZinkin: A reminder to myself to make sure it works as expected. I need to remove this comment.

gprofiler/profilers/python.py
```diff
@@ -36,6 +39,36 @@
...
+                return "({} ({}-{}))".format(module_name_match.group("module_info"), package_info[0], package_info[1])
+            return module_name_match.group()
+
+        new_stack = _module_name_in_stack.sub(_replace_module_name, stack)
+        new_stacks[new_stack] = stacks[stack]
+
+    return new_stacks
+
+
+def _add_versions_to_stacks(process_to_stack_sample_counters: ProcessToStackSampleCounters):
```
guybortnikov, line 64:
```suggestion
def _add_versions_to_stacks(process_to_stack_sample_counters: ProcessToStackSampleCounters) -> ProcessToStackSampleCounters:
```

tests/containers/python/Dockerfile
YishaiZinkin: Oh OK 😅 
I choose this package because it contains a module with a different name in it, which supposed to make the test somewhat harder.

gprofiler/metadata/py_package_version.py
```diff
@@ -0,0 +1,258 @@
...
+            continue
+
+        dist_info = path_to_dist.get(convert_to_proc_root_path(path, pid))
+        if dist_info is not None:
+            name = _get_package_name(dist_info)
+            if name is not None:
+                result[path] = (name, dist_info.version)
+
+    # Don't forget to restore the original implementation in case someone else uses this function
+    pkg_resources._normalize_cached = original__normalize_cache
```
guybortnikov, line 234: Put the whole function in `try... finally` block to make sure this restoration runs no matter what.

gprofiler/profilers/python.py
```diff
@@ -36,6 +39,36 @@
...
+
+def _add_versions_to_process_stacks(pid: int, stacks: StackToSampleCount) -> StackToSampleCount:
+    # TODO: Optimize the whole thing. Specifically, add cache in get_modules_versions
+    new_stacks: StackToSampleCount = Counter()
+    for stack in stacks:
+        modules_paths = (match.group("filename") for match in _module_name_in_stack.finditer(stack))
+        packages_versions = get_modules_versions(modules_paths, pid)
+
+        def _replace_module_name(module_name_match):
+            package_info = packages_versions.get(module_name_match.group("filename"))
```
guybortnikov, line 53: Unpack the tuple here and give indicative names for the items

tests/containers/python/lister.py
guybortnikov: meh ok

tests/test_sanity.py
```diff
@@ -110,6 +112,8 @@ def test_python_ebpf(
         assert_function_in_collapsed(
             "_PyEval_EvalFrameDefault_[pn]", process_collapsed, True
         )  # ensure native user stacks exist
+        assert_function_in_collapsed("setuptools-59.4.0", process_collapsed)  # ensure package info is presented
+        assert_function_in_collapsed("standard-library-3.6.15", process_collapsed)  # ensure Python version is presented
```
guybortnikov, line 116: How can we assure python is being run when the application is executable is `3.6.15`?

gprofiler/metadata/py_package_version.py
```diff
@@ -0,0 +1,258 @@
...
+                return match.group()
+    return None
+
+
+@functools.lru_cache(maxsize=128)
+def _get_python_full_version(pid: int, short_version: str) -> Optional[str]:
+    assert re.match(r"[23]\.\d\d?", short_version)
+
+    bin_file = _get_libpython_path(pid) or f"/proc/{pid}/exe"
+    full_version_string_pattern = re.compile(rb"(?<=\D)" + short_version.encode() + rb"\.\d\d?(?=\x00)")
```
guybortnikov, line 149: Use f-string instead of string concatenation

guybortnikov: How you've decided about this regex pattern? Are we sure it will suffice across all supported python versions?

gprofiler/metadata/py_package_version.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # Try to extract the version string from the binary
+    with open(convert_to_proc_root_path(bin_file, pid), "rb") as f:
+        for line in f.readlines():
+            match = full_version_string_pattern.search(line)
+            if match is not None:
+                return match.group().decode()
```
guybortnikov, line 156: Why splitting lines? Can just match against the whole bytes buffer

gprofiler/metadata/py_package_version.py
```diff
@@ -0,0 +1,258 @@
...
+    standard_lib_pattern = re.compile(r"/python(?P<version>\d\.\d\d?)/(?!.*(site|dist)-packages)")
+    py_version = None
+
+    for path in result:
+        match = standard_lib_pattern.search(path)
+        if match is not None:
+            if py_version is None:
+                try:
+                    py_version = _get_python_full_version(pid, match.group("version"))
+                except OSError:
```
guybortnikov, line 171: Why `_get_python_full_version` might raise `OSError`? Please `except` it there and either raise some more specific exception/return `None`

gprofiler/utils.py
```diff
@@ -446,3 +446,8 @@ def get_kernel_release() -> Tuple[int, int]:
     """Return Linux kernel version as (major, minor) tuple."""
     major_str, minor_str = os.uname().release.split(".", maxsplit=2)[:2]
     return int(major_str), int(minor_str)
+
+
+def convert_to_proc_root_path(path: str, pid: int) -> str:
```
guybortnikov, line 451: Can remove the function and just use: `resolve_host_path` https://github.com/Granulate/granulate-utils/blob/master/granulate_utils/linux/ns.py#L167

guybortnikov: Make sure to update `granulate-utils` submodule

gprofiler/metadata/py_package_version.py
```diff
@@ -0,0 +1,258 @@
...
+    metadata = _get_metadata(dist)
+    if metadata:
+        # The metadata should NEVER be missing the Name: key, but if it somehow
+        # does, fall back to the known canonical name.
+        return metadata.get("Name", dist.project_name)
+    return None
+
+
+def _get_libpython_path(pid: int) -> Optional[str]:
+    libpython_maps_pattern = re.compile(r"(?<=\s)/\S*libpython\S*\.so(\.\S+)?\Z")
```
guybortnikov, line 135: Add explanation/example of the regex pattern -> what you're searching for

gprofiler/metadata/py_package_version.py
```diff
@@ -0,0 +1,258 @@
...
+    # This function resolves symlinks and makes paths absolute for comparison purposes which isn't required
+    # for our usage.
+    if hasattr(pkg_resources, "_normalize_cached"):
+        original__normalize_cache = pkg_resources._normalize_cached
+        pkg_resources._normalize_cached = lambda path: path
+    else:
+        global _warned_no__normalized_cached
+        if not _warned_no__normalized_cached:
+            # Log only once so we don't spam the log
+            logger.warning("Cannot get modules version, pkg_resources has no '_normalize_cached' attribute")
```
guybortnikov, line 204: Why ever should this scenario happen? We're installing a specific version of setuptools which includes the **exact** `pkg_resources` module we're looking for.

guybortnikov (changes requested):
General comments:
1. Make the whole feature a feature flag (optional cli argument to the profiler that allows disabling collection)
2. Make sure the tests are passing consistently.

tests/test_sanity.py
YishaiZinkin: I just wanted to see if by any chance it'll work just like that because I saw we always use the same container image in the tests. I need to fix this as we discussed.

gprofiler/metadata/py_package_version.py
YishaiZinkin: I had in mind that I don't want to load the entire file at once because I was concerned about memory usage. Now that I think about it I'm not sure that `file.read()` actually loads the entire file like that.

gprofiler/metadata/py_package_version.py
YishaiZinkin: Just a safety measure. See [this comment](https://github.com/Granulate/gprofiler/pull/240#discussion_r761832382).

gprofiler/profilers/python.py
```diff
@@ -36,6 +39,36 @@
...
+            if package_info is not None:
+                return "({} ({}-{}))".format(module_name_match.group("module_info"), package_info[0], package_info[1])
```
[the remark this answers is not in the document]
YishaiZinkin: I think it's clearer to have the whole additional information for the frame (i.e. file, line number and package info) wrapped up together. It's merely cosmetic though, so this opinion isn't too strong.

gprofiler/metadata/py_package_version.py
YishaiZinkin: > Use f-string instead of string concatenation

You can't use f-string on bytes. No format either. (And the regex must be `bytes` to match `bytes` as read from the file.)

> How you've decided about this regex pattern? Are we sure it will suffice across all supported python versions?

Any Python release has this string defined:
```
/* Version as a string */
#define PY_VERSION              "3.10.0a7+"
```
This snippet was taken from `master`, but if it was from a release tag the `a7+` portion wouldn't be there. I could write a more complex regex, but decided to be more conservative and only support compilations of Python from tags. 
Generally in this module I prefer to fail in finding the version rather than to present wrong results.

gprofiler/metadata/py_package_version.py
YishaiZinkin: > Why _get_python_full_version might raise OSError?

It opens some files, so any exception that may occur during this process. Mostly file-not-found if the process doesn't exist anymore or permission error if gProfiler isn't privileged.

tests/containers/python/Dockerfile
YishaiZinkin: Changed to `PyYAML` instead

tests/utils.py
```diff
@@ -98,3 +98,21 @@ def assert_function_in_collapsed(function_name: str, collapsed: Mapping[str, int
...
+        if exit_code != 0:
+            return None
+    else:
+        r, w = os.pipe()
+        try:
+            subprocess.run("python --version".split(), check=True, stderr=w, stdout=w)
+            output = os.read(r, 32)
+        finally:
+            os.close(r)
+            os.close(w)
```
guybortnikov, line 115: can just use `subprocess.check_output`

guybortnikov: and add a comment that the python app is running on the host, so that's why it can be extracted that way

tests/utils.py
```diff
@@ -98,3 +98,21 @@ def assert_function_in_collapsed(function_name: str, collapsed: Mapping[str, int
     assert any(
         (function_name in record) for record in collapsed.keys()
     ), f"function {function_name!r} missing in collapsed data!"
+
+
+def get_python_version(application_docker_container: Container):
```
guybortnikov, line 103: can be a fixture, if you like can change

gprofiler/profilers/python.py
```diff
@@ -355,6 +422,7 @@ def __init__(
         stop_event: Event,
         storage_dir: str,
         python_mode: str,
+        python_add_versions: bool,
```
guybortnikov, line 425: IMO
```suggestion
        python_add_package_versions: bool,
```
is better

gprofiler/profilers/python.py
```diff
@@ -385,10 +455,11 @@ def _create_ebpf_profiler(
         duration: int,
         stop_event: Event,
         storage_dir: str,
+        add_versions: bool,
```
guybortnikov, line 458: same as https://github.com/Granulate/gprofiler/pull/240/files#r767750244

gprofiler/profilers/python.py
YishaiZinkin: I chose this name because it's not necessarily packages but also built-in modules. I think this name is more accurate.

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,266 @@
...
+    except Exception:
+        pass
+    finally:
+        # Don't forget to restore the original implementation in case someone else uses this function
+        pkg_resources._normalize_cached = original__normalize_cache
+
+    return result
+
+
+def get_modules_versions(modules_paths: Iterator[str], process: Process):
```
guybortnikov, line 246:
```suggestion
def get_modules_package_versions(modules_paths: Iterator[str], process: Process):
```

YishaiZinkin: [Same argument](https://github.com/Granulate/gprofiler/pull/240#discussion_r767754631)

tests/utils.py
YishaiZinkin: > and add a comment that the python app is running on the host, so that's why it can be extracted that way

Pretty self explanatory, but OK

tests/containers/python/Dockerfile
guybortnikov: noice

gprofiler/profilers/python.py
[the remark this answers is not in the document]
guybortnikov: can you provide please an example of a frame after the addition of version info?

tests/utils.py
YishaiZinkin: Good idea

gprofiler/profilers/python.py
guybortnikov: meh ok

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,266 @@
...
+    with f:
+        for line in f.readlines():
+            match = libpython_maps_pattern.search(line.strip())
+            if match is not None:
+                return match.group()
+    return None
+
+
+@functools.lru_cache(maxsize=128)
+def _get_python_full_version(process: Process, short_version: str) -> Optional[str]:
```
guybortnikov, line 152: Can't we remove the requirement of `short_version` which gets derived from the executable path, many times we'll profile python applications that don't have the python version in the path, for example `/usr/local/bin/gunicorn`

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,266 @@
...
+        files_iter = _files_from_record(dist) or _files_from_legacy(dist)
+        if files_iter is not None:
+            path_to_dist.update(dict.fromkeys((os.path.join(packages_path, file) for file in files_iter), dist))
+    return path_to_dist
+
+
+_warned_no__normalized_cached = False
+
+
+def _get_packages_versions(result: Dict[str, Optional[Tuple[str, str]]], process: Process):
```
guybortnikov, line 202: Any reason for getting the result as an argument and also returning it?
I don't see any reason getting it as an argument, you can just return it as a result and then merge all the results (the cost of this one is negligible

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,266 @@
...
+                # This module is (probably) not part of a package
+                continue
+            packages_path = resolve_host_path(process, packages_path)
+            paths_to_dists = _get_dists_files(packages_path)
+            dist_info = paths_to_dists.get(resolve_host_path(process, path))
+            if dist_info is not None:
+                name = _get_package_name(dist_info)
+                if name is not None:
+                    result[path] = (name, dist_info.version)
+    except Exception:
```
guybortnikov, line 237: log it? I know we don't want to spam the log but on the other hand we do want to monitor and get this kind of errors.

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,266 @@
...
+        for line in f.readlines():
+            match = full_version_string_pattern.search(line)
+            if match is not None:
+                return match.group().decode()
+    return None
+
+
+def _get_standard_libs_version(result: Dict[str, Optional[Tuple[str, str]]], process: Process):
+    # Standard library modules are identified by being under a pythonx.y dir and *not* under site/dist-packages
+    standard_lib_pattern = re.compile(r"/python(?P<version>\d\.\d\d?)/(?!.*(site|dist)-packages)")
```
guybortnikov, line 174: Extract to const out of function

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,266 @@
...
+            if py_version is None:
+                py_version = _get_python_full_version(process, match.group("version"))
+                if py_version is None:
+                    # No need to continue trying if we failed
+                    return None
+            result[path] = ("standard-library", py_version)  # type: ignore
+
+
+@functools.lru_cache(maxsize=128)
+def _get_dists_files(packages_path: str) -> Dict[str, pkg_resources.Distribution]:
```
guybortnikov, line 189: Why aren't we caching against the `Process` object here?

tests/test_sanity.py
```diff
@@ -49,10 +49,14 @@ def test_pyspy(
...
     assert_collapsed,
+    python_version: Optional[str],
 ) -> None:
-    with PySpyProfiler(1000, 1, Event(), str(tmp_path)) as profiler:
+    with PySpyProfiler(1000, 3, Event(), str(tmp_path), True) as profiler:
         process_collapsed = profiler.snapshot().get(application_pid)
         assert_collapsed(process_collapsed, check_comm=True)
+        assert_function_in_collapsed("PyYAML-6.0", process_collapsed)  # Ensure package info is presented
+        # Ensure Python version is presented
+        assert_function_in_collapsed(f"standard-library-{python_version}", process_collapsed)
```
guybortnikov, line 59: assert `python_version` is not None before?

gprofiler/metadata/py_module_version.py
YishaiZinkin: We can. I did it because searching for e.g. "3.4.x" in the binary is more precise than searching for "[23].x.y", but I don't actually believe we'll find any string that looks like that in the binary.
I'll change it.

gprofiler/metadata/py_module_version.py
YishaiZinkin: Mistake, should be `Process` + `packages_path` in the process' namespace

gprofiler/metadata/py_package_version.py
guybortnikov: I see that, have you checked this pattern isn't changed between python versions?

gprofiler/metadata/py_module_version.py
YishaiZinkin: I want to do it once per namespace, hence https://github.com/Granulate/granulate-utils/pull/17

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,274 @@
...
+    except OSError:
+        return None
+
+    with f:
+        for line in f.readlines():
+            match = _LIBPYTHON_MAPS_PATTERN.search(line.strip())
+            if match is not None:
+                return match.group()
+    return None
+
```
guybortnikov, line 150: Can just use psutil `memory_maps`, as @Jongy did in https://github.com/Granulate/gprofiler/blob/719adaf7d9ab027fa2c705622420900e0c503612/gprofiler/profilers/java.py#L242

gprofiler/metadata/py_module_version.py
guybortnikov: I see but the per namespace can wait to a later stage and we don't want to miss these errors

gprofiler/metadata/py_package_version.py
YishaiZinkin: Yes

gprofiler/metadata/py_package_version.py
YishaiZinkin: I'm aware that it's not a textual file. `readlines` is used here as an arbitrary way to split the file in a way that ensures that we won't cut `PY_VERSION`.
I'll change it though.

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,274 @@
...
+            if match is not None:
+                return match.group().decode()
+    return None
+
+
+# Standard library modules are identified by being under a pythonx.y dir and *not* under site/dist-packages
+_STANDARD_LIB_PATTERN = re.compile(r"/python\d\.\d\d?/(?!.*(site|dist)-packages)")
+
+
+def _get_standard_libs_version(result: Dict[str, Optional[Tuple[str, str]]], process: Process):
```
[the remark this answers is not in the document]
YishaiZinkin: It was wrapped with `except OSError` but you requested me to remove it in a previous comment.
AFAIK the only thing that may `raise` under this function are the calls to `open` which are appropriately wrapped with `try...except` so I don't think there's a risk here, but I'll restore it anyway.

gprofiler/profilers/python.py
```diff
@@ -36,12 +39,60 @@
...
     _BLACKLISTED_PYTHON_PROCS = ["unattended-upgrades", "networkd-dispatcher", "supervisord", "tuned"]
     _EXTRA_TIMEOUT = 10  # give py-spy some seconds to run (added to the duration)
 
+    def __init__(
+        self,
+        frequency: int,
+        duration: int,
+        stop_event: Optional[Event],
+        storage_dir: str,
+        add_versions: bool,
```
d3dave, line 91: Please make this a kwarg-only. It is unclear when it's only a `bool` specified without the name.

gprofiler/metadata/py_module_version.py
d3dave: The name `result` is very confusing in this context. 
1. It'd be better named `packages_versions`.
2. The function would be better named `populate_packages_versions` to signify that it modifies the parameter and does not return a value.

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,282 @@
...
+        return result
+
+    # We use the process only for its /proc/[pid]/root, so it's more efficient for caching purposes to use the
+    # ns ancestor
+    ancestor = get_mnt_ns_ancestor(process)
+
+    # Make sure to catch any exception. If something goes wrong just don't get the version, we shouldn't
+    # interfere with gProfiler
+    try:
+        for path in result:
```
d3dave, line 243:
```suggestion
        for module_path in result:
```

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,282 @@
...
+
+@functools.lru_cache(maxsize=128)
+def _get_packages_files(process: Process, packages_path: str) -> Dict[str, Tuple[str, str]]:
+    """Return a dict of filename: (package_name, package_version) for the packages in packages_path"""
+    # Transform packages_path to be relative to /proc/[pid]/root/
+    packages_host_path = resolve_host_path(process, packages_path)
+
+    path_to_package_info = {}
+    for dist in pkg_resources.find_distributions(packages_host_path):
+        files_iter = _files_from_record(dist) or _files_from_legacy(dist)
```
d3dave, line 204: > `_files_from_legacy(dist)`

Do we know of an example package that is using legacy format?

gprofiler/profilers/python.py
d3dave: The `add_versions`

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,282 @@
...
+                        (os.path.join(packages_path, file) for file in files_iter), (pacakge_name, dist.version)
+                    )
+                )
+    return path_to_package_info
+
+
+_warned_no__normalized_cached = False
+
+
+def _get_packages_versions(result: Dict[str, Optional[Tuple[str, str]]], process: Process):
```
d3dave, line 219: The name `result` is very confusing in this context. 
1. It'd be better named `packages_versions`.
2. The function would be better named `populate_packages_versions` to signify that it modifies the parameter and does not return a value.

gprofiler/metadata/py_module_version.py
YishaiZinkin: I'm not sure actually if I saw one or not. I definitely saw packages without the newer format info.
This code is based on pip's `show` implementation so if they choose to support it there is no reason for us not to.

gprofiler/profilers/python.py
```diff
@@ -36,12 +39,60 @@
...
+) -> ProcessToStackSampleCounters:
+    result: ProcessToStackSampleCounters = defaultdict(Counter)
+
+    for pid, stack_to_sample_count in process_to_stack_sample_counters.items():
+        try:
+            process = Process(pid)
+        except NoSuchProcess:
+            # The process doesn't exist anymore so we can't analyze versions
+            continue
+        result[pid] = _add_versions_to_process_stacks(process, stack_to_sample_count)
```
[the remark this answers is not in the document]
YishaiZinkin: I put the `try..except` in `get_modules_versions`. I think it's more accurate

gprofiler/metadata/py_module_version.py
d3dave: Well now I'm curious why not just call into `pip`s implementation?

YishaiZinkin: That was the approach at first, but `pip show` requires the name of the package so it's kind of the opposite of what we need, and `pip freeze` doesn't provide the information we want

d3dave: Ah ofc because our mapping is from module file to package and not a listing of packages. Ok 👍

gprofiler/profilers/python.py
```diff
@@ -332,14 +393,22 @@ def stop(self):
     "pyspy (always use py-spy), pyperf (always use PyPerf, and avoid py-spy even if it fails)"
     " or disabled (no runtime profilers for Python).",
     profiler_arguments=[
+        ProfilerArgument(
+            "--no-python-versions",
+            dest="python_add_versions",
+            action="store_false",
```
guybortnikov, line 399: IMO better to have a default here

guybortnikov (changes requested):
Looks really good, also tested locally and the overhead is invisible in the flamegraph! 
Let's just take care of https://github.com/Granulate/gprofiler/pull/240/files#r767906364, can just log everytime as I think the scenario is very rare and I don't want to miss these errors (I don't think it will be too spammy)
Also left another small comment so we'll just follow the convention with the arguments defaults.

Really really cool!

gprofiler/metadata/py_module_version.py
```diff
@@ -0,0 +1,276 @@
...
+            if package_info is not None:
+                packages_versions[module_path] = package_info
+    except Exception:
+        pass
+    finally:
+        # Don't forget to restore the original implementation in case someone else uses this function
+        pkg_resources._normalize_cached = original__normalize_cache
+
+
+def get_modules_versions(modules_paths: Iterator[str], process: Process):
```
d3dave, line 256: Return type annotation

d3dave (changes requested):

gprofiler/profilers/python.py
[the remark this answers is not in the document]
d3dave: Umm... I don't see a `try..except` in `get_modules_versions`...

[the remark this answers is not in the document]
YishaiZinkin: I forgot to push yesterday night, MB

d3dave (approved):

d3dave:
LGTM 👍

guybortnikov (approved):

PR #242, opened 2021-12-06
Feature/Application Names

guybortnikov:
## Description
This feature adds a signing on applications (based on simple heuristics) mechanism, this will allow further filtering based on the found application names.

## Motivation and Context
We'll like to increase the metadata collected by the gProfiler in order to provide more information to the Performance Studio for smarter and more efficient filtering.

## How Has This Been Tested?

**TODO - test**

## Checklist:


gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+
+from granulate_utils.linux.ns import resolve_host_path
+from psutil import NoSuchProcess, Process
+
+from gprofiler.log import get_logger_adapter
+
+logger = get_logger_adapter(__name__)
+
+
+class ApplicationSeparator(metaclass=ABCMeta):
```
YishaiZinkin, line 15: I don't understand the terminology here. How does it separates applications?

YishaiZinkin, line 15: Are theses classes part of the API of the module? If not they should be prefixed with `_`. It'll make it much easier to understand how this module is supposed to be used.
Also, it may fit better under `gprofiler/metadata/`.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+    for separator in APPLICATION_SEPARATORS:
+        try:
+            if separator.is_supported(process):
+                app_name = separator.get_application_name(process)
+                prefix = separator.app_prefix
+                return f"{prefix}-{app_name}"
+
+        except Exception:
+            logger.exception(
+                f"Application separator {separator} raised an exception while matching against process" f"{process}"
```
YishaiZinkin, line 189:
```suggestion
                f"Application separator {separator} raised an exception while matching against process " f"{process}"
```

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+    def get_application_name(self, process: Process) -> Optional[str]:
+        module_arg = get_positional_arg(process.cmdline(), "-m")
+        if module_arg is not None:
+            return module_arg
+
+        module_filename = process.cmdline()[1][:-3]  # Strip the ".py" (checked before)
+        return append_python_module_to_proc_wd(process, module_filename)
+
+
+# Please note that the order matter, so when adding new separators pay attention to the order.
```
YishaiZinkin, line 164: I think it's worth to also describe how does it matter

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+    def get_application_name(self, process: Process) -> Optional[str]:
+        pass
+
+    @property
+    @abstractmethod
+    def app_prefix(self) -> str:
+        pass
+
+
+IP_PORT_RE = re.compile(r"^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}(?:\:\d{2,5})?$")  # Matches against (ip(:port))
```
YishaiZinkin, line 30: Like I commented above, this variable should be prefixed with `_`

gprofiler/profilers/java.py
```diff
@@ -186,7 +186,7 @@ def locate_hotspot_error_file(self) -> Optional[str]:
     @functools.lru_cache(maxsize=1)
     def _is_musl(self) -> bool:
         # Is target process musl-based?
-        return any("ld-musl" in m.path for m in self.process.memory_maps())
+        return any("ld-musl" in m.path for m in self.process.memory_maps())  # type: ignore
```
YishaiZinkin, line 189: Why?

gprofiler/merge.py
```diff
@@ -238,6 +239,7 @@ def concatenate_profiles(
     add_container_names: bool,
     metadata: Metadata,
     metrics: Metrics,
+    add_application_names: bool = True,
```
YishaiZinkin, line 242: Did you mean
```suggestion
    add_application_name: bool = True,
```
?

YishaiZinkin: Oh I now see that there is a `add_container_names` parameter. I don't understand this name...

YishaiZinkin, line 242: I would've place it next to `add_container_names` and make it non-default, or vice versa. Just for sake of coherency.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+APPLICATION_SEPARATORS = [
+    GunicornApplicationSeparator(),
+    UwsgiApplicationSeparator(),
+    CeleryApplicationSeparator(),
+    PySparkApplicationSeparator(),
+    PythonModuleApplicationSeparator(),
+]
+
+
+def get_application_name(pid: int) -> Optional[str]:
```
YishaiZinkin, line 174: Add doc, including a note regarding the heuristic nature of this function.

YishaiZinkin:
Can you add in the description an example to how it'll look like?

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+class PySparkApplicationSeparator(ApplicationSeparator):
+    @property
+    def app_prefix(self) -> str:
+        return "pyspark"
+
+    def is_supported(self, process: Process) -> bool:
+        # We're looking for pythonXX -m pyspark.daemon
+        return process.cmdline() >= 3 and ["-m", "pyspark.daemon"] == process.cmdline()[1:]
```
YishaiZinkin, line 136: Don't we need to check `process.cmdline()[0]` as well?

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        return "generic pyspark.daemon"
+
+
+class PythonModuleApplicationSeparator(ApplicationSeparator):
+    @property
+    def app_prefix(self) -> str:
+        return "generic-python"
```
YishaiZinkin, line 145: Why not just `"Python"`?

YishaiZinkin: When I think about it "python" is added to the first frame anyway, isn't it? Isn't this app prefix redundant?

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+class CeleryApplicationSeparator(ApplicationSeparator):
+    @property
+    def app_prefix(self) -> str:
+        return "celery"
+
+    def is_supported(self, process: Process) -> bool:
+        if process.cmdline()[0] == "celery":
+            return True
+
+        return len(process.cmdline()) >= 3 and ["-m", "celery"] == process.cmdline()[1:]
```
YishaiZinkin, line 118: Same as in [here](https://github.com/Granulate/gprofiler/pull/242/files#r765956354)

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+    @property
+    @abstractmethod
+    def app_prefix(self) -> str:
+        pass
+
+
+IP_PORT_RE = re.compile(r"^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}(?:\:\d{2,5})?$")  # Matches against (ip(:port))
+
+
+def get_positional_arg(args: List[str], pos_arg_name: str) -> Optional[str]:
```
YishaiZinkin, line 33: Hmm I wouldn't call cmd arguments "positional", just args. Maybe call this function `get_cmd_arg_vlaue` and rename `pos_arg_name` to `arg_name`?

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+    if pos_arg_name in args:
+        return args[args.index(pos_arg_name) + 1]
+
+    return None
+
+
+def append_python_module_to_proc_wd(process: Process, module_name: str) -> str:
+    return process.cwd().replace("/", ".") + module_name
```
YishaiZinkin, line 41: `process.cwd()` returns an absolute path, so for a module named "my-module", if I ran it from `/home/myself` I would get `'.home.myselfmy-module'`. It doesn't seem right...

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+    @property
+    def app_prefix(self) -> str:
+        return "pyspark"
+
+    def is_supported(self, process: Process) -> bool:
+        # We're looking for pythonXX -m pyspark.daemon
+        return process.cmdline() >= 3 and ["-m", "pyspark.daemon"] == process.cmdline()[1:]
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        return "generic pyspark.daemon"
```
YishaiZinkin, line 139: Pyspark applications don't have their own name? If they don't I would say the application name should be just "pyspark".  The extra `"generic pyspark.daemon"` is confusing.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+class PythonModuleApplicationSeparator(ApplicationSeparator):
+    @property
+    def app_prefix(self) -> str:
+        return "generic-python"
+
+    def is_supported(self, process: Process) -> bool:
+        if "python" not in process.cmdline()[0]:
```
YishaiZinkin, line 148: Better to match against `r"^python([23](\.\d{1,2})?)?$"`

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+            return False
+
+        return (len(process.cmdline()) > 3 and process.cmdline()[1] == "-m") or (
+            len(process.cmdline()) >= 2 and process.cmdline()[1].endswith(".py")
```
YishaiZinkin, line 152: The ".py" file can be passed anywhere in the command line. I think you can just check if `any(arg.endswith(".py") for arg in process.cmdline()[1:])`

gprofiler/application_separators.py
YishaiZinkin, line 148: Save `process.cmdline()` in a member var instead of calling this function seven times.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+        )
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        module_arg = get_positional_arg(process.cmdline(), "-m")
+        if module_arg is not None:
+            return module_arg
+
+        module_filename = process.cmdline()[1][:-3]  # Strip the ".py" (checked before)
+        return append_python_module_to_proc_wd(process, module_filename)
```
YishaiZinkin, line 161: Running `python3.8 /usr/local/lib/python3.8/dist-packages/my-module.py` from `/home/myself` would've get `".home.myself/usr/local/lib/python3.8/dist-packages/my-module"`. I think `os.path.basename(process.cmdline()[1])[:-3]` would've sufficed here.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+    def is_supported(self, process: Process) -> bool:
+        if process.cmdline()[0] == "celery":
+            return True
+
+        return len(process.cmdline()) >= 3 and ["-m", "celery"] == process.cmdline()[1:]
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        app_name = get_positional_arg(process.cmdline(), "-A") or get_positional_arg(process.cmdline(), "--app")
+        if app_name is None:
+            logger.warning("Couldn't find positional argument -A or --app for application indication")
```
YishaiZinkin, line 123: Might spam the log

gprofiler/application_separators.py
guybortnikov: Can't do it since the api is stateless (`is_supported` might be called on a separator with some process and then on another before `get_application_name` is getting called), the cost of calling `cmdline()` is very small and negligible

gprofiler/application_separators.py
guybortnikov: Each type of application separator separates on its own, the interface is easy, you get a process and return whether you can tell which kind of application it is, and return it.

guybortnikov: They are not a part of the API of the module, but It's really weird to prefix classes with `_` it's common to have classes that are not API of the module without a prefix.

gprofiler/profilers/java.py
guybortnikov: because `process.memory_maps` attribute may be unavailable.

gprofiler/application_separators.py
guybortnikov: I can add but I'm sure that `-m pyspark.daemon` suffice

gprofiler/application_separators.py
guybortnikov: They do have their own name but unfortunately, extracting it is not simple. I will change to "pyspark"

gprofiler/application_separators.py
guybortnikov: nope, since it can be some `/usr/bin/python` without a number, actually it's very common (many applications are being run from just `python` executable -> in most of the cases this executable is a symlink to an executable with some version identifier in it's name)

gprofiler/application_separators.py
guybortnikov: You're right but  it can also be passed in non python application (some wrapper scripts and so..), here I'm checking for running python script in the most simple way as I don't want to get into more complicated scenarios and maybe produce false positives, as this case will cover most of the scenarios (the most common scenarios are running a python script with `-m` or with just python script as argv[1])

gprofiler/application_separators.py
guybortnikov: The idea was to add the cwd to the module name since we can have 2 different applications with the same basename in 2 different dirs.

gprofiler/application_separators.py
YishaiZinkin: I don't think it's weird. Actually, I believe it's the standard (also according to PEP8).
You can search for `class _` in cpython.

gprofiler/application_separators.py
YishaiZinkin: You're right, my bad.
You can still store it in a local variable in this function though.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,193 @@
...
+class PythonModuleApplicationSeparator(ApplicationSeparator):
+    @property
+    def app_prefix(self) -> str:
+        return "generic-python"
+
+    def is_supported(self, process: Process) -> bool:
+        if "python" not in process.cmdline()[0]:
+            return False
+
+        return (len(process.cmdline()) > 3 and process.cmdline()[1] == "-m") or (
```
YishaiZinkin, line 151: Shouldn't it be
```suggestion
        return (len(process.cmdline()) >= 3 and process.cmdline()[1] == "-m") or (
```
?

gprofiler/application_separators.py
YishaiZinkin: This regex will match `python`, `python3` and `python3.10`.
On the other hand, if a client will decide to run an application named "happythongs" for any reason, this condition specifically would match it too.
BTW, the suggested regex won't match against execution of Python with a full path (i.e. if you run `/use/bin/python` directly), but it can be expanded.

gprofiler/application_separators.py
YishaiZinkin: I think it's better to strive to be precise in these checks. Anyway in this case it's worth it even just for clarity.

gprofiler/application_separators.py
guybortnikov: CPython is not fully following PEP8, and PEP8 only talks about **private attributes** the convention of function names starting with an underscore is not a part of the PEP8 convention.

gprofiler/application_separators.py
guybortnikov: It's correct that it returns an absolute path, it had a bug that's why it wasn't seemed right, now I fixed it and the result of the function will be: `home.myself.my-module` which is what I wanted.
The intention of the function is to add the workdir since many applications run as as module `wsgi/app`  which is not indicative at all, but when adding the working directory it does add some indicative data.

gprofiler/application_separators.py
guybortnikov: kk changed

gprofiler/application_separators.py
guybortnikov: You're right but will be not so easy ratelimiting it as it per process and I do want to know about all the cases.

gprofiler/application_separators.py
YishaiZinkin: From [PEP 8](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces):
> Even with \_\_all\_\_ set appropriately, internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore.

gprofiler/profilers/java.py
YishaiZinkin: How didn't it failed the tests until now?
And if it may be unavailable shouldn't we check it?

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+_PYTHON_BIN_RE = re.compile(r"^python([23](\.\d{1,2})?)?$")
+
+
+def _is_python_bin(bin_name: str):
+    return _PYTHON_BIN_RE.match(os.path.basename(bin_name)) is not None
+
+
+class ApplicationSeparator(metaclass=ABCMeta):
+    @abstractmethod
+    def is_supported(self, process: Process) -> bool:
```
Jongy, line 24: Maybe merge it into `get_application_name`? In some of the impls, e.g `PythonModuleApplicationSeparator`, you are making assumptions in `get_application_name` based on what was checked in `is_supported`. Code is better understood if you make the checks then immediately use the new "assumptions".

gprofiler/merge.py
```diff
@@ -250,9 +252,11 @@ def concatenate_profiles(
 
     for pid, stacks in process_profiles.items():
         container_name = _get_container_name(pid, docker_client, add_container_names)
+        application_name = get_application_name(pid) or "" if add_application_names else ""
```
Jongy, line 255: Use consistent naming for the feature across the codebase. Here you have `add_application_names`. Previously I saw `separate_applications`. I'd go with the former.

gprofiler/profilers/java.py
Jongy: @guybortnikov you are missing `types-psutil`.

If it passes in the CI and doesn't pass on your box, problem is with your box.

gprofiler/profilers/java.py
```diff
@@ -239,7 +239,7 @@ def _check_disk_requirements(self) -> None:
     def _check_async_profiler_not_loaded(self) -> None:
         if not self._java_safemode:
             return
-        for mmap in self.process.memory_maps():
+        for mmap in self.process.memory_maps():  # type: ignore
```
Jongy, line 242: https://github.com/Granulate/gprofiler/pull/242/files#r771963620

Jongy:
Please update granulate-utils & open a PR for it

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+    def get_application_name(self, process: Process) -> Optional[str]:
+        wsgi = get_cli_arg_value(process.cmdline(), "-w")
+        if wsgi is not None:
+            return append_python_module_to_proc_wd(process, wsgi)
+
+        wsgi = self._find_wsgi_from_config_file(process)
+        if wsgi is not None:
+            return append_python_module_to_proc_wd(process, wsgi)
+
+        logger.warning("Couldn't find uwsgi wsgi module, both from cmdline and from config file")
```
Jongy, line 124: Not logging the process here. Also, as suggested on the gunicorn class, you should log the commandline arguments so we'll have *something* to debg.

gprofiler/merge.py
YishaiZinkin: I agree with the name suggestion of @Jongy. As I wrote [here](https://github.com/Granulate/gprofiler/pull/242#discussion_r763772060) applications "separation" isn't an indicative terminology. The module finds the name of applications, so something like `AppNameGetter` or `AppNameIdentifier` will more suitable.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+            return True
+
+        return len(process.cmdline()) >= 3 and ["-m", "celery"] == process.cmdline()[1:3]
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        app_name = get_cli_arg_value(process.cmdline(), "-A") or get_cli_arg_value(
+            process.cmdline(), "--app", check_for_equals_arg=True
+        )
+        if app_name is None:
+            logger.warning("Couldn't find positional argument -A or --app for application indication")
```
Jongy, line 144: Same comment about logging the process + its commandline here.

gprofiler/application_separators.py
Jongy: I suggest we start without ratelimiting so we don't miss anything indeed. Once this stabilizes, we can add.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+class PySparkApplicationSeparator(ApplicationSeparator):
+    @property
+    def app_prefix(self) -> str:
+        return "PySpark"
+
+    def is_supported(self, process: Process) -> bool:
+        # We're looking for pythonXX -m pyspark.daemon
+        return (
+            len(process.cmdline()) >= 3
+            and "python" in process.cmdline()
```
Jongy, line 159: It might be e.g `python3.7 -m pyspark.daemon`. The check should be `any(a.startswith("python") for a in process.cmdline())`, or be more specific and use `_is_python_bin`, probably better.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+            and ["-m", "pyspark.daemon"] == process.cmdline()[1:]
+        )
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        return "PySpark"
```
Jongy, line 164: I think it's nicer to maintain lowercase

Jongy: (everywhere)

gprofiler/application_separators.py
Jongy: It's okay to ignore other cases now. However if you can extract this logic (from here, PySpark and possibly others) to some function that will currently check only for `argv[0] == python && argv[1] == -m` for example, and later on we can make it smarter and skip other flags etc.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+
+class JavaJarApplicationSeparator(ApplicationSeparator):
+    def is_supported(self, process: Process) -> bool:
+        return "java" in os.path.basename(process.cmdline()[0]) and "-jar" in process.cmdline()
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        return get_cli_arg_value(process.cmdline(), "-jar")
+
+    @property
+    def app_prefix(self) -> str:
```
Jongy, line 197: `app_prefix` is the first function in your other classes, reorder

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+        return "Java-Jar"
```
Jongy, line 198: Just `java` would do

gprofiler/merge.py
```diff
@@ -250,9 +252,11 @@ def concatenate_profiles(
...
+        prefix = container_name + ";" if add_container_names else "" + application_name
```
Jongy, line 256: Order of operators here is not clear (on the left hand there's `container_name + ";"` that are in the `if`, and o the right side there is `"" + application_name` where you clearly meant that `""` is in the `if` and `application_name` is not. But either both are or both are not. Please fix and add parenthesis

Jongy, line 256: We'll be missing `;` for the `application_name`, because here you don't seem to replace the first frame (which is what I recall we have decided earlier). I think we should just replace the first frame name (i.e `python` `celery` whatever into your new app name)

Jongy (changes requested):
Done reviewing all

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+        return None
+
+
+class UwsgiApplicationSeparator(ApplicationSeparator):
+    @property
+    def app_prefix(self) -> str:
+        return "uwsgi"
+
+    def is_supported(self, process: Process) -> bool:
+        return "uwsgi" in process.cmdline()[0]
```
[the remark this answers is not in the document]
guybortnikov: nope, actually it might be a part of an absolute path `/usr/local/bin/uwsgi` - can do basename but I think this one will suffice

gprofiler/application_separators.py
```diff
@@ -0,0 +1,238 @@
...
+
+    @abstractmethod
+    def get_application_name(self, process: Process) -> Optional[str]:
+        pass
+
+    @property
+    @abstractmethod
+    def app_prefix(self) -> str:
+        pass
+
```
YishaiZinkin, line 35: As we talked yesterday, in many cases one of these functions is redundant so my suggestion is to have a single `get_application_name` method that'll perform the formatting internally

gprofiler/application_separators.py
YishaiZinkin: I don't see a reason not to stick to PEP 8. Conventions are important for readability, especially when they're so widely excepted like PEP 8 and in an open source project

gprofiler/application_separators.py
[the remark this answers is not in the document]
YishaiZinkin: You can still
```suggestion
        return process.cmdline()[0] == "uwsgi" or process.cmdline()[0].endswith("/uwsgi")
```
It's better to have false negatives of not identifying applications and add more cases in the future.

gprofiler/application_separators.py
YishaiZinkin: It's a good point. But I think that `home.myself.my-module` is not the correct way to display the path because it looks like the name of the module, which it isn't so it looks like a bug. I think we should display the module name and the path separately like in Python frames, i.e. `module_name (file_path`).

As a separated point, it's worth checking if the regexes that collapses generated paths to i.e. `\m\` when displayed in the UI work with these changes.

gprofiler/application_separators.py
```diff
@@ -0,0 +1,238 @@
...
+    @property
+    def app_prefix(self) -> str:
+        return "gunicorn"
+
+    def is_supported(self, process: Process) -> bool:
+        # Either gunicorn (for example: /usr/local/bin/gunicorn) or python that runs gunicorn
+        # (For example: /usr/local/bin/python /usr/local/bin/gunicorn)
+        return "gunicorn" in get_cli_arg_by_index(process.cmdline(), 0) or "gunicorn" in get_cli_arg_by_index(
+            process.cmdline(), 0
+        )
```
YishaiZinkin, line 75: It's the same condition in both sides of the `or`

gprofiler/application_separators.py
```diff
@@ -0,0 +1,238 @@
...
+        return None
+
+
+class CeleryApplicationSeparator(ApplicationSeparator):
+    @property
+    def app_prefix(self) -> str:
+        return "celery"
+
+    def is_supported(self, process: Process) -> bool:
+        if get_cli_arg_by_index(process.cmdline(), 0) == "celery":
```
YishaiZinkin, line 140: Can't it be an absolute path?

gprofiler/application_separators.py
```diff
@@ -0,0 +1,238 @@
...
+        try:
+            return config.get("uwsgi", "module")
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            pass
+
+        return None
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        wsgi = get_cli_arg_by_name(process.cmdline(), "-w")
+        if wsgi is not None:
```
YishaiZinkin, line 123: In failure `get_cli_arg_by_name` returns an empty string. It applies to other usages of this functions as well.
IMO it's more common to return `None` in case of failure and therefor clearer and more intuitive than returning `""`. This opinion isn't very strict, but I do think it might save us some bugs in the future.

YishaiZinkin: And BTW `mypy` sucks

gprofiler/application_separators.py
```diff
@@ -0,0 +1,238 @@
...
+class PySparkApplicationSeparator(ApplicationSeparator):
+    @property
+    def app_prefix(self) -> str:
+        return "PySpark"
+
+    def is_supported(self, process: Process) -> bool:
+        # We're looking for pythonXX -m pyspark.daemon
+        return (
+            len(process.cmdline()) >= 3
+            and "python" in process.cmdline()
```
YishaiZinkin, line 165: In the following line you check `["-m", "pyspark.daemon"] == process.cmdline()[1:]` which suggests you meant
```suggestion
            len(process.cmdline()) == 3
            and "python" in process.cmdline()[0]
```

gprofiler/application_separators.py
YishaiZinkin: [This comment](https://github.com/Granulate/gprofiler/pull/242#discussion_r774495062)

gprofiler/application_separators.py
guybortnikov: kk, agreed with @YishaiZinkin fixed it and moved to `metadata` dir

gprofiler/profilers/java.py
guybortnikov: it was the scandal with darwin, fixed...

gprofiler/application_separators.py
guybortnikov: added check

gprofiler/application_separators.py
guybortnikov: the comm maybe not just "python", the app can change it and many apps do.
Changed it to "python"

gprofiler/application_separators.py
guybortnikov: I don't think storing it is crucial, it will make the whole code less clean and I think that's a premature optimization

gprofiler/merge.py
guybortnikov: Changed to `ApplicationIdentifier` I think it's suits good

gprofiler/application_separators.py
guybortnikov: fixed

gprofiler/application_separators.py
guybortnikov: don't relevant anymore

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+        for arg in args:
+            arg_name, _, arg_val = arg.rpartition("=")
+            if arg_val is not None:
+                return arg_val
+
+    return None
+
+
+def append_python_module_to_proc_wd(process: Process, module_name: str) -> str:
+    return f'{process.cwd().replace("/", ".").strip(".")}.{module_name}'
```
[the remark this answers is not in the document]
guybortnikov: kk fixed

gprofiler/application_separators.py
guybortnikov: Agreed, fixed

gprofiler/application_separators.py
guybortnikov: fixed

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+
+class GunicornApplicationSeparator(ApplicationSeparator):
+    @property
+    def app_prefix(self) -> str:
+        return "gunicorn"
+
+    def is_supported(self, process: Process) -> bool:
+        # Either gunicorn (for example: /usr/local/bin/gunicorn) or python that runs gunicorn
+        # (For example: /usr/local/bin/python /usr/local/bin/gunicorn)
+        if "gunicorn" in process.cmdline()[0]:
```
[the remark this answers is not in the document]
guybortnikov: I know that, I prefer to keep it that way for now...

gprofiler/application_separators.py
guybortnikov: It's not just empty string it's a sentinel of string type, I did it so all the checks for string contains will be very intuitive (without the need to check for returned type), returning Optional or Union type is shit both for mypy and both for the developer.

gprofiler/merge.py
guybortnikov: Not relevant anymore..

guybortnikov: Not relevant anymore..

gprofiler/main.py
```diff
@@ -83,6 +83,7 @@ def __init__(
         client: APIClient,
         collect_metrics: bool,
         collect_metadata: bool,
+        separate_applications: bool,
```
Jongy, line 86: Rename it per the new name we have for it: `add_application_ids`?

gprofiler/merge.py
```diff
@@ -250,9 +254,14 @@ def concatenate_profiles(
 
     for pid, stacks in process_profiles.items():
         container_name = _get_container_name(pid, docker_client, add_container_names)
+        application_name = get_application_name(pid)
+        prefix = container_name + ";" if add_container_names else ""
         for stack, count in stacks.items():
+            if add_application_names and application_name is not None:
+                stack = STACK_COMM_REGEX.sub(application_name, stack)
```
Jongy, line 261: nitpick, but I think that just `stack.split(';', 1)` would be faster?

guybortnikov: You were right...
```
In [14]: def rep1(s, y):
    ...:     return STACK_COMM_REGEX.sub(y, s)

In [21]: def rep2(s, y):
    ...:     return f'{y};{s.split(";", maxsplit=1)[1]}'

In [22]: %timeit rep1(x, "bbb")
30.6 µs ± 51.1 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [23]: %timeit rep2(x, "bbb")
447 ns ± 5.45 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Jongy: Not a good enough check because the input data is short (test it with a larger stack).

Anyway I'm sure it'll be faster also with that

gprofiler/application_separators.py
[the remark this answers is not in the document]
Jongy: k

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,248 @@
...
+        ):
+            return None
+
+        for arg in process.cmdline():
+            # ":" might appear in ip:port or in wsgi app specification (module:func)
+            if ":" in arg:
+                if _IP_PORT_RE.match(arg):
+                    continue
+
+                return f"gunicorn-{_append_python_module_to_proc_wd(process, arg)}"
```
Jongy, line 89: I think it's better as `gunicorn: {...}` instead of `gunicorn-{...}`

gprofiler/merge.py
guybortnikov: How do you know the input data is short? 
Input data is `x`

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    Checks whether the process ran as "python -m ..." pattern.
+    """
+    args = process.cmdline()
+
+    # Just to be on the safe side, module name must be run after the "-m" argument, python enforces it.
+    if len(args) < 3:
+        return False
+
+    return _is_python_bin(args[0]) and args[1] == "-m"
```
Jongy, line 28: Why aren't you using `_get_cli_arg_by_name(process.cmdline(), "-m")` here? (Which is "stronger" as it works on args in other indices)

Jongy:
Backmerge master

gprofiler/merge.py
Jongy: Eh you're right, I read this function incorrectly

gprofiler/main.py
guybortnikov: Changed to `identify_applications` which suits good IMO

gprofiler/merge.py
```diff
@@ -236,6 +239,7 @@ def concatenate_profiles(
     process_profiles: ProcessToStackSampleCounters,
     docker_client: Optional[DockerClient],
     add_container_names: bool,
+    identify_applications: bool,
```
Jongy, line 242: Why here `identify_applications` and `add_application_names` in other sites....

gprofiler/merge.py
Jongy: So you're changing that?

gprofiler/application_separators.py
Jongy: Done, resolving.

gprofiler/application_separators.py
Jongy: Done, resolving.

gprofiler/application_separators.py
Jongy: I agree with @YishaiZinkin 's suggestion.

@guybortnikov , I don't see it pushed..

gprofiler/application_separators.py
Jongy: @guybortnikov , can you comment as a TODO:
1. That's it's hard and that's why we don't do it 
2. Any reference to a resource / ... on how to do it? Some pointers for the next guy working on it.

gprofiler/application_separators.py
Jongy: Done, resolving, the regex seems okay now:
```
In [2]: _PYTHON_BIN_RE = re.compile(r"^python([23](\.\d{1,2})?)?$")                                                                                                                                        

In [3]: _PYTHON_BIN_RE.match("python")                                                                                                                                                                     
Out[3]: <re.Match object; span=(0, 6), match='python'>

In [4]: _PYTHON_BIN_RE.match("python2")                                                                                                                                                                    
Out[4]: <re.Match object; span=(0, 7), match='python2'>

In [5]: _PYTHON_BIN_RE.match("python2.7")                                                                                                                                                                  
Out[5]: <re.Match object; span=(0, 9), match='python2.7'>

In [6]: _PYTHON_BIN_RE.match("python2.76")                                                                                                                                                                 
Out[6]: <re.Match object; span=(0, 10), match='python2.76'>
```

gprofiler/application_separators.py
Jongy: My suggestion was done, @YishaiZinkin 's point will not be addressed as part of the MVP; resolving.

gprofiler/application_separators.py
Jongy: Each call is a syscall and it's not currently cached. I actually opened an issue in psutil about it: https://github.com/giampaolo/psutil/issues/2036

I agree with @guybortnikov that it's a premature optimization at the moment. If we see it high in the profile, we'll address it (e.g by subclassing `Process` and making the our class implement the caching).

Closing this for now as premature opt.

gprofiler/application_separators.py
Jongy: Not pushed yet

gprofiler/application_separators.py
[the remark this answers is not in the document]
Jongy: Agree with @YishaiZinkin on this:

> It's better to have false negatives of not identifying applications and add more cases in the future.

It's better to fail in identification instead of identifying incorrectly.

Anyway, @YishaiZinkin 's suggestion is... equivalent to `basename`? So let's just change it to basename.

gprofiler/application_separators.py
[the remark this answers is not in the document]
Jongy: I don't see you translating the app name from gunicorn/etc to its file?

> Even for the gunicorn case, it is nicer to convert the app name to its Python file

Is it not required?

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+        return "uwsgi" in process.cmdline()[0]
+
+    @staticmethod
+    def _find_wsgi_from_config_file(process: Process) -> Optional[str]:
+        config_file = None
+        for arg in process.cmdline():
+            if arg.endswith(".ini"):
+                config_file = arg
+
+        if config_file is None:
```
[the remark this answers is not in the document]
Jongy: Not addressed?

gprofiler/application_separators.py
```diff
@@ -0,0 +1,232 @@
...
+        # As of gunicorn documentation the WSGI module name most probably will come from the cmdline and not from the
+        # config file / environment variables (they added the option to specify `wsgi_app` only in 20.1.0)
+        for arg in process.cmdline():
+            if ":" in arg:
+                if _IP_PORT_RE.match(arg):
+                    continue
+
+                return append_python_module_to_proc_wd(process, arg)
+
+        logger.warning(f"GunicornApplicationSeparator: matched against process {process} but couldn't find WSGI module")
```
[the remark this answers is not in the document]
Jongy: Not pushed?

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,248 @@
...
+        wsgi_arg = _get_cli_arg_by_name(process.cmdline(), "-w")
+        if wsgi_arg is not _NON_AVAILABLE_ARG:
+            return f"uwsgi-{_append_python_module_to_proc_wd(process, wsgi_arg)}"
+
+        wsgi_config = self._find_wsgi_from_config_file(process)
+        if wsgi_config is not None:
+            return f"uwsgi-{_append_python_module_to_proc_wd(process, wsgi_config)}"
+
+        _logger.warning(
+            "Couldn't find uwsgi wsgi module, both from cmdline and from config file",
```
Jongy, line 133: In `_GunicornApplicationIdentifier` you prefix your log message with the class name; here you don't. Be consistent.

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,248 @@
...
+        for arg in process.cmdline():
+            # ":" might appear in ip:port or in wsgi app specification (module:func)
+            if ":" in arg:
+                if _IP_PORT_RE.match(arg):
+                    continue
+
+                return f"gunicorn-{_append_python_module_to_proc_wd(process, arg)}"
+
+        _logger.warning(
+            f"GunicornApplicationSeparator: matched against process {process} but couldn't find WSGI module"
```
Jongy, line 92: Replace `GunicornApplicationSeparator` with `self.__class__.__name__` so it doesn't get outdated. Or remove altogether

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,248 @@
...
+    def get_application_name(self, process: Process) -> Optional[str]:
+        if not self.is_celery_process(process):
+            return None
+
+        app_name = _get_cli_arg_by_name(process.cmdline(), "-A") or _get_cli_arg_by_name(
+            process.cmdline(), "--app", check_for_equals_arg=True
+        )
+        if app_name is None:
+            _logger.warning(
+                "Couldn't find positional argument -A or --app for application indication",
```
Jongy, line 157: Same as https://github.com/Granulate/gprofiler/pull/242/files#r777152988

gprofiler/application_separators.py
Jongy: Done, resolving

gprofiler/application_separators.py
Jongy: Not addressed?

gprofiler/application_separators.py
Jongy: Not pushed

gprofiler/application_separators.py
Jongy: It's a bad sentinel tho, since CPython interns strings...
```
In [9]: str() is str()                                                                                                                                                                                     
Out[9]: True

In [10]: str() is ''                                                                                                                                                                                       
<>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
<ipython-input-10-bedcce6c87d1>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
  str() is ''
Out[10]: True
```

An empty argument will be considered your sentinel. That won't do.

I suggest the following:
1. We can make `_get_cli_arg_by_name` and `_get_cli_arg_by_index` raise some `NoSuchArg` exception in the missing case. Upon being raised it will be caught in the high-level `get_application_name` and ignored - moving to the next app identifier. That's a fair solution IMO, but it works only if you use those 2 functions to test for "positive" cases and never for "negative" cases. which I'm not sure is the situation now.
2. Replace that with a real sentinel. You can see how I did it (with mypy compatibility) in `run_in_ns`: https://github.com/Granulate/granulate-utils/blob/1cfa3870036cecd3d2dfe9e9936376b0704846ba/granulate_utils/linux/ns.py#L71. It's fine IMO, not too much of a hassle.

I don't like exception-based programming all that much, and I prefer explicitness... So personally I'm up with option 2. But up to you.

gprofiler/application_separators.py
Jongy: Done, resolving.

gprofiler/application_separators.py
Jongy: Missed that

Jongy: Done, resolving.

gprofiler/application_separators.py
Jongy: Yes pushed

Jongy: Done, resolving

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,248 @@
...
+            cmdline=process.cmdline(),
+            no_extra_to_server=True,
+        )
+        return None
+
+
+class _CeleryApplicationIdentifier(_ApplicationIdentifier):
+    @staticmethod
+    def is_celery_process(process: Process) -> bool:
+        if "celery" in os.path.basename(_get_cli_arg_by_index(process.cmdline(), 0)):
```
Jongy, line 143: `in` is too loose. Any reason to check `in basename` and not `basename == `?

gprofiler/application_separators.py
Jongy: Okay, it is pushed and fixed `_PythonModuleApplicationIdentifier`, not sure about the others tho. But here it's okay now. Resolving.

Jongy:
I went over all open discussions. Took the liberty to resolve @YishaiZinkin 's discussion when the resolution was obvious. Commented about all other discussions.

Jongy (changes requested):
.

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,248 @@
...
+    Tries to identify the application running in a given process, application identification is fully heuristic,
+    heuristics are being made on each application type available differ from each other and those their
+    "heuristic level".
+    """
+    try:
+        process = Process(pid)
+    except NoSuchProcess:
+        return None
+
+    for separator in _APPLICATION_SEPARATORS:
```
YishaiZinkin, line 231: Rename here as well

gprofiler/application_separators.py
YishaiZinkin: If returning a sentinel, why not just `None`?

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,248 @@
...
+    return os.path.join(process.cwd(), module_path)
+
+
+class _GunicornApplicationIdentifier(_ApplicationIdentifier):
+    def get_application_name(self, process: Process) -> Optional[str]:
+        # As of gunicorn documentation the WSGI module name most probably will come from the cmdline and not from the
+        # config file / environment variables (they added the option to specify `wsgi_app`
+        # in the config file only in version 20.1.0)
+
+        if "gunicorn" not in _get_cli_arg_by_index(process.cmdline(), 0) or "gunicorn" not in _get_cli_arg_by_index(
```
YishaiZinkin, line 78: [Same](https://github.com/Granulate/gprofiler/pull/242/files#r777154006).
Also, what's the case of gunicorn `in` the second arg?

gprofiler/metadata/application_identifiers.py
YishaiZinkin: IMO it should be removed. I WSGI can be passed in a config file this warning becomes a spam.

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,248 @@
...
+        config.read(resolve_host_path(process, config_file))
+        try:
+            return config.get("uwsgi", "module")
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            pass
+
+        return None
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        if "uwsgi" not in _get_cli_arg_by_index(process.cmdline(), 0):
```
YishaiZinkin, line 121: [Same](https://github.com/Granulate/gprofiler/pull/242/files#r777154006)

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,248 @@
...
+        return args[index]
+    except IndexError:
+        return _NON_AVAILABLE_ARG
+
+
+def _append_python_module_to_proc_wd(process: Process, module_path: str) -> str:
+    if os.path.isabs(module_path):
+        return module_path
+
+    return os.path.join(process.cwd(), module_path)
```
YishaiZinkin, line 69: Maybe use `realpath`? So it'll look like this:
```python
In [4]: os.path.realpath(os.path.join('/abs/path', './python_app.py'))
Out[4]: '/abs/path/python_app.py'
```
As opposed to like this:
```python
In [5]: os.path.join('/abs/path', './python_app.py')
Out[5]: '/abs/path/./python_app.py'
```

(Also in the `isabs` case)

gprofiler/application_separators.py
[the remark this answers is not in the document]
guybortnikov: not relevant anymore I've deleted the line..

gprofiler/application_separators.py
guybortnikov: @Jongy I'm up with option 2, implemented it.
@YishaiZinkin `None` isn't nice because you should check type before comparing with other strings

gprofiler/merge.py
guybortnikov: no pushed.. sorry 😢

gprofiler/metadata/application_identifiers.py
guybortnikov: I removed the log in `_GunicornApplicationIdentifier` so not relevant anymore.

gprofiler/metadata/application_identifiers.py
guybortnikov: not relevant anymore, the log was removed

gprofiler/metadata/application_identifiers.py
guybortnikov: done for both

gprofiler/application_separators.py
YishaiZinkin: But if you replace with a "real" sentinel you'll need to check the type anyway

guybortnikov: nope, I declared a sentinel that derives from `str` and it works

gprofiler/application_separators.py
Jongy: Done, resolving

gprofiler/application_separators.py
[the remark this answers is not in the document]
Jongy: Done now

gprofiler/application_separators.py
Jongy: Hmm you could use `None` I suppose. But whatever

gprofiler/metadata/application_identifiers.py
Jongy: Well, rest of the code assumes that for `_is_python_m_proc` the returned process has the module at `argv[2]`... so meh.

gprofiler/metadata/application_identifiers.py
Jongy: Done, resolving

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,256 @@
...
+
+class _JavaJarApplicationIdentifier(_ApplicationIdentifier):
+    def get_application_name(self, process: Process) -> Optional[str]:
+        if "java" != os.path.basename(_get_cli_arg_by_index(process.cmdline(), 0)) or "-jar" not in process.cmdline():
+            return None
+
+        return f"java: {_get_cli_arg_by_name(process.cmdline(), '-jar')}"
+
+
+# Please note that the order matter, because the FIRST matching separator will be used.
```
Jongy, line 214: Separators -> identifiers

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,256 @@
...
+            cmdline=process.cmdline(),
+            no_extra_to_server=True,
+        )
+        return None
+
+
+class _CeleryApplicationIdentifier(_ApplicationIdentifier):
+    @staticmethod
+    def is_celery_process(process: Process) -> bool:
+        if "celery" == os.path.basename(_get_cli_arg_by_index(process.cmdline(), 0)) or "celery" == os.path.basename(
```
Jongy, line 147: For celery you check argv0, argv1. For uwsgi you check only argv0. Is it never `python /path/to/uwsgi/uwsgi`? uwsgi is a binary compiled w/ cpython, yeah?

Jongy:
Went over all comments, I still want to go over the latest changes as well

gprofiler/merge.py
```diff
@@ -250,9 +252,14 @@ def concatenate_profiles(
 
     for pid, stacks in process_profiles.items():
         container_name = _get_container_name(pid, docker_client, add_container_names)
+        application_name = get_application_name(pid)
```
Jongy, line 255: Why calculate it if `identify_applications` is false?

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,256 @@
...
+    return _PYTHON_BIN_RE.match(os.path.basename(bin_name)) is not None
+
+
+def _get_cli_arg_by_name(args: List[str], arg_name: str, check_for_equals_arg: bool = False) -> str:
+    if arg_name in args:
+        return args[args.index(arg_name) + 1]
+
+    if check_for_equals_arg:
+        for arg in args:
+            arg_key, _, arg_val = arg.rpartition("=")
```
Jongy, line 53:
```suggestion
            arg_key, _, arg_val = arg.partition("=")
```

So for `--blabla=a=c=b` the value is `a=c=b`

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,256 @@
...
+        else:
+            return None
+
+        if not os.path.isabs(config_file):
+            config_file = os.path.join(process.cwd(), config_file)
+
+        config = configparser.ConfigParser()
+        config.read(resolve_host_path(process, config_file))
+        try:
+            return config.get("uwsgi", "module")
```
Jongy, line 116: What do the parameters mean here? It looks like a dict access with a default value, but you're catching some exceptions, so?

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,256 @@
...
+
+        return _is_python_m_proc(process) or (len(process.cmdline()) >= 2 and process.cmdline()[1].endswith(".py"))
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        if not _is_python_bin(_get_cli_arg_by_index(process.cmdline(), 0)):
+            return None
+
+        module_arg = _get_cli_arg_by_name(process.cmdline(), "-m")
+        if module_arg is not _NON_AVAILABLE_ARG:
+            return f"python: {_append_python_module_to_proc_wd(process, module_arg)}"
```
Jongy, line 197: This doesn't make much sense. If I have, for example, `gprofiler` installed and I run `python3 -m gprofiler`, I'm going to get it as if `gprofiler.py` is in my cwd.

uwsgi/gunicorn et al, do not take it from the Python path. `python` does. So here we want to skip translation I suppose.

tests/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN apt-get install -y -qq --no-install-recommends python3 python3-pip docker.io php
+
+RUN pip install -q --upgrade setuptools pip
+RUN pip install -q pytest docker
+
+ADD . /gprofiler
+RUN ./scripts/build.sh
+RUN pip install .
+CMD ["pytest", "-v", "tests/", "--ignore=tests/test_executable.py"]
```
Jongy, line 21: newline eof

gprofiler/metadata/application_identifiers.py
guybortnikov: Yep, it's the `ConfigParser` (builtin parser for ini files), added a comment

gprofiler/metadata/application_identifiers.py
guybortnikov: yep

gprofiler/metadata/application_identifiers.py
guybortnikov: kk I don't really care I think it's almost the same but agree

tests/Dockerfile
guybortnikov: removed it..

gprofiler/metadata/application_identifiers.py
Jongy: It's not the same.... it will be shown very badly.

gprofiler/metadata/application_identifiers.py
Jongy: kk

tests/conftest.py
```diff
@@ -285,6 +286,14 @@ def assert_collapsed(runtime: str) -> Callable[[Mapping[str, int], bool], None]:
     return partial(assert_function_in_collapsed, function_name)
 
 
+@fixture
+def assert_application_name(application_pid: int, runtime: str, in_container: bool) -> Generator:
+    desired_names = {"java": "java: /app/Fibonacci.jar", "python": "python: /app/lister.py"}
+    yield
+    if in_container and runtime in desired_names:
```
Jongy, line 293: Only `in_container`?

tests/containers/python/Dockerfile
```diff
@@ -5,4 +5,4 @@ ADD lister.py /app
 # Install some package so we can test that its info appears in the collapsed
 RUN pip install pyyaml==6.0
 
-CMD ["/app/lister.py"]
+CMD ["python", "lister.py"]
```
Jongy, line 8: I think there was a reason to run it like that - so the comm is `lister.py`.

Jongy: Ahh `Dockerfile.libpython` does that

Jongy (changes requested):
Reviewed all again

gprofiler/metadata/application_identifiers.py
Jongy: k

tests/conftest.py
guybortnikov: Yep, I think that's enough, I refactored the applications running in docker images

gprofiler/metadata/application_identifiers.py
```diff
@@ -194,7 +196,7 @@ def get_application_name(self, process: Process) -> Optional[str]:
 
         module_arg = _get_cli_arg_by_name(process.cmdline(), "-m")
         if module_arg is not _NON_AVAILABLE_ARG:
-            return f"python: {_append_python_module_to_proc_wd(process, module_arg)}"
+            return f"python -m {_append_python_module_to_proc_wd(process, module_arg)}"
```
Jongy: Revert back to `python:` and remove the `append_wd`

gprofiler/metadata/application_identifiers.py
YishaiZinkin: @guybortnikov Forgot to answer my question

gprofiler/metadata/application_identifiers.py
```diff
@@ -0,0 +1,262 @@
...
+    # Convert module name to module path, for example a.b -> a/b.py
+    if not module.endswith(".py"):
+        module = module.replace(".", "/") + ".py"
+
+    return _append_file_to_proc_wd(process, module)
+
+
+def _append_file_to_proc_wd(process: Process, file_path: str) -> str:
+    if os.path.isabs(file_path):
+        return file_path
```
YishaiZinkin, line 77: Maybe
```suggestion
        return os.path.realpath(file_path)
```
?

gprofiler/metadata/application_identifiers.py
guybortnikov: gunicorn may be run directly or through python, for example:
a. `/usr/local/bin/gunicorn` 
b. `/usr/local/bin/python /usr/local/bin/gunicorn`

gprofiler/metadata/application_identifiers.py
guybortnikov: Done

tests/test_sanity.py
```diff
@@ -38,6 +39,7 @@ def test_java_from_host(
         java_safemode=False,
         java_mode="ap",
     ) as profiler:
+        _ = assert_application_name  # Fixture only for asserting application name upon test teardown.
```
YishaiZinkin, line 42: Why is this line required?

tests/conftest.py
YishaiZinkin: I don't understand this answer, can you elaborate?

tests/conftest.py
```diff
@@ -285,6 +296,17 @@ def assert_collapsed(runtime: str) -> Callable[[Mapping[str, int], bool], None]:
     return partial(assert_function_in_collapsed, function_name)
 
 
+@fixture
+def assert_application_name(application_pid: int, runtime: str, in_container: bool) -> Generator:
+    desired_names = {
+        "java": "java: /app/Fibonacci.jar",
+        "python": "python: /app/lister.py",
+    }
+    yield
```
YishaiZinkin, line 305: Why `yield` before the check?

tests/conftest.py
Jongy: He means that he fixed the Dockerfiles of applications but not their execution when we run them not in a container. So he disables the check for non-containers

YishaiZinkin: Oh. It's pretty confusing, why not fix in the execution as well? Or at least add a `TODO` for clarification.

tests/conftest.py
guybortnikov: Hmm I would like to check it only after the tests are done so we don't miss maybe some required warm-up/wait, etc..

gprofiler/metadata/application_identifiers.py
Jongy: @YishaiZinkin done?

tests/conftest.py
Jongy: Not worth the time. Can add a TODO

tests/test_sanity.py
Jongy: Avoid mypy "unused variable" warning. It's a common idiom to force a fixture to be used

tests/conftest.py
Jongy: Explain in comment

gprofiler/metadata/application_identifiers.py
```diff
@@ -69,10 +69,14 @@ def _append_python_module_to_proc_wd(process: Process, module: str) -> str:
...
-        return module
+    return _append_file_to_proc_wd(process, module)
 
-    return os.path.realpath(os.path.join(process.cwd(), module))
+
+def _append_file_to_proc_wd(process: Process, file_path: str) -> str:
+    if os.path.isabs(file_path):
+        return file_path
+
+    return os.path.realpath(os.path.join(process.cwd(), file_path))
```
Jongy, line 79: `realpath` reads links but the path is in the process's namespace... How can it run?

tests/conftest.py
```diff
@@ -251,8 +252,17 @@ def output_directory(tmp_path: Path) -> Path:
...
 
 @fixture
-def application_pid(in_container: bool, application_process: subprocess.Popen, application_docker_container: Container):
-    return application_docker_container.attrs["State"]["Pid"] if in_container else application_process.pid
+def application_pid(
+    in_container: bool, application_process: subprocess.Popen, application_docker_container: Container
+) -> int:
+    pid = application_docker_container.attrs["State"]["Pid"] if in_container else application_process.pid
+
+    # Application might be run using "sh -c ...", we detect the case and return the "real" application pid
```
Jongy, line 260: Eh, why do you need this patch? To find the inner Java? Maybe `pgrep java` inside the container in your specific test?

gprofiler/metadata/application_identifiers.py
```diff
@@ -74,7 +74,7 @@ def _append_python_module_to_proc_wd(process: Process, module: str) -> str:
 
 def _append_file_to_proc_wd(process: Process, file_path: str) -> str:
     if os.path.isabs(file_path):
-        return file_path
+        return os.path.realpath(file_path)
```
Jongy, line 77: This realpath as well. Makes no sense

tests/test_sanity.py
YishaiZinkin: Maybe could make the comment more descriptive in this regard

gprofiler/metadata/application_identifiers.py
YishaiZinkin:
```python
In [2]: os.path.realpath("/not/a/real/./../path")
Out[2]: '/not/a/path'
```
🤷🏽

gprofiler/metadata/application_identifiers.py
guybortnikov: It does since: 

```python
In [2]: os.path.isabs('/a/./../b')
Out[2]: True

In [4]: os.path.realpath('/a/./../b')
Out[4]: '/b'
```

Jongy: Python docs lie

gprofiler/metadata/application_identifiers.py
Jongy: The docs say it reads links. Sigh

tests/conftest.py
guybortnikov: added TODO

tests/conftest.py
guybortnikov: Meh, it's more ugly IMO, why not just fixing it that way?

Jongy: Because this solution is ugly as well and introduces unwanted logic into this previously-simple fixture. But fuck it

tests/test_sanity.py
Jongy: Done

Jongy (approved):

tests/test_sanity.py
YishaiZinkin: @guybortnikov I suggest replacing this comment with something like `# Silence mypy - this fixture asserts application name upon test teardown`.

AFAIC this is my last comment on the PR.

YishaiZinkin: Oh I just saw the change. NVM, I approve.

YishaiZinkin (approved):
חגיגה

PR #245, opened 2021-12-16
Python: Improve packages versions cache

YishaiZinkin:
Cache Python packages versions by mount namespace.

## Description
Improve the cache of Python packages versions to be based on `/proc/[pid]/ns/mnt` instead of on ancestor process in the namespace.

Also, fix a potential bug by caching Python's version according mnt ns *and* `cmdline[0]` of the process to prevent collision in case several Python versions are executed in the same namespace.

## Motivation and Context
Getting ancestor process is much heavier than `readlink`ing `/proc/[pid]/ns/mnt`. Also, namespace-based cache is more accurate.

## Screenshots
[

## Checklist:

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,6 +3,7 @@ flake8==3.9.2
 black==21.5b2
 mypy==0.901
 isort==5.8.0
+types-cachetools==4.2.6
 types-requests==0.1.9
 types-dataclasses==0.1.5
 types-psutil==5.8.0
--- a/gprofiler/metadata/py_module_version.py
+++ b/gprofiler/metadata/py_module_version.py
@@ -9,24 +9,62 @@ in pip 21.3.1, as mentioned in the functions' documentation.
 """
 import csv
 import email.parser
-import functools
 import os
 import pathlib
 import re
-from typing import Dict, Iterator, Optional, Tuple
+from typing import Dict, Iterator, Optional, Tuple, Type, Union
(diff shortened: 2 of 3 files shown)


d3dave:
Awesome! I wonder, however, how the kernel determines the id for a new namespace. Is it guaranteed to be unique for the duration that gprofiler runs?

d3dave:
> how the kernel determines the id for a new namespace. Is it guaranteed to be unique for the duration that gprofiler runs?

The id is apparently just a new dynamic inode number. Considering that namespace inodes come and go, I'm not sure whether we can rely on this for caching.

Also, I think `cmdline[0]` may not be reliable enough to distinguish python versions. The version number in `libpythonX.Y.so` is definitely better.

Jongy:
As @d3dave correctly states, the inode for namespaces is a dynamic number that can be reused as namespaces are deleted and recreated. Using it as a cache key this way is not correct.

Technically, using the mnt ancestor is not entirely correct either (since processes may move between namespaces) but this is much less likely to happen ("normal" containers don't do that) comparing with ID recycling.

> Getting ancestor process is much heavier than readlinking /proc/[pid]/ns/mnt. Also, namespace-based cache is more accurate.

It's indeed much heavier and involves more than an order of magnitude of syscalls, but how often does it happen, anyway? Just once per Python PID per profiler session, no? It's not very often. Do we need to optimize that?

PR #246, opened 2021-12-16
Prevent registering of profiler arguments with a wrong dest name

YishaiZinkin:
## Description
In order to be passed to the relevant profiler, the `dest` of profiler-specific arguments must be prefixed with the profiler's name. This PR `assert`s they're named correctly.

## Motivation and Context
I tried to add an argument without knowing this and spent some good amount of time debugging that could've been saved with this `assert`.

## How Has This Been Tested?
Just ran gProfiler and saw that this `assert` isn't triggered, and then ran it again with a wrong arg `dest`.

## Checklist:

### DIFF
--- a/gprofiler/profilers/registry.py
+++ b/gprofiler/profilers/registry.py
@@ -72,6 +72,9 @@ def register_profiler(
 
     def profiler_decorator(profiler_class):
         assert profiler_name not in profilers_config, f"{profiler_name} is already registered!"
+        assert all(
+            arg.dest.startswith(profiler_name.lower()) for arg in profiler_arguments or []
+        ), f"{profiler_name}: Profiler args dest must be prefixed with the profiler name"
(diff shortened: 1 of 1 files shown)


alonbentamar (approved):
Make sure that all the current profilers are not broken by this change, especially the node profiler can be problematic.

YishaiZinkin:
> Make sure that all the current profilers are not broken by this change, especially the node profiler can be problematic.

@alonbentamar nodejs doesn't have it's own profiler and IIUC doesn't use the interface this PR modifies altogether. Am I missing something?

alonbentamar:
> > Make sure that all the current profilers are not broken by this change, especially the node profiler can be problematic.
> 
> @alonbentamar nodejs doesn't have its own profiler and IIUC doesn't use the interface this PR modifies altogether. Am I missing something?

Exactly, I just remembered it as an edge case that you should make sure is not broken

PR #248, opened 2021-12-19
java: Expand --java-safemode into a set of options

Jongy:
## Description
Now, passing plain --java-safemode implies all options.
Alternatively, one can pass --java-safemode=a,b,c to enable only a, b, c.

## Motivation and Context
Finer control on which features are enabled.

Note that, for most features that revolve around disabling the profiler - they are enabled anyway but do not cause disabling of the Java profiler if the flag is not enabled.

## How Has This Been Tested?
Test disabling each appropriate flag, and test it again with `--java-safemode=all`.
* [x] General OOM stops in `all`
* [x] General OOM DOES NOT stop if not given
* [x] General signaled stops in `all`
* [x] General signaled DOES NOT stop if not given
* [x] Profiled OOM stops by default
* [x] Profiled OOM stops in `all`
* [x] Profiled OOM DOES NOT stop if not given in safemode
* [ ] ~~Profiled signaled stops by default~~ not relevant for Java as it catches SIGSEGV (which is what gets logged and we grep for in the kernel log), will be tested with https://github.com/Granulate/gprofiler/pull/254 which adds proc_events support.
* [ ] ~~Profiled signaled stops in `all`~~ same ^^
* [ ] ~~Profiled signaled DOES NOT stop not given in safemode~~ same ^^
* [x] HS err stops by default
* [x] HS err stops in `all`
* [x] HS err DOES NOT stop if not given in safemode
* [x] PID in kernel messages stops in `all`
* [x] PID in kernel messages DOES NOT stop if not given

Additionally, those 2 are unrelated to disabling the Java profiler:
* [x] We do not profile if another libap is loaded in `all`
* [x] We do profile if another libap is loaded and `ap-loaded-check` is not given
* [x] Extensive Java version checks are enabled in `all`
* [x] Extensive Java version checks are disabled if `java-version-check` is not given.

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -544,7 +544,9 @@ class JavaProfiler(ProcessProfilerBase):
             )
 
         if java_safemode == JAVA_SAFEMODE_ALL:
-            assert self._ap_safemode == 127, "async-profiler safemode must be set to 127 in --java-safemode"
+            assert (
+                self._ap_safemode == 127
+            ), f"async-profiler safemode must be set to 127 in --java-safemode={JAVA_SAFEMODE_ALL} (or --java-safemode)"
 
     def _disable_profiling(self, cause: str):
         if self._should_profile and cause in self._java_safemode:
@@ -631,7 +633,8 @@ class JavaProfiler(ProcessProfilerBase):
             # then check with that instead (if exe isn't java)
             if process_basename != "java":
                 logger.warning(
-                    "Non-java basenamed process, skipping... (disable --java-safemode to profile it anyway)",
+                    "Non-java basenamed process, skipping... (disable "
+                    f" --java-safemode={JavaSafemodeOptions.JAVA_EXTENDED_VERSION_CHECKS} to profile it anyway)",
                     pid=process.pid,
                     exe=process.exe(),
                 )
@@ -645,7 +648,8 @@ class JavaProfiler(ProcessProfilerBase):
             if not self._is_jvm_version_supported(java_version_output):
                 logger.warning(
                     "Process running unsupported Java version, skipping..."
-                    " (disable --java-safemode to profile it anyway)",
+                    f" (disable --java-safemode={JavaSafemodeOptions.JAVA_EXTENDED_VERSION_CHECKS}"
+                    " to profile it anyway)",
                     pid=process.pid,
                     java_version_output=java_version_output,
                 )
@@ -667,7 +671,7 @@ class JavaProfiler(ProcessProfilerBase):
             if "libasyncProfiler.so" in mmap.path and not mmap.path.startswith(TEMPORARY_STORAGE_PATH):
                 logger.warning(
                     "Non-gProfiler async-profiler is already loaded to the target process."
-                    " Disable --java-safemode=ap-loaded-check to bypass this check.",
+                    f" Disable --java-safemode={JavaSafemodeOptions.AP_LOADED_CHECK} to bypass this check.",
                     pid=process.pid,
                     ap_path=mmap.path,
                 )
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -70,6 +70,11 @@ class JavaSafemodeOptions(str, Enum):
 
 
 JAVA_SAFEMODE_ALL_OPTIONS = [o.value for o in JavaSafemodeOptions]
+JAVA_SAFEMODE_DEFAULT_OPTIONS = [
+    JavaSafemodeOptions.PROFILED_OOM.value,
+    JavaSafemodeOptions.PROFILED_SIGNALED.value,
+    JavaSafemodeOptions.HSERR.value,
+]
 
 
 class JattachException(CalledProcessError):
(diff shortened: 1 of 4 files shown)


tests/test_java.py
```diff
@@ -215,10 +231,20 @@ def sap_and_crash(self, *args, **kwargs):
         return result
 
     monkeypatch.setattr(AsyncProfiledProcess, "start_async_profiler", sap_and_crash)
-    # To make sure it is reverted to True (the original value) after the test
-    monkeypatch.setattr(JavaProfiler, "_should_profile", True)
 
-    with JavaProfiler(1, 5, Event(), str(tmp_path), False, False, "cpu", 0, False, "ap") as profiler:
+    profiler = JavaProfiler(
```
IlayRosenberg, line 235: Why did you change the arguments here?

IlayRosenberg (changes requested):
Nice

gprofiler/profilers/java.py
```diff
@@ -469,38 +508,54 @@ def __init__(
...
             logger.warning("Java version checks are disabled")
         self._mode = java_async_profiler_mode
-        self._safemode = java_async_profiler_safemode
+        self._ap_safemode = java_async_profiler_safemode
+        self._init_java_safemode(java_safemode)
         self._saved_mlock: Optional[int] = None
-        self._java_safemode = java_safemode
-        if self._java_safemode:
-            logger.debug("Java safemode enabled")
+        self._should_profile = True
```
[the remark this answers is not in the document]
Jongy: I changed it because it was more convenient in the code now; and anyway, there's indeed a single instance (of `GProfiler` and all "child" classes), so it's alright.

gprofiler/profilers/java.py
```diff
@@ -54,6 +55,39 @@
...
+    # * we saw an exit code of signal in a proc_events event.
+    PROFILED_SIGNALED = "profiled-signaled"
+    # hs_err file was written for a profiled process
+    HSERR = "hserr"
+    # a process was OOM-killed and we saw it in the kernel log
+    GENERAL_OOM = "general-oom"
+    # a process was fatally signaled and we saw it in the kernel log
+    GENERAL_SIGNALED = "general-signaled"
+    # we saw the PID of a profiled process in the kernel logs
+    PID_IN_MESSAGES = "pid-in-messages"
```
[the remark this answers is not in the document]
Jongy: kk

tests/test_java.py
Jongy: I needed to change `False` to `'all'` now. Also added kwargs for clarity

IlayRosenberg: Isn't `False` supposed to be the same as `''`? `all` is the "new" `True` right?

Jongy: Yeah but this conversion happens in the `argparse` / `configparse` level. `JavaProfiler` now expects this value as a string, so I had to change it.

gprofiler/profilers/java.py
```diff
@@ -469,38 +508,54 @@ def __init__(
...
+
+        if self._java_safemode:
+            logger.debug("Java safemode enabled", safemode=self._java_safemode)
+
+        if JavaSafemodeOptions.JAVA_EXTENDED_VERSION_CHECK in self._java_safemode:
+            assert (
+                self._simple_version_check
+            ), f"Java version checks are mandatory in --java-safemode={JavaSafemodeOptions.JAVA_EXTENDED_VERSION_CHECK}"
+
+        if java_safemode == JAVA_SAFEMODE_ALL:
```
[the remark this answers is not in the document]
Jongy: Well, previously it was checked only if the boolean of safemode was set. I considered adding it as a separate safemode option, but it feels stupid to add an option that only checks for the existence of another option.

From my point of view - it's just helping to make sure that if you use the safemode simple dumb - you also enable the AP safemode. If a user picks their own options, they can also pick their own AP safemode options.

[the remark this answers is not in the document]
Jongy: We might see in the future that it bugs us (in situation where you want to disable AP safemode but keep full Java safemode), if that happens I'll remove this `assert` completely.

Jongy:
@IlayRosenberg - I have finished performing all tests mentioned in https://github.com/Granulate/gprofiler/pull/248#issue-1084190938, and made some changes. Most are cosmetic, few contain real logic change such as https://github.com/Granulate/gprofiler/pull/248/commits/9c763fc79f4bbaafdd9182a0358c9e4d3134eebd. I separated the commits well so you can go one by one.

gprofiler/profilers/java.py
```diff
@@ -665,10 +654,30 @@ def _is_profiling_supported(self, process: Process) -> bool:
...
+                )
+                return True
+
+        return False
+
     def _profile_process(self, process: Process) -> Optional[StackToSampleCount]:
         if not self._is_profiling_supported(process):
             return None
 
+        if self._check_async_profiler_loaded(process):
```
IlayRosenberg, line 686: Is there a special reason you moved it to `JavaProfiler` from `AsyncProfiledProcess`?

Jongy: I preferred the logging line + `return None` here approach, to the exception-based approach. In `AsyncProfiledProcess` I currently lack the API for the `return None`, and we already do some "compatibility" checks here, so this seemed like the right spot.

IlayRosenberg: Makes sense

Jongy:
Something is completely messed up with the CI.. sigh

IlayRosenberg (dismissed):
Let's make sure the CI failures are benign/make them pass

Jongy:
There was 1 real error, fixed

IlayRosenberg (approved):

Jongy:
Failing tests are irrelevant - eBPF shit & the known `dl` thing. I'll check the eBPF error in another PR and possibly revert the Python 3.10 change from https://github.com/Granulate/gprofiler/pull/255

PR #249, opened 2021-12-20
merge: Scale process counts using random instead of just rounding

DanielShaulov:
## Description
Re-using the same logic from dwarf scaling to process scaling - instead of rounding the samples to nearest integer we call `random` and `ceil` if the non-integer part is lower than the result, so over time the sample counts are more correct.
In addition - for cases where the result is 0, we remove the whole stack from the results. The behavior for dwarf scaling was to just leave the old value which is bogus, and for perf it was to leave the stack with a zero count which spams the backend and enlarges the output files.

## Related Issue
None

## How Has This Been Tested?
CI will test it

### DIFF
--- a/gprofiler/merge.py
+++ b/gprofiler/merge.py
@@ -8,7 +8,7 @@ import random
 import re
 from collections import Counter, defaultdict
 from pathlib import Path
-from typing import Iterable, Optional, Tuple
+from typing import Dict, Iterable, Optional, Tuple
 
 from gprofiler.docker_client import DockerClient
 from gprofiler.gprofiler_types import ProcessToStackSampleCounters, StackToSampleCount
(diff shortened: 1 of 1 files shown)


gprofiler/merge.py
```diff
@@ -148,32 +148,26 @@ def add_highest_avg_depth_stacks_per_process(
...
+    scaled_stacks: StackToSampleCount = StackToSampleCount()
+    for stack, count in stacks.items():
+        new_count = count * ratio
         # If we were to round all of the sample counts it could skew the results. By using a random factor,
         # we mostly solve this by randomly rounding up / down stacks.
         # The higher the fractional part of the new count, the more likely it is to be rounded up instead of down
-        new_count = math.ceil(new_count) if new_count - int(new_count) <= random.random() else math.floor(new_count)
-        if new_count == 0:
-            # TODO: For more accurate truncation, check if there's a common frame for the truncated stacks and combine
-            #  them
```
alonbentamar, line 172: Keep the `todo`

alonbentamar (changes requested):
We need more eyes on this, @Jongy care to review?

alonbentamar:
What kind of manual tests did you do?

gprofiler/merge.py
```diff
@@ -148,32 +148,27 @@ def add_highest_avg_depth_stacks_per_process(
...
-            continue
-        dwarf_collapsed_stacks_counters[stack] = new_count
-    # Note - returning the value is not necessary, but is done for readability
-    return dwarf_collapsed_stacks_counters
+        scaled_value = math.ceil(new_count) if new_count - int(new_count) <= random.random() else math.floor(new_count)
```
Jongy, line 167: `int` is used as `floor` here, right? Let's use `math.floor` directly so this expression remains simpler

Jongy, line 167: Umm... I'm either missing the point here, or, this `if` is swapped? For `new_count` with "large" fractions (e.g `.95`) this expression is more likely to pick the `floor` path instead of the `ceil` path. (and vice-versa for small fractions):

```
In [27]: new_count                                                                                                                                                                                         
Out[27]: 3.99

In [28]: math.ceil(new_count) if new_count - int(new_count) <= random.random() else math.floor(new_count)                                                                                                  
Out[28]: 3

In [29]: math.ceil(new_count) if new_count - int(new_count) <= random.random() else math.floor(new_count)                                                                                                  
Out[29]: 3

In [32]: new_count = 2.01                                                                                                                                                                                  

In [33]: math.ceil(new_count) if new_count - int(new_count) <= random.random() else math.floor(new_count)                                                                                                  
Out[33]: 3

In [34]: math.ceil(new_count) if new_count - int(new_count) <= random.random() else math.floor(new_count)                                                                                                  
Out[34]: 3

In [35]: math.ceil(new_count) if new_count - int(new_count) <= random.random() else math.floor(new_count)                                                                                                  
Out[35]: 3
```

I suppose the condition needs to be swapped.

Jongy (changes requested):
.

gprofiler/merge.py
DanielShaulov: `math.modf(new_count)[0]`?

DanielShaulov: Yea, should be `random <= fraction`
This logic existed before the PR for dwarf scaling, fixed now

Jongy: kk

Jongy: Mhm, was broken before.

Jongy (approved):

PR #253, opened 2021-12-27
python: pyperf: Bump bpf_get_stack_offset version to v0.0.3

Jongy:
## Description
v0.0.3 avoids BPF accesses to non-canonical addresses, which may trigger a `WARN_ONCE` that we prefer to avoid :)

See the fix in https://github.com/Jongy/bpf_get_stack_offset/commit/54b70ee65708cc8d3d7817277e82376d95205356.

## Motivation and Context
Avoid triggering kernel warnings during gProfiler's run.

## How Has This Been Tested?
CI.

I have tested the fixed `get_stack_offset` program on a 5.4 kernel which triggers the `WARN_ONCE` on v0.0.2; and v0.0.3 no longer triggers it.

I have also tested v0.0.3 on 4.15.0-72-generic, which is old enough to have the limit of 4096 instructions. The program still runs (that is, the instructions I added in the loop do not cause us to exceed the limit.

## Checklist:


d3dave (approved):

Jongy:
Both CI failures are irrelevant (same dl errors, and in some Java test...)

PR #254, opened 2021-12-28
java: Disable profiling (safemode) on proc_events signals

Jongy:
Based on https://github.com/Granulate/gprofiler/pull/248

## Description
Use detected "signaled profiled processes" from proc_events to disable profiling when the safemode option is on.

## Motivation and Context
Improve our safemode; specifically, this helps in the cases where the hs_err file is not available / will not be available (e.g because the target process runs in a container that'll exit once the process exits; so we won't see the file as the container went down).

## How Has This Been Tested?
* [x] Send SIGSEGV to a Java process running as PID 1 of a container (so the container exits immediately and we don't see the hs_err file) - profiling should be disabled based on the proc_events detection
* [x] Same ^^ with SIGKILL

## Checklist:

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -63,6 +63,7 @@ class JavaSafemodeOptions(str, Enum):
     PROFILED_OOM = "profiled-oom"
     # a profiled process was signaled:
     # * fatally signaled and we saw it in the kernel log
+    # * we saw an exit code of signal in a proc_events event.
     PROFILED_SIGNALED = "profiled-signaled"
     # hs_err file was written for a profiled process
     HSERR = "hserr"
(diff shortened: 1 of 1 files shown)


Jongy:
@IlayRosenberg short one after https://github.com/Granulate/gprofiler/pull/248

IlayRosenberg (approved):

Jongy:
All tests mentioned in https://github.com/Granulate/gprofiler/pull/254#issue-1090074423 have passed with this based on the current tip of #248 , b81719e.

I'll push this after #248 is merged so you can re-approve.

Jongy:
Interesting... @IlayRosenberg see - I have disabled the test that used to fail with the `dl` error consistently (`test_async_profiler_already_...`). Now another one fails:
```
FAILED tests/test_java.py::test_java_async_profiler_cpu_mode[True] - Assertio...
```

lol

IlayRosenberg (approved):

Jongy:
Only the eBPF failure, irrelevant as explained in https://github.com/Granulate/gprofiler/pull/248#issuecomment-1005501711

PR #255, opened 2022-01-02
python: Upgrade PyPerf & run the CI on Python 3.10

Jongy:
## Description
* Support Python 3.10 offsets - so PyPerf can profile Python 3.10 apps :)
* Fix native stacks that were broken for certain kernel versions since PyPerf v1.2.0;
  should now work for all versions.
* Enable CI on Python 3.10

## Motivation and Context
Support PyPerf with Python 3.10, and fix native stacks so it works everywhere :)

## How Has This Been Tested?
PyPerf was tested on kernel versions such as `4.15.0-72-generic` that have previously failed to get native stacks. Now they succeed. Also see more in https://github.com/Granulate/bcc/pull/32.

And also, was tested by CI.

### DIFF
--- a/.github/workflows/build-container-and-test.yml
+++ b/.github/workflows/build-container-and-test.yml
@@ -6,7 +6,7 @@ jobs:
   lint:
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10.1]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10.0]
 
     runs-on: ubuntu-latest
 
@@ -34,7 +34,7 @@ jobs:
     strategy:
       fail-fast: false  # helps detecting flakiness / errors specific to one Python version
       matrix:
-        python-version: [3.6, 3.7, 3.8.8, 3.9.2, 3.10.1]
+        python-version: [3.6, 3.7, 3.8.8, 3.9.2, 3.10.0]
 
     runs-on: ubuntu-latest
 
--- a/.github/workflows/build-container-and-test.yml
+++ b/.github/workflows/build-container-and-test.yml
@@ -34,7 +34,7 @@ jobs:
     strategy:
       fail-fast: false  # helps detecting flakiness / errors specific to one Python version
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-container-and-test.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   lint:
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10.1]
```
YishaiZinkin, line 9: Why you include the minor? Same for the matrix line below

Jongy: See commit message of https://github.com/Granulate/gprofiler/pull/255/commits/659d4f58d869522cfa980e869442f53ece68a829 (and see the CI status of https://github.com/Granulate/gprofiler/pull/255/commits/2254671565a76bba40d8c109edb6ea562bcb3da7, the previous commit)

YishaiZinkin: WTF lol

Jongy:
I downgraded to Python 3.10.0 because the tests fail with 3.10.1 and that's a real error. I have tried to debug (generated offsets for 3.10.1 and they seemed okay). I think it's something more deep. Leaving it for now.

YishaiZinkin:
> I downgraded to Python 3.10.0 because the tests fail with 3.10.1 and that's a real error. I have tried to debug (generated offsets for 3.10.1 and they seemed okay). I think it's something more deep. Leaving it for now.

Doesn't it better to declare 3.10 as not supported at the moment instead of choosing a minor that happen to pass the tests?

Jongy:
> Doesn't it better to declare 3.10 as not supported at the moment instead of choosing a minor that happen to pass the tests?

Since it works for this version - since it already works, let's continue testing it meanwhile so we test that e.g PyPerf changes do not break 3.10 support;

Jongy:
Errors are irrelevant, this is ready for merging

YishaiZinkin (approved):

PR #258, opened 2022-01-05
app ids: Keep original frame

Jongy:
## Description
1. Don't insert Java app id if no `-jar` arg is present (bugfix)
2. Keep the original first frame - insert the app id frame second.


guybortnikov (approved):

Jongy:
Unrelated `dl` crap

PR #262, opened 2022-01-08
java: Add --java-async-profiler-args, allowing to send arbitrary extra args to AP

Jongy:
## Description
AP itself is highly configurable. This option will allow trying out different AP configurations, even if they are not yet supported by gProfiler's interface.

## How Has This Been Tested?
Added AP option `dot` and saw that Java class names are now using `.` instead of `/`.

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -134,6 +134,7 @@ class AsyncProfiledProcess:
         buildids: bool,
         mode: str,
         ap_safemode: int,
+        ap_args: str,
     ):
         self.process = process
         # access the process' root via its topmost parent/ancestor which uses the same mount namespace.
@@ -178,6 +179,7 @@ class AsyncProfiledProcess:
         assert mode in ("cpu", "itimer"), f"unexpected mode: {mode}"
         self._mode = mode
         self._ap_safemode = ap_safemode
+        self._ap_args = ap_args
 
     def __enter__(self):
         os.makedirs(self._ap_dir_host, 0o755, exist_ok=True)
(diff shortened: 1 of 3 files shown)


Jongy:
In another PR I will organize the tests so we have 1 site creating `JavaProfiler` and we don't need to make the same change repeatedly in all sites.

gprofiler/profilers/java.py
```diff
@@ -303,7 +308,7 @@ def _get_stop_cmd(self, with_output: bool) -> List[str]:
             ap_params.append(self.OUTPUT_FORMAT)
             ap_params.append(self.FORMAT_PARAMS)
         ap_params.append(f"log={self._log_path_process}")
-        return self._get_base_cmd() + [",".join(ap_params)]
+        return self._get_base_cmd() + [",".join(ap_params) + self._get_extra_ap_args()]
```
DanielShaulov, line 311: hmmmm... we sure we want to pass those same arguments on stop? What is the use case here? Aren't most arguments we might pass only relevant to `start` cmd?

DanielShaulov (changes requested):

gprofiler/profilers/java.py
Jongy: Some arguments are relevant for `stop` (e.g the output file, output type, output format). In any case, AP doesn't care if you give arguments in start/stop; IIRC `profiler.sh` does that as well.

And actually, until [this](https://github.com/jvm-profiling-tools/async-profiler/pull/503) PR, AP wouldn't even complain on unsupported args.

Jongy: The idea of this change is to support everything we might want

DanielShaulov: I see, so even now it just silently ignores args that are inappropriate for the current action :thinking: , oh well, if we ever need it we will just separate it to 2 different args (extra_start and extra_stop

DanielShaulov (approved):

gprofiler/profilers/java.py
Jongy: I doubt we will

DanielShaulov (approved):

PR #263, opened 2022-01-09
java: Show a "profiling skipped" stack when profiling is skipped for a process

Jongy:
## Description
When profiling of a process is skipped by gProfiler for whatever reason, the following happens:
* In perf-less mode (`--perf-mode=none`) - we won't see its stacks and we won't know it existed.
* In standard mode, gProfiler will use the samples collected by `perf` (if any) when merging the stacks. So we'll see the native Java code, which is seldom interesting.

This PR amends the Java profiler to return a specialized stack per "skip" case (safemode, AP already loaded in a process, ...). This way, the user sees the cause of the error immediately, and doesn't have to skim through the logs in order to understand. Also, it makes the skip explicit ("gProfiler found this process and decided to skip", instead of "maybe gProfiler missed it..."). Explicitness is good.

## Motivation and Context
Make it easier to understand when & why gProfiler decides to skip Java profiling.

## How Has This Been Tested?
* [x] Profiling a process with AP loaded (from a different path) - results in `java;[Profiling skipped: async-profiler is already loaded]`.
* [x] Profiling a process with unsupported JVM version / type - results in `java;[Profiling skipped: profiling this JVM is not supported]`.
* [x] Profiling a process after safemode has disabled profiling completely - results in the profiling skipped stack, with the safemode reason in it. For example: `java;[Profiling skipped: disabled due to hserr]`

Automatic tests for all the above ^^

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -9,6 +9,7 @@ import os
 import re
 import shutil
 import signal
+from collections import Counter
 from enum import Enum
 from pathlib import Path
 from threading import Event
@@ -515,6 +516,8 @@ class JavaProfiler(ProcessProfilerBase):
         self._ap_safemode = java_async_profiler_safemode
         self._init_java_safemode(java_safemode)
         self._should_profile = True
+        # if set, profiling is disabled due to this safemode reason.
+        self._safemode_disable_reason: Optional[str] = None
         self._profiled_pids: Set[int] = set()
         self._pids_to_remove: Set[int] = set()
         self._kernel_messages_provider = get_kernel_messages_provider()
@@ -545,9 +548,15 @@ class JavaProfiler(ProcessProfilerBase):
             ), f"async-profiler safemode must be set to 127 in --java-safemode={JAVA_SAFEMODE_ALL} (or --java-safemode)"
 
     def _disable_profiling(self, cause: str):
-        if self._should_profile and cause in self._java_safemode:
+        if self._safemode_disable_reason is None and cause in self._java_safemode:
             logger.warning("Java profiling has been disabled, will avoid profiling any new java processes", cause=cause)
-            self._should_profile = False
+            self._safemode_disable_reason = cause
+
+    def _profiling_error_stack(self, reason: str) -> StackToSampleCount:
+        # return 1 sample, it will be scaled later in merge_profiles().
+        # if --perf-mode=none mode is used, it will not, but we don't have anything logical to
+        # do here in that case :/
+        return Counter({f"[Profiling error: {reason}]": 1})
 
     def _is_jvm_type_supported(self, java_version_cmd_output: str) -> bool:
         return all(exclusion not in java_version_cmd_output for exclusion in self.JDK_EXCLUSIONS)
@@ -676,6 +685,9 @@ class JavaProfiler(ProcessProfilerBase):
         return False
 
     def _profile_process(self, process: Process) -> Optional[StackToSampleCount]:
+        if self._safemode_disable_reason is not None:
+            return self._profiling_error_stack(f"disabled due to {self._safemode_disable_reason}")
+
         if not self._is_profiling_supported(process):
             return None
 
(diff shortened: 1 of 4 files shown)


gprofiler/profilers/java.py
```diff
@@ -676,11 +685,16 @@ def _check_async_profiler_loaded(self, process: Process) -> bool:
         return False
 
     def _profile_process(self, process: Process) -> Optional[StackToSampleCount]:
```
DanielShaulov, line 687: Return value is no longer `Optional`

gprofiler/profilers/java.py
```diff
@@ -676,11 +685,16 @@ def _check_async_profiler_loaded(self, process: Process) -> bool:
...
-        if not self._is_profiling_supported(process):
-            return None
+        comm = process_comm(process)
+
+        if self._safemode_disable_reason is not None:
+            return self._profiling_skipped_stack(f"disabled due to {self._safemode_disable_reason}", comm)
+
+        if not self._is_jvm_profiling_supported(process):
+            return self._profiling_skipped_stack("profiling this JVM is not supported", comm)
```
DanielShaulov, line 694: consider splitting to the sub reasons from the function (jvm version/jvm type/process name)

tests/utils.py
```diff
@@ -98,3 +101,9 @@ def assert_function_in_collapsed(function_name: str, collapsed: Mapping[str, int
     assert any(
         (function_name in record) for record in collapsed.keys()
     ), f"function {function_name!r} missing in collapsed data!"
+
+
+def snapshot_one_collaped(profiler: ProfilerInterface) -> StackToSampleCount:
+    result = profiler.snapshot()
+    assert len(result) == 1
+    return result[next(iter(result.keys()))]
```
DanielShaulov, line 109:
```suggestion
    return next(iter(result.values()))
```

DanielShaulov (changes requested):
Minor comments, looks good.
I think you are also missing a test for bad/incompatible java version, but I understand if it is too hard to do and not worth it

gprofiler/profilers/java.py
Jongy: Yup. Interface is still Optional, but will change it here

DanielShaulov: Why? what other runtimes have elective disablement? Should change the interface and do the same for everything IMO

Jongy: Uh sadly not precisely:
```
        if output is None:
            logger.warning(f"Profiled process {ap_proc.process.pid} exited before reading the output")
            return None
```

Some of these I have fixed in a later PR. Will change it a future PR.

gprofiler/profilers/java.py
Jongy: Not sure it's relevant. This is only a marker for the error - from that point you can inspect the log and see the exact version. In the stack itself - I think it is too much. So even if you see a stack "JVM version bla bla" or "JVM type bla bla" you still have to check the log :shrug:

gprofiler/profilers/java.py
Jongy: > Why? what other runtimes have elective disablement? Should change the interface and do the same for everything IMO

Not selective disablement, but other failure reasons, as I have found here.

DanielShaulov: Well - we do want all of those to also not just have native stacks, but an "error" stack, but if you want to delay to a later PR, it is OK

gprofiler/profilers/java.py
DanielShaulov: k

DanielShaulov (approved):

Jongy:
> Minor comments, looks good.
> I think you are also missing a test for bad/incompatible java version, but I understand if it is too hard to do and not worth it

We have `test_java_safemode_version_check` and `test_java_safemode_build_number_check`, but no test for unsupported JVM type because it's indeed hard to implement.

PR #268, opened 2022-01-15
Fix CI by reordering tests

Jongy:
## Description
Attempts to fix:
* [x] The weird dl error that we have been having in the last 3 weeks (for example https://github.com/Granulate/gprofiler/runs/4829130163?check_suite_focus=true). Exception is `AssertionError: unexpected 'grep' exit code: 127, stdout b'' stderr b"Inconsistency detected by ld.so: dl-open.c: 560: dl_open_worker: Assertion `_dl_debug_initialize (0, args->nsid)->r_state == RT_CONSISTENT' failed!\n"`.
* [x] ~~Occasional PyPerf errors (for example https://github.com/Granulate/gprofiler/runs/4829130238?check_suite_focus=true). Exception is `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xbc in position 1678190: invalid start byte` (as if PyPerf outputs invalid textual data... Happens only when testing against Python 3.10, so I feel like it's a bug in the Python 3.10 support)~~ with the added print I understood the source of the issue (see bad data in https://github.com/Granulate/gprofiler/runs/5258558141?check_suite_focus=true), will fix[ in another PR](https://github.com/Granulate/gprofiler/pull/300).

### DIFF
--- a/.github/workflows/build-container-and-test.yml
+++ b/.github/workflows/build-container-and-test.yml
@@ -72,4 +72,4 @@ jobs:
 
     # TODO: Add docker layer caching when GitHub Actions cache is stabilized and works good with "satackey/action-docker-layer-caching@v0.0.11"
     - name: Run gProfiler tests
-      run: which grep; ldd `which grep`; ldd `which python3`; NO_APT_INSTALL=1 ./tests/test.sh --ignore=tests/test_executable.py
+      run: NO_APT_INSTALL=1 ./tests/test.sh --ignore=tests/test_executable.py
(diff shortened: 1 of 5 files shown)


Jongy:
Heh, this seems to have fixed the issue.... Trying again.

Jongy:
This fixes the CI + adds a print of the failing data of PyPerf (when it fails)

adi-benz (approved):
I don't like that we're reordering tests instead of actually fixing the problem, but I guess it's better then nothing / better than investing too much time into it

Jongy:
> I don't like that we're reordering tests instead of actually fixing the problem, but I guess it's better then nothing / better than investing too much time into it

I don't like it much either, but we've already spent a couple of hours on it (and I have personally spent, haha) with no conclusion. I am interested about its source as I truly don't understand how / why t could happen. You can take a couple of hours to look at it if you'd like.

PR #269, opened 2022-01-16
linux: ns: get_process_nspid() support old kernels

adi-benz:
## Description
Use `ns.get_process_nspid()` from `granulate_utils`, it supported older kernels too (kernels older than 4.1 that doesn't expose the NSpid through `/proc/pid/status`).

## How Has This Been Tested?
Tested Java profiling correctly identifies hotspot's error file using the ns pid  
Tested on Ubuntu 14.04 `3.13.0-170-generic`

## Checklist:
- [x] Update the git submodule after merging https://github.com/Granulate/granulate-utils/pull/20


gprofiler/profilers/java.py
```diff
@@ -216,7 +215,6 @@ def _existing_realpath(self, path: str) -> Optional[str]:
     def locate_hotspot_error_file(self) -> Optional[str]:
         # nspid is required
         if self._nspid is None:
```
Jongy, line 217: This should never happen now, right? This can be an `assert`

gprofiler/profilers/java.py
```diff
@@ -596,7 +594,6 @@ def _get_java_version(process: Process) -> str:
             #    other JDK types.
             java_path = f"/proc/{nspid}/exe"
         else:
-            # TODO fix get_process_nspid() for all cases.
```
Jongy, line 599: This can be an `assert` as well?

Jongy (changes requested):
.

Jongy (approved):
Looks good, we'll just update granulate-utils once that's ready

Jongy (approved):

PR #276, opened 2022-01-26
java: Add the "AP DSO was loaded" into JattachException

Jongy:
## Description
It's easier to read & understand when the info is augmented inside the exception itself.

## How Has This Been Tested?
Still thinking how we can automate this into a test.

CC @YoniKF

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -90,16 +90,18 @@ JAVA_SAFEMODE_DEFAULT_OPTIONS = [
 
 
 class JattachException(CalledProcessError):
-    def __init__(self, returncode, cmd, stdout, stderr, target_pid: int, ap_log: str):
+    def __init__(self, returncode, cmd, stdout, stderr, target_pid: int, ap_log: str, is_loaded: bool):
         super().__init__(returncode, cmd, stdout, stderr)
         self._target_pid = target_pid
         self._ap_log = ap_log
+        self.is_loaded = is_loaded
 
     def __str__(self):
         ap_log = self._ap_log.strip()
         if not ap_log:
             ap_log = "(empty)"
-        return super().__str__() + f"\nJava PID: {self._target_pid}\nasync-profiler log:\n{ap_log}"
+        loaded_msg = f"async-profiler DSO was{'' if self.is_loaded else ' not'} loaded"
+        return super().__str__() + f"\nJava PID: {self._target_pid}\n{loaded_msg}\nasync-profiler log:\n{ap_log}"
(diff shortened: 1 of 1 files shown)


adi-benz (approved):
Nice.  
I don't think we should have a test for this change, unless we don't want to break logs

Jongy:
> unless we don't want to break logs

I want to break logs (I mean, the format) - this will be easier to read, is all

Jongy:
Tested with this diff:
```
diff --git a/gprofiler/profilers/java.py b/gprofiler/profilers/java.py
index 3ee742c..6cecd89 100644
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -318,6 +318,7 @@ class AsyncProfiledProcess:
         try:
             # kill jattach with SIGTERM if it hangs. it will go down
             run_process(cmd, stop_event=self._stop_event, timeout=self._JATTACH_TIMEOUT, kill_signal=signal.SIGTERM)
+            raise CalledProcessError(5, "aaaa bbb", "a", "b")
         except CalledProcessError as e:  # catches timeouts as well
             if os.path.exists(self._log_path_host):
                 log = Path(self._log_path_host)
```

I get:
```
[2022-01-26 13:03:35,845] ERROR: gprofiler.profilers.profiler_base: JavaProfiler: failed to profile process 2118741
  File "/app/gprofiler/profilers/profiler_base.py", line 111, in snapshot
    result = future.result()
    return self.__get_result()
    raise self._exception
  File "/app/gprofiler/profilers/java.py", line 740, in _profile_process
    return self._profile_ap_process(ap_proc, comm)
  File "/app/gprofiler/profilers/java.py", line 751, in _profile_ap_process
    ap_proc.stop_async_profiler(with_output=False)
  File "/app/gprofiler/profilers/java.py", line 371, in stop_async_profiler
    self._run_async_profiler(self._get_stop_cmd(with_output))
  File "/app/gprofiler/profilers/java.py", line 334, in _run_async_profiler
    raise JattachException(
gprofiler.profilers.java.JattachException: Command 'aaaa bbb' returned non-zero exit status 5. 
stdout: a
stderr: b
Java PID: 2118741
async-profiler DSO was loaded
async-profiler log:
(empty)
```

adi-benz:
> I want to break logs (I mean, the format) - this will be easier to read, is all

I understand, my reply was to "Still thinking how we can automate this into a test." - I meant there's no reason to test this change, because tests are good to make sure stuff doesn't break, and I don't think it's interesting to make sure logs don't break

PR #277, opened 2022-01-26
Use process_exe() everywhere

Jongy:
## Description

See comment on process_exe() - it correctly raises ZombieProcess (which is a subclass
of NoSuchProcess) for zombie processes.
gProfiler should treat this error as processes going down.

These are all invocations of `Process.exe()` in gProfiler.

See https://github.com/giampaolo/psutil/pull/2062 which solves the issue in `psutil`.

## Motivation and Context
Currently we don't handle it as "process is down", so we get unrelated errors like `Non-java basenamed process, skipping... (disable --java-safemode to profile it anyway)` with `exe=""`, which happens because the process is a zombie, so I actually want it ignored.

### DIFF
--- a/gprofiler/utils.py
+++ b/gprofiler/utils.py
@@ -29,6 +29,7 @@ from typing import Callable, Iterator, List, Optional, Tuple, Union
 import importlib_resources
 import psutil
 from granulate_utils.linux.ns import run_in_ns
+from granulate_utils.linux.process import process_exe
 from psutil import Process
 
 from gprofiler.exceptions import (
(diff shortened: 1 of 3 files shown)


gprofiler/utils.py
```diff
@@ -266,7 +267,7 @@ def pgrep_exe(match: str) -> List[Process]:
     procs = []
     for process in psutil.process_iter():
         try:
-            if pattern.match(process.exe()):
+            if pattern.match(process_exe(process)):
                 procs.append(process)
         except psutil.NoSuchProcess:  # process might have died meanwhile
```
adi-benz, line 272: `pgrep_exe()` is unused in the project, so either remove this method, or `except ZombieProcess` as well

adi-benz (changes requested):

gprofiler/utils.py
Jongy: Keeping it because we might use it later, no reason to remove.

And `ZombieProcess` extends `NoSuchProcess` so it's okay :)

adi-benz: Oops, missed it in the description

adi-benz (approved):

PR #279, opened 2022-01-29
Return profiling error stack in all error/skip cases of _profile_process

Jongy:
## Description
As discussed in https://github.com/Granulate/gprofiler/pull/263#discussion_r785871160. The cases fixed here are much less common, but let's have them covered as well, and I changed the interface.

~~We still don't handle exceptions raised in `_profile_process`. I'll think how we want them handled - perhaps, automatically converted to an error stack by `ProcessProfilerBase.snapshot`?~~

## How Has This Been Tested?
* CI
* [x] Java process in a container going down - results in the expected stack.

### DIFF
--- a/gprofiler/profilers/profiler_base.py
+++ b/gprofiler/profilers/profiler_base.py
@@ -144,7 +144,7 @@ class ProcessProfilerBase(ProfilerBase):
                     result = self._profiling_error_stack("error", "process went down during profiling", comm)
                 except Exception as e:
                     logger.exception(f"{self.__class__.__name__}: failed to profile process {pid} ({comm})")
-                    result = self._profiling_error_stack("error", str(e), comm)
+                    result = self._profiling_error_stack("error", f"exception {type(e).__name__}", comm)
 
                 results[pid] = result
 
--- a/gprofiler/profilers/profiler_base.py
+++ b/gprofiler/profilers/profiler_base.py
@@ -92,7 +92,7 @@ class ProcessProfilerBase(ProfilerBase):
     def _select_processes_to_profile(self) -> List[Process]:
         raise NotImplementedError
 
-    def _profile_process(self, process: Process) -> Optional[StackToSampleCount]:
+    def _profile_process(self, process: Process) -> StackToSampleCount:
         raise NotImplementedError
 
     def snapshot(self) -> ProcessToStackSampleCounters:
--- a/gprofiler/profilers/profiler_base.py
+++ b/gprofiler/profilers/profiler_base.py
@@ -15,6 +15,7 @@ from gprofiler.exceptions import StopEventSetException
 from gprofiler.gprofiler_types import ProcessToStackSampleCounters, StackToSampleCount
 from gprofiler.log import get_logger_adapter
 from gprofiler.utils import limit_frequency
+from gprofiler.utils.process import process_comm
 
 logger = get_logger_adapter(__name__)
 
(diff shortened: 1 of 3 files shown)


gprofiler/profilers/python.py
```diff
@@ -119,13 +119,14 @@ def _make_command(self, pid: int, output_path: str) -> List[str]:
...
-    def _profile_process(self, process: Process) -> Optional[StackToSampleCount]:
-        try:
-            logger.info(
-                f"Profiling process {process.pid} with py-spy", cmdline=process.cmdline(), no_extra_to_server=True
-            )
-        except NoSuchProcess:
-            return None
+    @staticmethod
+    def _profiling_error_stack(reason: str, comm: str) -> StackToSampleCount:
+        # like _profiling_skipped_stack of java.py
```
DanielShaulov, line 124: Just move this to `profiler_base`?

gprofiler/profilers/python.py
```diff
@@ -119,13 +119,14 @@ def _make_command(self, pid: int, output_path: str) -> List[str]:
...
+        return Counter({f"{comm};[Profiling error: {reason}]": 1})
+
+    def _profile_process(self, process: Process) -> StackToSampleCount:
+        logger.info(f"Profiling process {process.pid} with py-spy", cmdline=process.cmdline(), no_extra_to_server=True)
```
DanielShaulov, line 128: You removed the `try:` around this, so if `process.cmdline` raises "NoSuchProcess", it will be caught in `profiler_base.py` and just written as log but no indicating stack.

If you move `_profiling_error_stack` to `profiiler_base` than you can just catch all those errors there and write the appropriate stack. It will also catch the other exceptions thrown in this function that didn't return `None` but just raised an exception

DanielShaulov (changes requested):

DanielShaulov:
> We still don't handle exceptions raised in _profile_process. I'll think how we want them handled - perhaps, automatically converted to an error stack by ProcessProfilerBase.snapshot?

Oh - I just saw this after I commented suggesting exactly that, lol. Let's do it.

gprofiler/profilers/python.py
Jongy: k

gprofiler/profilers/profiler_base.py
```diff
@@ -120,22 +121,31 @@ def snapshot(self) -> ProcessToStackSampleCounters:
...
-                        f"{self.__class__.__name__}: process went down during profiling {futures[future]}",
+                        f"{self.__class__.__name__}: process went down during profiling {pid} ({comm})",
                         exc_info=True,
                     )
-                except Exception:
-                    logger.exception(f"{self.__class__.__name__}: failed to profile process {futures[future]}")
+                    result = self._profiling_error_stack("error", "process went down during profiling", comm)
+                except Exception as e:
+                    logger.exception(f"{self.__class__.__name__}: failed to profile process {pid} ({comm})")
+                    result = self._profiling_error_stack("error", str(e), comm)
```
DanielShaulov, line 147: `str(e)` might be too much, maybe `type(e).__name__` is enough?

DanielShaulov (dismissed):
Minor comment that you may ignore

gprofiler/profilers/profiler_base.py
Jongy: Hmm I suppose we can't know in advance, I'll fix

DanielShaulov (approved):

PR #280, opened 2022-01-29
Move proc_events run_in_ns to granulate-utils

Jongy:
## Description
Move logic to granulate-utils, see https://github.com/Granulate/granulate-utils/pull/26

PR #281, opened 2022-01-30
Profile spawning processes

Jongy:
Based on:
* https://github.com/Granulate/gprofiler/pull/280
* https://github.com/Granulate/gprofiler/pull/279

## Description
This PR aims to solve the "short processes" issue: gProfiler runs profiling sessions of 60 seconds (that's the default). Upon each session, the different profilers (`JavaProfiler`, `PySpyProfiler` etc) scan through all currently running processes and invoke their respective profilers for each relevant process.

This works nicely for processes running for long durations (and still, it has some disadvantages which I'll describe later). However, it fails miserably in environments where processes are spawning and exiting quickly. For example, for an app with average lifetime of 10 seconds, we'll only see profiling data from the respective profiler for less than 1/6 of the time! (because on average, we'll catch only one out of 6 processes, and that process might already be halfway through its 10 seconds...). The situation was even worse for Java before https://github.com/Granulate/gprofiler/pull/265, because we didn't read the output file if the process exited before the profiling session has ended. So by design - processes shorter than 60 seconds were not profiled... (py-spy & rb-spy do not suffer from this problem; PyPerf & phpspy, by design, do not suffer from it).

This PR is the first step towards the "final" design of gProfiler, as I see it. I'll explain what I did here, then explain the next steps.

The `SpawningProcessProfilerBase` class was added (name T.B.D). It extends `ProcessProfilerBase` which is the base class for process-based profilers - profilers invoked per process (AP, py-spy, rbspy).
This class extends the existing behavior of `ProcessProfilerBase` using `proc_events`. Here's how it works:
1. `SpawningProcessProfilerBase.snapshot()` is called
2. `ProcessProfilerBase.snapshot()` is called. Upon selecting the currently running processes to profile, it notifies its base classes with the selected processes. At this point, `SpawningProcessProfilerBase` starts listening for new execs via `proc_events` (had it been listening earlier, we could "double catch" processes). This section is racy; fixing it is not too hard, but it's irrelevant in the final solution, and the race is not very likely anyway. So I don't intend to do it now.
3. Upon new execs, we place the `psutil.Process` object of the new process into a queue. It will now be checked for "should we profile it" multiple times over 1.5 seconds (with some basic exponential backoff logic). The reason we need to check multiple times is that our "should profile" checks are based on parsing `/proc/pid/maps`. Once the exec completes, libraries are not yet loaded (it is `ld.so`'s job). Furthermore, in the case of Java, `libjvm.so` is not even an (in)direct dependency of `java` - it's loaded later by whatever. This may take a short while, also depending on the load on the machine... In the case of Python, we sometimes select processes to profile based on `site-packages` in their maps - that definitely doesn't show up in maps right after the exec.
4. Once a spawned process was deemed relevant for profiling, we submit it to `_profile_process` in a TPE, with the duration decreased by how much time has passed since this session started.
5. Upon `ProcessProfilerBase.snapshot()` return, we also stop listening for new execs, and terminate the TPE.
6. Results from `ProcessProfilerBase.snapshot()` are merged with the results from our TPE.

Issues with the current impl:
* If used for rbspy, py-spy and AP: we're cloning the scheduler thread 3 times, Not sure it's that bad.
* `SpawningProcessProfilerBase` contains lots of logic that should be split.

### Roadmap & end goal
The final idea of gProfiler is this:
* *all* profilers run continuously (like perf, PyPerf & phpspy are now).
* Process-based profilers (AP, py-spy, rbspy) can run indefinitely, and dump output upon signal / request / process exit. So upon each session, we just "ask" profilers to give their current state.
* gProfiler runs a single thread using `proc_events`, and decides for each PID if any of the profilers should get it (as an optimization - phpsy & PyPerf should be modified to accept PIDs this way, instead of scanning themselves - scanning has a cost on the system, and thus if we have multiple unrelated "scanners", they run at lower frequency compared to the effectiveness of `proc_events`.
  * for low-priv executions, where we can't use `proc_events`, gProfiler will continue scanning for processes, but at a much higher interval (e.g each 0.5 seconds)

Roadmap:
1. Implement an alternative to `proc_events` for low-priv.
2. Get rid of `_select_processes_to_profile` - that is, don't "select processes" upon each session start, just select processes based on `proc_events`, and start profiling until the current session end.
3. Add support for continuous profiling in AP, rbspy, py-spy. I think that AP might have it (just gotta use the `dump` action instead of `stop`), not sure about the others.
4. Run all profilers independently of the profiling sessions, and just call a "dump data from all currently running py-spys" method.

## How Has This Been Tested?
T.B.A

## Checklist:
* [ ] Decide on the final design & implement it
* [ ] Make sure it is enabled for rbspy & py-spy as well.
* [ ] Add at least one automatic test, so that it runs in CI.
* [ ] Document theory of operation (I plan to add that in README)
* [ ] Document my future plans (as noted shortly in this PR)

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -39,7 +39,7 @@ from gprofiler.gprofiler_types import ProcessToStackSampleCounters, StackToSampl
 from gprofiler.kernel_messages import get_kernel_messages_provider
 from gprofiler.log import get_logger_adapter
 from gprofiler.merge import parse_one_collapsed
-from gprofiler.profilers.profiler_base import ProcessProfilerBase
+from gprofiler.profilers.profiler_base import SpawningProcessProfilerBase
 from gprofiler.profilers.registry import ProfilerArgument, register_profiler
 from gprofiler.utils import (
     TEMPORARY_STORAGE_PATH,
@@ -514,7 +514,7 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
         ),
     ],
 )
-class JavaProfiler(ProcessProfilerBase):
+class JavaProfiler(SpawningProcessProfilerBase):
     JDK_EXCLUSIONS = ["OpenJ9", "Zing"]
     # Major -> (min version, min build number of version)
     MINIMAL_SUPPORTED_VERSIONS = {
@@ -733,7 +733,7 @@ class JavaProfiler(ProcessProfilerBase):
 
         return False
 
-    def _profile_process(self, process: Process) -> Optional[StackToSampleCount]:
+    def _profile_process(self, process: Process, duration: int) -> Optional[StackToSampleCount]:
         comm = process_comm(process)
 
         if self._safemode_disable_reason is not None:
@@ -756,9 +756,11 @@ class JavaProfiler(ProcessProfilerBase):
         with AsyncProfiledProcess(
             process, self._storage_dir, self._stop_event, self._buildids, self._mode, self._ap_safemode, self._ap_args
         ) as ap_proc:
-            return self._profile_ap_process(ap_proc, comm)
+            return self._profile_ap_process(ap_proc, comm, duration)
 
-    def _profile_ap_process(self, ap_proc: AsyncProfiledProcess, comm: str) -> Optional[StackToSampleCount]:
+    def _profile_ap_process(
+        self, ap_proc: AsyncProfiledProcess, comm: str, duration: int
+    ) -> Optional[StackToSampleCount]:
         started = ap_proc.start_async_profiler(self._interval, ap_timeout=self._ap_timeout)
         if not started:
             logger.info(f"Found async-profiler already started on {ap_proc.process.pid}, trying to stop it...")
@@ -775,7 +777,7 @@ class JavaProfiler(ProcessProfilerBase):
                 )
 
         try:
-            wait_event(self._duration, self._stop_event, lambda: not is_process_running(ap_proc.process), interval=1)
+            wait_event(duration, self._stop_event, lambda: not is_process_running(ap_proc.process), interval=1)
         except TimeoutError:
             # Process still running. We will stop the profiler in finally block.
             pass
@@ -828,6 +830,10 @@ class JavaProfiler(ProcessProfilerBase):
             # continue - _profile_process will return an appropriate error for each process selected for
             # profiling.
         return pgrep_maps(r"^.+/libjvm\.so$")
+
+    def _should_profile_process(self, pid: int):
+        # TODO use psutil or at least make this code raise NoSuchProcess when appropriate.
+        return any(line.endswith("/libjvm.so") for line in Path(f"/proc/{pid}/maps").read_text().splitlines())
 
     def start(self) -> None:
         super().start()
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -832,8 +832,7 @@ class JavaProfiler(ProcessProfilerBase):
     def start(self) -> None:
         super().start()
         try:
-            # needs to run in init net NS - see netlink_kernel_create() call on init_net in cn_init().
-            run_in_ns(["net"], lambda: proc_events.register_exit_callback(self._proc_exit_callback), 1)
+            proc_events.register_exit_callback(self._proc_exit_callback)
         except Exception:
             logger.warning("Failed to enable proc_events listener for exited Java processes", exc_info=True)
         else:
--- a/gprofiler/profilers/profiler_base.py
+++ b/gprofiler/profilers/profiler_base.py
@@ -233,7 +233,6 @@ class SpawningProcessProfilerBase(ProcessProfilerBase):
                 with self._submit_lock:
                     if self._is_profiling_spawning:
                         # TODO ensure > 0 etc
-                        print("starting at interval", interval)
                         assert self._start_ts is not None and self._threads is not None
                         duration = self._duration - (time.monotonic() - self._start_ts)
                         self._futures[self._threads.submit(self._profile_process, process, int(duration))] = process.pid
--- a/gprofiler/profilers/profiler_base.py
+++ b/gprofiler/profilers/profiler_base.py
@@ -3,10 +3,15 @@
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
 
-import concurrent.futures
-from threading import Event
-from typing import List, Optional
+import sched
+import time
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from concurrent.futures._base import Future
+from threading import Event, Lock, Thread
+from typing import Dict, List, Optional
 
+from granulate_utils.linux.proc_events import register_exec_callback, unregister_exec_callback
+from granulate_utils.linux.process import is_process_running
 from psutil import NoSuchProcess, Process
 
 from gprofiler.exceptions import StopEventSetException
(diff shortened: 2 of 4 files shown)


adi-benz:
Completely agree with the roadmap and end goals.

About the currently proposed implementation, I don't see a reason to create `SpawningProcessProfilerBase`, why not change `ProcessProfilerBase`'s implementation to the new one proposed? We won't need the old implementation, and it would be easier not to separate them using inheritance  
The separation seems "unnatural", I think that having both just makes  `ProcessProfilerBase` a class that's only purpose in life is to be a part of `SpawningProcessProfilerBase`, but it still needs to keep the `ProfilerInterface` API without any good reason (no one will inherit from it directly)

The name even still fits, `ProcessProfilerBase` - it handles profilers that run per process, and that class make it seem to the API like a system-wide profiler

Jongy:
> Completely agree with the roadmap and end goals.
> 
> About the currently proposed implementation, I don't see a reason to create `SpawningProcessProfilerBase`, why not change `ProcessProfilerBase`'s implementation to the new one proposed? We won't need the old implementation, and it would be easier not to separate them using inheritance The separation seems "unnatural", I think that having both just makes `ProcessProfilerBase` a class that's only purpose in life is to be a part of `SpawningProcessProfilerBase`, but it still needs to keep the `ProfilerInterface` API without any good reason (no one will inherit from it directly)
> 
> The name even still fits, `ProcessProfilerBase` - it handles profilers that run per process, and that class make it seem to the API like a system-wide profiler

The reason I had in mind is - I wanted some way to disable this feature with a flag; that could be implemented by changing the base class.. Kinda ugly tho :sweat_smile: I also liked that their logic is separate. But I suppose I could merge them.

Jongy:
I wanna be able to disable the proc_events logic. I'll see if I can do it with a flag in the class, or if I desire to keep them separate.

gprofiler/profilers/profiler_base.py
```diff
@@ -119,41 +156,131 @@ def _profiling_error_stack(
...
+        super().start()
+
+        self._sched_thread.start()
+
+        try:
+            register_exec_callback(self._proc_exec_callback)
+        except Exception:
+            logger.warning("Failed to enable proc_events listener for executed processes", exc_info=True)
+        else:
+            self._enabled_proc_events = True
```
ty93, line 236: This also exists in `JavaProfiler`, so it's overwriting this one. If the Base manages to register its callback but `JavaProfiler` doesn't (or vice versa), then it'll be in inconsistent state.

ty93:
> Get rid of _select_processes_to_profile - that is, don't "select processes" upon each session start, just select processes based on proc_events, and start profiling until the current session end.

Wouldn't we still need it for the processes that already exist at the beginning of the session?

Jongy:
* Track only execs
* Track forks for Python processes

ty93:
> Track only execs
> Track forks for Python processes

```elif event["what"] == self._PROC_EVENT_EXEC:```
Currently only filtering EXECs (and EXITs), do you want me to add support for FORKs for python in this PR?

**Comment by Jongy** [comment]
> > Track only execs
> > Track forks for Python processes
> 
> `elif event["what"] == self._PROC_EVENT_EXEC:` Currently only filtering EXECs (and EXITs), do you want me to add support for FORKs for python in this PR?

Nope - FORKs later, in another PR.

File: gprofiler/profilers/python.py [hunk 2]
```diff
@@ -245,27 +246,36 @@ def _select_processes_to_profile(self) -> List[Process]:
...
+    def _should_profile_process(self, process: Process) -> bool:
+        return any(
+            re.match(DETECTED_PYTHON_PROCESSES_REGEX, line) for line in read_proc_file(process, "maps").splitlines()
+        ) and not self._should_skip_process(process)
+
+    def _should_skip_process(self, process: Process) -> bool:
+        if process.pid == os.getpid():
+            return True
+
+        cmdline = " ".join(process.cmdline())
```
Jongy, line 265: Why's the `" ".join()` change? It's not required (all processes are the argv[0])

ty93: For example, if the cmdline is `/usr/bin/python3 /usr/bin/networkd-dispatcher --run-startup-triggers`, then checking if `networkd-dispatcher` is `in` the cmdline as a list will return false, while checking if it's `in` the cmdline as a string will return true.

gprofiler/profilers/profiler_base.py
Jongy: > Personally I think this is one of the reasons to use more compositions over inheritance here.

Makes sense. Next PR

gprofiler/profilers/java.py
```diff
@@ -1025,6 +1026,9 @@ def _select_processes_to_profile(self) -> List[Process]:
             # profiling.
         return pgrep_maps(DETECTED_JAVA_PROCESSES_REGEX)
 
+    def _should_profile_process(self, process: Process) -> bool:
+        return any(line.endswith("/libjvm.so") for line in read_proc_file(process, "maps").splitlines())
```
Jongy, line 1030: Use `DETECTED_JAVA_PROCESSES_REGEX` (and just `re.search()` on the file can be enough, no need for line split)

gprofiler/profilers/python.py
```diff
@@ -245,27 +246,36 @@ def _select_processes_to_profile(self) -> List[Process]:
...
+                if not self._should_skip_process(process):
+                    filtered_procs.append(process)
             except Exception:
                 logger.exception(f"Couldn't add pid {process.pid} to list")
 
         return filtered_procs
 
+    def _should_profile_process(self, process: Process) -> bool:
+        return any(
+            re.match(DETECTED_PYTHON_PROCESSES_REGEX, line) for line in read_proc_file(process, "maps").splitlines()
```
Jongy, line 258: I don't think you need the line split?

Although.. to mimic `pgrep_maps`, I suppose we do. Wdyt

gprofiler/utils/process.py
```diff
@@ -14,15 +14,19 @@ def ensure_running(process: Process) -> None:
...
+def read_proc_file(process: Process, file: str) -> str:
     try:
-        status = Path(f"/proc/{process.pid}/status").read_text()
-    except FileNotFoundError:
-        raise NoSuchProcess(process.pid)
+        data = Path(f"/proc/{process.pid}/{file}").read_text()
+    except FileNotFoundError as e:
+        raise NoSuchProcess(process.pid) from e
     else:
         # ensures we read the right comm (i.e PID was not reused)
```
Jongy, line 23:
```suggestion
        # ensures we read the right file (i.e PID was not reused)
```

tests/test_sanity.py
```diff
@@ -144,14 +149,60 @@ def test_from_container(
...
+    output_collapsed: Path,
+    assert_collapsed: AssertInCollapsed,
+    profiler_flags: List[str],
+    application_factory: Callable[[], _GeneratorContextManager],
+) -> None:
+    args = runtime_specific_args.copy()
+    if "-d" in args:
+        args[args.index("-d") + 1] = "10"
+    else:
+        args.extend(["-d", "10"])
```
Jongy, line 188:
```suggestion
    args.extend(["-d", "10"])
```

it's okay to supply it multiple times in any case

tests/test_sanity.py
```diff
@@ -144,14 +149,60 @@ def test_from_container(
...
+    )
+
+    try:
+        wait_event(30, Event(), lambda: b"starting profiling spawning processes" in container.logs())
+    except TimeoutError:
+        print(container.logs())
+        raise
+
+    with application_factory():
+        collapsed_text = wait_for_gprofiler_container(container, output_collapsed)
```
Jongy, line 206: What proves that it's the "spawning processes" mechanism that caught the process is - we waited for gprofiler to do the first `_select_processes_to_profile()`  and only then we spawned the app, and gprofiler runs non-continuous so if app results exist in the final collapsed, we know it's due to spawning. Right? Maybe comment

gprofiler/profilers/profiler_base.py
```diff
@@ -196,65 +185,79 @@ class SpawningProcessProfilerBase(ProcessProfilerBase):
...
     def _is_profiling_spawning(self) -> bool:
         return self._threads is not None
 
-    def _start_profiling_spawning(self) -> None:
+    def _start_profiling_spawning(self, processes: List[Process]) -> None:
+        logger.debug(f"{self.__class__.__name__}: starting profiling spawning processes")
         with self._submit_lock:
             self._start_ts = time.monotonic()
             self._threads = ThreadPoolExecutor()
+            self._preexisting_pids = list(map(lambda p: p.pid, processes))
```
Jongy, line 224:
```suggestion
            self._preexisting_pids = [p.pid for p in processes]
```

gprofiler/profilers/profiler_base.py
```diff
@@ -196,65 +185,79 @@ class SpawningProcessProfilerBase(ProcessProfilerBase):
...
-        except Exception:
-            logger.warning("Failed to enable proc_events listener for executed processes", exc_info=True)
-        else:
-            self._enabled_proc_events = True
+            try:
+                register_exec_callback(self._proc_exec_callback)
+            except Exception:
+                logger.warning("Failed to enable proc_events listener for executed processes", exc_info=True)
+            else:
+                self._enabled_proc_events = True
```
Jongy, line 249: Let's name `_enabled_proc_events` differently so there's no chance `java.py` could accidentally use it. (`_enabled_proc_events_spawning`? idk)

Jongy:
I went over all changes @ty93 . No major comment. Let's fix these and I'm good to merge this as an experimental feature, disabled by default.

gprofiler/profilers/profiler_base.py
ty93: I have Java brain-rot

Jongy: haha

ty93 (dismissed):

gprofiler/profilers/profiler_base.py
```diff
@@ -239,6 +238,7 @@ def start(self) -> None:
         super().start()
 
         if self._profile_spawned_processes:
+            logger.debug(f"{self.__class__.__name__}: starting profiling spawning processes")
```
ty93, line 241: Maybe the log message wasn't good enough, but it was in `_start_profiling_spawning` on purpose. I use this log message in the test to make sure that the profiling session started and that the profiler actually started detecting spawning processes. If the log is here then we can't know for sure that it won't catch the process as a preexisting process for the session.

Jongy: Hmm. It was confusing on that function because it kept happening forever and it was printed also when profiling spawned processes was disabled.

Maybe can be sovled with mocking? (or we run the container here...)

Jongy: I have another idea, pushing a patch

gprofiler/profilers/profiler_base.py
```diff
@@ -119,41 +157,162 @@ def _profiling_error_stack(
...
+            assert self._threads is not None
+            threads = self._threads
+            # delete it before blocking on the exit of all threads (to ensure no new work
+            # is added)
+            self._threads = None
+            threads.shutdown()  # waits (although - all are done by now)
+            self._preexisting_pids = None
+
+    def _proc_exec_callback(self, tid: int, pid: int) -> None:
+        with contextlib.suppress(NoSuchProcess):
```
glipner, line 241: Should check here if we're not stopping, otherwise we might still block forever in `self._sched.run()`
```suggestion
        if not self._is_profiling_spawning:
            return

        with contextlib.suppress(NoSuchProcess):
```

gprofiler/profilers/profiler_base.py
```diff
@@ -119,41 +157,162 @@ def _profiling_error_stack(
...
+        assert set(results).intersection(results_spawned) == set()
+        results.update(results_spawned)
         return results
+
+    def _sched_thread_run(self) -> None:
+        while not (self._stop_event.is_set() or self._sched_stop):
+            self._sched.run()
+            self._stop_event.wait(0.1)
+
+    def _check_process(self, process: Process, interval: float) -> None:
```
glipner, line 289: Clear `self._sched` if we're stopping, so that `self._sched.run()` exits faster.
```suggestion
    def _clear_shced(self):
        for event in self._sched.queue:
            self._sched.cancel(event)

    def _check_process(self, process: Process, interval: float) -> None:
        with contextlib.suppress(NoSuchProcess):
            if not self._is_profiling_spawning:
                self._clear_sched()
                return

            if not is_process_running(process) or process.ppid() == os.getpid():
```

glipner (dismissed):
.

gprofiler/profilers/profiler_base.py
```diff
@@ -119,41 +157,162 @@ def _profiling_error_stack(
...
+
+    def _stop_profiling_spawning(self) -> None:
+        with self._submit_lock:
+            self._start_ts = None
+            assert self._threads is not None
+            threads = self._threads
+            # delete it before blocking on the exit of all threads (to ensure no new work
+            # is added)
+            self._threads = None
+            threads.shutdown()  # waits (although - all are done by now)
```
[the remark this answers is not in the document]
Jongy: Why does it matter, tho? for safety let's add it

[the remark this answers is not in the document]
glipner: Doesn't really matter.
We're wasting time waiting rather than reading profiling output :-P

gprofiler/profilers/profiler_base.py
Jongy: Fixed it @ty93

gprofiler/profilers/profiler_base.py
Jongy: The `clear_sched` call needn't be in suppress

glipner: I agree

glipner:
```suggestion
    def _clear_shced(self):
        for event in self._sched.queue:
            self._sched.cancel(event)
            
    def _check_process(self, process: Process, interval: float) -> None:
        if not self._is_profiling_spawning:
            self._clear_sched()
            return
        
        with contextlib.suppress(NoSuchProcess):
            if not is_process_running(process) or process.ppid() == os.getpid():```

ty93: Premature optimization if I've ever seen one

Jongy:
Something is messed up with the tests :/

ty93 (approved):

PR #282, opened 2022-01-30
mypy: Enable mypy strict mode

adi-benz:
## Description
From now on, mypy will be set to `strict = True` it its config file.

## Motivation and Context
`pre-commit` and CI both run `lint.sh` that runs `mypy .`, but mypy wasn't strict, so it didn't check much, now it will.

## How Has This Been Tested?

* Made sure `pre-commit` hook works
* Made sure CI catches problems

## Checklist:
- [X] I have read the **CONTRIBUTING** document.
- [x] I have tested mypy fails CI when appropriate

### DIFF
--- a/gprofiler/client.py
+++ b/gprofiler/client.py
@@ -6,7 +6,7 @@ import datetime
 import gzip
 import json
 from io import BytesIO
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Any
+from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Any, cast
 
 import requests
 from requests import Session
@@ -63,7 +63,7 @@ class APIClient:
         self,
         method: str,
         path: str,
-        data: Optional[Dict],
+        data: Any,
         files: Dict = None,
         timeout: float = DEFAULT_REQUEST_TIMEOUT,
         api_version: str = None,
@@ -101,21 +101,21 @@ class APIClient:
                 raise APIError(resp.text)
         else:
             resp.raise_for_status()
-        return resp.json()
+        return cast(dict, resp.json())
 
-    def get(self, path: str, data: Optional[Dict] = None, **kwargs: Any) -> Dict:
+    def get(self, path: str, data: Any = None, **kwargs: Any) -> Dict:
         return self._send_request("GET", path, data, **kwargs)
 
-    def post(self, path: str, data: Optional[Dict] = None, **kwargs: Any) -> Dict:
+    def post(self, path: str, data: Any = None, **kwargs: Any) -> Dict:
         return self._send_request("POST", path, data, **kwargs)
 
-    def put(self, path: str, data: Optional[Dict] = None, **kwargs: Any) -> Dict:
+    def put(self, path: str, data: Any = None, **kwargs: Any) -> Dict:
         return self._send_request("PUT", path, data, **kwargs)
 
-    def patch(self, path: str, data: Optional[Dict] = None, **kwargs: Any) -> Dict:
+    def patch(self, path: str, data: Any = None, **kwargs: Any) -> Dict:
         return self._send_request("PATCH", path, data, **kwargs)
 
-    def delete(self, path: str, data: Optional[Dict] = None, **kwargs: Any) -> Dict:
+    def delete(self, path: str, data: Any = None, **kwargs: Any) -> Dict:
         return self._send_request("DELETE", path, data, **kwargs)
 
     def get_health(self) -> Dict:
--- a/gprofiler/client.py
+++ b/gprofiler/client.py
@@ -6,7 +6,7 @@ import datetime
 import gzip
 import json
 from io import BytesIO
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Any, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, cast
 
 import requests
 from requests import Session
--- a/gprofiler/client.py
+++ b/gprofiler/client.py
@@ -6,7 +6,7 @@ import datetime
 import gzip
 import json
 from io import BytesIO
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, cast, IO
 
 import requests
 from requests import Session
@@ -82,7 +82,7 @@ class APIClient:
             buffer = BytesIO()
             with gzip.open(buffer, mode="wt", encoding="utf-8") as gzip_file:
                 try:
-                    json.dump(data, gzip_file, ensure_ascii=False)  # type: ignore
+                    json.dump(data, cast(IO[str], gzip_file), ensure_ascii=False)
                 except TypeError:
                     # This should only happen while in development, and is used to get a more indicative error.
                     bad_json = str(data)
--- a/gprofiler/client.py
+++ b/gprofiler/client.py
@@ -40,7 +40,7 @@ class APIClient:
         self._init_session()
         logger.info(f"The connection to the server was successfully established (service {service!r})")
 
-    def _init_session(self):
+    def _init_session(self) -> None:
         self._session: Session = requests.Session()
         self._session.headers.update({"GPROFILER-API-KEY": self._key, "GPROFILER-SERVICE-NAME": self._service})
 
--- a/gprofiler/client.py
+++ b/gprofiler/client.py
@@ -6,7 +6,7 @@ import datetime
 import gzip
 import json
 from io import BytesIO
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, cast, IO
+from typing import IO, TYPE_CHECKING, Any, Dict, List, Optional, Tuple, cast
 
 import requests
 from requests import Session
--- a/gprofiler/client.py
+++ b/gprofiler/client.py
@@ -6,7 +6,7 @@ import datetime
 import gzip
 import json
 from io import BytesIO
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Any
 
 import requests
 from requests import Session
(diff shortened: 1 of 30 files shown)


gprofiler/client.py
```diff
@@ -63,7 +63,7 @@ def _send_request(
         self,
         method: str,
         path: str,
-        data: Optional[Dict],
+        data: Any,
```
Jongy, line 66: Why? Is it really not always a dict?

gprofiler/exceptions.py
```diff
@@ -28,11 +28,13 @@ def __str__(self):
 
 
 class CalledProcessTimeoutError(CalledProcessError):
-    def __init__(self, timeout: float, returncode: int, cmd: Union[str, List[str]], output=None, stderr=None):
+    def __init__(
+        self, timeout: float, returncode: int, cmd: Union[str, List[str]], output: Any = None, stderr: Any = None
```
Jongy, line 32: `output` is always `str`, no? And `stderr` as well

gprofiler/main.py
```diff
@@ -228,34 +233,34 @@ def start(self):
...
 
-    def _snapshot(self):
+    def _snapshot(self) -> None:
         local_start_time = datetime.datetime.utcnow()
         monotonic_start_time = time.monotonic()
         process_profilers_futures = []
         for prof in self.process_profilers:
             prof_future = self._executor.submit(prof.snapshot)
-            prof_future.name = prof.name
+            prof_future.name = prof.name  # type: ignore
```
Jongy, line 252: `cast`?

Jongy: Ahh, it's due to `prof_future.name` missing. Indeed a hack... lol. Can comment that next to the type ignore?

gprofiler/metadata/system_metadata.py
```diff
@@ -178,7 +181,7 @@ def get_static_system_info() -> SystemInfo:
         kernel_version=uname.version,
         system_name=uname.system,
         processors=cpu_count,
-        memory_capacity_mb=round(psutil.virtual_memory().total / 1024 / 1024),
+        memory_capacity_mb=round(psutil.virtual_memory().total / 1024 / 1024),  # type: ignore
```
Jongy, line 184: This line used to have `type ignore` and then we removed it. Now, why does it want that again?

gprofiler/client.py
adi-benz: yes :(

From `log.py:191`, a list is passed:
```python
            self._api_client.post(self._path, data=self._logs[:logs_count], api_version="v1")
```

gprofiler/exceptions.py
adi-benz: They are, but I wanted to match `CalledProcessError`'s `__init__`, and he allows `Any`

gprofiler/main.py
adi-benz: Indeed a hack

gprofiler/metadata/system_metadata.py
adi-benz: Yea, it's shit, `types-psutil` [doesn't specify a return type](https://github.com/python/typeshed/search?q=virtual_memory) for `virtual_memory()`, so now it warns:
```bash
gprofiler/metadata/system_metadata.py:184: error: Call to untyped function "virtual_memory" in typed context
```

Jongy: Blehh

gprofiler/exceptions.py
Jongy: But if we always pass `str` here, why not require it?

gprofiler/client.py
Jongy: Oof

gprofiler/profilers/java.py
```diff
@@ -833,7 +843,7 @@ def start(self) -> None:
         super().start()
         try:
             # needs to run in init net NS - see netlink_kernel_create() call on init_net in cn_init().
-            run_in_ns(["net"], lambda: proc_events.register_exit_callback(self._proc_exit_callback), 1)
+            run_in_ns(["net"], lambda: proc_events.register_exit_callback(self._proc_exit_callback), 1)  # type: ignore
```
Jongy, line 846: Why is it `type ignore` here?

Generally - I suggest that, wherever it's not obvious, we add a comment explaining the type ignore. So the reader doesn't have to wonder, and also so we can slowly eliminate them as things progress and we solve typing issues.

gprofiler/utils/__init__.py
```diff
@@ -206,8 +208,8 @@ def run_process(
     kill_signal: signal.Signals = signal.SIGKILL,
     communicate: bool = True,
     stdin: bytes = None,
-    **kwargs,
-) -> CompletedProcess:
+    **kwargs: Any,
+) -> "CompletedProcess[bytes]":
```
Jongy, line 212:
```suggestion
) -> CompletedProcess[bytes]:
```

(it's already declared)

gprofiler/profilers/python.py
```diff
@@ -335,7 +335,9 @@ def start(self):
             wait_event(self._POLL_TIMEOUT, self._stop_event, lambda: os.path.exists(self.output_path))
         except TimeoutError:
             process.kill()
-            logger.error(f"PyPerf failed to start. stdout {process.stdout.read()!r} stderr {process.stderr.read()!r}")
+            stdout = process.stdout.read()  # type: ignore
```
Jongy, line 338: Why is it needed? Because `stdout` might be None, or because it might not be `TextIOWrapper`?

If the former - then I think that adding `assert process.stdout is not None` is preferred.

mypy.ini
```diff
@@ -3,6 +3,11 @@ python_version = 3.6
 warn_unused_configs = True
 exclude = venv/|deploy/dataflow/|granulate-utils
 platform = linux
+strict = True
+no_implicit_optional = False
+disallow_any_generics = False
+warn_redundant_casts = True
```
Jongy, line 9: Cool, I wasn't aware of `warn_redundant_casts` and `warn_unused_ignores`

tests/conftest.py
```diff
@@ -255,7 +259,7 @@ def output_directory(tmp_path: Path) -> Path:
 def application_pid(
     in_container: bool, application_process: subprocess.Popen, application_docker_container: Container
 ) -> int:
-    pid = application_docker_container.attrs["State"]["Pid"] if in_container else application_process.pid
+    pid: int = application_docker_container.attrs["State"]["Pid"] if in_container else application_process.pid
```
Jongy, line 262: Why does it need `: int` here if we're assigning an `int` right away?

tests/test_java.py
```diff
@@ -60,9 +69,9 @@ def test_async_profiler_already_running(application_pid, assert_collapsed, tmp_p
             ap_safemode=0,
             ap_args="",
         ) as ap_proc:
-            ap_proc.status_async_profiler()
+            ap_proc.status_async_profiler()  # type: ignore
```
Jongy, line 72: It's our class :/ why `type ignore`

Jongy (changes requested):
Wonderful.

I wonder, did it catch any bugs now? I didn't notice anything, but I have manually caught at least 2 bugs in the tests that `mypy --strict` would've caught.

Anyway it's about time we'll have strict enabled.

Jongy:
> pre-commit and CI both run lint.sh that runs mypy ., but mypy basically checked nothing, now it will.

What do you mean "checked nothing"? Did it not look at its config file correctly? Or you just mean that we didn't have `strict` enabled?

gprofiler/utils/__init__.py
adi-benz: `CompletedProcess` isn't generic at runtime, only at type check time (it's generic in `typeshed`), so removing the double quotes results in `TypeError: 'type' object is not subscriptable`.

See [this](https://stackoverflow.com/questions/68089860/completedprocess-is-a-generic-for-mypy)

If we were to drop support for Python 3.6, we could remove the double quotes, and `from __future__ import annotations`

tests/conftest.py
adi-benz: `application_docker_container.attrs["State"]["Pid"]`, mypy can't be sure it's an `int`

tests/test_java.py
adi-benz: Glad you asked :)

The function `status_async_profiler` isn't defined in `AsyncProfiledProcess`, it's defined in `AsyncProfiledProcessForTests` that inherits from `AsyncProfiledProcess`, and `AsyncProfiledProcess::_enter__()` returns `AsyncProfiledProcess`.

I didn't want to make `AsyncProfiledProcess` generic because I felt it's a hack just got testing, but now I think we should.

Changed `__enter__()`'s return type to be `T` and removed `type: ignore`s

gprofiler/profilers/java.py
adi-benz: Removed some more ignores

adi-benz:
> Wonderful.
> 
> I wonder, did it catch any bugs now? I didn't notice anything, but I have manually caught at least 2 bugs in the tests that `mypy --strict` would've caught.
> 
> Anyway it's about time we'll have strict enabled.

Actually no, all errors were a result of miss typing (there was [this](https://github.com/Granulate/gprofiler/pull/282#discussion_r795185480), but it's a type error)

adi-benz:
> > pre-commit and CI both run lint.sh that runs mypy ., but mypy basically checked nothing, now it will.
> 
> What do you mean "checked nothing"? Did it not look at its config file correctly? Or you just mean that we didn't have `strict` enabled?

I meant it didn't have `strict` enabled

gprofiler/profilers/java.py
Jongy: Yay

gprofiler/utils/__init__.py
Jongy: Blehhh

We will drop 3.6 support at some point, but not in this PR :sweat_smile:

tests/conftest.py
Jongy: Hmm, I supposed that you need to use `cast` in that case

tests/test_java.py
Jongy: > The function status_async_profiler isn't defined in AsyncProfiledProcess, it's defined in AsyncProfiledProcessForTests that inherits from AsyncProfiledProcess, and AsyncProfiledProcess::_enter__() returns AsyncProfiledProcess.

Hmmm. I wonder what is the correct solution here.

gprofiler/profilers/java.py
```diff
@@ -846,7 +846,7 @@ def start(self) -> None:
         super().start()
         try:
             # needs to run in init net NS - see netlink_kernel_create() call on init_net in cn_init().
-            run_in_ns(["net"], lambda: proc_events.register_exit_callback(self._proc_exit_callback), 1)  # type: ignore
+            run_in_ns(["net"], lambda: cast(None, proc_events.register_exit_callback(self._proc_exit_callback)), 1)
```
Jongy: Why is it needed? o.O `run_in_ns` has type T

tests/test_java.py
adi-benz: You can't have the cake and eat it too.

`__enter__()` has to declare it returns the class (instead of the hard-coded `AsyncProfiledProcess`), and if we don't want that, then we shouldn't inherit from it and add functions.  
`AsyncProfiledProcess` can either support inheritance, or not, and in that case, don't inherit from it (tests shouldn't be an exception to that)

So I chose to support inheritance and treat the testing code as a legitimate use-case

gprofiler/profilers/java.py
Jongy: Removed later

adi-benz: > gprofiler/profilers/java.py:848: error: Returning Any from function declared to return "object"

I guess `T` is an `object` and it `Any` is not  
See [this issue](https://github.com/python/mypy/issues/9590)

Anyway, this code is irrelevant after merging with `master`

tests/test_java.py
Jongy: I'm satisfied with the solution here, per my understanding from https://stackoverflow.com/a/59937840/797390

Jongy (dismissed):
Merge from msater

Jongy (approved):
Fix tests

PR #284, opened 2022-01-31
utils: Raise NoSuchProcess from process_comm() if process is missing

Jongy:
Avoid the `FileNotFoundError` - `NoSuchProcess` is handled gracefully.

### DIFF
--- a/gprofiler/utils/process.py
+++ b/gprofiler/utils/process.py
@@ -1,0 +1,15 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+
+from pathlib import Path
+
+from psutil import Process
+
+
+def process_comm(process: Process) -> str:
+    status = Path(f"/proc/{process.pid}/status").read_text()
+    name_line = status.splitlines()[0]
+    assert name_line.startswith("Name:\t")
+    return name_line.split("\t", 1)[1]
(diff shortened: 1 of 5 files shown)


gprofiler/utils/process.py
```diff
@@ -0,0 +1,19 @@
...
+#
+
+from pathlib import Path
+
+from psutil import NoSuchProcess, Process
+
+
+def process_comm(process: Process) -> str:
+    try:
+        status = Path(f"/proc/{process.pid}/status").read_text()
```
adi-benz, line 13: Should we make sure we're checking the right pid by calling `process.is_running()` after reading the file?  
That's what we did in [get_process_nspid](https://github.com/Granulate/granulate-utils/pull/20/files#diff-47d7a7028052d61df24913a34657a7edc5efc625dfd980446d52aeb8de517d8aR141) PR

I'm slightly worried with the performance implications of performing this check in a function that might be called in a loop

WDUT?

adi-benz (changes requested):

gprofiler/utils/process.py
Jongy: Hmm a bit extreme (PID reuse this quick is not something I expect).

But... we have decided to do so in the past, you're right. Let's keep the standard, it's a small check anyway.

Jongy: I don't think the performance is meaningful here, and if it is - we'll see that in the profiler :) Judging by a quick look in psutil, it's just one syscall. Negligible.

adi-benz (approved):

PR #285, opened 2022-02-01
appids: Support gunicorns which use setproctitle()

Jongy:
## Description
Support another "flavor" of proc names that gunicorn uses, depending on its config.

## How Has This Been Tested?
I ran `Process(pid).cmdline()` on a system which has a gunicorn that uses `setproctitle()`. I took the output and ran it via this new class - and it returned the correct appid. This was the input:
```
['gunicorn: worker [myapp.wsgi:app]', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '']
```

### DIFF
--- a/gprofiler/metadata/application_identifiers.py
+++ b/gprofiler/metadata/application_identifiers.py
@@ -85,7 +85,14 @@ class _ApplicationIdentifier(metaclass=ABCMeta):
         pass
 
 
-class _GunicornApplicationIdentifier(_ApplicationIdentifier):
+class _GunicornApplicationIdentifierBase(_ApplicationIdentifier):
+    def gunicorn_to_application_name(self, wsgi_app_spec: str, process: Process) -> str:
+        # strips the :app
+        wsgi_app = wsgi_app_spec.split(":", maxsplit=1)[0]
+        return f"gunicorn: {_append_python_module_to_proc_wd(process, wsgi_app)}"
+
+
+class _GunicornApplicationIdentifier(_GunicornApplicationIdentifierBase):
     def get_application_name(self, process: Process) -> Optional[str]:
         # As of gunicorn documentation the WSGI module name most probably will come from the cmdline and not from the
         # config file / environment variables (they added the option to specify `wsgi_app`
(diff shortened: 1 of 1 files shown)


gprofiler/metadata/application_identifiers.py
```diff
@@ -101,6 +101,27 @@ def get_application_name(self, process: Process) -> Optional[str]:
...
+        title format:
+        https://github.com/benoitc/gunicorn/blob/60d0474a6f5604597180f435a6a03b016783885b/gunicorn/arbiter.py#L580
+    """
+
+    def get_application_name(self, process: Process) -> Optional[str]:
+        cmdline = process.cmdline()
+        # There should be one entry in the commandline, starting with "gunicorn: ",
+        # and the rest should be empty strings per Process.cmdline() (I suppose that setproctitle
+        # zeros out the arguments array).
+        if cmdline[0].startswith("gunicorn: ") and len(list(filter(lambda s: s, cmdline))) == 1:
```
guybortnikov, line 120: cmdline may be empty, use `_get_cli_arg_by_index`

gprofiler/metadata/application_identifiers.py
```diff
@@ -101,6 +101,27 @@ def get_application_name(self, process: Process) -> Optional[str]:
...
+            return cmdline[0]
```
guybortnikov, line 121: Let's normalize the `gunicorn: ` app id, so it will be the same in both `_GunicornTitleApplicationIdentifier` and `_GunicornApplicationIdentifier`. Currently in `_GunicornApplicationIdentifier` we return the following format: `gunicorn: <wsgi module>`

guybortnikov (changes requested):
small comments

gprofiler/metadata/application_identifiers.py
Jongy: Okay, better this way really.

guybortnikov (approved):
Trust you on the regex

PR #286, opened 2022-02-02
Bugfix/ uwsgi-config-file-parsing

guybortnikov:
## Description
Fix a small bug, when trying to identify applications, we may parse uwsgi config file which is an ini file, the file format allows overriding config options, although by default Python's [configparser](https://docs.python.org/3.9/library/configparser.html)
raises exception upon this scenario, this PR fixes this issue by disabling this behavior. 

## Related Issue
See explanation below.


Jongy (approved):

guybortnikov:
Tested locally, works, merging

PR #291, opened 2022-02-09
Python: Bugfix: Don't emit incorrect Python versions

YishaiZinkin:
## Description
Previously it was possible to get an incorrect Python versions in cases where the python versions regex matched a wrong string.
This PR prevents it by narrowing down the possible Python versions to expected ones only (2.7 and 3.5-3.12), and by not presenting a Python version in case the regex had more than a single match.

## How Has This Been Tested?
I tested it in a container with an application for which we previously got an incorrect version (conveniently, gprofiler itself...).

### DIFF
--- a/gprofiler/metadata/py_module_version.py
+++ b/gprofiler/metadata/py_module_version.py
@@ -147,7 +147,8 @@ def _get_libpython_path(process: Process) -> Optional[str]:
 
 
 # Matches PY_VERSION in Python's binary
-_PY_VERSION_STRING_PATTERN = re.compile(rb"(?<=\D)(?:2\.7|3\.1?\d)\.\d\d?(?=\x00)")
+# Notice that we'll match only 2.7 and 3.5-3.12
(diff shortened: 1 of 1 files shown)


gprofiler/metadata/py_module_version.py
```diff
@@ -147,7 +147,8 @@ def _get_libpython_path(process: Process) -> Optional[str]:
 
 
 # Matches PY_VERSION in Python's binary
-_PY_VERSION_STRING_PATTERN = re.compile(rb"(?<=\D)(?:2\.7|3\.1?\d)\.\d\d?(?=\x00)")
+# Notice that we'll match only 2.7 and 3.5-3.12
+_PY_VERSION_STRING_PATTERN = re.compile(rb"(?<=\D)(?:2\.7|3\.(?:[5-9]|1[0-2]))\.\d\d?(?=\x00)")
```
Jongy, line 151: Don't need parenthesis around both sides of the | ?

Jongy: Nope tested

Jongy (approved):

Jongy (approved):

PR #299, opened 2022-02-17
WIP Application metadata

Jongy:
## Description
MVP of application metadata: retrieve per-pid information like runtime version, file buildids, ...

Tasks:
* [ ] Update protocol version to v3
* [ ] Make sure that if v1 is given, no app metadata & no container names are sent (and strip the two `;`)
* [ ] Remove code dups
* [ ] Rethink of design with objects instead of classes?

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -22,12 +22,13 @@ from granulate_utils.linux.process import is_process_running
 from psutil import NoSuchProcess, Process
 from requests import RequestException, Timeout
 
-from gprofiler import __version__, merge
+from gprofiler import __version__
 from gprofiler.client import DEFAULT_UPLOAD_TIMEOUT, GRANULATE_SERVER_HOST, APIClient
 from gprofiler.docker_client import DockerClient
 from gprofiler.exceptions import APIError, SystemProfilerInitFailure
 from gprofiler.gprofiler_types import ProcessToStackSampleCounters, UserArgs, positive_integer
 from gprofiler.log import RemoteLogsHandler, initial_root_logger_setup
+from gprofiler.merge import EnrichmentOptions, concatenate_profiles, merge_profiles
 from gprofiler.metadata.metadata_collector import get_current_metadata, get_static_metadata
 from gprofiler.metadata.metadata_type import Metadata
 from gprofiler.metadata.system_metadata import get_hostname, get_run_mode, get_static_system_info
@@ -83,12 +84,11 @@ class GProfiler:
         client: Optional[APIClient],
         collect_metrics: bool,
         collect_metadata: bool,
-        identify_applications: bool,
+        enrichment_options: EnrichmentOptions,
         state: State,
         usage_logger: UsageLoggerInterface,
         user_args: UserArgs,
         duration: int,
-        include_container_names: bool = True,
         profile_api_version: Optional[str] = None,
         remote_logs_handler: Optional[RemoteLogsHandler] = None,
         controller_process: Optional[Process] = None,
@@ -102,7 +102,7 @@ class GProfiler:
         self._profile_api_version = profile_api_version
         self._collect_metrics = collect_metrics
         self._collect_metadata = collect_metadata
-        self._identify_applications = identify_applications
+        self._enrichment_options = enrichment_options
         self._stop_event = Event()
         self._static_metadata: Optional[Metadata] = None
         self._spawn_time = time.time()
@@ -127,7 +127,7 @@ class GProfiler:
         except SystemProfilerInitFailure:
             logger.exception("System profiler initialization has failed, exiting...")
             sys.exit(1)
-        if include_container_names and profile_api_version != "v1":
+        if self._enrichment_options.container_names and profile_api_version != "v1":
             self._docker_client: Optional[DockerClient] = DockerClient()
         else:
             self._docker_client = None
@@ -280,22 +280,20 @@ class GProfiler:
         metrics = self._system_metrics_monitor.get_metrics()
         if NoopProfiler.is_noop_profiler(self.system_profiler):
             assert system_result == {}, system_result  # should be empty!
-            merged_result, total_samples = merge.concatenate_profiles(
+            merged_result, total_samples = concatenate_profiles(
                 process_profiles,
                 self._docker_client,
-                self._profile_api_version != "v1",
-                self._identify_applications,
+                self._enrichment_options,
                 metadata,
                 metrics,
             )
 
         else:
-            merged_result, total_samples = merge.merge_profiles(
+            merged_result, total_samples = merge_profiles(
                 system_result,
                 process_profiles,
                 self._docker_client,
-                self._profile_api_version != "v1",
-                self._identify_applications,
+                self._enrichment_options,
                 metadata,
                 metrics,
             )
@@ -717,6 +715,8 @@ def main() -> None:
         if client is not None and remote_logs_handler is not None:
             remote_logs_handler.init_api_client(client)
 
+        enrichment_options = EnrichmentOptions(not args.disable_container_names, args.identify_applications)
+
         gprofiler = GProfiler(
             args.output_dir,
             args.flamegraph,
@@ -724,12 +724,11 @@ def main() -> None:
             client,
             args.collect_metrics,
             args.collect_metadata,
-            args.identify_applications,
+            enrichment_options,
             state,
             usage_logger,
             args.__dict__,
             args.duration,
-            not args.disable_container_names,
             args.profile_api_version,
             remote_logs_handler,
             controller_process,
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -540,6 +540,14 @@ def parse_cmd_args() -> Any:
     )
 
     parser.add_argument(
+        "--disable-application-metadata",
+        action="store_false",
+        default=True,
+        dest="application_metadata",
+        help="Disable collection of application metadata",
+    )
+
+    parser.add_argument(
         "--controller-pid",
         default=None,
         type=int,
@@ -715,7 +723,9 @@ def main() -> None:
         if client is not None and remote_logs_handler is not None:
             remote_logs_handler.init_api_client(client)
 
-        enrichment_options = EnrichmentOptions(not args.disable_container_names, args.identify_applications)
+        enrichment_options = EnrichmentOptions(
+            not args.disable_container_names, args.identify_applications, args.application_metadata
+        )
 
         gprofiler = GProfiler(
             args.output_dir,
--- a/gprofiler/merge.py
+++ b/gprofiler/merge.py
@@ -36,11 +36,13 @@ class EnrichmentOptions:
     Profile enrichment options:
     * container_names: Include container names for each stack in result profile
     * application_identifiers: Attempt to produce & include appid frames for each stack in result profile
+    * application_metadata: Include specialized metadata per application, e.g for Python - the Python version
     """
 
-    def __init__(self, container_names: bool, application_identifiers: bool):
+    def __init__(self, container_names: bool, application_identifiers: bool, application_metadata: bool):
         self.container_names = container_names
         self.application_identifiers = application_identifiers
+        self.application_metadata = application_metadata
 
 
 def parse_one_collapsed(collapsed: str, add_comm: Optional[str] = None) -> StackToSampleCount:
--- a/gprofiler/merge.py
+++ b/gprofiler/merge.py
@@ -31,6 +31,18 @@ SAMPLE_REGEX = re.compile(
 FRAME_REGEX = re.compile(r"^\s*[0-9a-f]+ (.*?) \((.*)\)$")
 
 
+class EnrichmentOptions:
+    """
+    Profile enrichment options:
+    * container_names: Include container names for each stack in result profile
+    * application_identifiers: Attempt to produce & include appid frames for each stack in result profile
+    """
+
+    def __init__(self, container_names: bool, application_identifiers: bool):
+        self.container_names = container_names
+        self.application_identifiers = application_identifiers
+
+
 def parse_one_collapsed(collapsed: str, add_comm: Optional[str] = None) -> StackToSampleCount:
     """
     Parse a stack-collapsed listing.
@@ -231,8 +243,7 @@ def _get_container_name(pid: int, docker_client: Optional[DockerClient], add_con
 def concatenate_profiles(
     process_profiles: ProcessToStackSampleCounters,
     docker_client: Optional[DockerClient],
-    add_container_names: bool,
-    identify_applications: bool,
+    enrichment_options: EnrichmentOptions,
     metadata: Metadata,
     metrics: Metrics,
 ) -> Tuple[str, int]:
@@ -246,13 +257,13 @@ def concatenate_profiles(
     lines = []
 
     for pid, stacks in process_profiles.items():
-        container_name = _get_container_name(pid, docker_client, add_container_names)
-        application_name = get_application_name(pid) if identify_applications else None
+        container_name = _get_container_name(pid, docker_client, enrichment_options.container_names)
+        application_name = get_application_name(pid) if enrichment_options.application_identifiers else None
         if application_name is not None:
             application_name = f"appid: {application_name}"
-        prefix = (container_name + ";") if add_container_names else ""
+        prefix = (container_name + ";") if enrichment_options.container_names else ""
         for stack, count in stacks.items():
-            if identify_applications and application_name is not None:
+            if enrichment_options.application_identifiers and application_name is not None:
                 # insert the app name between the first frame and all others
                 try:
                     first_frame, others = stack.split(";", maxsplit=1)
@@ -263,7 +274,7 @@ def concatenate_profiles(
             total_samples += count
             lines.append(f"{prefix}{stack} {count}")
 
-    lines.insert(0, _make_profile_metadata(docker_client, add_container_names, metadata, metrics))
+    lines.insert(0, _make_profile_metadata(docker_client, enrichment_options.container_names, metadata, metrics))
     return "\n".join(lines), total_samples
 
 
@@ -271,8 +282,7 @@ def merge_profiles(
     perf_pid_to_stacks_counter: ProcessToStackSampleCounters,
     process_profiles: ProcessToStackSampleCounters,
     docker_client: Optional[DockerClient],
-    add_container_names: bool,
-    identify_applications: bool,
+    enrichment_options: EnrichmentOptions,
     metadata: Metadata,
     metrics: Metrics,
 ) -> Tuple[str, int]:
@@ -299,6 +309,4 @@ def merge_profiles(
         # swap them: use the samples from the runtime profiler.
         perf_pid_to_stacks_counter[pid] = scaled_stacks
 
-    return concatenate_profiles(
-        perf_pid_to_stacks_counter, docker_client, add_container_names, identify_applications, metadata, metrics
-    )
+    return concatenate_profiles(perf_pid_to_stacks_counter, docker_client, enrichment_options, metadata, metrics)
--- a/gprofiler/merge.py
+++ b/gprofiler/merge.py
@@ -220,7 +220,11 @@ def _parse_perf_script(script: Optional[str]) -> ProcessToStackSampleCounters:
 
 
 def _make_profile_metadata(
-    docker_client: Optional[DockerClient], add_container_names: bool, metadata: Metadata, metrics: Metrics
+    docker_client: Optional[DockerClient],
+    add_container_names: bool,
+    metadata: Metadata,
+    metrics: Metrics,
+    application_metadata: List[Optional[Dict]],
 ) -> str:
     if docker_client is not None and add_container_names:
         container_names = docker_client.container_names
@@ -235,6 +239,7 @@ def _make_profile_metadata(
         "container_names_enabled": enabled,
         "metadata": metadata,
         "metrics": metrics.__dict__,
+        "application_metadata": application_metadata,
     }
     return "# " + json.dumps(profile_metadata)
 
@@ -292,7 +297,12 @@ def concatenate_profiles(
             total_samples += count
             lines.append(f"{application_prefix}{container_prefix}{stack} {count}")
 
-    lines.insert(0, _make_profile_metadata(docker_client, enrichment_options.container_names, metadata, metrics))
+    lines.insert(
+        0,
+        _make_profile_metadata(
+            docker_client, enrichment_options.container_names, metadata, metrics, application_metadata
+        ),
+    )
     return "\n".join(lines), total_samples
 
 
--- a/gprofiler/merge.py
+++ b/gprofiler/merge.py
@@ -8,12 +8,13 @@ import random
 import re
 from collections import Counter, defaultdict
 from pathlib import Path
-from typing import Iterable, Optional, Tuple
+from typing import Dict, Iterable, List, Optional, Tuple
 
 from gprofiler.docker_client import DockerClient
 from gprofiler.gprofiler_types import ProcessToStackSampleCounters, StackToSampleCount
 from gprofiler.log import get_logger_adapter
 from gprofiler.metadata.application_identifiers import get_application_name
+from gprofiler.metadata.application_metadata import get_application_metadata
 from gprofiler.metadata.metadata_type import Metadata
 from gprofiler.system_metrics import Metrics
 
@@ -257,13 +258,28 @@ def concatenate_profiles(
     """
     total_samples = 0
     lines = []
+    application_metadata: List[Optional[Dict]] = [None]
 
     for pid, stacks in process_profiles.items():
-        container_name = _get_container_name(pid, docker_client, enrichment_options.container_names)
+        # generate application name
         application_name = get_application_name(pid) if enrichment_options.application_identifiers else None
         if application_name is not None:
             application_name = f"appid: {application_name}"
-        prefix = (container_name + ";") if enrichment_options.container_names else ""
+
+        # generate metadata
+        if enrichment_options.application_metadata:
+            app_metadata = get_application_metadata(pid)
+        else:
+            app_metadata = None
+        if app_metadata not in application_metadata:
+            application_metadata.append(app_metadata)
+        idx = application_metadata.index(app_metadata)
+        application_prefix = (f"{idx};") if enrichment_options.application_metadata else ""
+
+        # generate container name
+        container_name = _get_container_name(pid, docker_client, enrichment_options.container_names)
+        container_prefix = (container_name + ";") if enrichment_options.container_names else ""
+
         for stack, count in stacks.items():
             if enrichment_options.application_identifiers and application_name is not None:
                 # insert the app name between the first frame and all others
@@ -274,7 +290,7 @@ def concatenate_profiles(
                     stack = f"{stack};{application_name}"
 
             total_samples += count
-            lines.append(f"{prefix}{stack} {count}")
+            lines.append(f"{application_prefix}{container_prefix}{stack} {count}")
 
     lines.insert(0, _make_profile_metadata(docker_client, enrichment_options.container_names, metadata, metrics))
     return "\n".join(lines), total_samples
--- a/gprofiler/metadata/application_metadata.py
+++ b/gprofiler/metadata/application_metadata.py
@@ -1,0 +1,12 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+
+from typing import Dict, Optional, Union
+
+from psutil import Process
+
+
+def get_application_metadata(process: Union[int, Process]) -> Optional[Dict]:
+    return None
--- a/gprofiler/metadata/application_metadata.py
+++ b/gprofiler/metadata/application_metadata.py
@@ -7,6 +7,13 @@ from typing import Dict, Optional, Union
 
 from psutil import Process
 
+from gprofiler.utils.elf import get_elf_buildid
+
 
 def get_application_metadata(process: Union[int, Process]) -> Optional[Dict]:
-    return None
+    pid = process if isinstance(process, int) else process.pid
+    try:
+        buildid = get_elf_buildid(f"/proc/{pid}/exe") if pid != 0 else None
+    except FileNotFoundError:
+        buildid = None
+    return {"build_id": buildid}
--- a/gprofiler/metadata/application_metadata.py
+++ b/gprofiler/metadata/application_metadata.py
@@ -9,22 +9,19 @@ from typing import Dict, Optional, Union
 from granulate_utils.linux.process import is_process_running
 from psutil import Process
 
-from gprofiler.utils.elf import get_elf_buildid
-
 
 def get_application_metadata(process: Union[int, Process]) -> Optional[Dict]:
-    pid = process if isinstance(process, int) else process.pid
-    try:
-        buildid = get_elf_buildid(f"/proc/{pid}/exe") if pid != 0 else None
-    except FileNotFoundError:
-        buildid = None
-    return {"build_id": buildid}
+    return ApplicationMetadata.get_metadata(process if isinstance(process, Process) else Process(process))
 
 
 class ApplicationMetadata:
     _CACHE_CLEAR_ON_SIZE = 16384
     _cache: Dict[Process, Optional[Dict]] = {}
     _cache_clear_lock = Lock()
+
+    @classmethod
+    def get_metadata(cls, process: Process) -> Optional[Dict]:
+        return cls._cache.get(process)
 
     @classmethod
     def _clear_cache(cls) -> None:
--- a/gprofiler/metadata/application_metadata.py
+++ b/gprofiler/metadata/application_metadata.py
@@ -7,11 +7,19 @@ from threading import Event, Lock
 from typing import Dict, Optional, Union
 
 from granulate_utils.linux.process import is_process_running
-from psutil import Process
+from psutil import NoSuchProcess, Process
 
 
 def get_application_metadata(process: Union[int, Process]) -> Optional[Dict]:
-    return ApplicationMetadata.get_metadata(process if isinstance(process, Process) else Process(process))
+    if process in (0, -1):  # funny values retrieved by perf
+        return None
+
+    try:
+        process = process if isinstance(process, Process) else Process(process)
+    except NoSuchProcess:
+        return None
+
+    return ApplicationMetadata.get_metadata(process)
 
 
 class ApplicationMetadata:
--- a/gprofiler/metadata/application_metadata.py
+++ b/gprofiler/metadata/application_metadata.py
@@ -3,8 +3,10 @@
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
 
+from threading import Event, Lock
 from typing import Dict, Optional, Union
 
+from granulate_utils.linux.process import is_process_running
 from psutil import Process
 
 from gprofiler.utils.elf import get_elf_buildid
@@ -17,3 +19,27 @@ def get_application_metadata(process: Union[int, Process]) -> Optional[Dict]:
     except FileNotFoundError:
         buildid = None
     return {"build_id": buildid}
+
+
+class ApplicationMetadata:
+    _CACHE_CLEAR_ON_SIZE = 16384
+    _cache: Dict[Process, Optional[Dict]] = {}
+    _cache_clear_lock = Lock()
+
+    @classmethod
+    def _clear_cache(cls) -> None:
+        with cls._cache_clear_lock:
+            for process in list(cls._cache.keys()):
+                if not is_process_running(process):
+                    del cls._cache[process]
+
+    @classmethod
+    def update_metadata(cls, process: Process, stop_event: Event) -> None:
+        if process not in cls._cache:
+            if len(cls._cache) > cls._CACHE_CLEAR_ON_SIZE:
+                cls._clear_cache()
+            cls._cache[process] = cls.make_application_metadata(process, stop_event)
+
+    @classmethod
+    def make_application_metadata(cls, process: Process, stop_event: Event) -> Optional[Dict]:
+        return None
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -15,7 +15,7 @@ from pathlib import Path
 from subprocess import CompletedProcess
 from threading import Event
 from types import TracebackType
-from typing import Any, List, Optional, Set, Type, TypeVar
+from typing import Any, Dict, List, Optional, Set, Type, TypeVar
 
 import psutil
 from granulate_utils.java import (
@@ -27,9 +27,9 @@ from granulate_utils.java import (
     java_exit_code_to_signo,
     locate_hotspot_error_file,
 )
-from granulate_utils.linux import ns, proc_events
+from granulate_utils.linux import proc_events
 from granulate_utils.linux.kernel_messages import KernelMessage
-from granulate_utils.linux.ns import get_proc_root_path, resolve_proc_root_links, run_in_ns
+from granulate_utils.linux.ns import get_proc_root_path, get_process_nspid, resolve_proc_root_links, run_in_ns
 from granulate_utils.linux.oom import get_oom_entry
 from granulate_utils.linux.process import is_process_running, process_exe
 from granulate_utils.linux.signals import get_signal_entry
@@ -41,6 +41,7 @@ from gprofiler.gprofiler_types import ProcessToStackSampleCounters, StackToSampl
 from gprofiler.kernel_messages import get_kernel_messages_provider
 from gprofiler.log import get_logger_adapter
 from gprofiler.merge import parse_one_collapsed
+from gprofiler.metadata.application_metadata import ApplicationMetadata
 from gprofiler.profilers.profiler_base import ProcessProfilerBase
 from gprofiler.profilers.registry import ProfilerArgument, register_profiler
 from gprofiler.utils import (
@@ -53,6 +54,7 @@ from gprofiler.utils import (
     touch_path,
     wait_event,
 )
+from gprofiler.utils.elf import get_elf_buildid
 from gprofiler.utils.perf import can_i_use_perf_events
 from gprofiler.utils.process import process_comm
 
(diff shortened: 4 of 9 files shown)


gprofiler/merge.py
```diff
@@ -31,6 +32,20 @@
 FRAME_REGEX = re.compile(r"^\s*[0-9a-f]+ (.*?) \((.*)\)$")
 
 
+class EnrichmentOptions:
```
guybortnikov, line 35: dataclass

gprofiler/profilers/python.py
```diff
@@ -82,6 +85,48 @@ def _add_versions_to_stacks(
     return result
 
 
+class PythonMetadta(ApplicationMetadata):
+    _PYTHON_VERSION_TIMEOUT = 3
+
+    @classmethod
+    def _get_python_version(cls, process: Process, stop_event: Event) -> str:
+        python_path = f"/proc/{get_process_nspid(process.pid)}/exe"
```
adi-benz, line 93: Regarding our conversation - how do you know that this is `python` exe?

adi-benz (commented):
.

gprofiler/merge.py
Jongy: It's a single instance. What it does matter for? shortening the code?

gprofiler/profilers/python.py
Jongy: I don't, this is TODO code, can't get the version if not `python` binary here :/ We'll need to figure it out.

Jongy:
CC @yish

gprofiler/metadata/application_metadata.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class ApplicationMetadata:
+    _CACHE_CLEAR_ON_SIZE = 16384
+    _cache: Dict[Process, Optional[Dict]] = {}
+    _cache_clear_lock = Lock()
+
+    @classmethod
+    def get_metadata(cls, process: Process) -> Optional[Dict]:
+        return cls._cache.get(process)
```
YishaiZinkin, line 34: Why not to acquire the lock here?

gprofiler/profilers/java.py
```diff
@@ -121,6 +123,58 @@ def get_ap_log(self) -> str:
...
+        # libjvm buildid
+        for m in process.memory_maps():
+            if "/libjvm" in m.path:
+                # don't need resolve_proc_root_links here - paths in /proc/pid/maps are normalized.
+                libjvm_buildid = get_elf_buildid(f"/proc/{process.pid}/root/{m.path}")
+                break
+        else:
+            libjvm_buildid = None
+
+        md = {"java_version": version, "libjvm_buildid": libjvm_buildid}
```
YishaiZinkin, line 173: md?

guybortnikov: Haha- metadata
Army conventions

gprofiler/profilers/java.py
```diff
@@ -121,6 +123,58 @@ def get_ap_log(self) -> str:
         return self._ap_log
 
 
+JAVA_VERSION_TIMEOUT = 5
```
YishaiZinkin, line 126: PEP8's convention of prefixing private variables/function applies to globals as well

gprofiler/profilers/java.py
YishaiZinkin: Ohhh I should've figured it out.
I believe `metadata` is a clearer name though. Same for `m`.

gprofiler/profilers/java.py
```diff
@@ -121,6 +123,58 @@ def get_ap_log(self) -> str:
...
+                java_path,
+                "-version",
+            ],
+            stop_event=stop_event,
+            timeout=JAVA_VERSION_TIMEOUT,
+        )
+
+    # doesn't work without changing PID NS as well (I'm getting ENOENT for libjli.so)
+    # Version is printed to stderr
+    return run_in_ns(["pid", "mnt"], _run_java_version, process.pid).stderr.decode().strip()
```
YishaiZinkin, line 157: Why did you remove the exception handling of the original function?

gprofiler/profilers/java.py
```diff
@@ -121,6 +123,58 @@ def get_ap_log(self) -> str:
...
+
+
+class JavaMetadta(ApplicationMetadata):
+    @classmethod
+    def make_application_metadata(cls, process: Process, stop_event: Event) -> Dict[str, Any]:
+        version = get_java_version(process, stop_event)
```
YishaiZinkin, line 163: It's better to be consistent in the `make`/`get` terminology IMO. Personally I think `get` is clearer.

gprofiler/profilers/python.py
YishaiZinkin: How PyPerf and py-spy know that a process is Python? Can't we rely on them?
And at least we can validate that the result from running the process makes sense, i.e. that it looks like "Python x.y.z".

gprofiler/profilers/python.py
```diff
@@ -82,6 +85,50 @@ def _add_versions_to_stacks(
     return result
 
 
+class PythonMetadta(ApplicationMetadata):
+    _PYTHON_VERSION_TIMEOUT = 3
+
+    @classmethod
+    def _get_python_version(cls, process: Process, stop_event: Event) -> str:
```
YishaiZinkin, line 92: Kind of duplicates the logic in `py_module_version.py`. If we execute Python anyway it's better to use this function there as well.

gprofiler/utils/elf.py
```diff
@@ -0,0 +1,69 @@
...
+from typing import Optional, cast
+
+from elftools.elf.elffile import ELFFile  # type: ignore
+from elftools.elf.sections import NoteSection  # type: ignore
+from psutil import NoSuchProcess, Process
+
+
+def get_elf_buildid(path: str) -> Optional[str]:
+    """
+    Gets the build ID embedded in an ELF file  section as an hex string,
```
YishaiZinkin, line 16:
```suggestion
    Gets the build ID embedded in an ELF file section as a hex string,
```

gprofiler/profilers/python.py
Jongy: MVP - I'll start with checking `basename(exe).startwith("python")`.

@YishaiZinkin - gProfiler discovers Python processes using the `DETECTED_PYTHON_PROCESSES_REGEX` regex. PyPerf is a bit simplified and only checks for python/libpython.

gprofiler/metadata/application_metadata.py
Jongy: Why yes? It either works, or doesn't work. No race here

gprofiler/profilers/java.py
Jongy: Just wanted a shorted name. Will change to full name

gprofiler/profilers/java.py
Jongy: Yup

gprofiler/profilers/java.py
Jongy: I mentioned in the commit message. We've added exceptions support to `run_in_ns` - i.e, if an exception is raised in the created thread, it will be propagated to the caller thread. So there's no need to maintain the manual return value / exceptions passing here.

Jongy:
I'll continue on it today

gprofiler/merge.py
guybortnikov: You can many benefits ootb but yeah, shortening code

gprofiler/metadata/application_metadata.py
YishaiZinkin: No risk of data corruption? I don't know what's considered "atomic" in Python, but I can imagine possible data corruption if `get`ting while updating for example.
What's the race you're preventing with the clear cache lock? Iterator invalidation in the `for` loop?

gprofiler/merge.py
Jongy: Made it a dataclass for easy printing

gprofiler/profilers/java.py
Jongy: `make` builds things
`get` queries things

That's what I had in mind, at least. `get_metadata` gets from cache. `make` builds it. `get_java_version` queries the java

gprofiler/profilers/python.py
Jongy: Mhm

Jongy:
This is ready for CR

gprofiler/merge.py
```diff
@@ -253,15 +267,30 @@ def concatenate_profiles(
...
+    application_metadata: List[Optional[Dict]] = [None]
 
     for pid, stacks in process_profiles.items():
-        container_name = _get_container_name(pid, container_names_client, add_container_names)
-        application_name = get_application_name(pid) if identify_applications else None
+        # generate application name
+        application_name = get_application_name(pid) if enrichment_options.application_identifiers else None
         if application_name is not None:
             application_name = f"appid: {application_name}"
-        prefix = (container_name + ";") if add_container_names else ""
```
Jongy, line 262: Previously - we'd always add `;` if protocol version required it; Now we don't. Same for app metadata. I'll clarify that with the backend

Jongy: I will keep that, and also keep protocol version v2 (i.e revert all changes relating to protocol version) as the backend will decide on existence of this field based on the `application_metadata: true` existence.

gprofiler/utils/elf.py
```diff
@@ -0,0 +1,97 @@
...
+    Gets an identifier for this ELF.
+    We prefer to use buildids. If a buildid does not exist for an ELF file,
+    we instead grab its SHA1.
+    """
+    buildid = get_elf_buildid(path)
+    if buildid is not None:
+        return f"buildid:{buildid}"
+
+    # hash in one chunk
+    return f"sha1:{hashlib.sha1(open(path, 'rb').read()).hexdigest()}"
```
YishaiZinkin, line 44: Close the file

Jongy: Brrrr

adi-benz:
Don't we want to add application metadata to `php` as well?

gprofiler/utils/elf.py
```diff
@@ -0,0 +1,98 @@
...
+    try:
+        with open(f"/proc/{process.pid}/mem", "rb", buffering=0) as mem:
+            mem.seek(addr)
+            return mem.read(size)
+    except FileNotFoundError:
+        raise NoSuchProcess(process.pid)
+
+
+def get_process_execfn(process: Process) -> str:
+    # reads process AT_EXECFN
```
adi-benz, line 95: Maybe rename the function to `read_process_execfn()`? Will also be clearer that it's not just a simple getter...

adi-benz:
Should we add a test to validate v1 protocol version returns stacks without the new metadata?

gprofiler/main.py
```diff
@@ -506,7 +504,7 @@ def parse_cmd_args() -> Any:
         dest="profile_api_version",
         default=None,
         choices=["v1"],
```
adi-benz, line 506: Add new versions?

adi-benz (changes requested):

Jongy:
I discussed the backend changes. We ended up with:
* For app metadata, the decision on "first frame is app metadata idx" is based on the existence of `application_metadata_enabled: true`. So if this field is missing (old profilers) or false (`--disable-application-metadata`) - there is no first frame and the backend won't expect it.
* container names - no change from current state (despite that's being a bit weird, because we always send it even when `container_names_enabled: false`.. but that's the state).

gprofiler/merge.py
Jongy: Discussed in https://github.com/Granulate/gprofiler/pull/299#issuecomment-1063493241

Jongy:
> Don't we want to add application metadata to `php` as well?

Nah, not in MVP, PHP is not important enough.

> Should we add a test to validate v1 protocol version returns stacks without the new metadata?

This code area is not built well enough for unit tests, and we don't do e2e yet. I manually tested it as mentioned [here](https://github.com/Granulate/gprofiler/pull/299#issuecomment-1063493241).

gprofiler/merge.py
```diff
@@ -253,15 +269,30 @@ def concatenate_profiles(
     """
     total_samples = 0
     lines = []
+    application_metadata: List[Optional[Dict]] = [None]
 
     for pid, stacks in process_profiles.items():
-        container_name = _get_container_name(pid, container_names_client, add_container_names)
```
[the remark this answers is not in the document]
Jongy: > I couldn't understand why this function called concatenate until I saw the last line

Lol.. Talk about god functions.

gprofiler/profilers/python.py
```diff
@@ -83,6 +86,54 @@ def _add_versions_to_stacks(
     return result
 
 
+class PythonMetadta(ApplicationMetadata):
```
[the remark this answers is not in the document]
Jongy: :sweat:

gprofiler/profilers/python.py
```diff
@@ -83,6 +86,54 @@ def _add_versions_to_stacks(
...
+    @classmethod
+    def make_application_metadata(cls, process: Process, stop_event: Event) -> Dict[str, Any]:
+        # python version
+        try:
+            version: Optional[str] = cls._get_python_version(process, stop_event)
+        except Exception:
+            version = None
+
+        # python id & libpython id, if exists
+        python_elfid = get_elf_id(f"/proc/{process.pid}/exe")
```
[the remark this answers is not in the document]
Jongy: No reason not to get the elfid for uwsgi etc as well. (note that this isn't where I run `python -V`)

gprofiler/profilers/ruby.py
```diff
@@ -5,23 +5,61 @@
...
+
+        return run_in_ns(["pid", "mnt"], _run_ruby_version, process.pid).stdout.decode().strip()
+
+    @classmethod
+    def make_application_metadata(cls, process: Process, stop_event: Event) -> Dict[str, Any]:
+        # ruby version
+        version = cls._get_ruby_version(process, stop_event)
+
+        # ruby elfid & libruby elfid, if exists
+        ruby_elfid = get_elf_id(f"/proc/{process.pid}/exe")
```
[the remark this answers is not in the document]
Jongy: Not sure. for `_get_ruby_version` - I'll make the same fix as did in Python. For others this one here, keeping it, same comment as in python https://github.com/Granulate/gprofiler/pull/299#discussion_r823211979

gprofiler/metadata/application_metadata.py
```diff
@@ -0,0 +1,52 @@
...
+    try:
+        process = process if isinstance(process, Process) else Process(process)
+    except NoSuchProcess:
+        return None
+
+    return ApplicationMetadata.get_metadata(process)
+
+
+class ApplicationMetadata:
+    _CACHE_CLEAR_ON_SIZE = 16384
```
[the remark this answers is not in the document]
Jongy: I spent some time searching for an image of someone licking their finger to express "arbitrariality". That's this number

gprofiler/profilers/python.py
```diff
@@ -371,6 +423,11 @@ def snapshot(self) -> ProcessToStackSampleCounters:
         parsed = parse_many_collapsed(collapsed_text)
         if self.add_versions:
             parsed = _add_versions_to_stacks(parsed)
+        for pid in parsed:
+            try:
+                PythonMetadta.update_metadata(Process(pid), self._stop_event)
+            except NoSuchProcess:
+                continue
```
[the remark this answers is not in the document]
Jongy: If app metadata is enabled for a profile, all stacks must precede with the index. That's why the app metadata list begins with a `None` entry - for all the PIDs where we've failed to collect it, they will get index `0`

gprofiler/utils/elf.py
Jongy: Yup

gprofiler/main.py
Jongy: See my comment in https://github.com/Granulate/gprofiler/pull/299#issuecomment-1063493241, no need for a new version here.

gprofiler/profilers/java.py
```diff
@@ -124,6 +126,53 @@ def get_ap_log(self) -> str:
...
+        # libjvm elfid
+        libjvm_elfid = get_mapped_dso_elf_id(process, "/libjvm")
+
+        metadata = {"java_version": version, "libjvm_elfid": libjvm_elfid}
```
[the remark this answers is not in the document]
Jongy: You mean `java_elfid`? Java is always a small binary that loads `libjvm.so`, we've never cared about it on its own.
In Python, sometimes there's no `libpython` so Python is interesting. I decided to retrieve both libpython and python in all cases, to avoid any chance for confusion (passing None if not found).

I'll document to clarify.

gprofiler/profilers/python.py
```diff
@@ -371,6 +423,11 @@ def snapshot(self) -> ProcessToStackSampleCounters:
         parsed = parse_many_collapsed(collapsed_text)
         if self.add_versions:
             parsed = _add_versions_to_stacks(parsed)
+        for pid in parsed:
+            try:
+                PythonMetadta.update_metadata(Process(pid), self._stop_event)
+            except NoSuchProcess:
```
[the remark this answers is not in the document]
Jongy: Well technically the `Process(pid)` will raise `NoSuchProcess`, that's what I had in mind :sweat_smile: 

Also, if we raise in `_get_python_version`, then I am catching it in `make_application_metadata`.

However, I do admit that I haven't done a pass on all executed code to ensure no exceptions leak. Looking at it now... we can also fail when reading the map files / exe, and that'd be just `FileNotFoundError`. Not sure what's the correct approach here - a catch-all at the `make_application_metadata` level, or trying to catch each of those exceptions where it might be raised... Wdyt?

gprofiler/main.py
```diff
@@ -720,19 +726,26 @@ def main() -> None:
         if client is not None and remote_logs_handler is not None:
             remote_logs_handler.init_api_client(client)
 
+        enrichment_options = EnrichmentOptions(
+            container_names=not args.disable_container_names and args.profile_api_version != "v1",
```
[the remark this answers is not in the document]
Jongy: Good point, it was this way forever and I never bothered changing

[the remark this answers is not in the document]
Jongy: > Why the and args.profile_api_version != "v1" here?

Because in "v1" we mustn't send them, anyway.

I think I understand your point - I will remove `container_names_in_protocol` and instead store the protocol version; then make those decisions based on it in `merge.py`.

gprofiler/merge.py
```diff
@@ -253,15 +270,33 @@ def concatenate_profiles(
...
+        else:
+            app_metadata = None
+        if app_metadata not in application_metadata:
+            application_metadata.append(app_metadata)
+        idx = application_metadata.index(app_metadata)
+        # we include the application metadata frame IFF application_metadata is enabled
+        application_prefix = (f"{idx};") if enrichment_options.application_metadata else ""
+
+        # generate container name
+        # to maintain compatibility with old profiler versions, we include the container name frame in any case
```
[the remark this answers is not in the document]
Jongy: "v1" is the old profile protocol version that we're maintaining, which does not support container names. So yeah, they should not be included as a prefix per stack.

gprofiler/merge.py
```diff
@@ -31,6 +33,18 @@
...
 
 
+@dataclass
+class EnrichmentOptions:
+    """
+    Profile enrichment options.
+    """
+
+    container_names: bool  # Include container names for each stack in result profile
+    container_names_in_protocol: bool  # Are container names a part of the protocol? (true iff not v1)
```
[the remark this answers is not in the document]
Jongy: I was expecting this comment :sad:
As mentioned in https://github.com/Granulate/gprofiler/pull/299#issuecomment-1063493241, I actually tried to break the current scheme of `container_names` which complicates this. But I decided not to, too many changes in one PR.

Anyway, this is indeed not an "enrichment option", but it does relate. Now I've placed the `profile_api_version` here.
I don't think it implies a bad design, since those options indeed work together. it's just bad naming. `EnrichmentContext`?

gprofiler/metadata/application_metadata.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @classmethod
+    def update_metadata(cls, process: Process, stop_event: Event) -> None:
+        if process not in cls._cache:
+            if len(cls._cache) > cls._CACHE_CLEAR_ON_SIZE:
+                cls._clear_cache()
+            cls._cache[process] = cls.make_application_metadata(process, stop_event)
+
+    @classmethod
+    def make_application_metadata(cls, process: Process, stop_event: Event) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
Jongy: Heh, I had this TODO `Rethink of design with objects instead of classes?` but decided to skip it.

Why I chose class based and not OOP?
1. Wasn't sure where to instantiate it, but now I see that it works just fine in each Profiler class
2. Wanted the cache to be shared (but now that I think of it, the `cls` of a `classmethod` of a superclass is not the superclass, but the descendant... so after all different `*Metadata` classes do not share the cache lol. Or actually, it is? Because I don't recreate it on each class, just on the superclass.

Spent 2 mins thinking about it more and didn't find any reason not to make it OOP, converting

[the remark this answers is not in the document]
Jongy: > This function doesn't use the stop_event, it's here because it's used by the descendants, that implies bad design.

I liked this rule, nice :+1:

gprofiler/metadata/application_metadata.py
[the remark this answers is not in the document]
Jongy: (and elaborated - not too low, not too high to consume too much memory)

gprofiler/metadata/application_metadata.py
```diff
@@ -0,0 +1,52 @@
...
+            for process in list(cls._cache.keys()):
+                if not is_process_running(process):
+                    del cls._cache[process]
+
+    @classmethod
+    def update_metadata(cls, process: Process, stop_event: Event) -> None:
+        if process not in cls._cache:
+            if len(cls._cache) > cls._CACHE_CLEAR_ON_SIZE:
+                cls._clear_cache()
+            cls._cache[process] = cls.make_application_metadata(process, stop_event)
```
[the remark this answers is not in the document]
Jongy: Good points. About short processes, I intentionally ignored them in the MVP, I don't think it makes sense to invest in them right now, let's get this feature first for the majority. Short processes' metadata will be handled alongside with https://github.com/Granulate/gprofiler/pull/281.

 About the design - I agree, basically. Not for the reason of PID reuse (that'll just never happen unless one is fork-bombing us) but for explicitness & cleaner design.

Let's discuss this f2f

YishaiZinkin:
@Jongy Can you please add a short stackcollapse example with the new format to the description so we can see it in front of our eyes?

gprofiler/metadata/application_metadata.py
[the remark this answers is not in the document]
YishaiZinkin: Lol this is oddly specific.
IMO it's worth a comment anyway. I also suggest you write this number in hex, I didn't immediately realized that its a power of 2.

Jongy:
> @Jongy Can you please add a short stackcollapse example with the new format to the description so we can see it in front of our eyes?

Sure, good idea :)
```
# {"containers": ["javas", "gprofiler"], "container_names_enabled": true, "metadata": {"hostname": "jong-XPS-15-7590"}, "metrics": {"cpu_avg": 24.5, "mem_avg": 94.5}, "application_metadata": [null, {"java_version": "openjdk version \"1.8.0_111\"\nOpenJDK Runtime Environment (build 1.8.0_111-8u111-b14-2~bpo8+1-b14)\nOpenJDK 64-Bit Server VM (build 25.111-b14, mixed mode)", "libjvm_elfid": "buildid:803cc02264a07cbee7de6ee1be15606c6dc36705", "exe": "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java", "execfn": "/usr/bin/java"}], "application_metadata_enabled": true}
0;gprofiler;perf;[unknown];__strncasecmp_avx;generic_start_main;main;run_builtin;cmd_record;evlist__enable;__ioctl;entry_SYSCALL_64_after_hwframe_[k];do_syscall_64_[k];__x64_sys_ioctl_[k];ksys_ioctl_[k];do_vfs_ioctl_[k];perf_ioctl_[k];_perf_ioctl_[k];perf_event_for_each_child_[k];_perf_event_enable_[k];event_function_call_[k];smp_call_function_single_[k];generic_exec_single_[k];remote_function_[k];event_function_[k];__perf_event_enable_[k];ctx_resched_[k];perf_pmu_enable.part.0_[k];x86_pmu_enable_[k];intel_pmu_enable_all_[k];native_write_msr_[k] 64
0;gprofiler;perf;[unknown];__strncasecmp_avx;generic_start_main;main;run_builtin;cmd_record;record__mmap_read_evlist.constprop.44;perf_mmap__push;record__pushfn;[/app/gprofiler/resources/perf];entry_SYSCALL_64_after_hwframe_[k];do_syscall_64_[k];__x64_sys_write_[k];ksys_write_[k];vfs_write_[k];__vfs_write_[k];new_sync_write_[k];ovl_write_iter_[k];ovl_real_fdget_meta_[k];ovl_inode_realdata_[k] 1
0;;swapper;secondary_startup_64_[k];x86_64_start_kernel_[k];x86_64_start_reservations_[k];start_kernel_[k];arch_call_rest_init_[k];rest_init_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];cpuidle_enter_state_[k];intel_idle_[k] 2
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];cpuidle_enter_state_[k];intel_idle_[k] 79
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];cpuidle_enter_state_[k];thermal_interrupt_[k];smp_thermal_interrupt_[k];intel_thermal_interrupt_[k];pkg_thermal_notify_[k];_raw_spin_lock_irqsave_[k];native_queued_spin_lock_slowpath_[k] 2
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];cpuidle_select_[k];menu_select_[k];tick_nohz_get_sleep_length_[k];hrtimer_next_event_without_[k];__hrtimer_next_event_base_[k];rb_next_[k] 1
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];cpuidle_enter_state_[k];apic_timer_interrupt_[k];smp_apic_timer_interrupt_[k];hrtimer_interrupt_[k];__hrtimer_run_queues_[k];tick_sched_timer_[k];tick_sched_handle_[k];update_process_times_[k];scheduler_tick_[k];perf_event_task_tick_[k] 1
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];cpuidle_select_[k];menu_select_[k] 4
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];apic_timer_interrupt_[k] 2
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];cpuidle_enter_state_[k];thermal_interrupt_[k];smp_thermal_interrupt_[k];intel_thermal_interrupt_[k];pkg_thermal_notify_[k] 1
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];cpuidle_enter_state_[k];apic_timer_interrupt_[k];smp_apic_timer_interrupt_[k];hrtimer_interrupt_[k];__hrtimer_run_queues_[k];tick_sched_timer_[k];tick_sched_handle_[k];update_process_times_[k] 1
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];cpuidle_enter_state_[k];apic_timer_interrupt_[k];smp_apic_timer_interrupt_[k];hrtimer_interrupt_[k];__hrtimer_run_queues_[k];tick_sched_timer_[k];tick_sched_handle_[k];update_process_times_[k];calc_global_load_tick_[k] 1
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];cpuidle_enter_state_[k];ret_from_intr_[k];do_IRQ_[k];_raw_spin_lock_[k] 1
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];cpuidle_enter_state_[k];ret_from_intr_[k];do_IRQ_[k];irq_exit_[k];tick_nohz_irq_exit_[k];sched_clock_idle_sleep_event_[k];sched_clock_cpu_[k];native_sched_clock_[k] 1
0;;swapper;secondary_startup_64_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];call_cpuidle_[k];cpuidle_enter_[k];cpuidle_enter_state_[k];apic_timer_interrupt_[k];smp_apic_timer_interrupt_[k];irq_exit_[k];__softirqentry_text_start_[k];run_rebalance_domains_[k];update_blocked_averages_[k] 1
0;;snapd;[/snap/core/12725/usr/lib/snapd/snapd];[/snap/core/12725/usr/lib/snapd/snapd];[/snap/core/12725/usr/lib/snapd/snapd];[/snap/core/12725/usr/lib/snapd/snapd];[/snap/core/12725/usr/lib/snapd/snapd];[/snap/core/12725/usr/lib/snapd/snapd];page_fault_[k];do_page_fault_[k];__do_page_fault_[k];do_user_addr_fault_[k];handle_mm_fault_[k];__handle_mm_fault_[k];do_fault_[k];__do_fault_[k];filemap_fault_[k];__do_page_cache_readahead_[k];read_pages_[k];squashfs_readpage_[k];squashfs_readpage_block_[k];squashfs_get_datablock_[k];squashfs_cache_get_[k];squashfs_read_data_[k];squashfs_decompress_[k];squashfs_xz_uncompress_[k];xz_dec_run_[k];xz_dec_lzma2_run_[k];dict_repeat_[k] 1
0;;gnome-terminal-;[unknown];[/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6400.6] 2
0;;gnome-shell;[unknown];g_type_check_instance_cast 1
0;;gnome-shell;[unknown];[unknown];[unknown];g_type_check_instance_is_a 1
0;;gnome-shell;[unknown];[/usr/lib/x86_64-linux-gnu/libatspi.so.0.0.1] 1
0;;gnome-shell;[unknown];g_main_context_check 1
0;;fdtransfer;__read_nocancel;entry_SYSCALL_64_after_hwframe_[k];do_syscall_64_[k];__x64_sys_read_[k];ksys_read_[k];vfs_read_[k];__vfs_read_[k];proc_reg_read_[k];seq_read_[k];s_show_[k];seq_printf_[k];seq_vprintf_[k];vsnprintf_[k];format_decode_[k] 1
0;;fdtransfer;__read_nocancel;entry_SYSCALL_64_after_hwframe_[k];do_syscall_64_[k];__x64_sys_read_[k];ksys_read_[k];vfs_read_[k];__vfs_read_[k];proc_reg_read_[k];seq_read_[k];s_next_[k];update_iter_[k];module_get_kallsym_[k] 1
1;javas;java;Target.main([Ljava/lang/String|)V_[j];Target.method3()V_[j];java/io/File.list()[Ljava/lang/String|_[j] 1
```

gprofiler/profilers/python.py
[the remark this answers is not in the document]
adi-benz: Then change the metadata key, right now it's `python_elfid`, maybe change to `exe_elfid`?

gprofiler/profilers/python.py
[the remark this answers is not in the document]
adi-benz: Can you document the `[None]` line (`merge.py:340`)

gprofiler/profilers/ruby.py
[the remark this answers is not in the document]
adi-benz: Then, same as for python, change the metadata key from `ruby_elfid` to `exe_elfid`

gprofiler/main.py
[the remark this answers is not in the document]
adi-benz: Yea, I couldn't think of any reason to specify them twice other than such optimization, but yea, that's was my point

gprofiler/profilers/python.py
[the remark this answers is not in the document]
adi-benz: > Well technically the Process(pid) will raise NoSuchProcess, that's what I had in mind 😅

Racy

I think we should except `NoSuchProcess` and `continue` (like now), and also except-all in here, specifically for `PythonEbpfProfiler` and log, so we won't interfere with the rest of the processes, this isn't a concern for `ProcessProfiler`s.

And to reduce the chance for a race and redundant logs - we should catch the exceptions where they might be raised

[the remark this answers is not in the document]
adi-benz: > Well technically the Process(pid) will raise NoSuchProcess, that's what I had in mind 😅

Racy

I think we should except `NoSuchProcess` and `continue` (like we do now), and also except-all in here, specifically for `PythonEbpfProfiler` and log, so we won't interfere with the rest of the processes, this isn't a concern for `ProcessProfiler`s.

Catching all in `make_application_metadata` will not allow us to `except NoSuchProcess` in `PythonEbpfProfiler` that we need...

And to reduce the chance for a race and redundant logs - we should catch the exceptions where they might be raised

gprofiler/profilers/python.py
```diff
@@ -83,6 +86,54 @@ def _add_versions_to_stacks(
...
+        except Exception:
+            version = None
+
+        # python id & libpython id, if exists.
+        # if libpython exists then the python binary itself is of less importance; however, to avoid confusion
+        # we collect them both here (then we're able to know if either exist)
+        python_elfid = get_elf_id(f"/proc/{process.pid}/exe")
+        libpython_elfid = get_mapped_dso_elf_id(process, "/libpython")
+
+        metadata = {"python_version": version, "python_elfid": python_elfid, "libpython_elfid": libpython_elfid}
```
Jongy, line 131: Add python2.7 m or mu

Jongy: Done

gprofiler/profilers/python.py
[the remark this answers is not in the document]
Jongy: Fair

gprofiler/profilers/ruby.py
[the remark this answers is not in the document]
Jongy: Done

gprofiler/profilers/python.py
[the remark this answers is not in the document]
Jongy: Mhm

gprofiler/merge.py
```diff
@@ -337,6 +337,8 @@ def concatenate_profiles(
     """
     total_samples = 0
     lines = []
+    # the metadata list always containers a "null" entry with index 0 - that's the index used for all
```
adi-benz, line 340:
```suggestion
    # the metadata list always contains a "null" entry with index 0 - that's the index used for all
```

gprofiler/profilers/python.py
```diff
@@ -89,46 +89,59 @@ def _add_versions_to_stacks(
...
+                    timeout=self._PYTHON_VERSION_TIMEOUT,
+                )
 
-        cp = run_in_ns(["pid", "mnt"], _run_python_version, process.pid)
-        version = cp.stdout.decode().strip()
-        if version:
-            return version
+            cp = run_in_ns(["pid", "mnt"], _run_python_version, process.pid)
+            output = cp.stdout.decode().strip()
+            if not output and args == ["-V"]:
```
adi-benz, line 113: Very specific...
Return `stdout, stderr`, and have one more function:
```python
def get_python_version(self, process):
    self. _run_process_python_python(process, ['-V'])
    if python2:

version = self.get_python_version()
sys_maxunicode, _ = self. _run_process_python_python(process, ['-S', '-c', ..])
```

tests/test_app_metadata.py
```diff
@@ -17,19 +17,58 @@
...
+            },
+        ),
+        (
+            True,
+            "java",
+            "ap",
+            {
+                "exe": "/usr/local/openjdk-8/bin/java",
+                "execfn": "/usr/local/openjdk-8/bin/java",
+                "java_version": 'openjdk version "1.8.0_275"\n'
```
adi-benz, line 54: We should strip `\n` from process stdout that goes into metadata

tests/utils.py
```diff
@@ -138,3 +139,27 @@ def make_java_profiler(
         java_safemode=java_safemode,
         java_mode=java_mode,
     )
+
+
+def run_gprofiler_container(
```
adi-benz, line 144:
```suggestion
def run_gprofiler_in_container_for_one_session(
```

adi-benz (changes requested):

gprofiler/profilers/python.py
Jongy: You're right but this code will be consolidated in a later PR

gprofiler/merge.py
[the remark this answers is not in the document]
Jongy: Decided f2f to keep it

gprofiler/metadata/application_metadata.py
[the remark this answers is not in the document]
Jongy: I'll document

gprofiler/metadata/application_metadata.py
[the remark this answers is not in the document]
adi-benz: Decided f2f to skip for now so we won't block other features. But will fix in a follow-up PR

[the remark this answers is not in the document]
Jongy: Discussed this - we have decided to move it to a later PR.

gprofiler/profilers/python.py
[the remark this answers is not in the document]
Jongy: I am swallowing all exceptions besides `NoSuchProcess` now.

tests/utils.py
Jongy: Mhm done

gprofiler/profilers/python.py
Jongy: Reorganized

gprofiler/metadata/application_metadata.py
```diff
@@ -0,0 +1,68 @@
...
+        if process not in self._cache:
+            if len(self._cache) > self._CACHE_CLEAR_ON_SIZE:
+                self._clear_cache()
+            try:
+                metadata = self.make_application_metadata(process)
+            except NoSuchProcess:
+                # let our caller handler this
+                raise
+            except Exception:
+                if self._metadata_exception_logs > self._MAX_METADATA_EXCEPTION_LOGS:
```
adi-benz, line 62: `self._metadata_exception_logs` is never updated

adi-benz, line 62: That's not rate limiting... we sure we want to make it quiet after 100 logs? document if so

gprofiler/profilers/python.py
```diff
@@ -83,6 +86,74 @@ def _add_versions_to_stacks(
...
+
+    def _run_process_python(self, process: Process, args: List[str]) -> Tuple[str, str]:
+        if not os.path.basename(process.exe()).startswith("python"):
+            # TODO: for dynamic executables, find the python binary that works with the loaded libpython, and
+            # check it instead. For static executables embedding libpython - :shrug:
+            raise NotImplementedError
+
+        python_path = f"/proc/{get_process_nspid(process.pid)}/exe"
+
+        def _run_python_version() -> "CompletedProcess[bytes]":
```
adi-benz, line 100:
```suggestion
        def _run_python_process_in_ns() -> "CompletedProcess[bytes]":
```

adi-benz (changes requested):

gprofiler/metadata/application_metadata.py
Jongy: Yeah, that's my intention. I think that in most environments, the same exception will be repeated all over again; so logging it only a couple of times is enough. Will document

gprofiler/profilers/python.py
Jongy: mhm

gprofiler/metadata/application_metadata.py
adi-benz: @Jongy

adi-benz (changes requested):
Missed one comment https://github.com/Granulate/gprofiler/pull/299#discussion_r825784777

gprofiler/metadata/application_metadata.py
Jongy: Argh

gprofiler/metadata/application_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+                self._clear_cache()
+            try:
+                metadata = self.make_application_metadata(process)
+            except NoSuchProcess:
+                # let our caller handler this
+                raise
+            except Exception:
+                # log only the first _MAX_METADATA_EXCEPTION_LOGS exceptions; I expect the same exceptions to
+                # be repeated again and again, so it's enough to log just a handful of them.
+                if self._metadata_exception_logs > self._MAX_METADATA_EXCEPTION_LOGS:
```
adi-benz, line 64: Flip the if... And give it a test run to make sure it's good

adi-benz (changes requested):
🙃

gprofiler/metadata/application_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+    return ApplicationMetadata.get_metadata(process)
+
+
+class ApplicationMetadata:
+    # chosen arbitrarily to be large enough to contain all processes we could possibly profile in one session; while
+    # not exploding memory too much.
+    _CACHE_CLEAR_ON_SIZE = 0x4000
+    _cache: Dict[Process, Optional[Dict]] = {}
+    _cache_clear_lock = Lock()
+    _metadata_exception_logs = 0
```
adi-benz, line 36:
```suggestion
    _metadata_exception_logs_count = 0
```

gprofiler/metadata/application_metadata.py
Jongy: So many errors in such simple code

Jongy: Testing

Jongy:
Applied this diff:
```
diff --git a/gprofiler/profilers/python.py b/gprofiler/profilers/python.py
index b6cdc98..313710e 100644
--- a/gprofiler/profilers/python.py
+++ b/gprofiler/profilers/python.py
@@ -149,6 +149,7 @@ class PythonMetadata(ApplicationMetadata):
             "libpython_elfid": libpython_elfid,
             "sys_maxunicode": maxunicode,
         }
+        raise 1
 
         metadata.update(super().make_application_metadata(process))
         return metadata
```

Got
```
[2022-03-14 12:57:51,140] ERROR: gprofiler.metadata.application_metadata: Exception while collecting metadata in PythonMetadata!
  File "/app/gprofiler/metadata/application_metadata.py", line 57, in update_metadata
    metadata = self.make_application_metadata(process)
  File "/app/gprofiler/profilers/python.py", line 152, in make_application_metadata
    raise 1
TypeError: exceptions must derive from BaseException
```

adi-benz (approved):

PR #300, opened 2022-02-19
python: Fix PyPerf for 3.10

Jongy:
BCC PR: https://github.com/Granulate/bcc/pull/35

Basically, upgrade PyPerf reference post the fixing PR, and add some tests that would catch the issue explicitly.

I found this using https://github.com/Granulate/gprofiler/pull/268 - which showed me that tests occasionally failed because Python 3.10 was running on the runner (e.g was running our pytest) and it had class names which were read incorrectly - reading random binary data which was later not decoded correctly. That's why it was statistical - because profiling is statistical :/ and sometimes stacks with class names were not encountered in the profile.

Lessons learned: try and test all flows, because some flows that you don't test might be "automatically" tested by your system-wide profiler doing its profiling on random crap on your system. Failing your tests unexpectedly. Argh

### DIFF
--- a/scripts/pyperf_build.sh
+++ b/scripts/pyperf_build.sh
@@ -5,7 +5,7 @@
 #
 set -e
 
-git clone --depth 1 -b v1.2.1 https://github.com/Granulate/bcc.git && cd bcc && git reset --hard 3f1d60180946852aa89e3644b786e4bf8934e7a9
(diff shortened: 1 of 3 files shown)


adi-benz:
Nice!  
Just don't forget to fix `pyperf_build.sh` after merging https://github.com/Granulate/bcc/pull/35

adi-benz (commented):
.

adi-benz (approved):

Jongy:
Will let `test_python_ebpf` pass before I merge

PR #307, opened 2022-03-03
utils: run_process(): Log when a process is killed due timeout

Jongy:


gprofiler/utils/__init__.py
```diff
@@ -241,6 +241,7 @@ def run_process(
                             raise
         except TimeoutExpired:
             returncode, stdout, stderr = _reap_process(process, kill_signal)
+            logger.debug(f"({process.args!r}) was killed with signal {kill_signal} due to timeout")
```
adi-benz, line 244: Move to `_reap_process()`

adi-benz, line 244:
```suggestion
            logger.debug(f"Killed ({process.args!r}) with signal {kill_signal} due to timeout")
```

This way it sound more like we killed it

adi-benz (changes requested):

gprofiler/utils/__init__.py
Jongy: Good call

Jongy: I want to keep the args first so it matches the other messages. I'll try make the message more clear

adi-benz (approved):

PR #308, opened 2022-03-03
java: Simplify complexity of atomic copies by ensuring sync

Jongy:
## Description
Title says it all

## How Has This Been Tested?
CI

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -221,8 +221,7 @@ class AsyncProfiledProcess:
         touch_path(self._output_path_host, self.OUTPUTS_MODE)
         self._recreate_log()
         # copy libasyncProfiler.so if needed
-        if not os.path.exists(self._libap_path_host):
-            self._copy_libap()
+        self._copy_libap()
 
         return self
 
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -52,7 +52,7 @@ from gprofiler.utils import (
     touch_path,
     wait_event,
 )
-from gprofiler.utils.fs import atomic_copy
+from gprofiler.utils.fs import safe_copy
 from gprofiler.utils.perf import can_i_use_perf_events
 from gprofiler.utils.process import process_comm
 
@@ -276,7 +276,7 @@ class AsyncProfiledProcess:
                 os.chmod(
                     libap_resource, 0o755
                 )  # make it accessible for all; needed with PyInstaller, which extracts files as 0700
-                atomic_copy(libap_resource, self._libap_path_host)
+                safe_copy(libap_resource, self._libap_path_host)
 
     def _recreate_log(self) -> None:
         touch_path(self._log_path_host, self.OUTPUTS_MODE)
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -2,18 +2,16 @@
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
-import errno
 import functools
 import json
 import os
 import re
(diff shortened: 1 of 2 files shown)


gprofiler/utils/fs.py
```diff
@@ -0,0 +1,13 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+
+import os
+import shutil
+
+
+def atomic_copy(src: str, dst: str) -> None:
```
adi-benz, line 10: This is not atomic, while two threads (and especially two processes) are running this function concurrently the file at `dst` might be corrupt.  
This is not a bug, since we use a lock when calling this function, but this is a misleading function

gprofiler/profilers/java.py
```diff
@@ -261,36 +261,22 @@ def _is_musl(self) -> bool:
...
+        with libap_copy_lock:
+            if not os.path.exists(self._libap_path_host):
+                # atomically copy it
+                libap_resource = resource_path(
+                    os.path.join("java", "musl" if self._is_musl() else "glibc", "libasyncProfiler.so")
+                )
+                os.chmod(
+                    libap_resource, 0o755
+                )  # make it accessible for all; needed with PyInstaller, which extracts files as 0700
+                atomic_copy(libap_resource, self._libap_path_host)
```
adi-benz, line 279: Why do we need an atomic copy? Why not just a regular copy? The lock protects it

adi-benz (changes requested):

adi-benz:
What was the motivation for this PR? Have we encountered any problems with the current libap copy?

gprofiler/utils/fs.py
```diff
@@ -0,0 +1,13 @@
...
+    dst_tmp = f"{dst}.tmp"
+    shutil.copy(src, dst)
```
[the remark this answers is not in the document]
Jongy: Heh

gprofiler/utils/fs.py
Jongy: I suppose `atomic` is misleading. This aims to be "atomic" at the filesystem level. I'll document and rename to `safe_copy`.

gprofiler/profilers/java.py
Jongy: I recall I had a concrete reason when I wrote it, Too bad I didn't document. Anyway, 2 new reasons from the top of my head - 
1. As a best practice when moving around executables. Linux behaves differently when you `open(.. ,"w")` a file and `os.rename` a file. The former edits the existing inode, the latter creates a new one (and existing references to the old inode are kept).
2. Besides the lock, we do protect this with `os.path.exists` so there's no chance we're `open`ing a file already loaded...

Anyway, this is still safer, when moving exes I strive to always use copy-rename

Jongy:
> What was the motivation for this PR? Have we encountered any problems with the current libap copy?

I saw some errors in a log of one of the profiler. Stuff like:
```
FileExistsError: [Errno 17] File exists: '/proc/1/root/tmp/gprofiler_tmp/async-profiler-v2.6g1/glibc.5596' -> '/proc/1/root/tmp/gprofiler_tmp/async-profiler-v2.6g1/glibc'
```

too much complexity for no good reason. It's been a while that I wanted to get rid of this code lol.

gprofiler/profilers/java.py
adi-benz: I much prefer the simpler lock and bool (instead of relying on FS exists):
```python
with lock:
    if not has_copied_libap:
        shutil.copy(src, dst)
        has_copied_libap = True
```

Less code, less stuff to wonder about when reading, but OK

adi-benz (approved):

PR #309, opened 2022-03-03
java: Fix _check_async_profiler_loaded() for cases where resolved storage path differs

Jongy:
If e.g `/tmp` is a symlink to `/var/tmp`, then this check would fail. Now it'll pass in either case.


adi-benz (commented):
Looks good, but CI problems

Jongy:
> Looks good, but CI problems

Check was reversed so even if container would build, I'd expect the tests to fail. Fixed now, let's see if CI is green

adi-benz (approved):

PR #311, opened 2022-03-08
Better java app id resolving

liorgorb:
Use the `sun.java.command` Java property as app ID for Java applications. This property contains the JAR if the program is run using `-jar`, or the provided main class otherwise. It also contains the arguments passed to the program.
`sun.java.command` is extracted by running `jattach` in `properties` mode for each Java process.

### DIFF
--- a/gprofiler/metadata/application_identifiers.py
+++ b/gprofiler/metadata/application_identifiers.py
@@ -11,7 +11,10 @@ from typing import List, Optional, TextIO, Tuple, Union
 from granulate_utils.linux.ns import resolve_host_path, resolve_proc_root_links
 from psutil import NoSuchProcess, Process
 
+from gprofiler.exceptions import CalledProcessError
 from gprofiler.log import get_logger_adapter
+from gprofiler.profilers.java import jattach_path
+from gprofiler.utils import run_process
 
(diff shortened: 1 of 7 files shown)


gprofiler/profilers/perf.py
```diff
@@ -9,10 +9,10 @@
 from threading import Event
 from typing import List, Optional
 
+import gprofiler.merge
```
Jongy, line 12: Remove this unnecessary diff everyhwere

gprofiler/metadata/application_identifiers.py
```diff
@@ -244,11 +247,16 @@ def get_application_name(self, process: Process) -> Optional[str]:
         if not any("libjvm.so" in m.path for m in process.memory_maps()):
             return None
 
-        jar_arg = _get_cli_arg_by_name(process.cmdline(), "-jar")
-        if jar_arg is _NON_AVAILABLE_ARG:
-            return None
+        try:
+            java_properties = run_process([jattach_path(), str(process.pid), "properties"]).stdout.decode()
```
Jongy, line 251: This doesn't return anything "private", otherwise pass `suppress_log=True`

Jongy (approved):

PR #318, opened 2022-03-13
Stripped arguments from Java app id

liorgorb:
Stripped arguments from Java app id


Jongy (dismissed):
Also update version to 1.2.17

Jongy (approved):

PR #321, opened 2022-03-16
appid: Remove full path from java appid

DanielShaulov:
## Description
The full path to java app ids might contain random elements, until we
find a good way to strip those, we want to remove this part of the appid.


Jongy:
Can the "command" include path somehow? Maybe the "randomized" component can include a part of the path which will still affect us? Maybe the "jar name" or "class name" which might appear as the "command", can contain random components as well?

liorgorb:
@Jongy it's possible but not prevalent. Let's merge this for now.

liorgorb (dismissed):

liorgorb (approved):

PR #325, opened 2022-03-29
docs: Pass --token, --service-name with =

Jongy:
This allows them to begin with a -, otherwise argparse doesn't like it.
Also, normalize the metavars in the docs.


YishaiZinkin (approved):
No modifications to `argparse` are required I assume?

Jongy:
> No modifications to `argparse` are required I assume?

Nope, `--token xxx` as 2 arguments and `--token="xxx"` as a single argument both work the same.

PR #331, opened 2022-03-31
Move run_process and friends to granulate-utils

Jongy:
Move `run_process` and related utils to granulate-utils, in preparation of moving other parts that are dependent on them; all in all, to make them all available for other projects.

Basically:
* `run_process` is now wrapped by `run_process_logged` which includes the gProfiler logger.
* This whole concept was transformed to a class `RunProcess`, to allow for easier extending. gProfiler adds `RunProcessStaticx` which adds the staticx-related logics.

This refactor came out larger than I anticipated, but it's a change in a good direction -
* It will help us break down and refactor `run_process` furthermore
* It will allow us to share other components in granulate-utils, that depend on `run_process`.

TODOs:
* [x] ~~Make sure that all `run_process_logged` invocations actually use `start_process_staticx` (the `via_staticx=False` part is important)~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -146,9 +146,13 @@ RUN pip3 install --upgrade pip
 # done separately from the 'pip3 install -e' below; so we don't reinstall all dependencies on each
 # code change.
 COPY requirements.txt ./
+# copy needed files from granulate-utils to run the pip install
 COPY granulate-utils/setup.py granulate-utils/requirements.txt granulate-utils/README.md granulate-utils/
-COPY granulate-utils/granulate_utils granulate-utils/granulate_utils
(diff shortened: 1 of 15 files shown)


kmaork:
Conflicts! (and CI failed, is it random?)

Jongy:
I don't like this approach - I think I will do it elseway (first refactor the function in this repo, then move repo)

Jongy:
Too many conflicts. The initial reasoning behind this was to move `system_metadata.py` to granulate-utils. There's only one usage of `run_process` - we'll deal without it and switch it to `subprocess`, after all we use `suppress_log=True there :shrug:

PR #332, opened 2022-03-31
Support celery appids with queue workers

kmaork:
Support appid for celery workers run like `celery -Q queue_name` or `celery --queues queue_name`

### DIFF
--- a/gprofiler/metadata/application_identifiers.py
+++ b/gprofiler/metadata/application_identifiers.py
@@ -18,7 +18,6 @@ from gprofiler.utils import run_process
 
 _logger = get_logger_adapter(__name__)
 
-
 _PYTHON_BIN_RE = re.compile(r"^python([23](\.\d{1,2})?)?$")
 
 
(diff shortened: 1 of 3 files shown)


tests/test_appids.py
```diff
@@ -110,7 +110,28 @@ def test_celery() -> None:
     assert "celery: /path/to/app3 (/path/to/app3.py)" == get_application_name(
         process_with_cmdline(["celery", "a", "b", "--app=/path/to/app3"])
     )
-    # No app
+
+
+def test_celery_with_queue() -> None:
```
Jongy, line 115: Also test with multiple queues? How does it look like? `-Q a,b,c` or `-Q a -Q b`?

Jongy (changes requested):
.

gprofiler/metadata/application_identifiers.py
```diff
@@ -204,6 +204,14 @@ def get_application_name(self, process: Process) -> Optional[str]:
...
             process.cmdline(), "--app", check_for_equals_arg=True
         )
+        if app_name is _NON_AVAILABLE_ARG:
+            queue_name = _get_cli_arg_by_name(process.cmdline(), "-Q") or _get_cli_arg_by_name(
+                process.cmdline(), "--queues", check_for_equals_arg=True
+            )
+            # TODO: One worker can handle multiple queues, it could be useful to encode that into the app id.
+            if queue_name is not _NON_AVAILABLE_ARG:
+                # The queue handler routing is defined in the directory where the worker is run
+                return f'celery queue: {queue_name} ({process.cwd()})'
```
Jongy, line 214: Does it make sense to add cwd?

We add it for script/file-based appids because the cwd makes a difference (i.e `python app.py` depends on cwd). Does it matter with queues?

kmaork: Yes, I wrote a comment in the line above

Jongy: Lol

tests/test_appids.py
kmaork: `-Q a,b,c`

PR #334, opened 2022-04-02
Bumped version to v1.2.19 + imports reordering

liorgorb:
Bumped version to v1.2.19 + imports reordering

PR #335, opened 2022-04-02
ci: Fail if isort modifies any files

Jongy:
## Description
Title.

See https://github.com/Granulate/gprofiler/pull/334#pullrequestreview-929610984


kmaork (approved):
Cool cool cool

PR #338, opened 2022-04-11
Upgrade granulate-utils and add a test for resolve_proc_root_links()

Jongy:
Upgrade granulate-utils, this includes 2 new commits:
1. Some changes to `resolve_proc_root_links()` - we didn't have a test that triggers the underlying logic in this function, so I hacked one now that tests it to full extent. CC @ty93 
2. `elf.py` moved to granulate-utils, use the copy from there.


Jongy:
Forgot to mention - I ran the new test with the old code (not the updated granulate-utils) and obv it fails.

tests/test_java.py
```diff
@@ -416,3 +416,47 @@ def test_java_noexec_dirs(
...
+    treatment is required for /proc/pid/root paths), and that profiling works eventually.
+    This basicaly tests the function resolve_proc_root_links().
+    """
+    caplog.set_level(logging.DEBUG)
+
+    # build this structure
+    # /run/final_tmp
+    # /run/step2 -> final_tmp
+    # /run/step1 -> step2
+    # /run/tmp -> /run/step1
```
adi-benz, line 440: `/run/tmpy`

tests/test_java.py
```diff
@@ -416,3 +416,47 @@ def test_java_noexec_dirs(
...
+    # /tmp -> /run/tmp
+    application_docker_container.exec_run(
+        [
+            "sh",
+            "-c",
+            "mkdir -p /run/final_tmp && "
+            "ln -s final_tmp /run/step2 && "
```
adi-benz, line 447:
```suggestion
            "ln -s final_tmp /run/step2 && "  # Test relative path
```

tests/test_java.py
```diff
@@ -416,3 +416,47 @@ def test_java_noexec_dirs(
...
+            "ln -s step2 /run/step1 && "
+            "ln -s /run/step1 /run/tmpy && "
```
adi-benz, line 449:
```suggestion
            "ln -s /run/step1 /run/tmpy && "  # Test absolute path
```

adi-benz (changes requested):

adi-benz (approved):

PR #346, opened 2022-04-25
Upgrade mypy to 0.942

shahar-gamliel:
Upgrade mypy version from 0.901 to 0.942, bumping granulate-utils to a version with the same change.

### DIFF
--- a/gprofiler/log.py
+++ b/gprofiler/log.py
@@ -75,30 +75,6 @@ class GProfilerLoggingAdapter(logging.LoggerAdapter):
         logging_kwargs["extra"] = extra
         return msg, logging_kwargs
 
-    def debug(self, msg: Any, *args: Any, no_server_log: bool = False, **kwargs: Any) -> None:
-        super().debug(msg, *args, no_server_log=no_server_log, **kwargs)
-
-    def info(self, msg: Any, *args: Any, no_server_log: bool = False, **kwargs: Any) -> None:
-        super().info(msg, *args, no_server_log=no_server_log, **kwargs)
-
-    def warning(self, msg: Any, *args: Any, no_server_log: bool = False, **kwargs: Any) -> None:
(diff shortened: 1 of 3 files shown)


gprofiler/log.py
```diff
@@ -75,30 +75,6 @@ def process(self, msg: Any, kwargs: MutableMapping[str, Any]) -> Tuple[Any, Muta
         logging_kwargs["extra"] = extra
         return msg, logging_kwargs
 
-    def debug(self, msg: Any, *args: Any, no_server_log: bool = False, **kwargs: Any) -> None:
```
Jongy, line 78: IIRC these were necessary for the `no_server_log: bool` parameter.

Jongy: Decided it's k because orig function accepts `kwargs` and we pop them anyway.

shahar-gamliel: The method dimply passes `no_server_log` as it is to the super method, which receives `kwargs` anyway, so this is redundant.
The motivation to this change is mypy errors similar to the one below:
``` bash
gprofiler/log.py:78: error: Signature of "debug" incompatible with supertype "LoggerAdapter"
gprofiler/log.py:78: note:      Superclass:
gprofiler/log.py:78: note:          def debug(self, msg: object, *args: object, exc_info: Union[None, bool, Union[Tuple[Type[BaseException], BaseException, Optional[TracebackType]], Tuple[None, None, None]], BaseException] = ..., stack_info: bool = ..., extra: Optional[Mapping[str, object]] = ..., **kwargs: object) -> None
gprofiler/log.py:78: note:      Subclass:
gprofiler/log.py:78: note:          def debug(self, msg: Any, *args: Any, no_server_log: bool = ..., **kwargs: Any) -> None
```

Jongy (approved):

PR #347, opened 2022-04-25
utils: fs: Use randomized path for test script

Jongy:
Function is called concurrently and possibly on the same mnt namespace, so threads collide writing/removing/executing the script.

Easiest way to avoid that is by randomizing its name.

### DIFF
--- a/gprofiler/utils/fs.py
+++ b/gprofiler/utils/fs.py
@@ -7,6 +7,7 @@ import errno
 import os
 import shutil
 from pathlib import Path
+from secrets import token_hex
 
 from gprofiler.utils import remove_path, run_process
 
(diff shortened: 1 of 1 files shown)


adi-benz (approved):
We actually saw that happen in production?

Jongy:
> We actually saw that happen in production?

Yes.

```
Traceback (most recent call last):   File "gprofiler/profilers/profiler_base.py", line 135, in snapshot   File "concurrent/futures/_base.py", line 425, in result   File "concurrent/futures/_base.py", line 384, in __get_result   File "concurrent/futures/thread.py", line 56, in run   File "gprofiler/profilers/java.py", line 788, in _profile_process   File "gprofiler/profilers/java.py", line 246, in __init__   File "gprofiler/profilers/java.py", line 276, in _find_rw_exec_dir   File "gprofiler/utils/fs.py", line 49, in is_rw_exec_dir   File "pathlib.py", line 1284, in unlink   File "pathlib.py", line 387, in wrapped FileNotFoundError: [Errno 2] No such file or directory: '/proc/1/root/tmp/gprofiler_tmp/t.sh'
```

PR #353, opened 2022-04-29
utils: perf: Catch another case of perf permissions error

Jongy:
[17:21:12] Couldn't create the Java profiler, continuing without it
	  File "gprofiler/profilers/factory.py", line 41, in get_profilers
	  File "gprofiler/profilers/java.py", line 637, in __init__
	  File "gprofiler/profilers/java.py", line 654, in _init_ap_mode
	  File "gprofiler/utils/perf.py", line 19, in can_i_use_perf_events
	  File "gprofiler/utils/__init__.py", line 265, in run_process
	gprofiler.exceptions.CalledProcessError: Command '['/tmp/_MEIYxgsVy/gprofiler/resources/perf', 'record', '-o', '/dev/null', '--', '/bin/true']' returned non-zero exit status 255.
	stdout: b''
	stderr: b'Permission error mapping pages.\nConsider increasing /proc/sys/kernel/perf_event_mlock_kb,\nor try again with a smaller value of -m/--mmap_pages.\n(current value: 4294967295,0)\n'

I don't know, maybe it's just best to catch all errors and treat them as "False" instead of failing the `JavaProfiler` completely. On the other hand, that might hide real errors. Well, keeping it this way for now.

### DIFF
--- a/gprofiler/utils/perf.py
+++ b/gprofiler/utils/perf.py
@@ -4,7 +4,10 @@
 #
 
 from gprofiler.exceptions import CalledProcessError
+from gprofiler.log import get_logger_adapter
 from gprofiler.utils import resource_path, run_process
+
+logger = get_logger_adapter(__name__)
 
 
 def perf_path() -> str:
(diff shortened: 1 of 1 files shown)


adi-benz:
> I don't know, maybe it's just best to catch all errors and treat them as "False" instead of failing the `JavaProfiler` completely. On the other hand, that might hide real errors. Well, keeping it this way for now.

What about changing it to `return False` and logging the error?

Jongy:
> > I don't know, maybe it's just best to catch all errors and treat them as "False" instead of failing the `JavaProfiler` completely. On the other hand, that might hide real errors. Well, keeping it this way for now.
> 
> What about changing it to `return False` and logging the error?

Yes, let's do that

gprofiler/utils/perf.py
```diff
@@ -19,12 +22,18 @@ def can_i_use_perf_events() -> bool:
         run_process([perf_path(), "record", "-o", "/dev/null", "--", "/bin/true"])
     except CalledProcessError as e:
         # perf's output upon start error (e.g due to permissions denied error)
-        if e.returncode == 255 and (
-            b"Access to performance monitoring and observability operations is limited" in e.stderr
-            or b"perf_event_open(..., PERF_FLAG_FD_CLOEXEC) failed with unexpected error" in e.stderr
+        if not (
+            e.returncode == 255
+            and (
+                b"Access to performance monitoring and observability operations is limited" in e.stderr
```
adi-benz, line 28: Why not log these errors? Even though we know these as possible errors we'll still be interested in knowing why perf couldn't run and why we're using `itimer` instead of `cpu`

adi-benz (changes requested):

gprofiler/utils/perf.py
Jongy: This error is legit - gProfiler is used in non-privileged settings as well. We're using `perf` as a generic mean to check for `perf_events` and mmaping pages etc. All "known" errors listed here - I am fine if they trigger, their cause is clear to me and no need to log. Other types of errors - may be other bugs, e.g `perf` failing to start for whatever reason, and those I want logged.

adi-benz (approved):

Issue #342, opened 2022-04-16
Aarch64 older than CentOS 8 (glibc 2.28) not supported


Jongy:
In https://github.com/Granulate/gprofiler/pull/314 we changed the Aarch64 to generate executables on CentOS 8.

However, I forgot that we don't use staticx on Aarch64 (see [here](https://github.com/Granulate/gprofiler/blob/9c7c02e3c39ac3658f887d06d7a7ce553eabc67b/pyi.Dockerfile#L213)) therefore all binaries we build (including the Python distribution used to run gProfiler itself) require the target system to provide libraries.
Now specifically, we build on CentOS 8 which uses glibc 2.28, and that's what our Python requires, then :( So we can't run on a system having older libcs, such as Amazon Linux 2 `4.14.273-207.502.amzn2.aarch64`.

We need to enable staticx for Aarch64 or use another packaging mechanism.

Jongy:
Used this diff to build on an Aarch64 machine:
```
diff --git a/pyi.Dockerfile b/pyi.Dockerfile
index 8f855fe..4e28f90 100644
--- a/pyi.Dockerfile
+++ b/pyi.Dockerfile
@@ -210,8 +210,7 @@ COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # we use list_needed_libs.sh to list the dynamic dependencies of *all* of our resources,
 # and make staticx pack them as well.
 # using scl here to get the proper LD_LIBRARY_PATH set
-# TODO: use staticx for aarch64 as well; currently it doesn't generate correct binaries when run over Docker emulation.
-RUN if [ $(uname -m) != "aarch64" ]; then source scl_source enable devtoolset-8 llvm-toolset-7 && libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gprofiler dist/gprofiler; fi
+RUN libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gprofiler dist/gprofiler
 
 FROM scratch AS export-stage
```

Results in:
```
ubuntu@ip-172-31-21-72:~/gprofiler$ sudo ./build/aarch64/gprofiler --version
gprofiler: dl-call-libc-early-init.c:37: _dl_call_libc_early_init: Assertion `sym != NULL' failed.
Aborted
```

requires more work :/ Perhaps staticx is not meant for Aarch64...

PR #287, opened 2022-02-04
python: PyPerf: Aarch64 support

Jongy:
## Description
Support PyPerf running on Aarch64.

This currently passes "sanity" - works on Python 2.7.18 on Ubuntu 20.04 kernel `5.13.0-1008-aws`. I am opening this PR to document what's left to be done & the tests we should run on it.

This depends on:
* BCC PR: https://github.com/Granulate/bcc/pull/34
* Aarch64 support for https://github.com/Jongy/bpf_get_stack_offset/tree/aarch64
* Aarch64 support for https://github.com/Jongy/bpf_get_fs_offset/tree/aarch64

## TODOs
* [ ] Figure out BCC build - with LLVM shared or not - so it works both in container & exe Aarch64 and doesn't break x86_64.
* [ ] Create final tag in https://github.com/Jongy/bpf_get_stack_offset
* [ ] Create final tag in https://github.com/Jongy/bpf_get_fs_offset
* [ ] Figure out if we really depend on https://lore.kernel.org/bpf/796ee46e948bc808d54891a1108435f8652c6ca4.1572649915.git.daniel@iogearbox.net/#r for reading usermode memory. If so, do not attempt to run PyPerf on lower kernel versions.
* [ ] Clean up BCC PR, run tests alongside, then merge it :)

### Testing
x86_64:
* [ ] x86_64 native stacks (as I have made changes in that area) when in user mode
* [ ] x86_64 native stacks (as I have made changes in that area) when in kernel mode
* [ ] x86_64 gProfiler as container PyPerf
* [ ] x86_64 gProfiler as exe PyPerf

Aarch64:
* [ ] Aarch64 gProfiler as container PyPerf
* [ ] Aarch64 gProfiler as exe PyPerf
* [ ] Aarch64 native stacks when in user mode
* [ ] Aarch64 native stacks when in kernel mode
* [ ] Aarch64 kernel stacks
* [ ] Aarch64 Python 2.7 Python stacks incl classes (glibc)
* [ ] Aarch64 Python 3.6 Python stacks incl classes (glibc)
* [ ] Aarch64 Python 3.7 Python stacks incl classes (glibc)
* [ ] Aarch64 Python 3.8 Python stacks incl classes (glibc)
* [ ] Aarch64 Python 3.9 Python stacks incl classes (glibc)
* [ ] Aarch64 Python 3.10 Python stacks incl classes (glibc)
* [ ] Aarch64 Python 3.8 Python stacks (musl based, one Python version is enough)

PR #355, opened 2022-05-01
Use staticx for Aarch64 exe builds

Jongy:
Closes: https://github.com/Granulate/gprofiler/issues/342

## Description
The problem starts with #342  - we needed to solve the glibc issue after upgrading the base image to CentOS 8.

staticx is the easiest solution because that's what we already do for x86_64. However, simply enabling staticx with this patch:
```
diff --git a/pyi.Dockerfile b/pyi.Dockerfile
index 8f855fe..952226b 100644
--- a/pyi.Dockerfile
+++ b/pyi.Dockerfile
@@ -211,7 +211,7 @@ COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # TODO: use staticx for aarch64 as well; currently it doesn't generate correct binaries when run over Docker emulation.
-RUN if [ $(uname -m) != "aarch64" ]; then source scl_source enable devtoolset-8 llvm-toolset-7 && libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gprofiler dist/gprofiler; fi
+RUN libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gprofiler dist/gprofiler
 
 FROM scratch AS export-stage
```
Results in this binary:
```
ubuntu@ip-172-31-21-72:~/gprofiler$ ./build/aarch64/gprofiler  --version
gprofiler: dl-call-libc-early-init.c:37: _dl_call_libc_early_init: Assertion `sym != NULL' failed.
Aborted (core dumped)
```

Not good.

gdb shows this trace:
```
gprofiler: dl-call-libc-early-init.c:37: _dl_call_libc_early_init: Assertion `sym != NULL' failed.

Program received signal SIGABRT, Aborted.
0x000000000040917c in raise ()
(gdb) bt
#0  0x000000000040917c in raise ()
#1  0x0000000000400380 in abort ()
#2  0x0000000000404f28 in __assert_fail_base ()
#3  0x0000000000404f90 in __assert_fail ()
#4  0x000000000045b9e0 in _dl_call_libc_early_init ()
#5  0x0000000000459b34 in dl_open_worker ()
#6  0x00000000004346c4 in _dl_catch_exception ()
#7  0x00000000004594c0 in _dl_open ()
#8  0x0000000000433e7c in do_dlopen ()
#9  0x00000000004346c4 in _dl_catch_exception ()
#10 0x0000000000434790 in _dl_catch_error ()
#11 0x0000000000433ecc in dlerror_run ()
#12 0x0000000000434368 in __libc_dlopen_mode ()
#13 0x00000000004305cc in __nss_lookup_function ()
#14 0x0000000000430664 in __nss_lookup ()
#15 0x000000000042b0fc in getpwnam_r ()
#16 0x000000000042adb4 in getpwnam ()
#17 0x00000000004027ac in th_get_uid ()
#18 0x00000000004019ec in tar_extract_file ()
#19 0x0000000000401d00 in tar_extract_all ()
#20 0x0000000000400f90 in extract_archive ()
#21 0x0000000000400510 in main ()
(gdb) 
```

Tar extraction calling into `getpwnam` which ends up loading DSOs... sigh.
We don't need this functionality.

So, this PR:
1. Adds a patch for latest staticx version that disables using `getpwnam` and `getgrnam`.
2. Builds staticx from source with the patch for Aarch64
3. Runs staticx also for Aarch64.

## Known issues
The exe build doesn't pass on x86_64 anymore.
```
 => ERROR [build-stage 53/53] RUN if [ $(uname -m) != "aarch64" ]; then source scl_source enable devtoolset-8 llvm-toolset-7 ; fi; libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gprof  0.7s
------
 > [build-stage 53/53] RUN if [ $(uname -m) != "aarch64" ]; then source scl_source enable devtoolset-8 llvm-toolset-7 ; fi; libs=$(./scripts/list_needed_libs.sh) && staticx $libs dist/gprofiler dist/gprofiler:
#106 0.690 ldd failed: ldd: exited with unknown exit code (139)
```

I suspect this to be a bug with the QEMU that emulates the usermode environment for Aarch64 binaries. It doesn't set up a correct environment for `ldd`.

Before we merge this PR, I plan to inspect if now after https://github.com/Granulate/gprofiler/pull/350 I can provision an Aarch64 runner and separate the Aarch64 build steps to run on it instead. This will also make the build faster :) but mostly it will not crash haha.
Alternatively we need to fix the build when running on x86_64. Maybe upgrade QEMU? We all run `multiarch/qemu-user-static:latest` so IDK....

## How Has This Been Tested?
All current binaries used in Aarch64 builds of gProfiler are static - only PyPerf is dynamically built, hence this PR is also in preparation for https://github.com/Granulate/gprofiler/pull/287 which finally uses PyPerf in Aarch64.
For the mean time, the only thing "tested" here is that gProfiler's Python starts - staticx provides us with libc. So the tests I have performed for Aarch64 are just `gprofiler --version`. For x86_64, I tested PyPerf on appropriate kernel versions as well, and not just `gprofiler --version`.

* [x] Aarch64 new glibc - > 2.28
* [x] Aarch64 old glibc - < 2.28
* [ ] x86_64 very old (CentOS 6)
* [ ] x86_64 gprofiler & PyPerf

### DIFF
--- a/scripts/staticx_patch.diff
+++ b/scripts/staticx_patch.diff
@@ -1,30 +1,28 @@
 diff --git a/libtar/decode.c b/libtar/decode.c
-index b34b1d9..5105cb0 100644
+index b34b1d9..d221c3a 100644
 --- a/libtar/decode.c
 +++ b/libtar/decode.c
-@@ -46,9 +46,9 @@ th_get_uid(const TAR *t)
-        int uid;
-        struct passwd *pw;
+@@ -44,11 +44,6 @@ uid_t
+ th_get_uid(const TAR *t)
+ {
+ 	int uid;
+-	struct passwd *pw;
+-
+-	pw = getpwnam(t->th_buf.uname);
+-	if (pw != NULL)
+-		return pw->pw_uid;
  
--       pw = getpwnam(t->th_buf.uname);
--       if (pw != NULL)
--               return pw->pw_uid;
-+       // pw = getpwnam(t->th_buf.uname);
-+       // if (pw != NULL)
-+       //      return pw->pw_uid;
+ 	/* if the password entry doesn't exist */
+ 	sscanf(t->th_buf.uid, "%o", &uid);
+@@ -60,11 +55,6 @@ gid_t
+ th_get_gid(const TAR *t)
+ {
+ 	int gid;
+-	struct group *gr;
+-
+-	gr = getgrnam(t->th_buf.gname);
+-	if (gr != NULL)
+-		return gr->gr_gid;
  
-        /* if the password entry doesn't exist */
-        sscanf(t->th_buf.uid, "%o", &uid);
-@@ -62,9 +62,9 @@ th_get_gid(const TAR *t)
-        int gid;
-        struct group *gr;
- 
--       gr = getgrnam(t->th_buf.gname);
--       if (gr != NULL)
--               return gr->gr_gid;
-+       // gr = getgrnam(t->th_buf.gname);
-+       // if (gr != NULL)
-+       //      return gr->gr_gid;
- 
-        /* if the group entry doesn't exist */
-        sscanf(t->th_buf.gid, "%o", &gid);
+ 	/* if the group entry doesn't exist */
+ 	sscanf(t->th_buf.gid, "%o", &gid);
--- a/scripts/staticx_patch.diff
+++ b/scripts/staticx_patch.diff
@@ -1,0 +1,30 @@
+diff --git a/libtar/decode.c b/libtar/decode.c
+index b34b1d9..5105cb0 100644
+--- a/libtar/decode.c
++++ b/libtar/decode.c
+@@ -46,9 +46,9 @@ th_get_uid(const TAR *t)
+        int uid;
+        struct passwd *pw;
+ 
+-       pw = getpwnam(t->th_buf.uname);
+-       if (pw != NULL)
+-               return pw->pw_uid;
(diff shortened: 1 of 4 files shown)


scripts/staticx_patch.diff
```diff
@@ -0,0 +1,28 @@
...
+ {
+ 	int gid;
+-	struct group *gr;
+-
+-	gr = getgrnam(t->th_buf.gname);
+-	if (gr != NULL)
+-		return gr->gr_gid;
+ 
+ 	/* if the group entry doesn't exist */
+ 	sscanf(t->th_buf.gid, "%o", &gid);
```
Jongy, line 28: @LiorMosk do you mind checking. I actually checked the unpacked files of staticx and they look alright. But I still want to verify that the contents of `t->th_buf.gid` (and same for `t->th_buf.uid`) are always legit in our case (i.e no chance that if `uname` is set to `"root"` then `uid` is not set because `root` is expected to always be processes by `getpwnam`, for example...)

Read the sources or maybe build staticx with prints and see?

LiorMosk: It seems to be ok, that way

LiorMosk (dismissed):

Jongy:
I think this will fix the build.

LiorMosk (approved):

PR #356, opened 2022-05-01
ci: Split container build so that Aarch64 runs on the Aarch64 runner

Jongy:
Build Aarch64 container on our Aarch64 runner, this will make the build & deploy much faster :)

Also, changed `ubuntu-20.04` runners to `ubuntu-latest`. The `ubuntu-20.04` runner was chosen when I debugged something which ended up being unrelated, so time to make it latest back.


adi-benz (changes requested):
CI problems

adi-benz (approved):

Jongy:
Hope it will run :)

PR #366, opened 2022-05-15
appids: Rework --app-id-args-filter

Jongy:
It will now:
* Search the entire arguments line (instead of individual arguments)
* Search arguments passed to Java as well (i.e from java_command)

Also, renamed it in the code to make it clear that it refers to Java only.

@liorgorb - the new scheme will work for your use case, lmk if you think anything else should be changed.

I will test after you approve this approach.

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -545,12 +545,14 @@ def parse_cmd_args() -> configargparse.Namespace:
         help="Disable identification of applications by heuristics",
     )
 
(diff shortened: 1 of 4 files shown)


Jongy:
@liorgorb - I tested it with:
* JVM arg like `-Xss[^ ]+` - matches e.g `-Xss10m`
* Java arg like `--blabla[= ]\d+` - matches `--blabla=555 and `--blabla 555` (2 separate arguments).

PR #371, opened 2022-06-02
java: Log "java -version" output

Jongy:


adi-benz (changes requested):

gprofiler/profilers/java.py
```diff
@@ -207,7 +207,9 @@ def _run_java_version() -> "CompletedProcess[bytes]":
 
     # doesn't work without changing PID NS as well (I'm getting ENOENT for libjli.so)
     # Version is printed to stderr
-    return run_in_ns(["pid", "mnt"], _run_java_version, process.pid).stderr.decode().strip()
+    version = run_in_ns(["pid", "mnt"], _run_java_version, process.pid).stderr.decode().strip()
+    logger.debug("java -version output", java_version_output=version, pid=process.pid)
```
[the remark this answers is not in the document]
Jongy: You are most certainly correct

gprofiler/profilers/java.py
```diff
@@ -830,7 +836,7 @@ def _is_jvm_profiling_supported(self, process: Process) -> bool:
                 return False
         else:
             if self._simple_version_check and process_basename == "java":
-                java_version_output = get_java_version(process, self._stop_event)
+                java_version_output = get_java_version_logged(process, self._stop_event)
```
adi-benz, line 839: Don't we want the version to be logged even if no checks at all?

If so - we can log the java version from outside this function if `basename == 'java'`

Jongy: I don't want to complicats the logic any more. This is seldom used anyway

adi-benz (approved):

PR #374, opened 2022-06-08
deploy: k8s: Fix "256m" to "256Mi"

Jongy:


Jongy:
Also prefer the Mi to M.

deploy/k8s/gprofiler.yaml
```diff
@@ -31,7 +31,7 @@ spec:
           resources:
             requests:
               cpu: 100m
-              memory: 256m
+              memory: 256Mi
```
adi-benz, line 34: Same for `k8s/helm-charts/values.yaml`

adi-benz (changes requested):

deploy/k8s/gprofiler.yaml
Jongy: Oh I actually checked another helm file and missed it.

Anyway it's retarded that this is cloned. Will fix in another PR

adi-benz (approved):

PR #378, opened 2022-06-14
Add metadata for golang processes

IlayRosenberg:
## Description

Collect the following metadata for golang processes:
* golang runtime version
* whether the binary is statically linked
* if not - whether it was linked with glibc or musl (currently assume these are the only 2 possibilities)

## How Has This Been Tested?

I added a test case to `test_app_metadata`

## Checklist:

### DIFF
--- a/gprofiler/profilers/perf.py
+++ b/gprofiler/profilers/perf.py
@@ -219,7 +219,7 @@ class SystemProfiler(ProfilerBase):
             metadata = None
             try:
                 if self._golang_metadata.is_golang_process(Process(k)):
-                    metadata = self._golang_metadata.make_application_metadata(Process(k))
+                    metadata = self._golang_metadata.get_metadata(Process(k))
             except (NoSuchProcess, FileNotFoundError):
                 pass
             profile_data_dict[k] = ProfileData(v, None, metadata)
@@ -243,7 +243,7 @@ class GolangMetadata(ApplicationMetadata):
         # 	str unsafe.Pointer
         # 	len int
         # }
-        addr, length = struct.unpack("<QQ", symbol_data)
+        addr, length = struct.unpack("QQ", symbol_data)
         golang_version_bytes = read_elf_va(elf_path, addr, length)
         if golang_version_bytes is None:
             return None
@@ -254,6 +254,8 @@ class GolangMetadata(ApplicationMetadata):
         metadata = {"golang_version": self._get_golang_version(process), "link": "static" if static else "dynamic"}
         if not static:
             metadata["libc"] = "musl" if is_musl(process) else "glibc"
+        else:
+            metadata["libc"] = None
 
         metadata.update(super().make_application_metadata(process))
         return metadata
--- a/gprofiler/profilers/perf.py
+++ b/gprofiler/profilers/perf.py
@@ -17,7 +17,7 @@ from psutil import NoSuchProcess, Process
 
 from gprofiler import merge
 from gprofiler.exceptions import StopEventSetException
-from gprofiler.gprofiler_types import ProcessToProfileData, ProfileData
+from gprofiler.gprofiler_types import AppMetadata, ProcessToProfileData, ProfileData
 from gprofiler.log import get_logger_adapter
 from gprofiler.metadata.application_metadata import ApplicationMetadata
 from gprofiler.profilers.profiler_base import ProfilerBase
@@ -165,7 +165,7 @@ class SystemProfiler(ProfilerBase):
     ):
         super().__init__(frequency, duration, stop_event, storage_dir)
         self._perfs: List[PerfProcess] = []
-        self._golang_metadata = GolangMetadata(stop_event)
+        self._metadata_collectors: List[PerfMetadata] = [GolangPerfMetadata(stop_event)]
 
         if perf_mode in ("fp", "smart"):
             self._perf_fp: Optional[PerfProcess] = PerfProcess(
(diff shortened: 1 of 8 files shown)


gprofiler/profilers/perf.py
```diff
@@ -202,11 +210,50 @@ def snapshot(self) -> ProcessToProfileData:
...
+        # 	str unsafe.Pointer
+        # 	len int
+        # }
+        addr, length = struct.unpack("<QQ", symbol_data)
+        golang_version_bytes = read_elf_va(elf_path, addr, length)
+        if golang_version_bytes is None:
+            return None
+        return golang_version_bytes.decode()
+
+    def make_application_metadata(self, process: Process) -> Dict[str, Any]:
```
YoniKF, line 252: Consider caching this one as well.

gprofiler/profilers/perf.py
```diff
@@ -202,11 +210,50 @@ def snapshot(self) -> ProcessToProfileData:
...
+        elf_path = f"/proc/{process.pid}/exe"
+        symbol_data = read_elf_symbol(elf_path, "runtime.buildVersion", 16)
+        if symbol_data is None:
+            return None
+        # Declaration of go string type:
+        # type stringStruct struct {
+        # 	str unsafe.Pointer
+        # 	len int
+        # }
+        addr, length = struct.unpack("<QQ", symbol_data)
```
Jongy, line 246:
```suggestion
        addr, length = struct.unpack("QQ", symbol_data)
```

gProfiler also runs on ARM, IDK if it's LE or BE

gprofiler/profilers/perf.py
Jongy: No caching until this is proven heavy

gprofiler/profilers/perf.py
```diff
@@ -202,11 +210,50 @@ def snapshot(self) -> ProcessToProfileData:
...
+
+        for k, v in merge.merge_global_perfs(
+            self._perf_fp.wait_and_script() if self._perf_fp is not None else None,
+            self._perf_dwarf.wait_and_script() if self._perf_dwarf is not None else None,
+        ).items():
+            metadata = None
+            try:
+                if self._golang_metadata.is_golang_process(Process(k)):
+                    metadata = self._golang_metadata.make_application_metadata(Process(k))
+            except (NoSuchProcess, FileNotFoundError):
```
Jongy, line 223: `FileNotFoundError`?

gprofiler/profilers/perf.py
```diff
@@ -158,6 +165,7 @@ def __init__(
     ):
         super().__init__(frequency, duration, stop_event, storage_dir)
         self._perfs: List[PerfProcess] = []
+        self._golang_metadata = GolangMetadata(stop_event)
```
Jongy, line 168: Move Golang related code to `SystemProfilerWithGolang(SystemProfiler)`, that's a good case for extending the class. (name is up to you :)

gprofiler/profilers/perf.py
Jongy: Eh what am I talking about. It's already cached. Just use the top-level API `get_metadata` which does the caching. (in `snapshot`)

Jongy (changes requested):
.

gprofiler/profilers/perf.py
IlayRosenberg: When trying to access `/proc/{pid}/exe`

gprofiler/profilers/perf.py
Jongy: (fix the usage in gProfiler to use the new `SystemProfilerWithGolang`)

Jongy: Actually, it's not very convenient with the profilers registry, so meh. If u find another way to encapsulate the golang code in another class, good. Otherwise meh

gprofiler/profilers/perf.py
IlayRosenberg: Although I now see `get_metadata` catches anything other than `NoSuchProcess`. Do you think I should catch `FileNotFoundError` and raise `NoSuchProcess` instead?

gprofiler/profilers/perf.py
```diff
@@ -204,9 +222,61 @@ def snapshot(self) -> ProcessToProfileData:
...
+    def relevant_for_process(self, process: Process) -> bool:
+        return False
+
+
+class GolangPerfMetadata(PerfMetadata):
+    def relevant_for_process(self, process: Process) -> bool:
+        version = self._get_golang_version(process)
+        return version is not None
+
+    @functools.lru_cache(maxsize=128)
```
Jongy, line 243: Can increase over 128 (128 processes is less than a typical system has). Use 4096

Jongy (changes requested):
Final

Jongy:
@IlayRosenberg - I am merging master into your branch, I solved a couple of CI issues

Jongy:
```
[2022-06-22 09:28:21,441] ERROR: gprofiler: Profiling run failed!
  File "/app/gprofiler/main.py", line 366, in run_continuous
    self._snapshot()
  File "/app/gprofiler/main.py", line 276, in _snapshot
    system_result = system_future.result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 444, in result
    return self.__get_result()
    raise self._exception
  File "/app/gprofiler/profilers/perf.py", line 223, in snapshot
    return {
  File "/app/gprofiler/profilers/perf.py", line 225, in <dictcomp>
    k: ProfileData(v, None, self._get_metadata(k))
  File "/app/gprofiler/profilers/perf.py", line 208, in _get_metadata
    process = Process(pid)
  File "/usr/local/lib/python3.8/dist-packages/psutil/__init__.py", line 326, in __init__
    self._init(pid)
  File "/usr/local/lib/python3.8/dist-packages/psutil/__init__.py", line 335, in _init
    raise ValueError('pid must be a positive integer (got %s)'
ValueError: pid must be a positive integer (got -1)
```

`perf` might report invalid PIDs as PIDs. See #326 where we had `get_application_metadata` that ignored specific PIDs by perf (`0`, `-1`).

PR #380, opened 2022-06-16
tests: Remove xfail

Jongy:
It seems to pass now and we're getting errors due to the XPASS


Jongy:
Should now pass and ready for merge

PR #384, opened 2022-06-19
tests: Print error logs from executed gProfiler container

Jongy:
Extracted from https://github.com/Granulate/gprofiler/pull/380

This PR makes sure that the gProfiler container we execute in the tests always prints its logs in entirety.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -38,6 +38,7 @@ def run_privileged_container(
     if volumes is None:
         volumes = {}
 
+    container = None
     try:
         container = docker_client.containers.run(
             image,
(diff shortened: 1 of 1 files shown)


Jongy:
The errors of the tests here are fixed in https://github.com/Granulate/gprofiler/pull/385, https://github.com/Granulate/gprofiler/pull/380

tests/utils.py
```diff
@@ -33,55 +33,58 @@ def run_privileged_container(
...
+        # let it finish
+        exit_status = container.wait()["StatusCode"]
+        # and read its log
+        logs = container.logs(stdout=True, stderr=True)
+
+        if exit_status != 0:
+            raise ContainerError(container, exit_status, command, image, logs)
+    finally:
+        if container is not None:
+            container.remove()
```
michelhe, line 64: you can pass `rm=Tru` to `docker_client.containers.run

Jongy: but then I won't have logs

michelhe (approved):

PR #385, opened 2022-06-19
Write output collapsed files & flamegraphs in utf-8

Jongy:
This combats failures I've seen in the tests: https://github.com/Granulate/gprofiler/runs/6954306861?check_suite_focus=true

Anyway it makes sense to write in utf-8 because symbols may have non-ascii letters.

This is extracted from https://github.com/Granulate/gprofiler/pull/380


Jongy:
The errors of the tests here are fixed by https://github.com/Granulate/gprofiler/pull/380

Jongy:
Failure is solved by another PR. Merging

PR #386, opened 2022-06-20
java: Log all Java exits & the exiting Java version

Jongy:
1. Cache java versions retrievals (a simple cache on `get_java_version` which is the core function)
2. Maintain pid-to-java-version mapping to be able to attach the exiting Java version to an exiting PID
3. Log exiting PID & Java version also for non-profiled processes (processes we would have profiled had the pre-checks passed)

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -763,8 +763,8 @@ class JavaProfiler(ProcessProfilerBase):
             logger.warning("Java profiling has been disabled, will avoid profiling any new java processes", cause=cause)
             self._safemode_disable_reason = cause
 
-    def _profiling_skipped_stack(self, reason: str, comm: str) -> StackToSampleCount:
-        return self._profiling_error_stack("skipped", reason, comm)
+    def _profiling_skipped_profile(self, reason: str, comm: str) -> ProfileData:
+        return ProfileData(self._profiling_error_stack("skipped", reason, comm), None, None)
 
     def _is_jvm_type_supported(self, java_version_cmd_output: str) -> bool:
         return all(exclusion not in java_version_cmd_output for exclusion in self.JDK_EXCLUSIONS)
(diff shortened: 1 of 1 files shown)


gprofiler/profilers/java.py
```diff
@@ -1017,18 +1033,33 @@ def stop(self) -> None:
...
+                    signal=signo,
+                    java_version_output=java_version_output,
+                )
 
-            if is_java_fatal_signal(signo):
-                self._disable_profiling(JavaSafemodeOptions.PROFILED_SIGNALED)
+                if is_java_fatal_signal(signo):
+                    self._disable_profiling(JavaSafemodeOptions.PROFILED_SIGNALED)
+            else:
+                # this is a process that we wanted to profile, but didn't profile due to safemode.
```
michelhe, line 1056: could also be short lived and exit by the time you try to start AP

michelhe:
merge from master

gprofiler/profilers/java.py
Jongy:
```suggestion
                # this is a process that we wanted to profile, but didn't profile due to safemode / any other reason.
```

michelhe (approved):

gprofiler/profilers/java.py
```diff
@@ -884,17 +895,34 @@ def _check_async_profiler_loaded(self, process: Process) -> bool:
...
+        if is_java_basename(process):
+            java_version_output: Optional[str] = get_java_version_logged(process, self._stop_event)
+        else:
+            java_version_output = None
+
+        if self._enabled_proc_events:
+            self._want_to_profile_pids.add(process.pid)
+            # there's no reliable way to get the underlying cache of get_java_version, otherwise
+            # I'd just use it.
+            if len(self._pid_to_java_version) > _JAVA_VERSION_CACHE_MAX:
```
Jongy, line 912: Too bad `popitem` pops the last one, not the first one

Jongy: Welp we `popitem` later

PR #388, opened 2022-06-26
Set critical logs to "CRITICAL"

Jongy:
1. CRITICAL logs are bad errors that prevent the profiler from operating correctly
2. ERROR logs are stuff we don't expect to happen and might indicate a bug on our code
3. WARN logs are errors we are aware of and can happen

Currently marked 2 bad errors as CRITICAL, more will be added

### DIFF
--- a/gprofiler/profilers/factory.py
+++ b/gprofiler/profilers/factory.py
@@ -42,7 +42,7 @@ def get_profilers(
         except Exception:
             if profiler_config.profiler_class is SystemProfiler:
                 raise SystemProfilerInitFailure("Could not create the system profiler")
-            logger.exception(f"Couldn't create the {profiler_name} profiler, continuing without it")
+            logger.critical(f"Couldn't create the {profiler_name} profiler, continuing without it", exc_info=True)
         else:
             if isinstance(profiler_instance, SystemProfiler):
                 system_profiler = profiler_instance
--- a/gprofiler/profilers/factory.py
+++ b/gprofiler/profilers/factory.py
@@ -1,7 +1,7 @@
 import sys
 from typing import TYPE_CHECKING, Any, List, Tuple, Union
 
-from gprofiler.exceptions import NoProfilersEnabledError, SystemProfilerInitFailure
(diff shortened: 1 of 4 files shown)


gprofiler/profilers/factory.py
```diff
@@ -42,7 +42,7 @@ def get_profilers(
         except Exception:
             if profiler_config.profiler_class is SystemProfiler:
                 raise SystemProfilerInitFailure("Could not create the system profiler")
-            logger.exception(f"Couldn't create the {profiler_name} profiler, continuing without it")
+            logger.critical(f"Couldn't create the {profiler_name} profiler, continuing without it", exc_info=True)
```
d3dave, line 45: This is the important one, why don't we exit with error on this, suggesting to run with `--no-{profiler_name}` if user wants to ignore?

gprofiler/profilers/perf.py
```diff
@@ -80,7 +80,7 @@ def start(self) -> None:
         except TimeoutError:
             process.kill()
             assert process.stdout is not None and process.stderr is not None
-            logger.error(f"perf failed to start. stdout {process.stdout.read()!r} stderr {process.stderr.read()!r}")
+            logger.critical(f"perf failed to start. stdout {process.stdout.read()!r} stderr {process.stderr.read()!r}")
```
Jongy, line 83: Same for this one, to be honest: https://github.com/Granulate/gprofiler/pull/388/files#r906863849

gprofiler/profilers/factory.py
Jongy: Agree, it should exit here probably. Will change

Jongy: Done

gprofiler/profilers/perf.py
Jongy: Done

Jongy: Although I didn't exit brutally, that's a decision for another PR

d3dave (approved):

PR #390, opened 2022-06-26
java: Cache is_musl() calls across AsyncProfiledProcess instances

Jongy:
I see it high in some profiles :)

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -83,6 +83,11 @@ def frequency_to_ap_interval(frequency: int) -> int:
     return int((1 / frequency) * 1_000_000_000)
 
 
+@functools.lru_cache(maxsize=1024)
+def is_musl_cached(process: Process) -> bool:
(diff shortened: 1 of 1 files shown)


michelhe (approved):
Have you confirmed that the cache is of any help at all? The highs you have seen might come with zero reentrance (new short lived processes)

Jongy:
> Have you confirmed that the cache is of any help at all? The highs you have seen might come with zero reentrance (new short lived processes)

Haven't confirmed but the cost of the cache is 0 compared to the cost of reading `/proc/pid/maps`, and I have saw this issue often with many services

michelhe:
> > Have you confirmed that the cache is of any help at all? The highs you have seen might come with zero reentrance (new short lived processes)
> 
> 
> 
> Haven't confirmed but the cost of the cache is 0 compared to the cost of reading `/proc/pid/maps`, and I have saw this issue often with many services

image

PR #393, opened 2022-06-29
Spark app name identifier collector

grariel:
## Description
Added collection of spark application name as an appid.

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [ ] I have updated the relevant documentation.
- [ ] I have added tests for new logic.

PR #394, opened 2022-06-30
Cache appid calculation per process

michelhe:


gprofiler/metadata/application_identifiers.py
```diff
@@ -317,6 +320,9 @@ def get_app_id(process: Process, identifiers: List[_ApplicationIdentifier]) -> O
     if not _ApplicationIdentifier.enrichment_options.application_identifiers:
         return None
 
+    assert runtime in _IDENTIFIERS_MAP
```
Jongy, line 323: What's the assert for

gprofiler/metadata/application_identifiers.py
```diff
@@ -289,25 +290,27 @@ def get_app_id(self, process: Process) -> Optional[str]:
...
+}
 
 
 def set_enrichment_options(enrichment_options: EnrichmentOptions) -> None:
     _ApplicationIdentifier.enrichment_options = enrichment_options
 
 
-def get_app_id(process: Process, identifiers: List[_ApplicationIdentifier]) -> Optional[str]:
+@functools.lru_cache(4096)  # TODO: arbitrary cache size
+def get_app_id(process: Process, runtime: Union[Literal["python"], Literal["java"]]) -> Optional[str]:
```
Jongy, line 313: This is gonna mess with https://github.com/Granulate/gprofiler/pull/393, but k, we'll solve that

Jongy (approved):

PR #46, opened 2021-04-28
Log the run mode (container / k8s container / ...)

Jongy:
## Description
Lets you know by the logs - how was gProfiler executed?

## Motivation and Context
Ease debugging.

## How Has This Been Tested?
Locally (container + standalone executable)

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] I have updated the relevant documentation.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #396, opened 2022-07-03
Upgrade py-spy v0.3.12, rbspy v0.12.1

Jongy:
I cleaned up the rust build environment a bit.

* Changed the base image to `rust:1.58-alpine3.15` (instead of `rust:xxx` which is based on Debian, we just use an Alpine image).
* Upgraded rust from 1.52 to 1.58. This was required for the new rbspy & py-spy. I tried to upgrade to newer rust (1.62) but the build fails (of either of the tools, I don't recall). We will need to upgrade them again if/when we upgrade rust.
* Removed duplicate logic for the libunwind builds - now using `libunwind_build.sh`. What I could otherwise do is get rid of `cfg(unwind)` because we don't currently use it.. but meh. It's nicer to keep the option, I guess :shrug:.
* The same PR upgrades both py-spy and rbspy, see [explaining comment](https://github.com/Granulate/gprofiler/pull/575#issuecomment-1352345345).

py-spy rebase diff - only the comment.
```
$ git range-diff v0.3.10..v0.3.10g1 v0.3.12..v0.3.12g1
1:  b23d457 ! 1:  10feb0f Don't gather thread activity for all threads when --nonblocking is provided
    @@ src/python_spy.rs: impl PythonSpy {
     -            thread_activity.insert(threadid, thread.active()?);
     -        }
      
    --        // Lock the process if appropiate. Note we have to lock AFTER getting the thread
    +-        // Lock the process if appropriate. Note we have to lock AFTER getting the thread
     -        // activity status from the OS (otherwise each thread would report being inactive always).
     -        // This has the potential for race conditions (in that the thread activity could change
     -        // between getting the status and locking the thread, but seems unavoidable right now
2:  f30e48f = 2:  8b041f9 Don't grab stack trace of non-GIL threads when --gil is provided
3:  480deec = 3:  b45bd56 Add a suffix to the stacks to easily identify Python stacks (#1)

```


Jongy:
```
#46 84.56 /usr/bin/ld: cannot find libgcc_s.so.1
```

Jongy:
> ```
> #46 84.56 /usr/bin/ld: cannot find libgcc_s.so.1
> ```

@michelhe from a quick look, any reason why this would happen once I upgrade the rust version?

michelhe:
@Jongy That didn't cut it as now `make install` is broken.
The difference is that the newer rust baseimage is based on a newer `buildpack-deps:buster` that uses gcc-10 and not gcc-8, and the libunwind build simply doesn't work properly on gcc-10

Jongy:
Upgrading libunwind might help?

michelhe:
> Upgrading libunwind might help?

No, tried from https://github.com/libunwind/libunwind/releases/download/v1.6.2/libunwind-1.6.2.tar.gz but same issue

michelhe:
@Jongy Have we re-iterated this one?

Jongy:
> @Jongy Have we re-iterated this one?

Not yet but I want to. Before the next big change in py-spy/pyperf I plan to do https://github.com/Granulate/gprofiler/pull/440#discussion_r945267173

Jongy:
```
#45 123.5 error: failed to run custom build command for `remoteprocess v0.4.9`
#45 123.5 
#45 123.5 Caused by:
#45 123.5   process didn't exit successfully: `/tmp/py-spy/target/release/build/remoteprocess-0485eadff44a6357/build-script-build` (exit status: 101)
#45 123.5   --- stdout
#45 123.5   cargo:rustc-cfg=use_libunwind
#45 123.5   cargo:rustc-link-search=native=/usr/local/lib
```

we don't really use libunwind in py-spy now. Not sure if rbspy requires it at the moment. If not, we can possibly disable this config altogether.

Jongy:
Now rbspy build fails - probably matter of Rust version compatibility. We'll need to upgrade it before we continue

Jongy:
Merged https://github.com/Granulate/gprofiler/pull/621 into it, for CI to pass This will be merged afterwards.

scripts/prepare_machine-unknown-linux-musl.sh
```diff
@@ -1,47 +1,39 @@
...
 
-apt-get update && apt-get install -y musl-dev musl-tools
+apk add --no-cache musl-dev make git curl  # git & curl used by next scripts
 
 mkdir builds && cd builds
 
-wget https://github.com/libunwind/libunwind/releases/download/v1.6.2/libunwind-1.6.2.tar.gz
-tar -xf libunwind-1.6.2.tar.gz
-pushd libunwind-1.6.2
-CC=musl-gcc ./configure --disable-minidebuginfo --enable-ptrace --disable-tests --disable-documentation
```
Jongy, line 27: Here we passed `--disable-minidebuginfo --enable-ptrace` but in the new build we no longer do that, is it necessary?

.github/workflows/build-container-and-deploy.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
Jongy, line 13: Merge #621 and get rid of this diff

scripts/rbspy_build.sh
```diff
@@ -1,11 +1,11 @@
...
 #
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
-set -euo pipefail
+set -eu
 
-git clone --depth 1 -b v0.8.1g2 https://github.com/Granulate/rbspy.git
-git -C rbspy reset --hard 2ce395058264d2d832d9a89589d52f26ccd55636
+git clone --depth 1 -b rbspy-upgrade https://github.com/Granulate/rbspy.git
```
Jongy, line 8: Tag rbspy instead of `rbspy-upgrade`

scripts/rbspy_build.sh
```diff
@@ -1,11 +1,11 @@
...
+git -C rbspy reset --hard 09fae41d424137af6b150efee3947f0f684a8128
```
Jongy, line 9: Comment the range-diff (or merge conflicts I had?) applying the update

Jongy: Diff between diffs, whatever

scripts/prepare_machine-unknown-linux-musl.sh
```diff
@@ -1,47 +1,39 @@
...
 #
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
-set -euo pipefail
+set -eu
 
 # prepares the environment for building py-spy:
 # 1. installs the rust target $(uname -m)-unknown-linux-musl - this can build static binaries
-# 2. downloads, builds & installs libunwind with musl
```
Jongy, line 10: Bring it back

scripts/prepare_machine-unknown-linux-musl.sh
```diff
@@ -1,47 +1,39 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 #
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
-set -euo pipefail
+set -eu
 
 # prepares the environment for building py-spy:
```
Jongy, line 8:
```suggestion
# prepares the environment for building py-spy & rbspy:
```

Jongy:
rbspy is broken:
```
E       stderr: b"error: Found argument '--on-cpu' which wasn't expected, or isn't valid in this context\n\n\tDid you mean '--on'?\n\n\tIf you tried to supply `--on-cpu` as a value rather than a flag, use `-- --on-cpu`\n\nUSAGE:\n    rbspy record --pid <PID> --silent --rate <RATE> --duration <DURATION> --nonblocking --on\n\nFor more information try --help\n"
```

Jongy:
Fix:
```
commit dd4518391de1cf41e0615174541c17542efc96b9 (HEAD -> rbspy-upgrade, origin/rbspy-upgrade)
Author: Yonatan Goldschmidt <yon.goldschmidt@gmail.com>
Date:   Thu Dec 15 02:37:40 2022 +0200

    --on-cpu -> --oncpu
    
    With the dash it's apparently parsed as 2 arg names (short and long): "-c, --on".
```

Jongy:
Need to disable this PHP lol

Jongy:
Moved the PHP tests flaky mark to https://github.com/Granulate/gprofiler/pull/632.

Jongy:
I verified that it builds on Aarch64.

Jongy:
py-spy range diff:
```
$ git range-diff origin/master..480deec8b5dde3cd331d1a793106981c1796d172 origin/master..b45bd56ee865835ed2738ed5c994a94590d307
....
 -:  ------- > 31:  fefb88c Fix typos discovered by codespell (#453)
 -:  ------- > 32:  385e7d2 Update tested python versions (#455)
 -:  ------- > 33:  9fd1949 Minor cleanup for python v3.10 (#456)
 -:  ------- > 34:  e1b117a Update dependencies (#457)
 -:  ------- > 35:  fee40b1 Warn about SYS_PTRACE when running in docker (#459)
 -:  ------- > 36:  f20219f Update tested python versions (#462)
 -:  ------- > 37:  5906cb7 Update to latest version of remoteprocess/proc-maps (#463)
 -:  ------- > 38:  693cba1 Update CHANGELOG for v0.3.11 release
 -:  ------- > 39:  0ee5daa Bump version: 0.3.10 → 0.3.11
 -:  ------- > 40:  73e12ef Update tested python versions (#464)
 -:  ------- > 41:  df209a9 Re-enable freebsd CI (#465)
 -:  ------- > 42:  4417b18 Update tested python versions (#467)
 -:  ------- > 43:  91e4eec Update clap & remoteprocess dependencies (#471)
 -:  ------- > 44:  1e971b9 Update README to note that we support python 3.10
 -:  ------- > 45:  39a9863 Update tested python versions (#474)
 -:  ------- > 46:  690008c Update remoteprocess to v0.4.9 (#487)
 -:  ------- > 47:  2a67ba7 Update tested python versions (#485)
 -:  ------- > 48:  232e3c1 Update clap dependency (#489)
 -:  ------- > 49:  ef380e5 Add a subprocesses option for the dump command (#492)
 -:  ------- > 50:  a408874 Add release-drafter workflow for generating release notes (#493)
 -:  ------- > 51:  6913531 Update release drafter workflow (#494)
 -:  ------- > 52:  1ca4e7f add workflow_dispatch trigger for release_drafter (#495)
 -:  ------- > 53:  a01ceeb Bump version: 0.3.11 → 0.3.12
31:  b23d457 ! 54:  10feb0f Don't gather thread activity for all threads when --nonblocking is provided
    @@ src/python_spy.rs: impl PythonSpy {
     -            thread_activity.insert(threadid, thread.active()?);
     -        }
      
    --        // Lock the process if appropiate. Note we have to lock AFTER getting the thread
    +-        // Lock the process if appropriate. Note we have to lock AFTER getting the thread
     -        // activity status from the OS (otherwise each thread would report being inactive always).
     -        // This has the potential for race conditions (in that the thread activity could change
     -        // between getting the status and locking the thread, but seems unavoidable right now
32:  f30e48f = 55:  8b041f9 Don't grab stack trace of non-GIL threads when --gil is provided
33:  480deec = 56:  b45bd56 Add a suffix to the stacks to easily identify Python stacks (#1) 
```

scripts/rbspy_build.sh
Jongy: Diff is new :shrug

scripts/prepare_machine-unknown-linux-musl.sh
Jongy: Well, this is here from #68 . If I didn't write in comment then it's not important enough.

michelhe (approved):

gprofiler/profilers/ruby.py
```diff
@@ -94,7 +94,7 @@ def _make_command(self, pid: int, output_path: str, duration: int) -> List[str]:
             "-d",
             str(duration),
             "--nonblocking",  # Don’t pause the ruby process when collecting stack samples.
-            "--on-cpu",  # only record when CPU is active
+            "--oncpu",  # only record when CPU is active
```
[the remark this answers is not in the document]
Jongy: https://github.com/Granulate/rbspy/commit/dd4518391de1cf41e0615174541c17542efc96b9

I think due to a change in the arguments parsing library in rbspy

PR #398, opened 2022-07-05
WIP: Feature/java short lived processes

glipner:


glipner:
@Jongy Please review latest changes, now mergable to master

gprofiler/gprofiler_types.py
```diff
@@ -56,3 +56,25 @@ def integer_range_check(value_str: str) -> int:
         return value
 
     return integer_range_check
+
+
+def sorted_positive_int_list(s: str) -> List[int]:
+    s = s.strip(" \t[]")
```
Jongy, line 62: What's the idea? To handle `[1,2,3]` and also separated with spaces/tabs? Please comment

Jongy: Let's remove thhe ` ` and `\t` from that - these are ignored when parsing a single int (acceptable) but will raise an error if the list was separated by spaces instead of commas for whatever reason:
```
In [3]: int("\t123 ")                                                                                                                                                                                              
Out[3]: 123

In [4]: int("\t123 4")                                                                                                                                                                                             
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-7cb2b6a777da> in <module>
----> 1 int("\t123 4")

ValueError: invalid literal for int() with base 10: '\t123 4'
```

gprofiler/gprofiler_types.py
```diff
@@ -56,3 +56,25 @@ def integer_range_check(value_str: str) -> int:
...
+            raise ValueError()
+
+        # Make sure the list is sorted in ascending order
+        for a, b in zip(parts, parts[1:]):
+            if a >= b:
+                raise ValueError()
+        return parts
+    except ValueError:
+        raise configargparse.ArgumentTypeError(
+            f"invalid argument {s}, " "expecting a sorted list of comma-separated integers"
```
Jongy, line 79:
```suggestion
            f"invalid argument {s},  expecting a sorted list of comma-separated integers"
```

gprofiler/gprofiler_types.py
```diff
@@ -56,3 +56,25 @@ def integer_range_check(value_str: str) -> int:
...
+def sorted_positive_int_list(s: str) -> List[int]:
+    s = s.strip(" \t[]")
+    if s == "":
+        return []
+    try:
+        parts = [int(x) for x in s.split(",")]
+
+        # Make sure the first element is positive. Next elements are required to be strictly ascending
+        if parts[0] <= 0:
+            raise ValueError()
```
Jongy, line 70: Give faulty value in message

gprofiler/gprofiler_types.py
```diff
@@ -56,3 +56,25 @@ def integer_range_check(value_str: str) -> int:
...
+
+        # Make sure the list is sorted in ascending order
+        for a, b in zip(parts, parts[1:]):
+            if a >= b:
+                raise ValueError()
```
Jongy, line 75: Give faulty values in message

gprofiler/profilers/java.py
```diff
@@ -274,6 +276,37 @@ def get_ap_version() -> str:
...
+        self._process = process
+        try:
+            self._create_time = process.create_time()
+        except psutil.NoSuchProcess:
+            self._create_time = None
+
+    def get_alive_time(self) -> Optional[float]:
+        if self._create_time is None:
+            return None
+        return time.time() - self._create_time
```
Jongy, line 293: Seems like psutil works with local time as well, any way to force it to work with absolute timestamps (and use `time.monotonic` as well, such that we're not affected by clock changes?)

gprofiler/profilers/java.py
```diff
@@ -966,11 +1011,22 @@ def _profile_process(self, process: Process, duration: int, spawned: bool) -> Pr
             self._jattach_timeout,
             self._ap_mcache,
         ) as ap_proc:
-            stackcollapse = self._profile_ap_process(ap_proc, comm, duration)
+            stackcollapse, process_uptime_str = self._profile_ap_process(ap_proc, comm, duration)
+
+        if process_uptime_str is not None:
+            if appid:
+                appid += ", "
+            appid = str(appid) + process_uptime_str
```
Jongy, line 1019: If `appid` was None then it's `str(None)` here, I guess in this case use `java: {process_uptime_str}`?

gprofiler/profilers/java.py
```diff
@@ -693,6 +726,16 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
             default=False,
             help="In case of Spark application executor process - add the name of the Spark application to the appid.",
         ),
+        ProfilerArgument(
+            "--java_collect_process_run_times",
```
Jongy, line 730:
```suggestion
            "--java-collect-process-run-times",
```

gprofiler/profilers/java.py
```diff
@@ -693,6 +726,16 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+            dest="java_collect_process_run_times",
```
Jongy, line 731: Don't need this

gprofiler/profilers/java.py
```diff
@@ -693,6 +726,16 @@ def parse_jvm_version(version_string: str) -> JvmVersion:
...
+            metavar="comma separated list of time bins, in seconds",
+            type=sorted_positive_int_list,
+            const=ProcessUptimeCalculator.DEFAULT_BINS,
+            nargs="?",
+            help="Split stack traces by total process run time, to account for transient phenomena "
```
Jongy, line 736: Making sure - this is disabled by default, right?

gprofiler/profilers/java.py
```diff
@@ -274,6 +276,37 @@ def get_ap_version() -> str:
...
+            return None
+
+        if not bins:  # None or an empty list -> use default
+            bins = self.DEFAULT_BINS
+        prev_bin_time = 0
+        for bin_time in bins:
+            if bin_time >= alive_time:
+                return f"run_time: {prev_bin_time}-{bin_time} seconds"
+            prev_bin_time = bin_time
+        return f"run_time: longer than {prev_bin_time} seconds"
```
Jongy, line 307:
```suggestion
        return f"run_time: {prev_bin_time}+ seconds"
```
Matter of taste but I think it'll be easier to read

gprofiler/profilers/java.py
```diff
@@ -274,6 +276,37 @@ def get_ap_version() -> str:
...
+class ProcessUptimeCalculator:
+    DEFAULT_BINS = [5, 10, 30, 60, 300]
+
+    def __init__(self, process: Process):
+        self._create_time: Optional[float]
+        self._process = process
+        try:
+            self._create_time = process.create_time()
+        except psutil.NoSuchProcess:
+            self._create_time = None
```
Jongy, line 288: Why is this value nullable? This makes the code a tad bit more complex. This object is constructed before we start profiling - if the process is dead by then, let the `NoSuchProcess` exception raise (just like we use `process_comm` etc and we don't catch its `NoSuchProcess`)

gprofiler/profilers/java.py
```diff
@@ -992,6 +1048,8 @@ def _profile_ap_process(self, ap_proc: AsyncProfiledProcess, comm: str, duration
             # Process still running. We will stop the profiler in finally block.
             pass
         else:
+            if self._collect_process_run_times is not None:
+                process_uptime_str = uptime_calculator.get_alive_time_bin(self._collect_process_run_times)
```
Jongy, line 1052: Don't we want `process_uptime_str` to be set also when the process doesn't exit? Such that we see what happened in its early lifetime

gprofiler/profilers/java.py
```diff
@@ -966,11 +1011,22 @@ def _profile_process(self, process: Process, duration: int, spawned: bool) -> Pr
             self._jattach_timeout,
             self._ap_mcache,
         ) as ap_proc:
-            stackcollapse = self._profile_ap_process(ap_proc, comm, duration)
+            stackcollapse, process_uptime_str = self._profile_ap_process(ap_proc, comm, duration)
```
Jongy, line 1014: Is it possible to do all "collect uptime" logic here, without changing `AsyncProfiledProcess` itself? (which tries to remain about async-profiler and not other stuff). Also will simplify the code if we manage to

Jongy (changes requested):
.

Jongy:
(Upgrade from master as well please, we've added new tests since)

Issue #383, opened 2022-06-18
.NET support


Jongy:
.NET is not supported at the moment. I don't know if .NET has any attach-profiling mechanism (i.e a program that gets a PID and extracts stacktraces from that PID, like py-spy or rbspy). We should look it up and if one exists it will be fairly easy to incorporate it into gProfiler.

Jongy:
Implemented in #412 with dotnet-trace.

PR #412, opened 2022-07-19
dotnet profiler addittion

pfilipko1:
## Checklist:


tests/test_sanity.py
```diff
@@ -94,6 +95,18 @@ def test_rbspy(
...
 
 
+@pytest.mark.parametrize("runtime", ["dotnet"])
+def test_dotnet_trace(
+    tmp_path: Path,
+    application_pid: int,
+    assert_collapsed: AssertInCollapsed,
+    gprofiler_docker_image: Image,
+) -> None:
+    with RbSpyProfiler(1000, 3, Event(), str(tmp_path), False, "dotnet-trace") as profiler:
```
Jongy, line 105: This needs to be `DotnetProfiler`

tests/containers/ruby/fibonacci.rb
```diff
@@ -4,11 +4,13 @@
 #
 
 def fibonacci( n )
+    puts "SIUR"
```
Jongy, line 7: Undo the changes here?

tests/containers/dotnet/dotnet.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
Jongy, line 1: Is there any reason to keep the autogenerated files here? If we can, keep only the actual source (i.e `Fibonacci.cs`) and avoid committing extra files that need to be CRed and maintained (instead of simply being autogenerated correctly during tests run)

tests/containers/dotnet/Fibonacci.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+namespace dotnet
+{
+    class Program
+    {
+        public static int Fibonacci(int n)
+        {
+            if(n==0 || n==1){
```
Jongy, line 8:
```suggestion
            if (n==0 || n==1) {

```

Dockerfile
```diff
@@ -145,18 +154,22 @@ RUN ./burn_build.sh
...
     apt-get install --no-install-recommends -y python3-pip && \
+    apt-get -y install sudo && \
+    apt-get install -qqq libicu66 && \
     if [ "$(uname -m)" = "aarch64" ]; then \
       apt-get install -y --no-install-recommends build-essential python3.8-dev; \
     fi && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+RUN useradd -m docker && echo "docker:docker" | chpasswd && adduser docker sudo
```
Jongy, line 172: Why is this required?

Dockerfile
```diff
@@ -112,6 +120,7 @@ RUN ./pyperf_build.sh
 
 # phpspy
 FROM ubuntu${PHPSPY_BUILDER_UBUNTU} AS phpspy-builder
+
```
Jongy, line 123:
```suggestion
```

Dockerfile
```diff
@@ -58,7 +67,6 @@ RUN ./perf_build.sh
 
 # pyperf (bcc)
 FROM ubuntu${PYPERF_BUILDER_UBUNTU} AS bcc-builder-base
-
```
Jongy, line 61: Bring it back please :)

Dockerfile
```diff
@@ -43,6 +45,13 @@ COPY scripts/rbspy_build.sh .
 RUN ./rbspy_build.sh
 RUN mv "/tmp/rbspy/target/$(uname -m)-unknown-linux-musl/release/rbspy" /tmp/rbspy/rbspy
 
+# dotnet-trace
+FROM mcr.microsoft.com/dotnet/sdk${DOTNET_BUILDER_UBUNTU} as dotnet-builder
+RUN apt-get update && \
+  dotnet tool install --global dotnet-trace
```
Jongy, line 51: Not all versions of packages/programs are pinned. We do pin the base Docker images, and most importantly we pin the profiler versions. Since we currently download dotnet-trace install of building it - make sure to download a pinned version from `dotnet tool install` if possible (to request a specific version and preferably to verify its hash against what you expect to receive here)

Dockerfile
```diff
@@ -43,6 +45,13 @@ COPY scripts/rbspy_build.sh .
...
+RUN dotnet --version
```
Jongy, line 52: Can be removed?

Dockerfile
```diff
@@ -8,6 +8,8 @@ ARG PYPERF_BUILDER_UBUNTU=@sha256:cf31af331f38d1d7158470e095b132acd126a7180a54f2
 ARG PERF_BUILDER_UBUNTU=@sha256:d7bb0589725587f2f67d0340edb81fd1fcba6c5f38166639cf2a252c939aa30c
 # phpspy - ubuntu:20.04
 ARG PHPSPY_BUILDER_UBUNTU=@sha256:cf31af331f38d1d7158470e095b132acd126a7180a54f263d386da88eb681d93
+# dotnet builder ubuntu:20.04
```
Jongy, line 11: It's not really `ubuntu:20.04` - the image is `mcr.microsoft.com/dotnet/sdk` and what tag did you use to get the hash from?

Dockerfile
```diff
@@ -181,6 +194,9 @@ COPY --from=async-profiler-builder-glibc /tmp/async-profiler/build/fdtransfer gp
 
 COPY --from=rbspy-builder /tmp/rbspy/rbspy gprofiler/resources/ruby/rbspy
 
+COPY --from=dotnet-builder /usr/share/dotnet /usr/share/dotnet
+COPY --from=dotnet-builder /tmp/.dotnet gprofiler/resources/dotnet
```
Jongy, line 198: What's actually required from here? Can you copy only the dotnet-trace binary itself? Does it require any libraries? We want to keep the image size small.

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+from psutil import Process
+from gprofiler.exceptions import ProcessStoppedException, StopEventSetException
+from gprofiler.gprofiler_types import ProfileData
+from gprofiler.metadata.application_metadata import ApplicationMetadata
+from gprofiler.log import get_logger_adapter
+from gprofiler.profilers.profiler_base import ProcessProfilerBase
+from gprofiler.profilers.registry import register_profiler
+from gprofiler.utils import pgrep_maps, random_prefix, removed_path, resource_path, run_process
+from gprofiler.utils.speedscope import load_speedscope_as_collapsed
+logger = get_logger_adapter(__name__)
```
Jongy, line 23: Blank before this line

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+class DotnetMetadata(ApplicationMetadata):
+
```
Jongy, line 27:
```suggestion
```

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+    _DOTNET_VERSION_TIMEOUT = 3
+
+    @functools.lru_cache(4096)
+    def _get_dotnet_version(self, process: Process) -> str:
+        if not os.path.basename(process_exe(process)).startswith("dotnet"):
+            # TODO: for dynamic executables, find the ruby binary that works with the loaded libruby, and
+            # check it instead. For static executables embedding libruby - :shrug:
```
Jongy, line 34: Comment is irrelevant here (it's for ruby).

I'm actually not sure if it's relevant for dotnet - i.e if a program can be named `abc` and have `libdotnet.so` loaded like sometimes happens for java/python.

You can remove the comment but keep the logic raising `NotImplementedError`, incase

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+            raise NotImplementedError
+        version = self.get_exe_version(process)  # not using cached version here since this wrapper is a cache
+        return version
+
+    def make_application_metadata(self, process: Process) -> Dict[str, Any]:
+        # ruby version
```
Jongy, line 40:
```suggestion
        # dotnet version
```

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+        metadata = {"dotnet_version": version}
+
+        metadata.update(super().make_application_metadata(process))
+        return metadata
+
+
+@register_profiler(
+    "dotnet",
+    possible_modes=["dotnet-trace", "disabled"],
+    supported_archs=["x86_64", "aarch64"],
```
Jongy, line 52: Doesn't support `aarch64` currently. No need to support it in MVP. Remove from this list, and please update the `Architecture support` list in readme - add dotnet and mark as x86 :heavy_check_mark: and aarch64 :x:

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+        socket_list = result.stdout.decode("utf-8").split('\n')
+        tempdir = f"TMPDIR=/proc/{process.pid}/root/tmp"
+        internal_pid = str(process.pid)
+        metadata = None
+        if int(socket_list[0].split("-")[2]) != 1:
+            # we are not in docker, pid different than 1
+            metadata = self._metadata.get_metadata(process)
+        else:
+            # we are in docker, pid is 1
+            internal_pid = "1"
```
Jongy, line 93: A few notes on the logic here.
1. In a container - the inner PID is not necessarily 1. The container may start with a shell script that then runs `dotnet`, and thus the shell script is PID 1 and dotnet is PID e.g 9.
2. To get the PID inside a container, you can use `get_process_nspid`. This gets you the PID of the process as the process sees it. So for PID 999 that runs not in a container, you get 999. For PID 1001 which runs in a container as PID 5, you get 5.
3. Even when in a container - we are able to get the version - `get_exe_version` works also if the target is containerized (works for Python, Java, Ruby so should also work for dotnet)

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+        return [
+            "sudo",
```
Jongy, line 95: No need for sudo - we run as root

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+            tempdir,
+            resource_path(self.RESOURCE_PATH),
+            "collect",
+            "--format",
+            "speedscope",
+            "--process-id",
+            internal_pid,
```
Jongy, line 102: This needs to be the host PID, not the PID in container, no?

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def _profile_process(self, process: Process, duration: int, spawned: bool) -> ProfileData:
+        logger.info(
+            f"Profiling{' spawned' if spawned else ''} process {process.pid} with dotnet-trace",
+            cmdline=" ".join(process.cmdline()),
+            no_extra_to_server=True,
+        )
+        appid = None
+        app_metadata = None
+        local_output_path = os.path.join(self._storage_dir, f"dotnet-trace.{random_prefix()}.{process.pid}.col")
```
Jongy, line 116: Why assign it if you override later? Is there no way to ask dotnet-trace to write to this path?

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+                    kill_signal=signal.SIGKILL,
+                )
+                local_output_path = result.stdout.decode("utf-8").split("\t")[1].split("\n")[0]
+            except ProcessStoppedException:
+                raise StopEventSetException
+            logger.info(f"Finished profiling process {process.pid} with dotnet")
+            return ProfileData(load_speedscope_as_collapsed(Path(local_output_path), self._frequency),
+                               appid, app_metadata)
+
+    def _parse_duration(self, duration: int):
```
Jongy, line 133: This translates the duration in seconds to HH:MM:ss timestamp, right?

It can be replaced with `str(datetime.timedelta(duration))`, just ensure `duration < 3600 * 24` because then its format changes (more than 1 day). That's not an acceptable usage of gProfiler anyway.

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+            tempdir,
+            resource_path(self.RESOURCE_PATH),
+            "collect",
+            "--format",
+            "speedscope",
+            "--process-id",
+            internal_pid,
+            "--duration",
+            self._parse_duration(duration),
+            #           str(self._frequency), - TODO: frequency handling to be determined
```
Jongy, line 105: We need to discuss the frequency indeed. I don't understand what frequency dotnet-trace currently uses for its sampling

gprofiler/profilers/dotnet.py
Jongy: (this refers to the logic from line 76 to 93)

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+        result = run_process(
+            f"ls /proc/{process.pid}/root/tmp | grep diagnostic ",
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            shell=True,
+            suppress_log=True,
+            check=False,
+        )
+        socket_list = result.stdout.decode("utf-8").split('\n')
+        tempdir = f"TMPDIR=/proc/{process.pid}/root/tmp"
```
Jongy, line 85: Document this trick :)

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,157 @@
...
+        metadata = None
+        if int(socket_list[0].split("-")[2]) != 1:
+            # we are not in docker, pid different than 1
+            metadata = self._metadata.get_metadata(process)
+        else:
+            # we are in docker, pid is 1
+            internal_pid = "1"
+        return [
+            "sudo",
+            tempdir,
```
Jongy, line 96: This doesn't work because we don't use `shell=True` here, so it's not passed as an environment variable.

Instead - pass `{"TMPDIR": ...}` as `env=` parameter of `run_process`.

Jongy (changes requested):
Done reviewing first round :)

Dockerfile
pfilipko1: `docker inspect --format='{{index .RepoDigests 0}}' mcr.microsoft.com/dotnet/sdk:6.0-focal
mcr.microsoft.com/dotnet/sdk@sha256:749439ff7a431ab4bc38d43cea453dff9ae1ed89a707c318b5082f9b2b25fa22`

Jongy: Cool - can you please modify the comment with that tag? So we know how to update it in the future.

Jongy:
Closes: https://github.com/Granulate/gprofiler/issues/383

Dockerfile
pfilipko1: I'm not sure, but dotnet-trace version may be linked to the dotnet version - in this case 6.0. I purposedly have different versions of dotnet on my machine and in other dotnet containers that I test it on. It seems to work properly on all of them.

Dockerfile
```diff
@@ -147,10 +157,12 @@ FROM ubuntu${GPROFILER_BUILDER_UBUNTU}
 WORKDIR /app
 
 # for Aarch64 - it has no .whl file for psutil - so it's trying to build from source.
+
 RUN set -e; \
     apt-get update && \
     apt-get upgrade -y && \
     apt-get install --no-install-recommends -y python3-pip && \
+    apt-get install -y libicu66 && \
```
Jongy, line 165: Why is this one required? for dotnet-trace?

Dockerfile
```diff
@@ -181,6 +193,10 @@ COPY --from=async-profiler-builder-glibc /tmp/async-profiler/build/fdtransfer gp
 
 COPY --from=rbspy-builder /tmp/rbspy/rbspy gprofiler/resources/ruby/rbspy
 
+COPY --from=dotnet-builder /usr/share/dotnet/host /usr/share/dotnet/host
```
Jongy, line 196: What are we copying from these 2 directories? i.e `gprofiler/resources/dotnet` is the `dotnet-trace` app itself, and what are these 2 directories?

Jongy:
@pfilipko1 I ran the CI

Jongy:
@pfilipko1 you can run the linters locally with `./lint.sh` (personally I added it as a git commit hook, with `./enable_hook.sh`)

Dockerfile
```diff
@@ -8,6 +8,8 @@ ARG PYPERF_BUILDER_UBUNTU=@sha256:cf31af331f38d1d7158470e095b132acd126a7180a54f2
 ARG PERF_BUILDER_UBUNTU=@sha256:d7bb0589725587f2f67d0340edb81fd1fcba6c5f38166639cf2a252c939aa30c
 # phpspy - ubuntu:20.04
 ARG PHPSPY_BUILDER_UBUNTU=@sha256:cf31af331f38d1d7158470e095b132acd126a7180a54f263d386da88eb681d93
+# dotnet builder - mcr.microsoft.com/dotnet/sdk:6.0-focal
```
Jongy, line 11: Please make the necessary changes in `pyi.Dockerfile` - that is the Dockerfile used to build the gProfiler executable (this Dockerfile is used to build the image - we build both and distribute both).

All resources under `gprofiler/resources/dotnet` are packed into the executable (and their dependency libraries) so it should be pretty straight forward and similar to what you've added here.

Jongy:
@pfilipko1 remaining issues:
* CR notes
* Add dotnet-trace in `pyi.Dockerfile`
* Ensure lint passes & tests pass - you can try running locally. (can follow what the CI does, basically it's `./scripts/build_x86_64_container.sh -t gprofiler; ./scripts/copy_resources_from_image.sh gprofiler; ./tests/test.sh --ignore=tests/test_executable.py` to run the standard tests.
* Decide on the frequency issue (or the lack of it) - I think it's okay for MVP to have a constant frequency without the ability to change it. @pfilipko1 can you please check if dotnet-trace has any similar concept to samping frequency that is configurable?

Jongy:
@pfilipko1 you can also merge from master as we have merged the speedscope parser

Dockerfile
pfilipko1: Yes, it is required by dotnet-trace. Reference:
https://stackoverflow.com/questions/59119904/process-terminated-couldnt-find-a-valid-icu-package-installed-on-the-system-in
setting the ENV variable was not enough

Dockerfile
pfilipko1: These directories contain libraries needed to run the dotnet-trace tool. I determined them by brute force - removing the directories from the /usr/share/dotnet and checking whether it would still work. It reduced the size by more than 2/3 and I hope that's ok.

Dockerfile
Jongy: I see. For `pyi.Dockerfile`, we build the binary in the image but the image itself isn't used, only the binary. We use `staticx` to pack all necessary libraries into the binary. Need to ensure that the result binary actually gets the library & is able to profile dotnet applications.

Dockerfile
Jongy: Which libraries precisely are required? Can it be determined with `ldd`? (i.e `ldd dotnet-trace`?)

Dockerfile
Jongy: (if `gprofiler` binary is able to profiler dotnet apps when started in an environment where `libicu66` is not isntalled - then that settles it)

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,127 @@
...
+@register_profiler(
+    "dotnet",
+    possible_modes=["dotnet-trace", "disabled"],
+    supported_archs=["x86_64"],
+    default_mode="dotnet-trace",
+)
+class DotnetProfiler(ProcessProfilerBase):
+
+    RESOURCE_PATH = "dotnet/tools/dotnet-trace"
+    MAX_FREQUENCY = 100
```
Jongy, line 60: Unused?

Jongy: can remove

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,127 @@
...
+        logger.info(
+            f"Profiling{' spawned' if spawned else ''} process {process.pid} with dotnet-trace",
+            cmdline=" ".join(process.cmdline()),
+            no_extra_to_server=True,
+        )
+        appid = None
+        app_metadata = self._metadata.get_metadata(process)
+        # had to change the dots for minuses because of dotnet-trace removing the last part in other case
+        local_output_path = os.path.join(self._storage_dir, f"dotnet-trace-{random_prefix()}-{process.pid}")
+        tempdir = f"/proc/{process.pid}/root/tmp"
```
Jongy, line 107: Document this part? (this causes dotnet-trace to lookup the socket in the mount namespace of the target process)

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,127 @@
...
+            try:
+                run_process(
+                    self._make_command(process, duration, local_output_path),
+                    env={"TMPDIR": tempdir},
+                    stop_event=self._stop_event,
+                    timeout=self._duration + self._EXTRA_TIMEOUT,
+                    kill_signal=signal.SIGKILL,
+                )
+                local_output_path = local_output_path + ".speedscope.json"
+
```
Jongy, line 118:
```suggestion
```

tests/containers/dotnet/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM  mcr.microsoft.com/dotnet/sdk@sha256:749439ff7a431ab4bc38d43cea453dff9ae1ed89a707c318b5082f9b2b25fa22
+
+WORKDIR /app
+ 
+ADD Fibonacci.dll /app
+ADD Fibonacci.deps.json /app
+ADD Fibonacci.runtimeconfig.json /app
+# ENV DOTNET_CLI_HOME = "/tmp"
```
Jongy, line 8: Is it needed? If not, please remove

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,127 @@
...
+            except ProcessStoppedException:
+                raise StopEventSetException
+            logger.info(f"Finished profiling process {process.pid} with dotnet")
+            return ProfileData(
+                load_speedscope_as_collapsed(Path(local_output_path), self._frequency), appid, app_metadata
```
Jongy, line 123: Other profilers use `merge.parse_one_collapsed` which optionally adds the `comm` of the profiled process. I'd like the dotnet profiler to do so as well, otherwise as the first frame I get this
image

which isn't coherent with other top-level frames that are process `comm`s.

Let's add the same logic to `load_speedscope_as_collapsed` - pass `add_comm: str` which is prepended as first frame for each stack.

gprofiler/profilers/dotnet.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+@register_profiler(
+    "dotnet",
+    possible_modes=["dotnet-trace", "disabled"],
+    supported_archs=["x86_64"],
+    default_mode="dotnet-trace",
+)
+class DotnetProfiler(ProcessProfilerBase):
+
```
Jongy, line 58:
```suggestion
```

Jongy (changes requested):
How is `Fibonacci.dll` rebuilt? Ideally we won't commit binary files, but the source files used.

Jongy:
I started the CI

Dockerfile
pfilipko1: ldd output did not really help:
`        linux-vdso.so.1 (0x00007ffdc2cb4000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f6d79aa0000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f6d7989c000)
        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f6d79513000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f6d79175000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f6d78f5d000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f6d78b6c000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f6d79cbf000)`
        
I determined dotnet-trace libraries manually by moving each one and checking if it still works. Here they are:

`3.3M    dotnet/shared/Microsoft.NETCore.App/6.0.8/libclrjit.so 
7.1M    dotnet/shared/Microsoft.NETCore.App/6.0.8/libcoreclr.so
376K    dotnet/shared/Microsoft.NETCore.App/6.0.8/libhostpolicy.so
92K     dotnet/shared/Microsoft.NETCore.App/6.0.8/libSystem.Native.so
812K    dotnet/shared/Microsoft.NETCore.App/6.0.8/Microsoft.CSharp.dll
60K     dotnet/shared/Microsoft.NETCore.App/6.0.8/Microsoft.NETCore.App.deps.json
16K     dotnet/shared/Microsoft.NETCore.App/6.0.8/Microsoft.Win32.Primitives.dll
48K     dotnet/shared/Microsoft.NETCore.App/6.0.8/mscorlib.dll
92K     dotnet/shared/Microsoft.NETCore.App/6.0.8/netstandard.dll
224K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Collections.Concurrent.dll
256K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Collections.dll
8.0K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.ComponentModel.dll
64K     dotnet/shared/Microsoft.NETCore.App/6.0.8/System.ComponentModel.Primitives.dll
176K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Console.dll
16K     dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Core.dll
2.9M    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Data.Common.dll
8.0K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Data.DataSetExtensions.dll
16K     dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Data.dll
252K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Diagnostics.Process.dll
40K     dotnet/shared/Microsoft.NETCore.App/6.0.8/System.dll
16K     dotnet/shared/Microsoft.NETCore.App/6.0.8/System.IO.FileSystem.dll
520K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Linq.dll
3.8M    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Linq.Expressions.dll
164K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Memory.dll
10M     dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Private.CoreLib.dll
32K     dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Runtime.dll
516K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Text.RegularExpressions.dll
116K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Threading.Channels.dll
72K     dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Threading.dll
8.0K    dotnet/shared/Microsoft.NETCore.App/6.0.8/System.Threading.Overlapped.dll`

It is 31M in total, and 37M can be removed. Is it ok? Hopefully there's a way to copy all these dependencies as a list determined elsewhere, otherwise the Dockerfile would look atrocius

Jongy (changes requested):

Dockerfile
Jongy: Yes, that's fine, let's copy just them like you did :)

tests/conftest.py
```diff
@@ -94,7 +111,12 @@ def command_line(runtime: str, java_command_line: List[str]) -> List[str]:
         "python": ["python3", str(CONTAINERS_DIRECTORY / "python/lister.py")],
         "php": ["php", str(CONTAINERS_DIRECTORY / "php/fibonacci.php")],
         "ruby": ["ruby", str(CONTAINERS_DIRECTORY / "ruby/fibonacci.rb")],
-        "dotnet": ["dotnet", str(CONTAINERS_DIRECTORY / "dotnet/Fibonacci.dll")],
+        "dotnet": dotnet_command_line,
+        # "dotnet": [
```
Jongy, line 115: Can be removed

tests/conftest.py
```diff
@@ -81,7 +81,24 @@ def java_command_line(tmp_path: Path, java_args: List[str]) -> List[str]:
 
 
 @fixture
-def command_line(runtime: str, java_command_line: List[str]) -> List[str]:
+def dotnet_command_line(tmp_path: Path) -> List[str]:
+    class_path = tmp_path / "dotnet"
+    class_path.mkdir()
+    class_path = class_path / "Fibonacci"
+    class_path.mkdir()
```
Jongy, line 88:
```suggestion
    class_path = tmp_path / "dotnet" / "Fibonacci"
    class_path.mkdir(parents=True)
```

tests/conftest.py
```diff
@@ -81,7 +81,24 @@ def java_command_line(tmp_path: Path, java_args: List[str]) -> List[str]:
...
+    chmod_path_parts(class_path, stat.S_IRGRP | stat.S_IROTH | stat.S_IXGRP | stat.S_IXOTH)
+    subprocess.run(["cp", str(CONTAINERS_DIRECTORY / "dotnet/Fibonacci.cs"), class_path])
+    subprocess.run(["dotnet", "new", "console", "--force"], cwd=class_path)
+    subprocess.run(["rm", "Program.cs"], cwd=class_path)
+    subprocess.run(["dotnet", "build", "-c", "Release"], cwd=class_path)
+    # dotnet building server (Roslyn) is waiting for its parent process to comple, which is pytest, it must be killed
+    # not to interfere with the actual test
+    subprocess.run(["pkill","-f","Roslyn"])
```
Jongy, line 96: Preferably we will avoid `pkill -f` on the user machine - if they have Roslyn running, for example. Or anothing that matches it.
Why does it interfere with the test? Because Roslyn is also also profiled by dotnet-trace? Does it affect the test results?

Jongy:
@pfilipko1 I ran gProfiler exe from the last commit (bdba8e06a49ccd8035995473309b658ad2be7fa0) and I get this:
```
[2022-09-05 18:48:28,249] ERROR: gprofiler.profilers.profiler_base: DotnetProfiler: failed to profile process 1032935 (dotnet)
  File "gprofiler/profilers/profiler_base.py", line 119, in _wait_for_profiles
    result = future.result()
  File "concurrent/futures/_base.py", line 425, in result
  File "concurrent/futures/_base.py", line 384, in __get_result
  File "concurrent/futures/thread.py", line 56, in run
  File "gprofiler/profilers/dotnet.py", line 114, in _profile_process
    kill_signal=signal.SIGKILL,
  File "gprofiler/utils/__init__.py", line 265, in run_process
    raise CalledProcessError(retcode, process.args, output=stdout, stderr=stderr)
gprofiler.exceptions.CalledProcessError: Command '['/tmp/_MEIAyTTB6/gprofiler/resources/dotnet/tools/dotnet-trace', 'collect', '--format', 'speedscope', '--process-id', '1', '--profile', 'cpu-sampling', '--duration', '0:00:05', '--output', '/tmp/gprofiler_tmp/tmp1o0qgnex/dotnet-trace-UYwYIgaCfZnCUZNL-1032935']' returned non-zero exit status 131. 
stdout: b''
stderr: b'You must install .NET to run this application.\n\nApp: /tmp/_MEIAyTTB6/gprofiler/resources/dotnet/tools/dotnet-trace\nArchitecture: x64\nApp host version: 6.0.7\n.NET location: Not found\n\nLearn about runtime installation:\nhttps://aka.ms/dotnet/app-launch-failed\n\nDownload the .NET runtime:\nhttps://aka.ms/dotnet-core-applaunch?missing_runtime=true&arch=x64&rid=ubuntu.22.04-x64&apphost_version=6.0.7\n'
```

tests/conftest.py
pfilipko1: By design, these project generators or code compilators seem to wait for the parent process to end (in our case - pytest), and then the test requires to have one matching dotnet process, so  AssertionError occurs. If test reqiures no other dotnet processes to pass, I thought that kind of solution will be viable. I'll try to think of something else.

Jongy:
You can add ".NET" to https://github.com/Granulate/gprofiler#architecture-support. (only for x86_64). And mark as experimental

README.md
```diff
@@ -359,15 +359,16 @@ The runtime stacks are then merged into the data collected by `perf`, substituti
...
+| Runtime                    | x86_64                            | Aarch64                           |
+|----------------------------|-----------------------------------|-----------------------------------|
+| perf (native, Golang, ...) | :heavy_check_mark:                | :heavy_check_mark:                |
+| Java (async-profiler)      | :heavy_check_mark:                | :heavy_check_mark:                |
+| Python (py-spy)            | :heavy_check_mark:                | :heavy_check_mark:                |
+| Python (PyPerf eBPF)       | :heavy_check_mark:                | :x:                               |
+| Ruby (rbspy)               | :heavy_check_mark:                | :heavy_check_mark:                |
+| PHP (phpspy)               | :heavy_check_mark:                | :heavy_check_mark: (experimental) |
+| NodeJS (perf)              | :heavy_check_mark:                | :heavy_check_mark:                |
+| .NET (dotnet-trace)S       | :heavy_check_mark: (experimental) | :x:                               |
```
Jongy, line 371:
```suggestion
| .NET (dotnet-trace)       | :heavy_check_mark: (experimental) | :x:                               |               
```

Jongy (approved):
Looks good. I will run some tests on it and then merge.

PR #420, opened 2022-07-23
Add perf tests

Jongy:
These sanity test FP, DWARF and smart modes on 2 applications: one built with FP only, and one built without
FP but with DWARF.

### DIFF
--- a/tests/test_perf.py
+++ b/tests/test_perf.py
@@ -1,0 +1,51 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+
+from pathlib import Path
+from threading import Event
+
+import pytest
+
+from gprofiler.profilers.perf import DEFAULT_PERF_DWARF_STACK_SIZE, SystemProfiler
+from tests.utils import assert_function_in_collapsed, is_function_in_collapsed, snapshot_pid_collapsed
+
+
+@pytest.mark.parametrize("runtime", ["native_fp", "native_dwarf"])
+@pytest.mark.parametrize("perf_mode", ["fp", "dwarf", "smart"])
+@pytest.mark.parametrize("in_container", [True])  # tests depend on the version we use in the image
+def test_perf(
+    tmp_path: Path,
+    application_pid: int,
+    runtime: str,
+    perf_mode: str,
+) -> None:
+    """ """
+    with SystemProfiler(
+        99,
+        3,
+        Event(),
+        str(tmp_path),
+        False,
+        perf_mode=perf_mode,
+        perf_inject=False,
+        perf_dwarf_stack_size=DEFAULT_PERF_DWARF_STACK_SIZE,
+    ) as profiler:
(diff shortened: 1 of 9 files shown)


Jongy:
The failure is flakiness:
```
>           assert is_function_in_collapsed(";_start;__libc_start_main;main;", process_collapsed) ^ bool(
                perf_mode not in ("dwarf", "smart")
            )
E           AssertionError: assert (False ^ False)
E            +  where False = is_function_in_collapsed(';_start;__libc_start_main;main;', Counter({'native;[unknown];__libc_start_main;main;recursive;recursive;recursive;recursive;recursive;recursive;recursiv...ist_rcv_[k];ip_sublist_rcv_[k];ip_rcv_finish_core.isra.0_[k];tcp_v4_early_demux_[k];__inet_lookup_established_[k]': 1}))
E            +  and   False = bool('smart' not in ('dwarf', 'smart'))
```

the thing is, kernel stacks are taken into account when calculating the average, which introduces flakiness because kernel stacks are not expected.
I changed the code so kernel stacks are not taken into account

Jongy:
Tests passed first time, running again to ensure no flakiness.

Jongy:
2nd time all passed, running again.

Jongy:
Screenshot from 2022-07-24 10-39-05

tests/test_perf.py
```diff
@@ -0,0 +1,51 @@
...
+        else:
+            # app is built with FP and without DWARF info, but DWARF mode is able to do FP unwinding,
+            # so it should always succeed.
+            assert runtime == "native_fp"
+            assert_function_in_collapsed(";recursive;recursive;recursive;recursive;", process_collapsed)
+
+        # expect to see libc stacks only when collecting with DWARF or smart.
+        assert is_function_in_collapsed(";_start;__libc_start_main;main;", process_collapsed) ^ bool(
+            perf_mode not in ("dwarf", "smart")
+        )
```
michelhe, line 51: Confusing test matrix lol, I would split it, if you resort to using `xor` that means the other check should probably be in another test

michelhe (changes requested):
.

tests/test_perf.py
Jongy: I figured the comment helps.

Indeed it can be another test but I don't want to introduce additional slowdown. It requires the exact matrix of `(native_dwarf/native_fp, fp/dwarf/smart)` so it's a waste to re-run all.

I can elaborate on the comment?

michelhe: Comment helps but it doesn't change the fact that this test matrix is complicated.
Meh

michelhe (approved):

tests/test_perf.py
Jongy: Complicated or complex. Because it is complex, but that's a fact because it's testing something complex.

Any way you see to simplify it besides splitting this part (which I won't do due to aforementioned reason)? More comments?

michelhe: No, aside from splitting it new comments will not elaborate further than its already is.

Jongy: K, merging

PR #422, opened 2022-07-26
Add PoC speedscope -> collapsed converter

Jongy:
PoC for https://github.com/Granulate/gprofiler/pull/412

### DIFF
--- a/gprofiler/utils/speedscope.py
+++ b/gprofiler/utils/speedscope.py
@@ -4,7 +4,8 @@
 #
 
 # speedscope -> collapsed converter, aimed to work for dotnet-trace.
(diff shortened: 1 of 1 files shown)


Jongy:
Should pass now. @michelhe let's merge this - it's just a utils file that is unused in this PR, I have written it for https://github.com/Granulate/gprofiler/pull/412

michelhe (approved):
Generally I would prefer to only merge such things along with a unittest, but as long as nothing uses it, whatever.

The code is slightly unreadable due to the deep nested for-loops, I consider we should split  it up but up to you

Jongy:
We'll add tests once this is actually used

Jongy:
Not doing any refactor now

Issue #417, opened 2022-07-20
k8s Daemonset pods stuck in "CrashLoopBackOff"


brijean:
just installed gprofiler daemonset on eks k8s cluster.  Pods keep restarting:

[cloudshell-user@ip-10-0-20-221 ~]$ kubectl get pods | grep granulate
granulate-gprofiler-t7dwh                0/1     CrashLoopBackOff   19 (75s ago)   74m
granulate-gprofiler-t855m                0/1     CrashLoopBackOff   19 (96s ago)   74m
granulate-gprofiler-w4x96                0/1     CrashLoopBackOff   19 (82s ago)   74m
granulate-maestro-6df6db74c9-t26bj       2/2     Running            0              73m

[cloudshell-user@ip-10-0-20-221 ~]$ kubectl describe pods granulate-gprofiler-t7dwh
Name:         granulate-gprofiler-t7dwh
Namespace:    default
Priority:     0
Node:         ip-192-168-17-133.us-east-2.compute.internal/192.168.17.133
Start Time:   Wed, 20 Jul 2022 19:02:24 +0000
Labels:       app=granulate-gprofiler
              controller-revision-hash=57dbd966bd
              pod-template-generation=1
Annotations:  kubernetes.io/psp: eks.privileged
Status:       Running
IP:           192.168.34.230
IPs:
  IP:           192.168.34.230
Controlled By:  DaemonSet/granulate-gprofiler
Containers:
  granulate-gprofiler:
    Container ID:  docker://90d9dcab2ac6423596e6553fb12b9c651f72e65e5c7bf31bc2b8ef39dbc30d1d
    Image:         index.docker.io/granulate/gprofiler:latest
    Image ID:      docker-pullable://granulate/gprofiler@sha256:ab80eda157f96962e0dcbfd4e6f878a185802f5346b5c93c447f085de59203b8
    Port:          <none>
    Host Port:     <none>
    Args:
      -cu
      --token
      $(GPROFILER_TOKEN)
      --service-name
      $(GPROFILER_SERVICE)
    State:          Waiting
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       Error
      Exit Code:    2
      Started:      Wed, 20 Jul 2022 20:45:46 +0000
      Finished:     Wed, 20 Jul 2022 20:45:47 +0000
    Ready:          False
    Restart Count:  25
    Limits:
      cpu:     500m
      memory:  1Gi
    Requests:
      cpu:     100m
[...]

PR #424, opened 2022-08-03
Pass --token, --service-name with = everywhere

Jongy:
PR #325 did it partially. Finish the job here.

Fixes the issue from https://github.com/Granulate/gprofiler/issues/417.

I manually tested the new k8s DaemonSet and it works fine.

The pattern `--token[^=]` doesn't have any other matches :)


michelhe (approved):
lgtm

PR #425, opened 2022-08-03
java: Read and log AP log also upon successful operation

Jongy:
For extra verbosity, we'd like to collect AP log at all times.

### DIFF
--- a/gprofiler/metadata/system_metadata.py
+++ b/gprofiler/metadata/system_metadata.py
@@ -13,7 +13,7 @@ from dataclasses import dataclass
 from functools import lru_cache
 from typing import Any, Dict, Optional, Tuple, cast
 
-import distro
+import distro  # type: ignore
 import psutil
 from granulate_utils.linux.ns import run_in_ns
 
--- a/gprofiler/metadata/system_metadata.py
+++ b/gprofiler/metadata/system_metadata.py
@@ -13,7 +13,7 @@ from dataclasses import dataclass
 from functools import lru_cache
 from typing import Any, Dict, Optional, Tuple, cast
 
-import distro  # type: ignore
+import distro
 import psutil
 from granulate_utils.linux.ns import run_in_ns
 
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -473,21 +473,24 @@ class AsyncProfiledProcess:
             + self._get_extra_ap_args()
         ]
 
+    def _read_ap_log(self) -> str:
+        if not os.path.exists(self._log_path_host):
+            return "(log file doesn't exist)"
+
+        log = Path(self._log_path_host)
+        ap_log = log.read_text()
+        # clean immediately so we don't mix log messages from multiple invocations.
+        # this is also what AP's profiler.sh does.
+        log.unlink()
+        self._recreate_log()
+        return ap_log
+
     def _run_async_profiler(self, cmd: List[str]) -> None:
         try:
             # kill jattach with SIGTERM if it hangs. it will go down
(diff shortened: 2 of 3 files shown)


gprofiler/metadata/system_metadata.py
```diff
@@ -13,7 +13,7 @@
 from functools import lru_cache
 from typing import Any, Dict, Optional, Tuple, cast
 
-import distro  # type: ignore
+import distro
```
Jongy, line 16: Mypy told me I can kick it, let's see

tests/test_java.py
```diff
@@ -536,3 +536,28 @@ def test_java_attach_socket_missing(
...
+    with make_java_profiler(
+        storage_dir=str(tmp_path),
+        duration=1,
+    ) as profiler:
+        snapshot_one_profile(profiler)
+
+        log_records = list(filter(lambda r: r.message == "async-profiler log", caplog.records))
+        assert len(log_records) == 2  # start,stop
+        # start
+        assert log_records[0].gprofiler_adapter_extra["ap_log"] == ""  # type: ignore
```
Jongy, line 561: Both of them print empty log, I wonder if that's enough for the tests or if I should find something else that triggers real logs written so that I could test it

gprofiler/profilers/java.py
```diff
@@ -473,21 +473,24 @@ def _get_stop_cmd(self, with_output: bool) -> List[str]:
             + self._get_extra_ap_args()
         ]
 
+    def _read_ap_log(self) -> str:
+        if not os.path.exists(self._log_path_host):
```
michelhe, line 477: what if the log is created inside a container?

tests/test_java.py
Jongy: Improved

gprofiler/profilers/java.py
Jongy: `_log_path_host` includes the `/proc/pid/root`

gprofiler/profilers/java.py
```diff
@@ -498,6 +501,8 @@ def _run_async_profiler(self, cmd: List[str]) -> None:
                 raise JattachSocketMissingException(*args) from None
             else:
                 raise JattachException(*args) from None
+        else:
+            logger.debug("async-profiler log", jattach_cmd=cmd, ap_log=self._read_ap_log())
```
michelhe, line 505: Why we only capture the AP log after `success`? 
And also, do we need the entire log each time? won't it be better to log the diff lines from before and after the profiling session? and filter the log lines by PID ?

Jongy: What do you mean "after `success`"?

The log is read in the `else` clause upon success, or in the `except` caluse before we raise an exception.

> And also, do we need the entire log each time? won't it be better to log the diff lines from before and after the profiling session? and filter the log lines by PID ?

The log is per-pid and is deleted after every async-profiler operation (so each log we read is the log for this single run)

michelhe (approved):

gprofiler/profilers/java.py
```diff
@@ -473,21 +473,24 @@ def _get_stop_cmd(self, with_output: bool) -> List[str]:
...
 
+    def _read_ap_log(self) -> str:
+        if not os.path.exists(self._log_path_host):
+            return "(log file doesn't exist)"
+
+        log = Path(self._log_path_host)
+        ap_log = log.read_text()
+        # clean immediately so we don't mix log messages from multiple invocations.
+        # this is also what AP's profiler.sh does.
+        log.unlink()
```
michelhe, line 484: small race here for pid recycling, I think, meh

Jongy: Too small

PR #411, opened 2022-07-18
Upgrade perf to base on Linux 5.18

Jongy:

PR #426, opened 2022-08-04
perf: Parse new event family names

Jongy:
This field used to be "cpu-clock:", with the new perf (and on a recent kernel) I get "cpu_core/cycles/",
so add _ and / to the pattern.

It's probably for the better to remove this complex regex and remain only with parsing the stack itself. I didn't want to make a larger change at this point, given that this already works, and that we don't have good tests coverage here.


Jongy:
Fixes #411

michelhe (approved):

Issue #418, opened 2022-07-21
How to improve the profiling of NodeJS applications


hrvacadaIM:
Hi.

First of all thank you for this nice tool. We have used the https://github.com/Granulate/gprofiler/blob/master/deploy/ecs/gprofiler_task_definition.json file to create task on our ECS clusters. We are getting some metrics in GProfiler Studio, but we would like to have also better insight into our own code and not just underlying code of NodeJS. Is there a possibility to get better insight into that part with passing some arguments? For example we found this option, but I am not sure, if this is needed at all or it can be passed to ECS gprofiler task:

> NodeJS profiling options
> --nodejs-mode: Controls which profiler is used for NodeJS.
> - none - (default) no profiler is used.
> - perf - augment the system profiler (perf) results with jitdump files generated by NodeJS. This requires running your node processes with --perf-prof (and for Node >= 10, with --interpreted-frames-native-stack). See this [NodeJS page](https://nodejs.org/en/docs/guides/diagnostics-flamegraph/) for more information.

Or, if it is possible to know from GProfiler Studio which part of the own code is causing the memory and cpu usage to increase. We have selected `My code` and set a filter for particular Service running in ECS, but that didn't give us any better insight in the issue.

I am thankful for any valuable insight.

Jongy:
Hi @hrvacadaIM :) I'm glad to hear gProfiler is useful for you.

If you're using NodeJS and not passing `--nodejs-mode=perf`, then I figure you're getting a native profile of NodeJS/V8 itself and not of your code.
To gain visibility into your NodeJS code itself (i.e to see JavaScript function names) you'll need to pass `--nodejs-mode=perf` to the profiler; and yes - it can be done in the ECS task definition, just add it to the `command` array like this:
```
            "command": [
                "-cu",
                "--token",
                "<TOKEN>",
                "--service-name",
                "<SERVICE NAME>",
                "--nodejs-mode=perf"
            ],
```
Additionally, you need to add the `--perf-prof` flag to your NodeJS application itself (i.e change `node myscript.js` to `node --perf-prod myscript.js`) as described [here](https://github.com/Granulate/gprofiler#nodejs-profiling-options). This is required to have NodeJS generate debugging information that gProfiler will use.

After you do that - gProfiler will show you a profile of your JavaScript code, and you'll be able to figure out what parts of your code are consuming CPU time.
[...]

PR #438, opened 2022-08-09
Add script which injects DSO into NodeJS process

kwisniewski98:
## Description
Script assumes that you have already compiled [linux-perf](https://www.npmjs.com/package/linux-perf) module and takes path to it as argument. Script can be easily modified to load any module, it is focused on that particular module only by running `start()` on it. In future this should be part of NodeJS profiler, but for now it is only standalone script.

## Related Issue
https://github.com/Granulate/gprofiler/issues/418

## Motivation and Context
In current state, gProfiler cannot profile NodeJS without restarting it. Functions from that script will allow it to attach to NodeJS process and enable generating debug symbols.

## How Has This Been Tested?
Tested manually on app with CommonJS entry script

## Checklist:

### DIFF
--- a/gprofiler/profilers/node.py
+++ b/gprofiler/profilers/node.py
@@ -1,0 +1,122 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+"""
+This script allows user to inject node-linux-perf module and start it to generate perf maps at runtime.
+To prepare module run npm install --prefix <location> linux-perf or clone it from https://github.com/mmarchini-oss/node-linux-perf and run npm install 
+This only works for applications, that have CommonJS as entry script.
+
+It can be also easly modified to utilize any other module by changing expression in CDP request.
+"""
+
+import asyncio
+import websockets
+import json
+import requests
+import os
+import signal
+import time
+import psutil
+import argparse
+
+EXPECTED_CDP_RESPONSE = {
+    "id": 1,
+    "result": {"result": {"type": "boolean", "value": True}},
+}
+
+
+class NodeDebuggerUrlNotFound(Exception):
+    pass
+
+
+class NodeDebuggerUnexpectedResponse(Exception):
+    pass
+
+
+def retry(exception, retries, waitTime):
+    def wrap(f):
+        def wrapped(*args):
+            for _ in range(0, retries):
+                try:
+                    return f(*args)
+                except exception:
+                    time.sleep(waitTime)
+
+        return wrapped
+
+    return wrap
+
+
+def __start_debugger(pid):
+    os.kill(pid, signal.SIGUSR1)
+
+
+@retry(NodeDebuggerUrlNotFound, 5, 1)
+def __get_debugger_url(pid):
+    process = psutil.Process(pid)
+    possible_ports = [connection.laddr.port for connection in process.connections()]
+    debugger_url = ""
+    for port in possible_ports:
+        possible_port_response = requests.get(f"http://127.0.0.1:{port}/json/list")
+        if (
+            possible_port_response.status_code != 200
+            or not "application/json"
+            in possible_port_response.headers.get("Content-Type")
+        ):
+            continue
+
+        response_json = possible_port_response.json()
+        if (
+            not isinstance(response_json, list)
+            or len(response_json) == 0
+            or not isinstance(response_json[0], dict)
+            or not "webSocketDebuggerUrl" in response_json[0].keys()
+        ):
+            continue
+
+        debugger_url = response_json[0]["webSocketDebuggerUrl"]
+    if not debugger_url:
+        raise NodeDebuggerUrlNotFound
+    return debugger_url
+
+
+@retry(NodeDebuggerUnexpectedResponse, 5, 1)
+async def __load_dso(sock, module_path):
+    cdp_request = {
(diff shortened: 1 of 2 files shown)


gprofiler/profilers/node.py
```diff
@@ -0,0 +1,122 @@
...
+
+class NodeDebuggerUrlNotFound(Exception):
+    pass
+
+
+class NodeDebuggerUnexpectedResponse(Exception):
+    pass
+
+
+def retry(exception, retries, waitTime):
```
Jongy, line 37: Naming-wise - use wait_time and not waitTime (not camelCase)

Jongy, line 37: Can use https://pypi.org/project/retry/ ?

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,122 @@
...
+                    return f(*args)
+                except exception:
+                    time.sleep(waitTime)
+
+        return wrapped
+
+    return wrap
+
+
+def __start_debugger(pid):
```
Jongy, line 51: Avoid double `__` - use a single `_` to prefix private functions, `__` is not the convention in Python and can also have adverse effects because some reserved names are prefixed with `__`.

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,122 @@
...
+    os.kill(pid, signal.SIGUSR1)
+
+
+@retry(NodeDebuggerUrlNotFound, 5, 1)
+def __get_debugger_url(pid):
+    process = psutil.Process(pid)
+    possible_ports = [connection.laddr.port for connection in process.connections()]
+    debugger_url = ""
+    for port in possible_ports:
```
Jongy, line 60: Isn't the port known before? i.e do we need to do this guessing here?

Preferably we'll avoid accessing the app's ports randomly - this may affect the application in undesired ways.

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,122 @@
...
+        possible_port_response = requests.get(f"http://127.0.0.1:{port}/json/list")
```
Jongy, line 61: This works only if the application is in the same network namespace, i.e not containerized.

For that we have `run_in_ns`. If you'd run this chunk of code inside `run_in_ns(["net"], target_pid=pid)` then it will work for containers as well. All networking operations you do with the target app must be done this way

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,122 @@
...
+
+    return wrap
+
+
+def __start_debugger(pid):
+    os.kill(pid, signal.SIGUSR1)
+
+
+@retry(NodeDebuggerUrlNotFound, 5, 1)
+def __get_debugger_url(pid):
```
Jongy, line 56: Please add typing everywhere, so here - 
```suggestion
def __get_debugger_url(pid: int) -> str:
```

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,122 @@
...
+    process = psutil.Process(pid)
+    possible_ports = [connection.laddr.port for connection in process.connections()]
+    debugger_url = ""
+    for port in possible_ports:
+        possible_port_response = requests.get(f"http://127.0.0.1:{port}/json/list")
+        if (
+            possible_port_response.status_code != 200
+            or not "application/json"
+            in possible_port_response.headers.get("Content-Type")
```
Jongy, line 65: If `Content-Type` doesn't exist then you have `"application/json" in None` which will raise

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,122 @@
...
+            "replMode": True,
+        },
+    }
+    await sock.send(json.dumps(cdp_request))
+    message = await sock.recv()
+    if json.loads(message) != EXPECTED_CDP_RESPONSE:
+        raise NodeDebuggerUnexpectedResponse(json.loads(message))
+
+
+async def generate_perf_map(pid, module_path):
```
Jongy, line 100: Since no code in gProfiler currently uses async, I think it's best if we don't mix it in, can this code be written w/o async?

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,122 @@
...
+    return debugger_url
+
+
+@retry(NodeDebuggerUnexpectedResponse, 5, 1)
+async def __load_dso(sock, module_path):
+    cdp_request = {
+        "id": 1,
+        "method": "Runtime.evaluate",
+        "params": {
+            "expression": f'process.mainModule.require("{module_path}").start()',
```
Jongy, line 90: The module path here is in the context of the script, while the app may be containerized.

Here, though, unlike [my comment](https://github.com/Granulate/gprofiler/pull/438#r944920412) about `run_in_ns`, you need to make the DSO accessible for the target process, because the target process will try to access the path (or alternatively you can encode the entire file in the `Runtime.evaluate` call but that's a bit dirty).

Logically it's the same as what we do in `_copy_libap` - where we copy the `libasyncProfiler.so` DSO to the namespace of the target process, if needed.

Jongy:
@kwisniewski98 as discussed, we should try to get a full flow working (with the DSO from node-linux-perf) and test it on container & non container nodejs apps. Then once we know that all parts connect well, we can focus on creating the jitdump DSO.

Jongy:
Superseded by #451, all comments were resolved / moved there. Closing this

PR #439, opened 2022-08-10
tests: Add test_controller_pid()

Jongy:


tests/utils.py
```diff
@@ -56,20 +57,28 @@ def start_privileged_container(
     )
 
 
+def wait_for_log(container: Container, log: str, timeout: int = 60) -> None:
+    try:
+        wait_event(timeout, Event(), lambda: re.search(log.encode(), container.logs(), re.DOTALL) is not None)
+    except TimeoutError:
+        print(container.logs())
+        raise
+
```
michelhe, line 66: Not a good idea, if you want to `wait_for_log` twice for the same message, in the second time onward it will always return immediately.

This `wait_for_log` function needs to be stateful and always start from the offset where the previous call found the log

tests/test_controller_pid.py
```diff
@@ -0,0 +1,49 @@
...
+        proc.kill()
+        proc.wait()
+
+        # wait for gprofiler to identify it
+        wait_for_log(container, f"Controller process {proc.pid} has exited; gProfiler stopping...")
+
+        wait_for_container(container)
+    finally:
+        if container is not None:
+            container.remove(force=True)
```
michelhe, line 49: Use of try finally suggests this needs to be in a fixture and not part of the test

tests/test_controller_pid.py
```diff
@@ -0,0 +1,49 @@
...
+            output_collapsed,
+            [],
+            ["-d", "3", "-c", "--controller-pid", str(proc.pid), "--no-java"],
+        )
+
+        cycle_log = r"INFO: gprofiler: Saved flamegraph to "
+        # wait one gprofiler cycle, to ensure gprofiler starts
+        wait_for_log(container, cycle_log)
+        # wait another to ensure it doesn't go down
+        wait_for_log(container, rf"{cycle_log}.*{cycle_log}")
```
michelhe, line 37: See above comment on `wait_for_log`, if the utility function is more resilient then we won't have to do stuff like this.

tests/utils.py
Jongy: My current use actually checks that the message appears multiple times to handle that. I agree that it makes more sense to follow the log and check from the last offset, yeah

tests/test_controller_pid.py
Jongy: Yup I'll change and fix other tests

tests/test_controller_pid.py
Jongy: Yes but not easy to move `start_gprofiler_in_container_for_one_session` to a fixture so not doing it now

michelhe:
```python
@fixture
def cleanup_containers():
   containers = []
   yield containers
   for c in containers:
      c.remove(force=True)

def some_test(cleanup_containers):
   c  = create_containers()
   cleanup_containers.append(c)
```

Jongy: It's ugly

michelhe (approved):

PR #440, opened 2022-08-13
tests: Add perf tests on changing comms

Jongy:
I liked writing them because they helped me grasp the current state of affairs.

An unrelated change - we no longer delete the application docker images, it's a
waste of time.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,7 @@
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
+import glob
 import os
 import stat
 import subprocess
@@ -81,6 +82,11 @@ def java_command_line(tmp_path: Path, java_args: List[str]) -> List[str]:
 
 @fixture
 def command_line(runtime: str, java_command_line: List[str]) -> List[str]:
+    # these do not have non-container application - so it will result in an error if the command
+    # line is used.
+    if runtime.startswith("native"):
+        return ["/bin/false"]
+
     return {
         "java": java_command_line,
         # note: here we run "python /path/to/lister.py" while in the container test we have
@@ -94,10 +100,6 @@ def command_line(runtime: str, java_command_line: List[str]) -> List[str]:
             "--interpreted-frames-native-stack",
             str(CONTAINERS_DIRECTORY / "nodejs/fibonacci.js"),
         ],
-        # these do not have non-container application - so it will result in an error if the command
-        # line is used.
-        "native_fp": ["/bin/false"],
-        "native_dwarf": ["/bin/false"],
     }[runtime]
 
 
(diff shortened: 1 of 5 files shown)


tests/test_perf.py
```diff
@@ -49,3 +54,97 @@ def test_perf(
...
+        collapsed = snapshot_pid_collapsed(profiler, application_pid)
+        assert not is_function_in_collapsed("oative;", collapsed)
+        assert is_function_in_collapsed("native;", collapsed)
+
+
+@pytest.mark.xfail(reason="https://github.com/Granulate/gprofiler/issues/431")
+@pytest.mark.parametrize("runtime", ["native_thread_comm"])
+@pytest.mark.parametrize("perf_mode", ["fp"])  # only fp is enough
+@pytest.mark.parametrize("in_container", [True])  # native app is built only for container
+def test_perf_thread_comm_is_process_comm(
```
Jongy, line 107: Hmm this ended up being the exact same test code :dumb: only different comments and parametrization. I'll merge somehow

tests/conftest.py
```diff
@@ -81,6 +82,11 @@ def java_command_line(tmp_path: Path, java_args: List[str]) -> List[str]:
 
 @fixture
 def command_line(runtime: str, java_command_line: List[str]) -> List[str]:
+    # these do not have non-container application - so it will result in an error if the command
+    # line is used.
+    if runtime.startswith("native"):
+        return ["/bin/false"]
+
```
michelhe, line 89: Looks like you are just piggybacking the `application`s family of fixture whilst you actually want a fixture of your own..

Jongy: You're right. I plan to address that later when I add tests for py-spy and PyPerf across the full matrix of (musl, python_version)

tests/test_perf.py
```diff
@@ -49,3 +55,101 @@ def test_perf(
...
+    """
+    with system_profiler as profiler:
+        # running perf & script now with --show-task-events would show:
+        #   pative 1925947 [010] 987095.272656: PERF_RECORD_COMM: pative:1925904/1925947
+        # our perf will prefer to use the exec comm, OR oldest comm available if exec
+        # one is not present; see logic in thread__exec_comm().
+
+        # first run - we get the changed name, because the app started before perf began recording.
+        # note that we still pick the process name and not thread name! (thread native is 'pative')
+        _assert_comm_in_profile(profiler, application_pid, False)
```
Jongy, line 140: Check also no `pative`

tests/test_perf.py
```diff
@@ -49,3 +55,101 @@ def test_perf(
...
+        #   native 1925904 [006] 987095.271786: PERF_RECORD_COMM exec: native:1925904/1925904
+        #   oative 1925904 [006] 987095.272430: PERF_RECORD_COMM: oative:1925904/1925904
+        #   oative 1925904 [006] 987095.272545: PERF_RECORD_FORK(1925904:1925947):(1925904:1925904)
+        #   pative 1925947 [010] 987095.272656: PERF_RECORD_COMM: pative:1925904/1925947
+        # we take the exec comm for all threads so we remain with the first, "native".
+
+        # this fails - because in thread__fork(), perf takes only the current comm of the forking thread
+        # and attaches it to the forked one. thus the exec comm (which is not current, but last) is lost.
+        # the fix can be to copy all comms in thread__fork().
+        _assert_comm_in_profile(profiler, application_pid, True)
```
Jongy, line 155: Check also no `pative`

tests/test_perf.py
```diff
@@ -49,3 +55,106 @@ def test_perf(
...
+    """
+    with system_profiler as profiler:
+        # first run - we get the changed name, because the app started before perf began recording.
+        _assert_comm_in_profile(profiler, application_pid, False)
+
+        # now, while perf is still running & recording, we restart the app
+        application_pid = _restart_app_container(application_docker_container)
+
+        # second run - we get the original name, because the app started after perf began recording.
+        _assert_comm_in_profile(profiler, application_pid, True)
```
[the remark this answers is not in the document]
Jongy: Nope. `snapshot` is - run gProfiler for a single session. Now for `perf` and `PyPerf` and `phpspy`, since they're continuous, if you:
```
snapshot()
xxx()
snapshot()
```
then it's racy because after `snapshot()` returns, they continue collecting samples until `xxx()` happens.

BUT here, we use `snapshot_pid_collapsed` to get the samples from a single PID, and we actually restart the app and switch to use the new PID. So nope, not racy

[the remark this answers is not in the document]
michelhe: Because all the invocations are under the same `with` - it appeared confusing. thx.

michelhe (approved):

PR #442, opened 2022-08-18
GProfiler with Py-Spy on Windows

teoluwaf:
PR provides batch scripts to build py-spy and gprofiler using py-spy on Windows

## Description
Provided batch scripts to build py-spy and gprofiler from source on Windows. This PR contains a build.bat file, a build-pyspy.bat file, py-spy (v0.3.10g1) submodule, granulate-utils submodule along with some Windows specific dependencies. In addition to the changes made to the gprofiler repo (which is tracked), the other new additions can all be found in the deps, py-spy, granulate-utils subfolders.

## Motivation and Context
This PR addresses Granulates initial foray into providing Windows support. To be specific, this PR provides python profiling on Windows using py-spy.


Jongy:
@teoluwaf can you please update to master & resolve conflicts?

I will go over the PR and suggest parts that we can split out as individual PRs.

gprofiler/main.py
```diff
@@ -177,7 +181,7 @@ def _generate_output_files(
     ) -> None:
         start_ts = get_iso8601_format_time(local_start_time)
         end_ts = get_iso8601_format_time(local_end_time)
-        base_filename = os.path.join(self._output_dir, "profile_{}".format(end_ts))
+        base_filename = os.path.join(self._output_dir, "profile_{}".format(end_ts.replace(':','-')))
```
Jongy, line 184: Why is that? Windows doesn't allow `:` in names?

gprofiler/main.py
```diff
@@ -121,12 +121,16 @@ def __init__(
...
         # files unnecessarily.
         self._temp_storage_dir = TemporaryDirectoryWithMode(dir=TEMPORARY_STORAGE_PATH, mode=0o755)
+
         try:
             self.system_profiler, self.process_profilers = get_profilers(
                 user_args,
                 storage_dir=self._temp_storage_dir.name,
                 stop_event=self._stop_event,
             )
+            # Added by Tope
```
Jongy, line 131: Can be removed

gprofiler/main.py
```diff
@@ -121,12 +121,16 @@ def __init__(
         # the latter can be root only. the former can not. we should do this separation so we don't expose
         # files unnecessarily.
         self._temp_storage_dir = TemporaryDirectoryWithMode(dir=TEMPORARY_STORAGE_PATH, mode=0o755)
+
```
Jongy, line 124: Can be removed

build-pyspy.bat
```diff
@@ -0,0 +1,35 @@
+@echo off
+WHERE cargo
+IF %ERRORLEVEL% NEQ 0 (
+	ECHO cargo wasn't found. Attempting to install...
+	.\dep\rustup-init.exe -y
```
Jongy, line 5: Instead of comitting it to the repo - download it during build as decribed in https://forge.rust-lang.org/infra/other-installation-methods.html#other-ways-to-install-rustup. Eventually this bat file is meant to run in the CICD on Windows, so I assume there IS a way to download & install external dependencies (i.e either CIs come with e.g `wget` installed, or you have some way to specify URLs you wish to be downloaded.

I checked [micropython](https://github.com/micropython/micropython) which has a [Windows CI](https://github.com/micropython/micropython/blob/master/ports/windows/.appveyor.yml), we'll target something like that, you can check if AppVeyor images have wget or curl and then we can depend on that.

Same for other dependencies - refrain from committing them and prefer to get them during CI / this script

exe-requirements.txt
```diff
@@ -11,3 +12,4 @@ pyinstaller==4.10; platform.machine == "aarch64"
 # During startup program terminated with signal SIGSEGV, Segmentation fault.
 # staying with 0.12.1 for the mean time...
 staticx==0.12.1; platform.machine == "x86_64"
+WMI==1.5.1; platform.machine == "AMD64"
```
Jongy, line 15: Use `sys_platform` and install it only if Windows

gprofiler/metadata/application_metadata.py
```diff
@@ -55,4 +55,7 @@ def get_metadata(self, process: Process) -> Optional[Dict]:
         return metadata
 
     def make_application_metadata(self, process: Process) -> Dict[str, Any]:
-        return {"exe": process.exe(), "execfn": read_process_execfn(process)}
+        # Disabled for lack of Windows support
+        #return {"exe": process.exe(), "execfn": read_process_execfn(process)}
+        return {"exe": process.exe(), "execfn": None}
```
Jongy, line 60: Add a helper function `is_windows()` that checks the platform, then pass `None` if Windows and otherwise call `read_process_execfn` as the code does now

gprofiler/metadata/system_metadata.py
```diff
@@ -93,9 +93,10 @@ def get_local_ip() -> str:
 
 
 def get_mac_address() -> str:
+    return "unknown"
```
Jongy, line 96: We'll need to implement it for Windows, but for MVP we can add

```
if is_windows():
  return "unknown"
```

gprofiler/profilers/php.py
```diff
@@ -2,7 +2,9 @@
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
-import fcntl
+# Disabled code block - Not supported on Windows - Alternative required
```
Jongy, line 5: This profiler doesn't support Windows now - so instead of modifying it, you can edit `gprofiler/profilers/__init__.py` to skip importing it at all. If it's not imported - it won't be enabled (as `register_profiler` won't be called)

gprofiler/profilers/profiler_base.py
```diff
@@ -128,6 +128,7 @@ def snapshot(self) -> ProcessToProfileData:
                 try:
                     comm = process_comm(process)
                 except NoSuchProcess:
+                    logger.debug("No such process: {process.pid}")
```
Jongy, line 131: Let's remove the debug prints :)

gprofiler/utils/__init__.py
```diff
@@ -18,6 +19,8 @@
 import subprocess
 import sys
 import time
+import pythoncom
```
Jongy, line 22: Please put Windows-specific imports under `if is_windows()`

gprofiler/profilers/python.py
```diff
@@ -496,7 +511,7 @@ def stop(self) -> None:
     # we build pyspy for both, pyperf only for x86_64.
     # TODO: this inconsistency shows that py-spy and pyperf should have different Profiler classes,
     # we should split them in the future.
-    supported_archs=["x86_64", "aarch64"],
+    supported_archs=["x86_64", "aarch64", "AMD64"],
```
Jongy, line 514: `AMD64` and `x86_64` are the same, is this because Windows reports `AMD64`?

gprofiler/utils/__init__.py
```diff
@@ -378,6 +403,10 @@ def grab_gprofiler_mutex() -> bool:
     on filesystem structure (as happens with file-based locks).
     In order to see who's holding the lock now, you can run "sudo netstat -xp | grep gprofiler".
     """
+    # Bypass due to lack of Windows support
+    return True
+    # End Edit
```
Jongy, line 408: Under `if is_windows()` and add a TODO for us to fix later (we'll implement the mutex in some other way)

gprofiler/utils/__init__.py
```diff
@@ -429,6 +461,10 @@ def atomically_symlink(target: str, link_node: str) -> None:
     """
     tmp_path = link_node + ".tmp"
     os.symlink(target, tmp_path)
+    # Added to fix bug where gprofiler crashes if link_node already exists.
```
Jongy, line 464: On Windows, you cannot `os.rename` a file if the target already exists?

gprofiler/utils/__init__.py
```diff
@@ -125,20 +134,25 @@ def start_process(
...
-        preexec_fn=cur_preexec_fn,
-        env=env,
-        **kwargs,
-    )
+                cmd,
+                stdout=kwargs.pop("stdout", subprocess.PIPE),
+                stderr=kwargs.pop("stderr", subprocess.PIPE),
+                stdin=subprocess.PIPE,
+                # Disabled for lack of Windows support
+                #preexec_fn=cur_preexec_fn,
```
Jongy, line 151: Windows doesn't support `preexec_fn`? Or it's our preexec function that's not Windows-compatible? We can avoid calling it in that case, or refrain from passing it only if it's Windows (and `assert` that no preexec function was passed)

gprofiler/utils/__init__.py
```diff
@@ -207,7 +221,7 @@ def run_process(
     via_staticx: bool = False,
     check: bool = True,
     timeout: int = None,
-    kill_signal: signal.Signals = signal.SIGKILL,
+    kill_signal: signal.Signals = signal.SIGTERM,
```
Jongy, line 224: Revert?

gprofiler/utils/__init__.py
```diff
@@ -265,11 +279,23 @@ def run_process(
         raise CalledProcessError(retcode, process.args, output=stdout, stderr=stderr)
     return result
 
+def pgrep_keyword_exe(match: str) -> List[Process]:
```
Jongy, line 282: Put this definition under `if is_windows()`

Jongy: Why `keyword` in the name?

gprofiler/profilers/python.py
```diff
@@ -243,7 +251,8 @@ def _profile_process(self, process: Process) -> ProfileData:
 
     def _select_processes_to_profile(self) -> List[Process]:
```
Jongy, line 252: Instead of modifying this function, add `if is_windows()` and call your new function in that case, otherwise fall-through the old behavior

gprofiler/profilers/python.py
```diff
@@ -281,7 +291,10 @@ class PythonEbpfProfiler(ProfilerBase):
     _EVENTS_BUFFER_PAGES = 256  # 1mb and needs to be physically contiguous
     # 28mb (each symbol is 224 bytes), but needn't be physicall contiguous so don't care
     _SYMBOLS_MAP_SIZE = 131072
-    _DUMP_SIGNAL = signal.SIGUSR2
```
Jongy, line 284: Undo those changes - we don't need the `PythonEbpfProfiler` on Windows, so this class shouldn't even be defined. We can move it to another file or we can put it under `if windows()`. I think another file makes more sense. Let's start with avoiding the import of `java.py`, `php.py` on Windows and then decide what we do with `python.py` which is mixed with some things that we need on Windows, and some that we don't.

Jongy:
In preparation, I added https://github.com/Granulate/gprofiler/pull/465

exe-requirements.txt
```diff
@@ -1,5 +1,6 @@
 # requirements for the standalone executable
-pyinstaller==4.0; platform.machine == "x86_64"
+pyinstaller==5.1; platform.machine == "AMD64"
+wheel==0.37.0; platform.machine == "AMD64"
```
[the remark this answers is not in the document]
teoluwaf: I'm not sure why I included wheel. This has been quite a long time ago. So, I'll try removing the package to see if it breaks anything. Also I'll see if pyinstaller==4.0 works.

build-pyspy.bat
teoluwaf: Ok, makes sense. I thought the batch script had to be self-contained without expecting the user to install any dependency. I can strip out the binaries for Rust and Windows SDK/Visual C++ build tools and assume they are already installed in your CI environment. Are you fine with this?

gprofiler/main.py
Jongy: (I meant the space :) I commented on the space line)

build-pyspy.bat
Jongy: The script needn't be self container. It's fair to download dependencies here (and it makes no sense to commit them - best to download and by that the user sees exactly what gets packed inside)

exe-requirements.txt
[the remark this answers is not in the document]
Jongy: Sounds good

build-pyspy.bat
teoluwaf: For Rust, you'll need to install a wget executable to then pull the Rust exe. For visual studio the executable that is included in the dep folder is what installs the needed components. At some point, we have to make assumptions about what dependencies are pre-installed in your CI environment - whether it is wget/rust/visual studio build tools or all of them. By the way, these dependencies are only needed to build the py-spy.exe. Can we assume py-spy.exe is built and available in your CI environment?

gprofiler/profilers/python.py
teoluwaf: Yes, Windows reports AMD64.

gprofiler/main.py
teoluwaf: Ah! Done.

gprofiler/main.py
teoluwaf: Yes, Windows doesn't allow : in names

exe-requirements.txt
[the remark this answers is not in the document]
teoluwaf: Removed wheel from the requirements as I don't seem to need it. Not sure why I had included it in the first place. It must have been needed at some point.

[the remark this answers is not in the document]
teoluwaf: Also reverted to v4.0 for pyinstaller.

gprofiler/profilers/profiler_base.py
teoluwaf: Done

gprofiler/utils/__init__.py
teoluwaf: Done

gprofiler/profilers/php.py
teoluwaf: Done

gprofiler/utils/__init__.py
teoluwaf: Done

gprofiler/utils/__init__.py
teoluwaf: Yes, you cannot. You'll get something like the following error, if you don't first delete the symbolic link.

This will only work one time because the symbolic link doesn't exist.

tmp_path = "tma-shortcut.tmp"
os.symlink("some-file", tmp_path)
os.rename(tmp_path, "tma-shortcut.txt")

If that codepath is called again, you'll get an error: 
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'tma-shortcut.tmp' -> 'tma-shortcut.txt'

gprofiler/utils/__init__.py
teoluwaf: Your preexec function (os.setpgrp) is not Windows compatible. I have the is_windows() check to call it without one if it is running on a Windows host system.

gprofiler/utils/__init__.py
teoluwaf: No. Windows does not support SIGKILL. I have added is_windows() check to select between both options.

gprofiler/utils/__init__.py
teoluwaf: Done. Removed the keyword as well.

gprofiler/profilers/python.py
teoluwaf: Done

gprofiler/profilers/python.py
teoluwaf: The PythonProfiler class in gprofiler.profilers.python imports PythonEbpfProfiler and has _create_ebpf_profiler defined as a member function. What I have done now is to move the PythonEbpfProfiler and PythonEbpfError to its own file and import that module into the python.py file. I also still have to use the is_windows helper function to switch between signal.SIGUSR2 and signal.SIGFPE as signal.SIGUSR2 is not supported on Windows.

.gitmodules
```diff
@@ -1,3 +1,7 @@
 [submodule "granulate-utils"]
 	path = granulate-utils
-	url = git@github.com:Granulate/granulate-utils.git
+	url = git@github.com:teoluwaf/granulate-utils.git
+	branch = py-win
+[submodule "py-spy"]
```
Jongy, line 5: Remove? (we don't need it)

exe-requirements.txt
```diff
@@ -1,8 +1,7 @@
 # requirements for the standalone executable
-pyinstaller==4.0; platform.machine == "x86_64"
```
Jongy, line 2: Add it back with `platform.machine == "x86_64" and not sys.platform == "win32"`

gprofiler/main.py
```diff
@@ -177,7 +177,7 @@ def _generate_output_files(
     ) -> None:
         start_ts = get_iso8601_format_time(local_start_time)
         end_ts = get_iso8601_format_time(local_end_time)
-        base_filename = os.path.join(self._output_dir, "profile_{}".format(end_ts))
+        base_filename = os.path.join(self._output_dir, "profile_{}".format(end_ts.replace(':','-')))
```
Jongy, line 180: `:` not allowed on Windows?

gprofiler/metadata/application_metadata.py
```diff
@@ -55,4 +56,7 @@ def get_metadata(self, process: Process) -> Optional[Dict]:
         return metadata
 
     def make_application_metadata(self, process: Process) -> Dict[str, Any]:
-        return {"exe": process.exe(), "execfn": read_process_execfn(process)}
+        if is_windows():
+            return {"exe": process.exe(), "execfn": None}
+        else:
+            return {"exe": process.exe(), "execfn": read_process_execfn(process)}
```
Jongy, line 62:
```suggestion
        return {"exe": process.exe(), "execfn": None if is_windows() else read_process_execfn(process)}
```

gprofiler/metadata/system_metadata.py
```diff
@@ -133,6 +138,7 @@ def get_mac_address() -> str:
         return address
 
     return "unknown"
+
```
Jongy, line 141:
```suggestion
```

gprofiler/profilers/__init__.py
```diff
@@ -1,8 +1,13 @@
...
-from gprofiler.profilers.php import PHPSpyProfiler
 from gprofiler.profilers.python import PythonProfiler
-from gprofiler.profilers.ruby import RbSpyProfiler
+from granulate_utils import is_windows
+if not is_windows():
+    from gprofiler.profilers.java import JavaProfiler
+    from gprofiler.profilers.perf import SystemProfiler
+    from gprofiler.profilers.php import PHPSpyProfiler
+    from gprofiler.profilers.ruby import RbSpyProfiler
+from gprofiler.profilers.python_ebpf import PythonEbpfProfiler
```
Jongy, line 9: This one needs to be under `if is_windows()` as well, no?

gprofiler/profilers/__init__.py
```diff
@@ -1,8 +1,13 @@
...
 
-__all__ = ["JavaProfiler", "SystemProfiler", "PHPSpyProfiler", "PythonProfiler", "RbSpyProfiler"]
+__all__ = ["JavaProfiler", "SystemProfiler", "PHPSpyProfiler", "PythonProfiler", "RbSpyProfiler"] if not is_windows() else ["SystemProfiler", "PythonProfiler"]
```
Jongy, line 11: I'd structure it as `__all__ = '["SystemProfiler", "PythonProfiler"]` and under the `is_windows()` branch, `__all__ += [...]`

gprofiler/profilers/php.py
```diff
@@ -42,7 +42,10 @@
 )
 class PHPSpyProfiler(ProfilerBase):
     PHPSPY_RESOURCE = "php/phpspy"
+    # Modified dump_signal as signal.SIGUSR2 is not supported on Windows
```
Jongy, line 45: Remove edits from this file :)

gprofiler/profilers/profiler_base.py
```diff
@@ -155,5 +155,6 @@ def snapshot(self) -> ProcessToProfileData:
                     )
 
                 results[pid] = result
+        logger.debug(f"Result snapshot: {results}")
```
Jongy, line 158: Remove once we're done

gprofiler/profilers/python.py
```diff
@@ -211,7 +221,7 @@ def _profile_process(self, process: Process) -> ProfileData:
                     self._make_command(process.pid, local_output_path),
                     stop_event=self._stop_event,
                     timeout=self._duration + self._EXTRA_TIMEOUT,
-                    kill_signal=signal.SIGKILL,
+                    kill_signal=signal.SIGTERM if is_windows() else signal.SIGKILL,
```
Jongy, line 224: `SIGTERM` on windows does `TerminateProcess()` and kills it?

gprofiler/profilers/python.py
```diff
@@ -190,7 +200,7 @@ def _make_command(self, pid: int, output_path: str) -> List[str]:
             "--format",
             "raw",
             "-F",
-            "--gil",
+            #"--gil", Not supported on python 3.9.13. Error: Cannot detect GIL holding in version '3.9.13' on the current platform (reason: failed to find _PyRuntime symbol)
```
Jongy, line 203: We need to upgrade our py-spy :) That'll be done. Reminding to revert this

gprofiler/utils/__init__.py
```diff
@@ -369,7 +404,6 @@ def assert_program_installed(program: str) -> None:
     else:
         raise ProgramMissingException(program)
 
-
```
Jongy, line 372: Bring back

gprofiler/utils/process.py
```diff
@@ -15,14 +15,16 @@ def ensure_running(process: Process) -> None:
...
-        ensure_running(process)
+        ensure_running(process)'''
+    ensure_running(process)
+    return process.name()
 
-    name_line = status.splitlines()[0]
+    '''name_line = status.splitlines()[0]
     assert name_line.startswith("Name:\t")
-    return name_line.split("\t", 1)[1]
+    return name_line.split("\t", 1)[1]'''
```
Jongy, line 30: Instead - you can provide this implementation of the function under `if is_windows()`, and provide a Windows alternative that does just `process.name()`

gprofiler/utils/__init__.py
Jongy: I see. The idea with the `rename` is to have it atomic - i.e at any given point, `last_profile` is ALWAYS a link to the last profile. So let's put this part under `if is_windows() and os.path.exists(link_node)`, and also add a comment explaining that Windows cannot override a link with `rename`.

gprofiler/profilers/python.py
teoluwaf: SIGTERM does not exist on Windows

gprofiler/utils/__init__.py
```diff
@@ -125,20 +133,32 @@ def start_process(
...
+                    stdin=subprocess.PIPE,
+                    env=env,
+                    **kwargs,)
+    else:
+        popen = Popen(
+                    cmd,
+                    stdout=kwargs.pop("stdout", subprocess.PIPE),
+                    stderr=kwargs.pop("stderr", subprocess.PIPE),
+                    stdin=subprocess.PIPE,
+                    preexec_fn=cur_preexec_fn,
```
Jongy, line 158: Is this change needed given that you only set `cur_preexec_fn` to `os.setpgrp` in the `not is_windows()` case? (I refer to this `is_windows()` branch from line 144)

gprofiler/profilers/python.py
Jongy: It's good to move to a separate file. It doesn't make sense though to change the signal because PyPerf requires SIGUSR2 (and PyPerf itself is a Linux only application). So we just need to make sure that this code (i.e the definition of `_DUMP_SIGNAL`) doesn't run on Windows :)

gprofiler/main.py
teoluwaf: Yes, not allowed on Windows

gprofiler/profilers/python.py
Jongy: Yeah I understand, I'm asking what's the logical substitute. Is it `SIGTERM`?

gprofiler/profilers/__init__.py
teoluwaf: No, because PythonProfiler(ProfilerInterface) uses it and is required for both Windows and Linux

gprofiler/profilers/python.py
teoluwaf: I included a check of is_windows in the definition of _DUMP_SIGNAL. PythonEbfProfiler is used by the class PythonProfiler. So, I can't choose to only import PythonEbfProfiler on Linux - Windows also requires it.

gprofiler/profilers/python.py
teoluwaf: I think SIGTERM is the closest substitute for Windows. Here's what's supported (https://docs.microsoft.com/en-us/previous-versions/xdkz3x12(v=vs.140)):

SIGABRT | Abnormal termination
SIGFPE | Floating-point error
SIGILL | Illegal instruction
SIGINT | CTRL+C signal
SIGSEGV | Illegal storage access
SIGTERM | Termination request

gprofiler/utils/__init__.py
teoluwaf: Sounds good! Done

gprofiler/utils/process.py
```diff
@@ -11,16 +11,22 @@
...
     return re.search(pattern, read_proc_file(process, "maps").decode(), re.MULTILINE)
 
 
 def process_comm(process: Process) -> str:
-    status = read_proc_file(process, "status").decode()
-    name_line = status.splitlines()[0]
-    assert name_line.startswith("Name:\t")
-    return name_line.split("\t", 1)[1]
+    if is_windows():
+        # TODO: Check if process is running
```
Jongy, line 23: Will it raise if the process is not running?

Jongy: Commenting on the https://github.com/Granulate/gprofiler/pull/515 PR

.gitmodules
Jongy: Remove the submodule - instead, clone the repository in a build script and pin it to the same hash as we use in the Linux build script. (actually, better, to ensure they both share the hash - add a file `scripts/pyspy_tag` and `scripts/pyspy_revision` which represent the git tag & revision used to build pyspy; to avoid cloning it)

build-pyspy.bat
Jongy: We can assume that wget is installed (i.e require the user to install it). We can't assume py-spy is available - we need to build it, mostly because we build it from a specific revision which includes our changes (see `scripts/pyspy_build.sh`).

Let's prepare that list of required dependencies? If VS SDK & Rust can be installed automatically once you have wget - that's good, and let's require only wget.

build-pyspy.bat
```diff
@@ -0,0 +1,56 @@
...
+)
+
+start /wait .\dep\vs_BuildTools.exe --passive --add Microsoft.VisualStudio.Workload.VCTools --add Microsoft.VisualStudio.Component.Windows10SDK.19041 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64
+
+SET LINK_LOC="C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\bin\Hostx86\x86\link.exe"
+
+SET SDK_LOC="C:\Program Files (x86)\Windows Kits\10\Lib\10.0.19041.0\um\x86\advapi32.lib"
```
Jongy, line 12: Can't these 2 paths change? (e.g `14.33.31629` looks like a minor version that can be changed without any change on our side?)

build-pyspy.bat
```diff
@@ -0,0 +1,56 @@
...
+ECHO "Done installing dependencies."
+
+ECHO "Current Path: %PATH%"
+SET PATH=%PATH%;%USERPROFILE%\.cargo\bin;%CD%\deps
+ECHO "Modified Path: %PATH%"
+CD py-spy
+rustup default stable
+rustup target add x86_64-pc-windows-gnu
+
+REM rustup toolchain install stable-x86_64-pc-windows-gnu
```
Jongy, line 36: What does REM do?

build-pyspy.bat
```diff
@@ -0,0 +1,56 @@
+@echo off
```
Jongy, line 1: Let's place the Windows scripts under `scripts/windows/`

build.bat
```diff
@@ -0,0 +1,51 @@
...
+) ELSE (
+	ECHO Building py-spy executable...
+	CALL build-pyspy.bat
+)
+CD app
+
+WHERE pip
+IF %ERRORLEVEL% NEQ 0 (
+	ECHO pip wasn't found.
+	ECHO Install pip for current and re-run build.bat
```
Jongy, line 34: Install pip in the script here? (e.g https://phoenixnap.com/kb/install-pip-windows, via `get-pip.py`)

exe-requirements.txt
```diff
@@ -11,3 +11,5 @@ pyinstaller==4.10; platform.machine == "aarch64"
 # During startup program terminated with signal SIGSEGV, Segmentation fault.
 # staying with 0.12.1 for the mean time...
 staticx==0.12.1; platform.machine == "x86_64"
+WMI==1.5.1; sys.platform == "win32"
```
Jongy, line 14: Don't need it here - we have in requirements.txt

Jongy: Same for `netifaces`

gprofiler/profilers/php.py
```diff
@@ -43,6 +43,7 @@
 class PHPSpyProfiler(ProfilerBase):
     PHPSPY_RESOURCE = "php/phpspy"
     dump_signal = signal.SIGUSR2
+
```
Jongy, line 46: undo the 2 changes in this file

gprofiler/profilers/python.py
```diff
@@ -189,7 +192,7 @@ def _make_command(self, pid: int, output_path: str, duration: int) -> List[str]:
             "--format",
             "raw",
             "-F",
-            "--gil",
+            # "--gil",
```
Jongy, line 195: Not working on Windows? In any case, disable only for the Windows branch, not for Linux

gprofiler/profilers/python.py
```diff
@@ -242,7 +245,13 @@ def _profile_process(self, process: Process, duration: int, spawned: bool) -> Pr
 
     def _select_processes_to_profile(self) -> List[Process]:
         filtered_procs = []
-        for process in pgrep_maps(DETECTED_PYTHON_PROCESSES_REGEX):
+        all_processes = []
```
Jongy, line 248:
```suggestion
```

Don't need to assign it, both sides of the `if` assign it

gprofiler/profilers/python.py
Jongy: I think it makes more sense here, to keep `supported_archs` for Linux, and add `supported_windows_archs` for the Windows support.

Then, for py-spy, we add `supported_windows_archs=["AMD64"]`. For all other profilers, we don't add, and thus only the Python one can be used now. Then we just have to fix the code in `get_profilers` to use `supported_windows_archs` when appropriate, i.e on Windows, otherwise the old `supported_archs`.

build-pyspy.bat
Jongy: Ah I see, comment. So, if there are not required, remove them please

build.bat
```diff
@@ -0,0 +1,51 @@
...
+	EXIT /B -1
+)
+Rem pip install --no-cache-dir --user --upgrade pip
+MKDIR granulate-utils 2> NUL
```
Jongy, line 38: What are you doing in the next 10-15 lines? Files are copied around and you re-create `granulate-utils`, what's the need for it? Can't we work in the existing source directory?

gprofiler/main.py
```diff
@@ -183,7 +183,7 @@ def _generate_output_files(
     ) -> None:
         start_ts = get_iso8601_format_time(local_start_time)
         end_ts = get_iso8601_format_time(local_end_time)
-        base_filename = os.path.join(self._output_dir, "profile_{}".format(end_ts))
+        base_filename = os.path.join(self._output_dir, "profile_{}".format(end_ts.replace(":", "-")))
```
Jongy, line 186: Let's do the replace only if Windows, easy trick is:
```suggestion
        base_filename = os.path.join(self._output_dir, "profile_{}".format(end_ts.replace(":", "-" if is_windows() else ":")))
```

teoluwaf:
### **Dependencies**

> **Options**

1.  [Recommended]: Provide prebuilt executable for py-spy on Windows and build gprofiler executable with provided build script. You control the build environment for the py-spy and the correct tags
2. Assume that the following prerequisites are available on target machine:
- Python (>=3.9.13)
- Pip (>=22.2.2)
- Rust Compiler
- py-spy source with appropriate tag and version
- VS Build Tools executable
> Using VS Build Tools install following components
- Microsoft.VisualStudio.Workload.VCTools
- Microsoft.VisualStudio.Component.Windows10SDK.19041
- Microsoft.VisualStudio.Component.VC.Tools.x86.x64
> Build py-spy and gprofiler executables with provided build script
3. Assume that the following prerequisites are available on target machine:
- Python (>=3.9.13)
- Pip (>=22.2.2)
- py-spy source with appropriate tag and version
- Wget
> Using wget download and install:
- Rust compiler
- VS Build Tools executable 
> Using VS Build Tools executable, install the following components:
- Microsoft.VisualStudio.Workload.VCTools
- Microsoft.VisualStudio.Component.Windows10SDK.19041
- Microsoft.VisualStudio.Component.VC.Tools.x86.x64
> Build py-spy and gprofiler executables with provided build script
4. Assume that the following prerequisites are available on target machine:
- Python (>=3.9.13)
- Pip (>=22.2.2)
- Git bash (https://community.chocolatey.org/packages/git)
- Wget
> Using Git, clone py-spy source with appropriate tag and version
> Using wget, download and install:
- Rust compiler
- VS Build Tools executable
> Using VS Build Tools executable, install the following components:
- Microsoft.VisualStudio.Workload.VCTools
- Microsoft.VisualStudio.Component.Windows10SDK.19041
- Microsoft.VisualStudio.Component.VC.Tools.x86.x64
> Build py-spy and gprofiler executables with provided build script

.gitmodules
teoluwaf: See note on dependencies recommendations

gprofiler/profilers/python.py
teoluwaf: Done

build-pyspy.bat
teoluwaf: See dependencies recommendation below

> ### **Dependencies**
> > **Options**
> 
> 1. [Recommended]: Provide prebuilt executable for py-spy on Windows and build gprofiler executable with provided build script. You control the build environment for the py-spy and the correct tags
> 2. Assume that the following prerequisites are available on target machine:
> 
> * Python (>=3.9.13)
> * Pip (>=22.2.2)
> * Rust Compiler
> * py-spy source with appropriate tag and version
> * VS Build Tools executable
> 
> > Using VS Build Tools install following components
> 
> * Microsoft.VisualStudio.Workload.VCTools
> * Microsoft.VisualStudio.Component.Windows10SDK.19041
> * Microsoft.VisualStudio.Component.VC.Tools.x86.x64
> 
> > Build py-spy and gprofiler executables with provided build script
> 
> 3. Assume that the following prerequisites are available on target machine:
> 
> * Python (>=3.9.13)
> * Pip (>=22.2.2)
> * py-spy source with appropriate tag and version
> * Wget
> 
> > Using wget download and install:
> 
> * Rust compiler
> * VS Build Tools executable
> 
> > Using VS Build Tools executable, install the following components:
> 
> * Microsoft.VisualStudio.Workload.VCTools
> * Microsoft.VisualStudio.Component.Windows10SDK.19041
> * Microsoft.VisualStudio.Component.VC.Tools.x86.x64
> 
> > Build py-spy and gprofiler executables with provided build script
> 
> 4. Assume that the following prerequisites are available on target machine:
> 
> * Python (>=3.9.13)
> * Pip (>=22.2.2)
> * Git bash (https://community.chocolatey.org/packages/git)
> * Wget
> 
> > Using Git, clone py-spy source with appropriate tag and version
> > Using wget, download and install:
> 
> * Rust compiler
> * VS Build Tools executable
> 
> > Using VS Build Tools executable, install the following components:
> 
> * Microsoft.VisualStudio.Workload.VCTools
> * Microsoft.VisualStudio.Component.Windows10SDK.19041
> * Microsoft.VisualStudio.Component.VC.Tools.x86.x64
> 
> > Build py-spy and gprofiler executables with provided build script

gprofiler/profilers/python.py
teoluwaf: Done

gprofiler/profilers/php.py
teoluwaf: Done

exe-requirements.txt
teoluwaf: Done

gprofiler/profilers/php.py
teoluwaf: Done

build-pyspy.bat
teoluwaf: Done

build.bat
teoluwaf: I had to re-create granulate-utils because the requirements.txt specifies a relative path for granulate-utils as follows. If it isn't at that location it fails. Also, I tried to put all build artifacts in the app directory so as to leave the source directory untouched. Doing so, doesn't permit me to use the relative path defined below in the requirements.txt file without making a copy of granulate-utils. I hope that makes sense

./granulate-utils/

gprofiler/profilers/python.py
teoluwaf: Done

gprofiler/profilers/factory.py
```diff
@@ -30,9 +31,14 @@ def get_profilers(
...
-            logger.warning(f"Disabling {profiler_name} because it doesn't support this architecture ({arch})")
-            continue
+        if is_windows():
+            if arch not in profiler_config.supported_windows_archs:
+                logger.warning(f"Disabling {profiler_name} because it doesn't support this architechture ({arch})")
+                continue
+        else:
+            if arch not in profiler_config.supported_archs:
+                logger.warning(f"Disabling {profiler_name} because it doesn't support this architecture ({arch})")
+                continue
```
Jongy, line 41:
```suggestion
        if arch not in (profiler_config.supported_windows_archs if is_windows() else profiler_config.supported_archs):
            logger.warning(f"Disabling {profiler_name} because it doesn't support this architechture ({arch})")
            continue
```

gprofiler/profilers/python.py
Jongy: We've since completed this split. Resolving this

gprofiler/profilers/__init__.py
Jongy: Was split out, resolving

build-pyspy.bat
teoluwaf: I made some changes to search for advapi32.lib and link.exe instead of hardcoding the path

build-pyspy.bat
teoluwaf: Done

Jongy:
@teoluwaf - regarding dependencies, as we discussed, I was referring to build-machine dependencies. Run-time dependencies should be none - only the output executable from PyInstaller.

Now, going over the depedencies in https://github.com/Granulate/gprofiler/pull/442#issuecomment-1289207125, let's assume that the build machine has:
1. Python installed (3.6+)
2. wget
3. Git
(So this is roughly approach 4 listed in your comment)

Using wget you can install pip if it's missing, and then install the VS SDK tools & Rust. Then, clone py-spy from our fork and build it, and finally build gProfiler with PyInstaller.
About py-spy - please remove the submodule of it, and add logic to clone the exact commit like we have in `pyspy_build.sh`. It's best if we don't need to maintain 2 sites that have the tag & git revision - so one solution is to move out the `git clone` command to a file that can be invoked both in Linux & Windows (IDK if the `&&` syntax used in the `git clone --depth 1 -b v0.3.10g1 https://github.com/Granulate/py-spy.git && git -C py-spy reset --hard 480deec8b5dde3cd331d1a793106981c1796d172` command can work on Batch?) or otherwise, store the hash & tag in files: `scripts/pyspy_tag` and `scripts/pyspy_revision`. Then the Linux command becomes `git clone --depth 1 -b $(cat pyspy_tag) https://github.com/Granulate/py-spy.git && git -C py-spy reset --hard $(cat pyspy_revision)` and a similar one in Windows, and we just have to update the 2 files).

Jongy:
The existing `dep/` directory can be removed

build.bat
Jongy: Addressed it. You can install pip with `get-pip.py` on the build machine, for MVP also okay to assume pip is installed.

build-pyspy.bat
Jongy: Not pushed?

.gitmodules
Jongy: Closing the discussion here, we continue in comments

build-pyspy.bat
Jongy: Closing the discussion here, we continue in comments

build-pyspy.bat
teoluwaf: It should be pushed now

build-pyspy.bat
teoluwaf: Done

teoluwaf:
> @teoluwaf - regarding dependencies, as we discussed, I was referring to build-machine dependencies. Run-time dependencies should be none - only the output executable from PyInstaller.
> 
> Now, going over the depedencies in [#442 (comment)](https://github.com/Granulate/gprofiler/pull/442#issuecomment-1289207125), let's assume that the build machine has:
> 
> 1. Python installed (3.6+)
> 2. wget
> 3. Git
> 
> (So this is roughly approach 4 listed in your comment)
> 
> Using wget you can install pip if it's missing, and then install the VS SDK tools & Rust. Then, clone py-spy from our fork and build it, and finally build gProfiler with PyInstaller. About py-spy - please remove the submodule of it, and add logic to clone the exact commit like we have in `pyspy_build.sh`. It's best if we don't need to maintain 2 sites that have the tag & git revision - so one solution is to move out the `git clone` command to a file that can be invoked both in Linux & Windows (IDK if the `&&` syntax used in the `git clone --depth 1 -b v0.3.10g1 https://github.com/Granulate/py-spy.git && git -C py-spy reset --hard 480deec8b5dde3cd331d1a793106981c1796d172` command can work on Batch?) or otherwise, store the hash & tag in files: `scripts/pyspy_tag` and `scripts/pyspy_revision`. Then the Linux command becomes `git clone --depth 1 -b $(cat pyspy_tag) https://github.com/Granulate/py-spy.git && git -C py-spy reset --hard $(cat pyspy_revision)` and a similar one in Windows, and we just have to update the 2 files. That's what I suggested [here](https://github.com/Granulate/gprofiler/pull/442#discussion_r1002794554)).

Done

teoluwaf:
> The existing `dep/` directory can be removed

Done

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "granulate-utils"]
 	path = granulate-utils
-	url = git@github.com:Granulate/granulate-utils.git
+	url  = https://github.com/Granulate/granulate-utils.git
```
Jongy, line 3: Revert please. By the way if `git submodule update` with the SSH URL fails for you, configuring Git SSH access will solve it.

exe-requirements.txt
```diff
@@ -2,7 +2,7 @@
 pyinstaller==4.0; platform.machine == "x86_64"
 # aarch64 requires a later version due to the use of a newer centos version.
 # see https://github.com/pyinstaller/pyinstaller/issues/5540
-pyinstaller==4.10; platform.machine == "aarch64"
+pyinstaller==4.10; platform.machine == "aarch64" or sys.platform == "win32"
```
Jongy, line 5: Line 2 is `pyinstaller==4.0; platform.machine == "x86_64"`, do we need to add `sys.platform == "linux"` to it? (otherwise Windows x86_64 matches both line 2 and line 5)

build-pyspy.bat
Jongy: Scripts are still under `scripts/` and not `scripts/windows/` now
image

gprofiler/profilers/factory.py
```diff
@@ -30,8 +31,8 @@ def get_profilers(
         if profiler_mode in ("none", "disabled"):
             continue
 
-        if arch not in profiler_config.supported_archs:
-            logger.warning(f"Disabling {profiler_name} because it doesn't support this architecture ({arch})")
+        if arch not in (profiler_config.supported_windows_archs if is_windows() else profiler_config.supported_archs):
+            logger.warning(f"Disabling {profiler_name} because it doesn't support this architechture ({arch})")
```
Jongy, line 35:
```suggestion
            logger.warning(f"Disabling {profiler_name} because it doesn't support this architecture ({arch})")
```

scripts/build-pyspy.bat
```diff
@@ -0,0 +1,63 @@
+@echo off
+ECHO Y | RMDIR /S py-spy
+
+git clone --depth 1 -b v0.3.10g1 https://github.com/Granulate/py-spy.git && git -C py-spy reset --hard 480deec8b5dde3cd331d1a793106981c1796d172
```
Jongy, line 4: The tag & revision are cloned here. Over time they will deviate from those in `pyspy_build.sh`. Dependencies need to be managed in a single site, not duplicated. Let's do what I suggested here:
> store the hash & tag in files: scripts/pyspy_tag and scripts/pyspy_revision. Then the Linux command becomes git clone --depth 1 -b $(cat pyspy_tag) https://github.com/Granulate/py-spy.git && git -C py-spy reset --hard $(cat pyspy_revision) and a similar one in Windows, and we just have to update the 2 files.

build.bat
Jongy: I see. I don't think it matters much if the source directory remains untouched - after all, it's just a build machine which runs it, so I think we can do the same as done in the Linux build scripts (i.e install everything in-place). Up to you

Jongy:
I ran it on Windows and got the message:
```
Please run me in the init PID namespace! In Docker, make sure you pass '--pid=host'. In Kubernetes, add 'hostPID: true' in the Pod spec.
```

make sure the check only runs on Linux? It is irrelevant on Windows.

Jongy:
Another Windows error:
```
[2022-11-05 00:55:05,125] ERROR: gprofiler: Failed to write pid to '/var/run/gprofiler.pid', continuing anyway
  File "gprofiler\main.py", line 753, in main
    init_pid_file(args.pid_file)
  File "gprofiler\main.py", line 729, in init_pid_file
    Path(pid_file).write_text(str(os.getpid()))
  File "pathlib.py", line 1255, in write_text
  File "pathlib.py", line 1222, in open
  File "pathlib.py", line 1078, in _opener
FileNotFoundError: [Errno 2] No such file or directory: '\\var\\run\\gprofiler.pid'
```

We can skip the pidfile feature for Windows as well - it's not needed here

Jongy:
3rd Windows error:
```
[2022-11-05 00:55:18,267] DEBUG: gprofiler.utils: Running command: (C:\Users\ADMINI~1\AppData\Local\Temp\2\_MEI69042\gprofiler\resources\burn convert --type=folded)
[2022-11-05 00:55:18,269] INFO: gprofiler: Stopping ...
[2022-11-05 00:55:18,269] ERROR: gprofiler: Unexpected error occurred
  File "gprofiler\main.py", line 858, in main
    gprofiler.run_single()
  File "gprofiler\main.py", line 360, in run_single
    self._snapshot()
  File "gprofiler\main.py", line 317, in _snapshot
    self._generate_output_files(merged_result, local_start_time, local_end_time)
  File "gprofiler\main.py", line 204, in _generate_output_files
    run_process(
  File "gprofiler\utils\__init__.py", line 233, in run_process
    with start_process(cmd, via_staticx, **kwargs) as process:
  File "gprofiler\utils\__init__.py", line 148, in start_process
    popen = Popen(
  File "subprocess.py", line 858, in __init__
  File "subprocess.py", line 1311, in _execute_child
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

We try to run `burn` to convert the collapsed file to a flamegraph. I don't want to invest time in building it for Windows in MVP (IDK if it even supports Windows). This feature can be disabled if the flag `--no-flamegraph` is passed; let's add `if is_windows(): args.no_flamegraph = True` in `parse_args()` so this is never invoked in Windows

Jongy:
I tried running it in `-u` mode (upload results) and got this error:
```
[2022-11-05 14:55:37,355] ERROR: gprofiler: Unexpected error occurred
  File "gprofiler\main.py", line 805, in main
  File "gprofiler\metadata\system_metadata.py", line 283, in get_hostname
AssertionError: hostname not initialized!
```

build.bat
teoluwaf: I'll keep it the way it is for now as it helps to be able to clean up between runs.

.gitmodules
teoluwaf: Done

teoluwaf:
> I tried running it in `-u` mode (upload results) and got this error:
> 
> ```
> [2022-11-05 14:55:37,355] ERROR: gprofiler: Unexpected error occurred
> Traceback (most recent call last):
>   File "gprofiler\main.py", line 805, in main
>   File "gprofiler\metadata\system_metadata.py", line 283, in get_hostname
> AssertionError: hostname not initialized!
> ```

Fixed

scripts/windows/commit.txt
```diff
@@ -0,0 +1 @@
+480deec8b5dde3cd331d1a793106981c1796d172
```
Jongy, line 1: My intention is to have this hash shared with `pyspy_build.sh`. The PR doesn't do it - it's still cloned. Please modify `pyspy_build.sh` as well, to use the commit & tag files. You can place both in `scripts/pyspy_commit` and `scripts/pyspy_tag`.

gprofiler/metadata/system_metadata.py
```diff
@@ -228,7 +231,8 @@ def get_windows_network_details() -> Tuple[str, str]:
 
 def get_static_system_info() -> SystemInfo:
     if is_windows():
-        hostname = platform.node()
+        global hostname
```
Jongy, line 234: If you add `global` here - it affects the entire function which is not intended for the second branch of Linux.

let's move the Windows code to a function `_initialize_system_info_windows`, defined under `if is_windows()`?

gprofiler/main.py
```diff
@@ -58,11 +58,11 @@
 
 logger: logging.LoggerAdapter
 
-DEFAULT_LOG_FILE = "/var/log/gprofiler/gprofiler.log"
+DEFAULT_LOG_FILE = "/var/log/gprofiler/gprofiler.log" if is_linux() else "./gprofiler.log"
 DEFAULT_LOG_MAX_SIZE = 1024 * 1024 * 5
 DEFAULT_LOG_BACKUP_COUNT = 1
 
-DEFAULT_PID_FILE = "/var/run/gprofiler.pid"
+DEFAULT_PID_FILE = "/var/run/gprofiler.pid" if is_linux() else "./gprofiler.pid"
```
Jongy, line 65: Indeed don't need a PID file in Windows, so we can remove this `if`

gprofiler/main.py
```diff
@@ -609,8 +610,8 @@ def parse_cmd_args() -> configargparse.Namespace:
 
     args = parser.parse_args()
 
-    args.perf_inject = args.nodejs_mode == "perf"
-    args.perf_node_attach = args.nodejs_mode == "attach-maps"
+    args.perf_inject = args.nodejs_mode == "perf" and is_linux()
```
Jongy, line 613: having many different branches with `is_linux()` makes it harder to read the function. I prefer if all "windows" changing are done in one site under a single `if`.

So, after `parse_args` you can set `args.flamegraph = False` because that's not supported as we saw.
As for the other perf arguments - do we really need that? I figured just `perf_mode = "disabled"` would be enough? Or maybe even that's not required because the `SystemProfiler` doesn't have any `supported_windows_archs` so it is never selected?

pyi.Dockerfile
```diff
@@ -26,6 +26,8 @@ RUN ./prepare_machine-unknown-linux-musl.sh
 FROM pyspy-rbspy-builder-common AS pyspy-builder
 WORKDIR /tmp
 COPY scripts/pyspy_build.sh .
+COPY scripts/pyspy_commit.txt .
```
Jongy, line 29: Note that this also needs to be added in `Dockerfile`

Jongy:
@teoluwaf the PR updated granulate-utils - please revert that, I figure it was unintentional

gprofiler/metadata/system_metadata.py
```diff
@@ -281,6 +281,35 @@ def get_hostname() -> str:
     return hostname
 
 
+def _initialize_system_info_windows() -> Any:
+
```
Jongy, line 285:
```suggestion
```

gprofiler/main.py
Jongy: Looks better now :)

Jongy (approved):

PR #445, opened 2022-08-25
Change java metadata['java_version'] to parsed JvmVersion

uriariel:
Change java metadata['java_version'] to parsed JvmVersion

## Description
Change java metadata['java_version'] to parsed JvmVersion

## Checklist:

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -242,7 +242,7 @@ def get_java_version_logged(process: Process, stop_event: Event) -> str:
 class JavaMetadata(ApplicationMetadata):
     def make_application_metadata(self, process: Process) -> Dict[str, Any]:
         if is_java_basename(process):
-            version = get_java_version(process, self._stop_event)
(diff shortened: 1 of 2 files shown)


Jongy:
The rationale in @uriariel 's change is to allow easy access to parsed `JvmVersion`s in our backend.

What makes more sense instead is to move the parsing logic to granulate-utils which makes it usable by other components:
1. To avoid changing the protocol format (existing users of `java_version` in the metadata, and the need to accommodate old gProfiler versions which still send it the old way)
2. `parse_jvm_version` has changed multiple times. If we keep gProfiler sending the raw data, then once we update the code we have the most recent parsing available (instead of storing parsed data, which means we could store data that was parsed incorrectly, in case we fix the code).

TL;DR @uriariel we will move the code to granulate-utils if we need access to `JvmVersion`.

Jongy:
Superseded by https://github.com/Granulate/granulate-utils/pull/65

PR #447, opened 2022-08-26
tests: Use openjdk images repository instead of java

Jongy:
It seems like "java" was removed, I get:
"Error response from daemon: manifest for java:latest not found: manifest unknown: manifest unknown"

Also, heading to https://hub.docker.com/_/java redirects you to https://hub.docker.com/_/openjdk so I suppose that's the official successor.

### DIFF
--- a/tests/containers/java/musl.Dockerfile
+++ b/tests/containers/java/musl.Dockerfile
@@ -1,5 +1,5 @@
 # see test_java_async_profiler_musl_and_cpu
-FROM openjdk:alpine
+FROM openjdk@sha256:d49bf8c44670834d3dade17f8b84d709e7db47f1887f671a0e098bafa9bae49f
 
 WORKDIR /app
 ADD Fibonacci.java /app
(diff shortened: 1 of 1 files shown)


Jongy:
The existing test runs `assert is_musl(psutil.Process(application_pid))` so we can be sure the new image is musl-based as well.

Jongy:
The old image (which I had on my box, luckily) gives:
```
openjdk version "1.8.0_111-internal"
OpenJDK Runtime Environment (build 1.8.0_111-internal-alpine-r0-b14)
OpenJDK 64-Bit Server VM (build 25.111-b14, mixed mode)
```
The `openjdk:alpine` one is a much newer build and it does use libgcc + libsrdc++ so the test that ensures they don't get loaded doesn't pass. I do want to test it in case we continue to encounter the older build.

Luckily, pulling by the hash I had on my box seems to work... so I changed it by hash and it gives the old image we had at `java:alpine` :)

Jongy:
Flaky tests - I created https://github.com/Granulate/gprofiler/issues/448 which describes one problem there, but the core issue appears to be flakiness in the `get_stack_offset` program.

michelhe (approved):
dockerhub is so unreliable :/

PR #451, opened 2022-08-29
Make NodeJS profiling attachable

kwisniewski98:
## Description
Adds `--nodejs-mode=attachable`, which injects DSO into already running NodeJS processes. 

## Related Issue
https://github.com/Granulate/gprofiler/issues/418

## Motivation and Context
At current state, resolving function addresses in NodeJS requires restarting each process, which might be very hard if not impossible in production. This features allows generating perf maps at runtime.

## How Has This Been Tested?
Tested manually in container and as executable. NodeJS targets were tested in both same namespace as gprofiler and in separate namespace (docker container)

## Checklist:


Dockerfile
```diff
@@ -1,5 +1,4 @@
 # these need to be defined before any FROM - otherwise, the ARGs expand to empty strings.
-
```
Jongy, line 2: Bring it back :)

Dockerfile
```diff
@@ -136,6 +139,23 @@ COPY scripts/async_profiler_build_shared.sh .
 COPY scripts/async_profiler_build_musl.sh .
 RUN ./async_profiler_build_shared.sh /tmp/async_profiler_build_musl.sh
 
+# node-package-builder-musl
+FROM alpine${NODE_PACKAGE_BUILDER_MUSL} AS node-package-builder-musl
+WORKDIR /tmp
+COPY scripts/build_node_package.sh .
```
Jongy, line 145: Place it right before the `RUN ./build_node_package.sh` - this `COPY` stage is very cheap, while the next `RUN` stage is not, as it installs many packages. If we change the script, we don't need the `apk add` to run again.

Dockerfile
```diff
@@ -136,6 +139,23 @@ COPY scripts/async_profiler_build_shared.sh .
...
+RUN apk add --no-cache curl g++ python3 make gcc git bash nodejs npm
+RUN ./build_node_package.sh
+
+# node-package-builder-glibc
+FROM ubuntu${NODE_PACKAGE_BUILDER_GLIBC} AS node-package-builder-glibc
+WORKDIR /tmp
+COPY scripts/build_node_package.sh .
```
Jongy, line 152: Same as I commented for the alpine part

Dockerfile
```diff
@@ -136,6 +139,23 @@ COPY scripts/async_profiler_build_shared.sh .
...
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt update -y && apt install -y curl g++ python3 make gcc git
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
```
Jongy, line 155: Btw why do we need this here but not for the alpine? Because the alpine has recent `nodejs` package, while the ubuntu one does not?

gprofiler/main.py
```diff
@@ -615,8 +617,8 @@ def parse_cmd_args() -> configargparse.Namespace:
     if args.perf_mode in ("dwarf", "smart") and args.frequency > 100:
         parser.error("--profiling-frequency|-f can't be larger than 100 when using --perf-mode 'smart' or 'dwarf'")
 
-    if args.nodejs_mode == "perf" and args.perf_mode not in ("fp", "smart"):
-        parser.error("--nodejs-mode perf requires --perf-mode 'fp' or 'smart'")
+    if (args.nodejs_mode == "perf" or args.nodejs_mode == "attachable") and args.perf_mode not in ("fp", "smart"):
```
Jongy, line 620:
```suggestion
    if args.nodejs_mode in ("perf", "attachable") and args.perf_mode not in ("fp", "smart"):
```

gprofiler/metadata/application_metadata.py
```diff
@@ -4,16 +4,15 @@
 #
 
 import functools
-from subprocess import CompletedProcess
 from threading import Event, Lock
 from typing import Any, Dict, Optional
+from . import get_exe_version
```
Jongy, line 9:
```suggestion
from gprofiler.metadata import get_exe_version
```
Prefer absolute imoports - that's our convention

gprofiler/profilers/perf.py
```diff
@@ -203,14 +205,18 @@ def __init__(
             self._perfs.append(self._perf_dwarf)
         else:
             self._perf_dwarf = None
-
```
Jongy, line 206: Bring it back

gprofiler/profilers/perf.py
```diff
@@ -203,14 +205,18 @@ def __init__(
...
+        self.perf_node_attach = perf_node_attach
         assert self._perf_fp is not None or self._perf_dwarf is not None
 
     def start(self) -> None:
+        if self.perf_node_attach:
```
Jongy, line 212: This runs only once. If new nodejs processes spawn during the lifetime of gProfiler, we want them to get the DSO loaded as well. I suggest you can `generate_map_for_node_processes()` upon every `snapshot()` call?

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+import shutil
+import traceback
+from typing import Callable
+from gprofiler.metadata import get_exe_version
+from gprofiler.log import get_logger_adapter
+from granulate_utils.linux.ns import get_proc_root_path, is_same_ns, NsType, run_in_ns
+from granulate_utils.linux.process import is_musl
+
+
+from . import add_permission_dir, pgrep_maps, resource_path, pgrep_exe
```
Jongy, line 11: Absolute imports please :)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+
+import json
```
Jongy, line 13: Please run `lint.sh` - it runs `black`, `isort` and other linters. Specifically here I can see that `isort` would reorder things.

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+import requests
+import os
+import signal
+import psutil
+import stat
+
+from retry import retry
```
Jongy, line 20: This part is from your previous PR https://github.com/Granulate/gprofiler/pull/438, so this PR supersedes it completely? If so, you can close that other PR.

Jongy: Actually, since I have already began reviewing it, maybe it's best if we complete it and then base this PR on top of it? To avoid losing the comments.

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+    dest = os.path.join(proc_root, dest_inside_container)
+    if os.path.exists(dest):
+        return "/" + dest_inside_container
+    src = resource_path(os.path.join("node", "module", libc, version))
+    shutil.copytree(src, dest)
+    add_permission_dir(dest, stat.S_IROTH, stat.S_IXOTH | stat.S_IROTH)
+    return "/" + dest_inside_container
+
+
+def _create_symlink_for_map(process: psutil.Process, pid_inside_ns: int):
```
Jongy, line 133: Is this required? As far as I recall, `perf` is smart enough to look for `/tmp/perf-{pid}.map` file with the NSpid, not the host PID.

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+    proc_root = get_proc_root_path(process)
+    map_file_name = f"perf-{process.pid}.map"
+    src = os.path.join(proc_root, "tmp", f"perf-{pid_inside_ns}.map")
+    dest = os.path.join("/tmp", map_file_name)
+    os.symlink(src, dest)
+
+
+@lru_cache(1024)
+def _get_pid_inside_ns(pid: int) -> int:
```
Jongy, line 142: Use `get_process_nspid` (in `granulate-utils`)

pyi.Dockerfile
```diff
@@ -80,6 +82,23 @@ WORKDIR /tmp
 COPY scripts/burn_build.sh .
 RUN ./burn_build.sh
 
+# node-package-builder-musl
+FROM alpine${NODE_PACKAGE_BUILDER_MUSL} AS node-package-builder-musl
+WORKDIR /tmp
+RUN apk add --no-cache curl g++ python3 make gcc git bash nodejs npm
```
Jongy, line 88: If you find yourself repeating these - prefer to put them in a script `node_builder_musl_env.sh` that's used in `Dockerfile` and `pyi.Dockerfile`

pyi.Dockerfile
```diff
@@ -80,6 +82,23 @@ WORKDIR /tmp
...
+COPY scripts/build_node_package.sh .
+RUN ./build_node_package.sh
+
+# node-package-builder-glibc
+FROM ubuntu${NODE_PACKAGE_BUILDER_GLIBC} AS node-package-builder-glibc
+WORKDIR /tmp
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt update -y && apt install -y curl g++ python3 make gcc git
```
Jongy, line 96: Same as I commented for the alpine part - share it with the `Dockerfile` defintions by creating an `env` script.

scripts/build_aarch64_executable.sh
```diff
@@ -18,6 +18,7 @@ GOLANG_VERSION=@sha256:f7d3519759ba6988a2b73b5874b17c5958ac7d0aa48a8b1d84d66ef25
 # alpine 3.14.2
 ALPINE_VERSION=@sha256:b06a5cf61b2956088722c4f1b9a6f71dfe95f0b1fe285d44195452b8a1627de7
 
+
```
Jongy, line 21:
```suggestion
```

scripts/build_node_package.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+MODULE_PATH=/tmp/module
+BUILD_TARGET_DIR=/tmp/module_build
+
+git clone https://github.com/mmarchini-oss/node-linux-perf.git $MODULE_PATH
```
Jongy, line 8: Pin the hash by running `git reset --hard ..` to the revision that you expect (can be the latest `master`)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+                logger.debug(cp.stdout.decode().strip())
+            if hasattr(cp, "stderr"):
+                logger.debug(cp.stderr.decode().strip())
+            if not is_same_ns(process, NsType.mnt.name):
+                os.remove(os.path.join("/tmp", f"perf-{process.pid}.map"))
+        except Exception as e:
+            logger.warn(
+                f"Could not clean up debug symbols for pid {process.pid}. Reason: {e}"
+            )
+            logger.debug(traceback.format_exc())
```
Jongy, line 223:
```suggestion
            logger.warn(
                f"Could not clean up debug symbols for pid {process.pid}", exc_info=True
            )
```

Jongy: Same for that construct in `generate_map_for_node_processes`

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+        debugger_url = _get_debugger_url()
+        sock = create_connection(debugger_url)
+        _stop_dso(sock, module_path)
+        shutil.rmtree(os.path.join("/tmp", "node_module"))
+        os.remove(os.path.join("/tmp", f"perf-{pid}.map"))
+
+    return clean_up
+
+
+def generate_map_for_node_processes():
```
Jongy, line 169: As I commented earlier, this function will run multiple times in gProfiler's lifetime. Therefore it needs to accommodate with processes that are already loaded - I suggest that before attaching, you check if the DSO exists in the target process' maps. Similar to `_check_async_profiler_loaded`.

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+        "result" not in message.keys()
+        or "result" not in message["result"].keys()
+        or "type" not in message["result"]["result"].keys()
+        or message["result"]["result"]["type"] != "boolean"
+    ):
+        raise NodeDebuggerUnexpectedResponse(message)
+
+
+def _copy_module_into_process_ns(process: psutil.Process, musl: bool, version: str):
+    proc_root = get_proc_root_path(process)
```
Jongy, line 121: Use `resolve_proc_root_links` to resolve the actualy path. If `/tmp` is an absolute symlink in the container, your current code won't work - the path generated in line 124 will eventually point the host. `resolve_proc_root_links` handles that for you

Jongy (changes requested):
Looks very good @kwisniewski98 !

I left some comments (did not complete reviewing, will continue this week).

Please add a test like `test_nodejs` that uses the new attach mode. Also, the target nodejs app should not be started with the jitdump flags like we do now - we can add another Dockerfile, or just control the extra flags passed to node via a parameter when starting the container (so the existing `test_nodejs` can pass the flags; and your new test won't pass them)

Dockerfile
kwisniewski98: Ubuntu installs node 10, which is too old for node-gyp. In case of alpine default one is 14.

gprofiler/utils/node.py
kwisniewski98: All changes suggested by you are in this PR, so I think we can just close previous one

Jongy: Okie

Jongy: I will follow the comments from that PR as well.

Jongy:
I ran the CI

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+
+@lru_cache(1024)
+def _get_pid_inside_ns(pid: int) -> int:
+    with open(f"/proc/{pid}/status") as f:
+        for line in f.readlines():
+            if line.startswith("NSpid"):
+                return line.split("\t")[-1].strip("\n")
+
+
+def _generate_perf_map_wrapper(module_path: str) -> Callable:
```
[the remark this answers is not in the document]
Jongy: Same for other wrappers

gprofiler/utils/node.py
Jongy, line 133: `-> None:` type

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+    if (
+        "result" not in message.keys()
+        or "result" not in message["result"].keys()
+        or "type" not in message["result"]["result"].keys()
+        or message["result"]["result"]["type"] != "boolean"
+    ):
+        raise NodeDebuggerUnexpectedResponse(message)
+
+
+def _copy_module_into_process_ns(process: psutil.Process, musl: bool, version: str):
```
Jongy, line 120:
```suggestion
def _copy_module_into_process_ns(process: psutil.Process, musl: bool, version: str) -> str:
```

same for all others

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+        message = json.loads(message)
+    except json.JSONDecodeError:
+        raise NodeDebuggerUnexpectedResponse(message)
+
+    if (
+        "result" not in message.keys()
+        or "result" not in message["result"].keys()
+        or "type" not in message["result"]["result"].keys()
+        or message["result"]["result"]["type"] != "boolean"
+    ):
```
Jongy, line 116: This repeats itself - extract to a function `validate_response`?

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+
+class NodeDebuggerUrlNotFound(Exception):
+    pass
+
+
+class NodeDebuggerUnexpectedResponse(Exception):
+    pass
+
+
+def _start_debugger(pid):
```
Jongy, line 35: Typnig

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+
+@retry(NodeDebuggerUrlNotFound, 5, 1)
+def _get_debugger_url() -> str:
+    # when killing process with SIGUSR1 it will open new debugger session on port 9229,
+    # so it will always the same
+    port = 9229
+    debugger_url_response = requests.get(f"http://127.0.0.1:{port}/json/list")
+    if (
+        debugger_url_response.status_code != 200
+        or not "application/json" in debugger_url_response.headers.get("Content-Type")
```
Jongy, line 51: Copying from https://github.com/Granulate/gprofiler/pull/438#discussion_r944920950

> If Content-Type doesn't exist then you have "application/json" in None which will raise

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+class NodeDebuggerUnexpectedResponse(Exception):
+    pass
+
+
+def _start_debugger(pid):
+    os.kill(pid, signal.SIGUSR1)
+
+
+def get_node_processes():
+    return pgrep_maps(r"(?:^.+/node[^/]*$)")
```
Jongy, line 40: I think for `node` you could just use `pgrep_exe("node")`. The idea of `pgrep_maps` is for Java/Python where the exe is not necessarily `java` or `python` but can be e.g `uwsgi` and load `libpython.so`. In node - it's always `node` binary.

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+        raise NodeDebuggerUrlNotFound
+
+    response_json = debugger_url_response.json()
+    if (
+        not isinstance(response_json, list)
+        or len(response_json) == 0
+        or not isinstance(response_json[0], dict)
+        or not "webSocketDebuggerUrl" in response_json[0].keys()
+    ):
+        raise NodeDebuggerUrlNotFound
```
Jongy, line 62: Include the response JSON in the exception? (so that we can debug it)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+        or not "application/json" in debugger_url_response.headers.get("Content-Type")
+    ):
+        raise NodeDebuggerUrlNotFound
+
+    response_json = debugger_url_response.json()
+    if (
+        not isinstance(response_json, list)
+        or len(response_json) == 0
+        or not isinstance(response_json[0], dict)
+        or not "webSocketDebuggerUrl" in response_json[0].keys()
```
Jongy, line 60:
```suggestion
        or "webSocketDebuggerUrl" not in response_json[0]
```

pyi.Dockerfile
```diff
@@ -237,6 +256,8 @@ COPY --from=async-profiler-builder-glibc /tmp/async-profiler/build/async-profile
 COPY --from=async-profiler-builder-glibc /tmp/async-profiler/build/libasyncProfiler.so gprofiler/resources/java/glibc/libasyncProfiler.so
 COPY --from=async-profiler-builder-musl /tmp/async-profiler/build/libasyncProfiler.so gprofiler/resources/java/musl/libasyncProfiler.so
 COPY --from=async-profiler-builder-glibc /tmp/async-profiler/build/fdtransfer gprofiler/resources/java/fdtransfer
+COPY --from=node-package-builder-musl /tmp/module_build gprofiler/resources/node/module/musl
```
Jongy, line 259:
```
root@e050c04ca431:/app# find gprofiler/resources/node/
gprofiler/resources/node/
gprofiler/resources/node/module
gprofiler/resources/node/module/musl
gprofiler/resources/node/module/musl/11
gprofiler/resources/node/module/musl/11/linux-perf.js
gprofiler/resources/node/module/musl/11/binding.gyp
gprofiler/resources/node/module/musl/11/test
gprofiler/resources/node/module/musl/11/test/test-linux-perf-start.js
gprofiler/resources/node/module/musl/11/test/test-linux-perf-stop.js
gprofiler/resources/node/module/musl/11/test/test-linux-perf-restart.js
gprofiler/resources/node/module/musl/11/linux-perf.h
gprofiler/resources/node/module/musl/11/test.sh
gprofiler/resources/node/module/musl/11/build
gprofiler/resources/node/module/musl/11/build/Release
gprofiler/resources/node/module/musl/11/build/Release/linux-perf.node
gprofiler/resources/node/module/musl/11/build/Release/.deps
gprofiler/resources/node/module/musl/11/build/Release/.deps/Release
gprofiler/resources/node/module/musl/11/build/Release/.deps/Release/obj.target
gprofiler/resources/node/module/musl/11/build/Release/.deps/Release/obj.target/linux-perf
gprofiler/resources/node/module/musl/11/build/Release/.deps/Release/obj.target/linux-perf/linux-perf.o.d
gprofiler/resources/node/module/musl/11/build/Release/.deps/Release/obj.target/linux-perf/linux-perf-listener.o.d
gprofiler/resources/node/module/musl/11/build/Release/.deps/Release/obj.target/linux-perf.node.d
gprofiler/resources/node/module/musl/11/build/Release/.deps/Release/linux-perf.node.d
gprofiler/resources/node/module/musl/11/build/Release/obj.target
gprofiler/resources/node/module/musl/11/build/Release/obj.target/linux-perf.node
gprofiler/resources/node/module/musl/11/build/Release/obj.target/linux-perf
gprofiler/resources/node/module/musl/11/build/Release/obj.target/linux-perf/linux-perf.o
gprofiler/resources/node/module/musl/11/build/Release/obj.target/linux-perf/linux-perf-listener.o
gprofiler/resources/node/module/musl/11/build/config.gypi
```

Many files get copied inside - copy only the ones actually necessary in runtime?

Dockerfile
Jongy: I see, please comment on the Dockerfile :)

scripts/build_node_package.sh
```diff
@@ -0,0 +1,35 @@
...
+npm install -g node-gyp
+
+cd $MODULE_PATH
+curl -L https://github.com/nodejs/nan/archive/refs/tags/v2.16.0.tar.gz -o nan.tar.gz
+tar -vxzf nan.tar.gz -C /tmp
+export NAN_PATH=$(realpath /tmp/nan-*)
+sed -i 's/node \-e \\"require('\''nan'\'')\\"/echo $NAN_PATH/g' binding.gyp
```
Jongy, line 15: What's this change here?

gprofiler/main.py
```diff
@@ -431,9 +431,10 @@ def parse_cmd_args() -> configargparse.Namespace:
         "--nodejs-mode",
         dest="nodejs_mode",
         default="disabled",
-        choices=["perf", "disabled", "none"],
-        help="Select the NodeJS profiling mode: perf (run 'perf inject --jit' on perf results, to augment them"
-        " with jitdump files of NodeJS processes, if present) or disabled (no runtime-specific profilers for NodeJS)",
+        choices=["attachable", "perf", "disabled", "none"],
```
Jongy, line 434: Let's call it `attach` not `attachable`, and maybe `attach-maps` (and later we'll add `attach-jit` for jitdumps)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+
+    return generate_perf_map
+
+
+def _clean_up_wrapper(module_path: str, pid: int) -> Callable:
+    def clean_up():
+        debugger_url = _get_debugger_url()
+        sock = create_connection(debugger_url)
+        _stop_dso(sock, module_path)
+        shutil.rmtree(os.path.join("/tmp", "node_module"))
```
Jongy, line 163: We should not remove the files as they're already loaded in target processes, this ends up like:
```
7fd8bd3b4000-7fd8bd3b5000 r--p 00006000 fd:01 114954256                  /tmp/node_module/glibc/10/build/Release/linux-perf.node (deleted)
```
and can be confusing upon the next profiling sessions, as the file is put there back again

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+def clean_up_node_maps():
+    """Stops generating perf maps for each NodeJS process and cleans up generated maps"""
+    processes = get_node_processes()
+    for process in processes:
+        try:
+            node_version = get_exe_version(process)
+            node_major_version = node_version[1:].split(".")[0]
+            pid_inside_ns = _get_pid_inside_ns(process.pid)
+            libc = "musl" if is_musl(process) else "glibc"
+            dest = os.path.join("/tmp", "node_module", libc, node_major_version)
```
Jongy, line 207: It's best if the final file path is versioned. This way, if in next gProfiler version we make a change to the DSO, and we want existing processes in existing deployments to get the new file, then it will be re-loaded because it's a different path. Similar to how it's done with async-profiler - we encode the async-profiler tag in the DSO path.

I suggest here that we encode e.g the 7 digits git revision from the repo of which we build the DSO. This way if we commit a fix, then the path changes and the DSO is re-loaded to existing processes

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+                ["pid", "mnt", "net"],
+                _clean_up_wrapper(dest, pid_inside_ns),
+                process.pid,
+            )
+            if hasattr(cp, "stdout"):
+                logger.debug(cp.stdout.decode().strip())
+            if hasattr(cp, "stderr"):
+                logger.debug(cp.stderr.decode().strip())
+            if not is_same_ns(process, NsType.mnt.name):
+                os.remove(os.path.join("/tmp", f"perf-{process.pid}.map"))
```
Jongy, line 218: Why? I don't think this file is copied to the host from the inner namespace (if that's what you meant to do here)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+            musl = is_musl(process)
+            node_version = get_exe_version(process)
+            node_major_version = node_version[1:].split(".")[0]
+            dest = _copy_module_into_process_ns(process, musl, node_major_version)
+            _start_debugger(process.pid)
+            pid_inside_ns = _get_pid_inside_ns(process.pid)
+            cp = run_in_ns(
+                ["pid", "mnt", "net"], _generate_perf_map_wrapper(dest), process.pid
+            )
+            if hasattr(cp, "stdout"):
```
Jongy, line 185: Are these relevant? Why would it have the `stdout` attribute?

Jongy:
The attach causes this to be printed:
```
Debugger attached.
Debugger listening on ws://127.0.0.1:9229/9e541850-fa6d-4e4c-b566-97bf6c786205
For help, see: https://nodejs.org/en/docs/inspector
```

Not too bad but if it can be avoided somehow, that's best.

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+
+def _copy_module_into_process_ns(process: psutil.Process, musl: bool, version: str):
+    proc_root = get_proc_root_path(process)
+    libc = "musl" if musl else "glibc"
+    dest_inside_container = os.path.join("tmp", "node_module", libc, version)
+    dest = os.path.join(proc_root, dest_inside_container)
+    if os.path.exists(dest):
+        return "/" + dest_inside_container
+    src = resource_path(os.path.join("node", "module", libc, version))
+    shutil.copytree(src, dest)
```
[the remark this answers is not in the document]
kwisniewski98: In current state I'm importing `linux-perf.js` file, but I can check how much I can shrink this directory

scripts/build_node_package.sh
kwisniewski98: Normally it would be build by npm, which will install `nan` in `.node_modules`, but since I didn't find a way to build it via npm for other versions than npm is installed for, I'm using `node-gyp`. This means I cannot use require (because we don't have `.node_modules`) and I'm providing this module to compiler via path.

gprofiler/utils/node.py
```diff
@@ -0,0 +1,223 @@
...
+    return generate_perf_map
+
+
+def _clean_up_wrapper(module_path: str, pid: int) -> Callable:
+    def clean_up():
+        debugger_url = _get_debugger_url()
+        sock = create_connection(debugger_url)
+        _stop_dso(sock, module_path)
+        shutil.rmtree(os.path.join("/tmp", "node_module"))
+        os.remove(os.path.join("/tmp", f"perf-{pid}.map"))
```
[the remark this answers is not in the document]
kwisniewski98: I've verified that with `lsof`. Looks like `_stop_dso` closes this file descriptor

gprofiler/utils/node.py
kwisniewski98: This was symlink in host namespace, but since we don't need it, this line can be removed

gprofiler/utils/node.py
kwisniewski98: It looks like both `cp.stdout` and `cp.stderr` are empty, so we can remove it.

gprofiler/utils/node.py
Jongy: kk

gprofiler/utils/node.py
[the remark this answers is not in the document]
Jongy: So `linux-perf.js` and the `.node` file?

pyi.Dockerfile
kwisniewski98: Now script copies only `linux-perf.js` and `linux-perf.node` to `/tmp/module_build` directory, so changes in Dockerfile are not required

kwisniewski98: It would be possible to copy `linux-perf.js` only once and then add proper code to gprofiler, but I don't think it is worth changing it since it is only 113 bytes and we will be saving only 15 blocks of disk (~60KB)

gprofiler/profilers/perf.py
Jongy: I see that you added in `snapshot` - there's no need to keep this in `start` now, only in `snapshot` suffices (and clean_up in `stop`)

gprofiler/utils/node.py
Jongy: The usage now (`resolve_proc_root_links(root, "/")`) will always result in returning `root` because `/` is never a link. But if `dest_inside_container` is a link, the code won't work. Need to call `resolve_proc_root_links(root, dest_inside_container)` and `dest_inside_container` needs to be an absolute path. This way, `resolve_proc_root_links` will follow any symlinks in `dest_inside_container` and correctly re-attach the `/proc/pid/root` part.

gprofiler/utils/node.py
Jongy: Well, this will raise but the exception is caught in `generate_map_for_node_processes` so maybe that's okay..

gprofiler/utils/node.py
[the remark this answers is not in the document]
Jongy: Great :)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,172 @@
...
+    copies node-linux-perf module into process' namespace, loads module and starts it.
+    After that it links it into gProfiler namespace's /tmp. This lets perf load it"""
+    for process in processes:
+        try:
+            musl = is_musl(process)
+            node_version = get_exe_version(process)
+            node_major_version = node_version[1:].split(".")[0]
+            dest = _copy_module_into_process_ns(process, musl, node_major_version)
+            _start_debugger(process.pid)
+            run_in_ns(["pid", "mnt", "net"], lambda: _generate_perf_map(dest), process.pid)
```
Jongy, line 153:
```suggestion
            run_in_ns(["pid", "mnt", "net"], lambda: _generate_perf_map(dest), process.pid, passthrough_exception=True)
```

This makes `run_in_ns` re-raise the exception raised by `_generate_perf_map` (if any). Otherwise, it relies on you checking the return value.

It really makes sense that we change this default value... I'll try to get that done soon

gprofiler/utils/node.py
```diff
@@ -0,0 +1,172 @@
...
+
+def clean_up_node_maps(processes: List[psutil.Process]) -> None:
+    """Stops generating perf maps for each NodeJS process and cleans up generated maps"""
+    for process in processes:
+        try:
+            node_version = get_exe_version(process)
+            node_major_version = node_version[1:].split(".")[0]
+            pid_inside_ns = get_process_nspid(process.pid)
+            dest = os.path.join("/", get_dest_inside_container(is_musl(process), node_major_version))
+            run_in_ns(
```
Jongy, line 166: Same as https://github.com/Granulate/gprofiler/pull/451/files#r969968737 - add `passthrough_exception=True`

scripts/node_builder_glibc_env.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+set -eu
+
+export DEBIAN_FRONTEND=noninteractive
+apt update -y && apt install -y curl g++ python3 make gcc git
```
Jongy, line 9:
```suggestion
apt update -y && apt install -y --no-install-recommends curl g++ python3 make gcc git
```

best practice is to add these everywhere, to reduce downloaded & installed size

pyi.Dockerfile
Jongy: Not worth changing indeed.

scripts/build_node_package.sh
```diff
@@ -0,0 +1,41 @@
...
+NAN_PATH=$(realpath /tmp/nan-*)
+export NAN_PATH
+sed -i 's/node \-e \\"require('\''nan'\'')\\"/echo $NAN_PATH/g' binding.gyp
+rm -rf nan.tar.gz
+mkdir $BUILD_TARGET_DIR
+node_versions=( "10.10.0" "11.0.0" )
+for node_version in "${node_versions[@]}"; do
+    node-gyp configure --target="$node_version"  --build_v8_with_gn=false
+    node-gyp build --target="$node_version"
+    t=("${node_version//./ }")
```
Jongy, line 24: This gets messed up on 10 & 11, I see these values:
```
gprofiler/resources/node/module/glibc/11 0 0
gprofiler/resources/node/module/glibc/11 0 0/linux-perf.js
gprofiler/resources/node/module/glibc/11 0 0/build
gprofiler/resources/node/module/glibc/11 0 0/build/Release
gprofiler/resources/node/module/glibc/11 0 0/build/Release/linux-perf.node
gprofiler/resources/node/module/glibc/14
gprofiler/resources/node/module/glibc/14/linux-perf.js
gprofiler/resources/node/module/glibc/14/build
gprofiler/resources/node/module/glibc/14/build/Release
gprofiler/resources/node/module/glibc/14/build/Release/linux-perf.node
gprofiler/resources/node/module/glibc/10 10 0
gprofiler/resources/node/module/glibc/10 10 0/linux-perf.js
gprofiler/resources/node/module/glibc/10 10 0/build
gprofiler/resources/node/module/glibc/10 10 0/build/Release
gprofiler/resources/node/module/glibc/10 10 0/build/Release/linux-perf.node
```

gprofiler/utils/node.py
Jongy: Uh my bad, I mixed the meaning here, it's passing the exception in any case, but wrapped in another one. Doesn't matter much.

gprofiler/utils/node.py
kwisniewski98: I'm a little bit worried about nexe / electron, but I don't think that they are common use cases and I don't even know if this method of generating symbols will work for them, so I think we can use `pgrep_exe`

kwisniewski98:
@Jongy can you rerun CI?

Dockerfile
Jongy: Ping @kwisniewski98 :)

Jongy: Ah done

gprofiler/metadata/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+from granulate_utils.linux.ns import get_process_nspid, run_in_ns
+from psutil import Process
+
+from gprofiler.utils import run_process
+
+
+def get_exe_version(
+    process: Process,
+    stop_event: Event = None,
+    get_version_timeout: int = None,
```
Jongy, line 13: These 2 should be required - let's remove the default value for them, so that you cannot call this function without passing them (I don't want it to be called without passing both of them)

gprofiler/metadata/__init__.py
```diff
@@ -0,0 +1,31 @@
+from subprocess import CompletedProcess
```
Jongy, line 1: Let's place at `gprofiler/metadata/versions.py`, this needn't be in the package level

gprofiler/profilers/perf.py
```diff
@@ -173,12 +174,13 @@ def __init__(
         perf_mode: str,
         perf_dwarf_stack_size: int,
         perf_inject: bool,
+        perf_node_attach: bool,
     ):
         super().__init__(frequency, duration, stop_event, storage_dir)
         _ = profile_spawned_processes  # Required for mypy unused argument warning
         self._perfs: List[PerfProcess] = []
         self._metadata_collectors: List[PerfMetadata] = [GolangPerfMetadata(stop_event), NodePerfMetadata(stop_event)]
-
```
Jongy, line 181: Bring it back :) (blank)

gprofiler/profilers/perf.py
Jongy: Ping @kwisniewski98 :)

gprofiler/profilers/perf.py
```diff
@@ -229,6 +237,11 @@ def _get_metadata(self, pid: int) -> Optional[AppMetadata]:
         return None
 
     def snapshot(self) -> ProcessToProfileData:
+        if self.perf_node_attach:
+            new_processes = [process for process in get_node_processes() if process not in self._node_processes]
+            generate_map_for_node_processes(new_processes)
+            self._node_processes.extend(new_processes)
```
Jongy, line 243: Prune dead processes, otherwise the list grows forever. I suppose the logic can be:
1. remove all processes where `not is_process_running()`
1. filter new from `get_node_processes()`
3. generate map for them
4. add to `_node_processes`

gprofiler/utils/__init__.py
```diff
@@ -273,7 +273,7 @@ def pgrep_exe(match: str) -> List[Process]:
         try:
             if pattern.match(process_exe(process)):
                 procs.append(process)
-        except psutil.NoSuchProcess:  # process might have died meanwhile
+        except (psutil.NoSuchProcess, MissingExePath):  # process might have died meanwhile
```
Jongy, line 276: `MissingExePath` is what we get for kernel threads, per the documentation of `process_exe`. We should not try to query them to begin with.

I think you started getting it now because `pgrep_exe` was unused for a while, and lately we added the `MissingExePath` exception.

The `except MissingExePath` is a bit confusing though, ideally we should skip kernel threads in this loop. Please see if it's possible with psutil API. Otherwise, you can add a comment on `MissingExePath` explaining that this is what you get for kernel threads, and we ignore them

gprofiler/utils/__init__.py
```diff
@@ -457,3 +457,12 @@ def is_pyinstaller() -> bool:
 
 def get_staticx_dir() -> Optional[str]:
     return os.getenv("STATICX_BUNDLE_DIR")
+
+
+def add_permission_dir(path: str, permission_for_file: int, permission_for_dir: int) -> None:
```
Jongy, line 462: Instead of this approach - you can use `shutil.copytree(copy_function=...)` and as your copy function, pass a function that calls `copy2` which is the one used by default, and then follows by the needed chmod.

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+from websocket import create_connection
+from websocket._core import WebSocket
+
+from gprofiler.log import get_logger_adapter
+from gprofiler.metadata import get_exe_version
+from gprofiler.utils import add_permission_dir, pgrep_exe, resource_path
+
+logger = get_logger_adapter(__name__)
+
+DSO_GIT_REVISION = "20eb88a"
```
Jongy, line 22: This requires manual updating which could be forgotten and then we cause harm by editing the file despite being loaded in some processes. We need to ensure gProfiler either uses the exact git revision OR refuses to use the file if the git revision is "incorrect".

I think it's easier to use the exact one - so like the AP build script, you can write the reivision to a file that is copied into our resources, and then add a function like `get_ap_version` which reads it?

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def _start_debugger(pid: int) -> None:
+    # for windows: in shell node -e "process._debugProcess(PID)"
+    os.kill(pid, signal.SIGUSR1)
+
+
+@retry(NodeDebuggerUrlNotFound, 5, 1)
+def _get_debugger_url() -> str:
+    # when killing process with SIGUSR1 it will open new debugger session on port 9229,
```
Jongy, line 45: Btw if you already found the source code in nodejs which performs this behavior (e.g the constant port, ..) it's nice a put a permalink code comment here to the github sources :) I'd love to take a look

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+def _clean_up(module_path: str, pid: int) -> None:
+    debugger_url = _get_debugger_url()
+    sock = create_connection(debugger_url)
+    _stop_dso(sock, module_path)
+    sock.settimeout(2)
+    os.remove(os.path.join("/tmp", f"perf-{pid}.map"))
+
+
+def get_node_processes() -> List[psutil.Process]:
+    return pgrep_exe(r"(?:^.+/node[^/]*$)")
```
Jongy, line 142:
```suggestion
    return pgrep_exe(r"(?:^.*/node[^/]*$)")
```

(to catch `/node` as well)

Also - what are the `(?: ... )` for (parenthesis)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+def _start_debugger(pid: int) -> None:
+    # for windows: in shell node -e "process._debugProcess(PID)"
+    os.kill(pid, signal.SIGUSR1)
+
+
+@retry(NodeDebuggerUrlNotFound, 5, 1)
+def _get_debugger_url() -> str:
+    # when killing process with SIGUSR1 it will open new debugger session on port 9229,
+    # so it will always the same
+    port = 9229
```
Jongy, line 47: By the way - if the port is already caught by another process, what will happen? No debug port will be opened?

We possibly need to add special handling for that here. Can you ensure the server you're talking with is actually the PID we desire? What if it already has the module loaded from e.g a previous run of gProfiler?

I think the best structure would be to add sanity checks before invoking the debugger, and after you have a debugger socket, ensure you're working with the desired process. The simplest way I can think of (which isn't very simple) is to compare the NSpid of the process bundled with its `/proc/self/ns/pid` link name (which corresponds to the PID namespace identifier of the process). Together these 2 are a unique identifier for a PID, seen both from the inner NS (process) and the outer NS (gprofiler)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+    debugger_url_response = requests.get(f"http://127.0.0.1:{port}/json/list")
+    if (
+        debugger_url_response.status_code != 200
+        or not debugger_url_response.headers.get("Content-Type")
+        or "application/json" not in debugger_url_response.headers.get("Content-Type")  # type: ignore
```
Jongy, line 52:
```suggestion
        or "application/json" not in debugger_url_response.headers.get("Content-Type", "")  # type: ignore
```
simpler :)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+    response_json = debugger_url_response.json()
+    if (
+        not isinstance(response_json, list)
+        or len(response_json) == 0
+        or not isinstance(response_json[0], dict)
+        or "webSocketDebuggerUrl" not in response_json[0]
+    ):
+        raise NodeDebuggerUrlNotFound(response_json)
+
+    return response_json[0]["webSocketDebuggerUrl"]  # type: ignore
```
Jongy, line 67: Instead of `type: ignore`, you can use `typing.cast` which lets you the programmer imply what was your typing intention

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+@retry(NodeDebuggerUnexpectedResponse, 5, 1)
```
Jongy, line 70: Why are retries needed?

Jongy: For example, preferably I don't want retries if `_load_dso` fails. What are the retries for?

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+    src = resource_path(os.path.join("node", "module", libc, version))
+    shutil.copytree(src, dest)
+    add_permission_dir(dest, stat.S_IROTH, stat.S_IXOTH | stat.S_IROTH)
+    return dest_inside_container
+
+
+def _generate_perf_map(module_path: str) -> None:
+    debugger_url = _get_debugger_url()
+    sock = create_connection(debugger_url)
+    sock.settimeout(2)
```
Jongy, line 129: These 3 lines can be moved to a function and reused?

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+    _load_dso(sock, module_path)
+
+
+def _clean_up(module_path: str, pid: int) -> None:
+    debugger_url = _get_debugger_url()
+    sock = create_connection(debugger_url)
+    _stop_dso(sock, module_path)
+    sock.settimeout(2)
```
Jongy, line 137: Socket code can be reused as suggested [here](https://github.com/Granulate/gprofiler/pull/451/files#r975785496), and also, why is settime after we're done with the socket haha

gprofiler/utils/node.py
Jongy: (for example, we might attempt to load the DSO of incorrect version to a process)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+            dest = _copy_module_into_process_ns(process, musl, node_major_version)
+            _start_debugger(process.pid)
+            run_in_ns(["pid", "mnt", "net"], lambda: _generate_perf_map(dest), process.pid, passthrough_exception=True)
+        except Exception as e:
+            logger.warning(f"Could not create debug symbols for pid {process.pid}. Reason: {e}", exc_info=True)
+
+
+def clean_up_node_maps(processes: List[psutil.Process]) -> None:
+    """Stops generating perf maps for each NodeJS process and cleans up generated maps"""
+    for process in processes:
```
Jongy, line 163: Check if process is running (`is_process_running`) before trying

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+        try:
+            node_version = get_exe_version(process)
+            node_major_version = node_version[1:].split(".")[0]
```
Jongy, line 166: What manipulation are you doing on the version here? (can explain in comment the expected value of `node_version` and the end result you try to get)

Jongy, line 166: Duplicated code from `generate_map_for_node_processes`, you can share the node version lines

Jongy: `v16.3.2` -> `16` is a good such comment IMO

gprofiler/utils/node.py
```diff
@@ -0,0 +1,176 @@
...
+            pid_inside_ns = get_process_nspid(process.pid)
```
Jongy, line 167: nspid is the term we try to follow :) (the linux term)

tests/conftest.py
```diff
@@ -125,14 +125,18 @@ def command_line(runtime: str, java_command_line: List[str], dotnet_command_line
             "--interpreted-frames-native-stack",
             str(CONTAINERS_DIRECTORY / "nodejs/fibonacci.js"),
         ],
+        "nodejs-attach": [
```
Jongy, line 128: I don't like adding it as a different runtime, it requires maintenance in many sites.

Instead, I suggest:
1. For the Dockerfile, I will get https://github.com/Granulate/gprofiler/pull/477 merged soon, this lets you easily add a separate image for `nodejs`. It could be implemented by passing a build arg that's used as an env var, and if the env var is present, the Dockerfile won't add the perf jit flags.
2. For `runtime_specific_args`, what makes more sense is to require the `profiler_type` fixture instead of using a separate runtime. `profiler_type` is not defined by default but I checked and all users of `runtime_specific_args` also define `profiler_type` in the test - it's a simple change to require `profiler_type` in `runtime_specific_args`, and to use its value to decide for `nodejs` which profiler mode is used (it's actually the exact value).
3. For `command_line`, I think it's used also in cases where `profiler_type` is not defined and that's just a "bug" in our test infra (there are tests which use it but don't define `profiler_type` and actually hardcode the value in the test, e.g this one: https://github.com/Granulate/gprofiler/blob/43e1e16042c758a987b1e4efb987829b86a93c74/tests/test_sanity.py#L87. I don't suggest we fix the test infra in this PR - as a fair workaround, we can parametrize the fixture in your test `test_nodejs_attach_maps` to the needed value. I'm opening a ticket for the long term fix in the test infra.

Jongy: Created https://github.com/Granulate/gprofiler/issues/484

tests/containers/nodejs-attach/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# node:10.24.1
```
Jongy, line 1: As I suggested in https://github.com/Granulate/gprofiler/pull/451/files#r975877845, we don't need a separate Docker image, please remove it and after https://github.com/Granulate/gprofiler/pull/477 is merged, build the `tests/containers/nodejs/` image with a build argument that prevents it from using the jitdump file

scripts/build_node_package.sh
Jongy: Understood :) Can you please comment that above the `sed` line?

scripts/build_node_package.sh
```diff
@@ -0,0 +1,41 @@
...
+node_versions=( "10.10.0" "11.0.0" )
+for node_version in "${node_versions[@]}"; do
+    node-gyp configure --target="$node_version"  --build_v8_with_gn=false
+    node-gyp build --target="$node_version"
+    t=(${node_version//./ })
+    node_major_version=${t[0]}
+    mkdir "$BUILD_TARGET_DIR/$node_major_version"
+    cp "$MODULE_PATH/linux-perf.js" "$BUILD_TARGET_DIR/$node_major_version/."
+    mkdir -p "$BUILD_TARGET_DIR/$node_major_version/build/Release"
+    cp "$MODULE_PATH/build/Release/linux-perf.node" "$BUILD_TARGET_DIR/$node_major_version/build/Release/linux-perf.node"
```
Jongy, line 29: Any reason to keep the `build/Release/` in the path? We can just copy `linux-perf.node` to `$BUILD_TARGET_DIR/$node_major_version/`?

Jongy (changes requested):
Done reviewing again :)

scripts/build_node_package.sh
Jongy: (I expect it to be just `11` for 11, not `11 0 0`, no? Otherwise the loading code won't find it, because it searches for the major only)

gprofiler/utils/node.py
kwisniewski98: If port is already opened it won't open. We can have one of two solutions for having multiple node processes in same network namespace:

1.  Close debugger after generating perf maps and starting it each time we want to do any operation on debugger. In this method no two debuggers will be open at the same time. This will also be useful when someone started container with i. e. `--network=host`. (I've started implementing that)
2.  We can change debug port, but that can only be done inside process (we need to change `process.debugPort`). This means we need to start it on port 9229 anyway, then change `debugPort` and restart debugger.

In both methods we have to start it at 9229 port and I didn't find any solution for that.

Answering your other questions:
If module is already loaded, CDP request will respond with False when loading module
I'll add this sanity check

gprofiler/utils/node.py
kwisniewski98: I.e. debugger might not start yet

scripts/build_node_package.sh
kwisniewski98: This is the path that `linux-perf.js` [requires](https://github.com/mmarchini/node-linux-perf/blob/master/linux-perf.js#L3). I was trying to import `linux-perf.node` directly in CDP request, but I was getting errors, so I decided to leave it as it is. If you want I can work on that a little bit more

Jongy: Nope that's a good reason, let's just leave a comment. If it annoys us we'll be able to fork the project and change `linux-perf.js` to use another path.

Jongy: You can leave a comment explaining that on this line

gprofiler/utils/node.py
Jongy: In that case won't we get a connection refused error in the socket level? The current `NodeDebuggerUnexpectedResponse` is an applicative error - an endpoint exists but it responded invalid, I don't understand why trying again would help

gprofiler/utils/node.py
Jongy: The 2 options you suggested are viable but I think we'll leave the decision for a future PR. I am meanwhile opening a ticket from it.

Jongy: We have 2 issues remaining here:
1. How do we know we're talking with the process we think we're talking with (nevermind if another process in the namespace could not use the port - we just need to know we're talking not with the expected one).
2. How do we check if the module is already loaded (and ideally, if possible, allow to reload once we change version in new gprofiler release, as we do for async-profiler).

gprofiler/utils/node.py
Jongy: I don't know if they're used in production, I have never seen them used so I think we can skip them for now.

gprofiler/utils/__init__.py
Jongy: I see. Okay, let's keep this function for now. Leaving the other comment I had on it.

gprofiler/utils/node.py
kwisniewski98: 1. I've already added logic that checks if `/proc/<pid>/ns/pid` in gprofiler ns is equal to `/proc/<nspid>/ns/pid` inside given namespace. Tried to do it with `/proc/self` but i was getting no such file or directory. This should be enough to check if we are in same namespace. If we want to also check if it is the same process in given namespace we can query debugger for it's pid (i.e. reading `process.pid`) and compare it with expected pid inside namespace.

2. Start function of module returns true if module is loaded and false if it isn't. We can possibly utilize that. Second option is to fork that  module and add our own method, shouldn't be that hard since it is only checking, if handler is nullptr. [ref](https://github.com/mmarchini/node-linux-perf/blob/master/linux-perf.cc#L41)

Jongy: 1. Sounds good - we compare both the `ns/pid` and the NSPID, if both match it's the same process.
2. Cases that we want to check: first: already running module of the same version - we detect it and log it. second: upon upgrade of linux-perf, gprofiler uses the new version of linux-perf (2 are loaded in the memory space)

gprofiler/utils/node.py
```diff
@@ -0,0 +1,243 @@
+import json
```
Jongy, line 1: This file is under `utils/` but logically it's like the rest of the profilers under `gprofiler/profilers/`, let's move as `gprofiler/profilers/node.py`?

tests/containers/nodejs/Dockerfile
```diff
@@ -7,4 +9,6 @@ WORKDIR /tmp
 RUN mkdir /app
 ADD fibonacci.js /app
 
-CMD ["node", "--perf-prof", "--interpreted-frames-native-stack", "/app/fibonacci.js"]
+ENV NODE_RUNTIME_FLAGS ${NODE_RUNTIME_FLAGS}
```
Jongy, line 12: Is this needed? the CMD line could just be `node ${NODE_RUNTIME_FLAGS} /app/fibonacci.js`, can't it?

gprofiler/utils/node.py
```diff
@@ -0,0 +1,243 @@
...
+    pass
+
+
+def _get_node_major_version(process: psutil.Process) -> str:
+    node_version = get_exe_version(process, Event(), 3)
+    # i. e. v16.3.2 -> 16
+    return node_version[1:].split(".")[0]
+
+
+def _get_dso_git_rev() -> str:
```
Jongy, line 39: Can add `@functools.lru_cache(maxsize=1)` on this function

gprofiler/utils/node.py
```diff
@@ -0,0 +1,243 @@
...
+        musl_dso_ver = f.readline()
+    # with no build errors this should always be the same
+    assert libc_dso_ver == musl_dso_ver
+    return libc_dso_ver
+
+
+@lru_cache()
+def _get_dest_inside_container(musl: bool, node_version: str) -> str:
+    libc = "musl" if musl else "glibc"
+    return os.path.join("/", "tmp", "node_module", _get_dso_git_rev(), libc, node_version)
```
Jongy, line 54: Please import `GPROFILER_DIRECTORY_NAME` and use it as a path component here (to avoid placing files outside of gprofiler directories)

scripts/build_node_package.sh
Jongy: Looks good now:
```
/app/gprofiler/resources/node/module/glibc/20eb88a
/app/gprofiler/resources/node/module/glibc/20eb88a/11
/app/gprofiler/resources/node/module/glibc/20eb88a/11/linux-perf.js
/app/gprofiler/resources/node/module/glibc/20eb88a/11/build
/app/gprofiler/resources/node/module/glibc/20eb88a/11/build/Release
/app/gprofiler/resources/node/module/glibc/20eb88a/11/build/Release/linux-perf.node
/app/gprofiler/resources/node/module/glibc/20eb88a/10
/app/gprofiler/resources/node/module/glibc/20eb88a/10/linux-perf.js
/app/gprofiler/resources/node/module/glibc/20eb88a/10/build
/app/gprofiler/resources/node/module/glibc/20eb88a/10/build/Release
/app/gprofiler/resources/node/module/glibc/20eb88a/10/build/Release/linux-perf.node
/app/gprofiler/resources/node/module/glibc/20eb88a/12
/app/gprofiler/resources/node/module/glibc/20eb88a/12/linux-perf.js
/app/gprofiler/resources/node/module/glibc/20eb88a/12/build
/app/gprofiler/resources/node/module/glibc/20eb88a/12/build/Release
/app/gprofiler/resources/node/module/glibc/20eb88a/12/build/Release/linux-perf.node
/app/gprofiler/resources/node/module/glibc/20eb88a/13
/app/gprofiler/resources/node/module/glibc/20eb88a/13/linux-perf.js
/app/gprofiler/resources/node/module/glibc/20eb88a/13/build
/app/gprofiler/resources/node/module/glibc/20eb88a/13/build/Release
/app/gprofiler/resources/node/module/glibc/20eb88a/13/build/Release/linux-perf.node
/app/gprofiler/resources/node/module/glibc/20eb88a/14
/app/gprofiler/resources/node/module/glibc/20eb88a/14/linux-perf.js
/app/gprofiler/resources/node/module/glibc/20eb88a/14/build
/app/gprofiler/resources/node/module/glibc/20eb88a/14/build/Release
/app/gprofiler/resources/node/module/glibc/20eb88a/14/build/Release/linux-perf.node
/app/gprofiler/resources/node/module/glibc/20eb88a/15
/app/gprofiler/resources/node/module/glibc/20eb88a/15/linux-perf.js
/app/gprofiler/resources/node/module/glibc/20eb88a/15/build
/app/gprofiler/resources/node/module/glibc/20eb88a/15/build/Release
/app/gprofiler/resources/node/module/glibc/20eb88a/15/build/Release/linux-perf.node
/app/gprofiler/resources/node/module/glibc/20eb88a/16
/app/gprofiler/resources/node/module/glibc/20eb88a/16/linux-perf.js
/app/gprofiler/resources/node/module/glibc/20eb88a/16/build
/app/gprofiler/resources/node/module/glibc/20eb88a/16/build/Release
/app/gprofiler/resources/node/module/glibc/20eb88a/16/build/Release/linux-perf.node
```

tests/test_sanity.py
```diff
@@ -116,12 +116,75 @@ def test_nodejs(
...
+        perf_mode="fp",
+        perf_inject=True,
+        perf_dwarf_stack_size=0,
+        perf_node_attach=False,
     ) as profiler:
         process_collapsed = snapshot_pid_collapsed(profiler, application_pid)
         assert_collapsed(process_collapsed)
 
 
+@pytest.mark.parametrize("profiler_type", ["attach-maps"])
```
Jongy, line 133: Move these 2 tests to `test_node.py`

tests/test_sanity.py
```diff
@@ -116,12 +116,75 @@ def test_nodejs(
...
+@pytest.mark.parametrize("command_line", [["node", f"{CONTAINERS_DIRECTORY}/nodejs/fibonacci.js"]])
+def test_nodejs_attach_maps_from_container(
+    docker_client: DockerClient,
+    application_pid: int,
+    runtime_specific_args: List[str],
+    gprofiler_docker_image: Image,
+    output_directory: Path,
+    output_collapsed: Path,
+    assert_collapsed: AssertInCollapsed,
+    assert_app_id: Callable,
```
Jongy, line 176: We can keep it although it doesn't do much as we don't have appid support for nodejs yet (https://github.com/Granulate/gprofiler/issues/504)

Jongy:
@kwisniewski98 , I closed all done comments. Still need to do another full review post your changes

tests/containers/nodejs/Dockerfile
kwisniewski98: build args are available only at build stage, they aren't available when running container, that's why we need to save it to i. e. env

gprofiler/utils/node.py
kwisniewski98: You are right, I've removed that

gprofiler/metadata/versions.py
```diff
@@ -0,0 +1,31 @@
+from subprocess import CompletedProcess
```
Jongy, line 1: Add project license

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,244 @@
+import json
```
Jongy, line 1: Add project license

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+@lru_cache(maxsize=1)
+def _get_dso_git_rev() -> str:
+    libc_dso_version_file = resource_path(os.path.join("node", "module", "glibc", "version"))
+    musl_dso_version_file = resource_path(os.path.join("node", "module", "musl", "version"))
+    with open(libc_dso_version_file) as f:
+        libc_dso_ver = f.readline()
+    with open(musl_dso_version_file) as f:
+        musl_dso_ver = f.readline()
```
Jongy, line 46: Can use `Path(musl_dso_version_file).read_text()`. Shorter and in case the file does contain more than one line for whatever reason.. This is more robust and gets you notified of the error

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,244 @@
...
+    try:
+        message = json.loads(message)
+    except json.JSONDecodeError:
+        raise NodeDebuggerUnexpectedResponse(message)
+
+    if (
+        "result" not in message.keys()
+        or "result" not in message["result"].keys()
+        or "type" not in message["result"]["result"].keys()
+        or message["result"]["result"]["type"] != "boolean"
```
Jongy, line 105: No further checks on the results need to be done here, btw? (I figured, if you check that the type is `boolean`, then we should probably check that the value is `true`, no?)

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,244 @@
...
+    ):
+        raise NodeDebuggerUnexpectedResponse(message)
+
+
+def _get_node_pid(sock: WebSocket) -> int:
```
Jongy, line 110:
```suggestion
def _get_node_nspid(sock: WebSocket) -> int:
```

Makes it more clear

tests/containers/nodejs/Dockerfile
Jongy: Hmm understood

tests/test_node.py
```diff
@@ -0,0 +1,66 @@
+from pathlib import Path
```
Jongy, line 1: Add project license

scripts/list_needed_libs.sh
```diff
@@ -10,7 +10,14 @@ set -euo pipefail
 # (staticx knows to pack the libraries used by the executable we're packing. it doesn't know
 # which executables are to be used by it)
 
-BINS=$(find gprofiler/resources -executable -type f)
+EXCLUDED_DIRECTORIES=('"gprofiler/resources/node/*"')
```
Jongy, line 13: Do we need the exclude for an array of dirs and not just `find gprofiler/resources -executable -type f -not -path gprofiler/resources/node/`?

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,244 @@
...
+def _get_debugger_url() -> str:
+    # when killing process with SIGUSR1 it will open new debugger session on port 9229,
+    # so it will always the same. When another debugger is opened in same NS it will not open new one.
+    # REF: Inspector agent initialization uses host_port
+    # https://github.com/nodejs/node/blob/5fad0b93667ffc6e4def52996b9529ac99b26319/src/inspector_agent.cc#L668
+    # host_port defaults to 9229
+    # ref: https://github.com/nodejs/node/blob/2849283c4cebbfbf523cc24303941dc36df9332f/src/node_options.h#L90
+    # in our case it won't be changed
+    port = 9229
+    debugger_url_response = requests.get(f"http://127.0.0.1:{port}/json/list")
```
Jongy, line 73: Let's add timeouts to all requests to prevent the profiler from being stuck if any process misbehaves. I think 3 seconds would suffice?

Jongy: In `create_debugger_socket` we already have a timeout, so only for this `requests.get` here it would suffice I suppose

gprofiler/profilers/perf.py
Jongy: On a second thought, it's okay to leave it here I suppose - it will affect immediately before perf starts recording. In `snapshot` it's too late for that

gprofiler/utils/node.py
Jongy: 2. was chcked - we're good

Jongy: 1. you implemented a check but I'm not sure it's enough, I'll comment on the new check.

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,244 @@
...
+        "params": {
+            "expression": f'process.mainModule.require("{os.path.join(module_path, "linux-perf.js")}").stop()',
+            "replMode": True,
+        },
+    }
+    _send_socket_request(sock, cdp_request)
+
+
+def validate_ns(nspid: int, expected_ns_link_name: str) -> None:
+    actual_nspid_name = os.readlink(os.path.join("/proc", str(nspid), "ns", "pid"))
```
Jongy, line 156: Continuing our discussion from https://github.com/Granulate/gprofiler/pull/451#discussion_r984097131 -

This validation is definitely an improvement - it catches the case where multiple node processes are running on the host, and multiple node processes are running in the same container.

It doesn't catch, however, these 2 cases:
* 2 node processes running in 2 different containers in the same k8s Pod - in k8s, the network namespace is shared between containers in the same Pod, but the PID namespace is not.
* 2 node processes running in 2 different containers with `--net=host`.
These 2 cases are much more exotic, I'd say, but if it doesn't add too much complexity then we can support them as well. For that - it'd require to run this logic `os.readlink(os.path.join("/proc", str(nspid), "ns", "pid"))` via `Runtime.evaluate` (i.e, it's JS equivalent).

I'm okay with leaving it for a future PR though - so you can try it out, and if it turns out not straight forward, we can leave a TODO in code (can refer to the link of thise comment) & I'll open a ticket for later.

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,244 @@
...
+    assert (
+        actual_nspid_name == expected_ns_link_name
+    ), f"Wrong namespace, expected {expected_ns_link_name}, got {actual_nspid_name}"
+
+
+def validate_pid(expected_pid: int, sock: WebSocket) -> None:
+    actual_pid = _get_node_pid(sock)
+    assert expected_pid == actual_pid, f"Wrong pid, expected  {expected_pid}, actual {actual_pid}"
```
Jongy, line 164:
```suggestion
    assert expected_pid == actual_pid, f"Wrong pid, expected {expected_pid}, actual {actual_pid}"
```

gprofiler/profilers/node.py
```diff
@@ -0,0 +1,244 @@
...
+        "method": "Runtime.evaluate",
+        "params": {
+            "expression": f'process.mainModule.require("{os.path.join(module_path, "linux-perf.js")}").start()',
+            "replMode": True,
+        },
+    }
+    _send_socket_request(sock, cdp_request)
+
+
+def _stop_dso(sock: WebSocket, module_path: str) -> None:
```
Jongy, line 143: This function is exactly like `_load_dso`, only `start` became `stop`. Coalesce them?

Jongy (changes requested):
Finished 4th round

gprofiler/profilers/node.py
kwisniewski98: Checking value will only show us if module was started/stopped before, which isn't necessary in my opinion. Seeing it returned boolean lets us know, that module is now in wanted state.

scripts/list_needed_libs.sh
kwisniewski98: We can, but I wanted it to be easy to add another directory if someone needs to do that in future. If you want I can just add it as one argument to command

Jongy: Fair. Let's keep it

gprofiler/profilers/node.py
```diff
@@ -123,53 +126,52 @@ def _get_node_pid(sock: WebSocket) -> int:
...
     assert (
         actual_nspid_name == expected_ns_link_name
     ), f"Wrong namespace, expected {expected_ns_link_name}, got {actual_nspid_name}"
 
 
-def validate_pid(expected_pid: int, sock: WebSocket) -> None:
-    actual_pid = _get_node_pid(sock)
-    assert expected_pid == actual_pid, f"Wrong pid, expected  {expected_pid}, actual {actual_pid}"
+def _validate_ns_node(sock: WebSocket, nspid: str, expected_ns_link_name: str) -> None:
+    command = f'const fs = process.mainModule.require("fs"); fs.readlinkSync("/proc/{nspid}/ns/pid")'
```
Jongy, line 155: This js code runs in the target process so you can use self instead of nspid

gprofiler/profilers/node.py
Jongy: I see. Since the module cannot fail (and return value is only "did I turn it on/off now"), you're right

gprofiler/profilers/node.py
```diff
@@ -123,53 +126,52 @@ def _get_node_pid(sock: WebSocket) -> int:
...
+    actual_ns_link_name = cast(str,_execute_js_command(sock, command))
+    logger.debug(f"Found link by node: {actual_ns_link_name}")
```
Jongy, line 157: Can be removed

gprofiler/profilers/node.py
```diff
@@ -123,53 +126,52 @@ def _get_node_pid(sock: WebSocket) -> int:
...
+
+
+def _validate_pid(expected_pid: int, sock: WebSocket) -> None:
+    actual_pid = cast(int, _execute_js_command(sock, "process.pid"))
+    assert expected_pid == actual_pid, f"Wrong pid, expected {expected_pid}, actual {actual_pid}"
 
 
 def create_debugger_socket(nspid: int, ns_link_name: str) -> WebSocket:
-    validate_ns(nspid, ns_link_name)
+    _validate_ns(nspid, ns_link_name)
```
Jongy, line 169: I don't think this check is required at all - this `create_debugger_socket` function is called by us in the PID namespace of the target process. There's no way that the PID namespace won't match here. Only in `_validate_ns_node` there's a chance for failure, because we could be communicating with a Node process that's in the same net NS as us, but different PID NS.

Jongy: Let's remove it

gprofiler/profilers/node.py
Jongy: Implemented, resolving this

Jongy:
Reviewed again. Left 3 final comments. Please fix the lint errors as well, and the one [comment](https://github.com/Granulate/gprofiler/pull/451#discussion_r984096450) from earlier that wasn't fixed yet as far as I can tell, and we're good to merge this :)

Jongy (dismissed):
LGTM.

Lint errors are what I described in https://github.com/Granulate/gprofiler/pull/517#issuecomment-1270023442, so I ignore them. I will merge after tests pass.

Jongy:
Also, you can merge from master, I fixed the flake8 issue

Jongy (approved):

Jongy:
At last :) Congrats @kwisniewski98 !

PR #339, opened 2022-04-14
java: Upgrade async-profiler (regression in build scripts, not using -static-* flags)

Jongy:
Before:
```
root@3994adfa7dcf:/app# for f in  gprofiler/resources/java/*/libasyncProfiler.so ; do echo $f; ldd $f; done
gprofiler/resources/java/glibc/libasyncProfiler.so
	linux-vdso.so.1 (0x00007ffe815cf000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f078b380000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f078b35d000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f078b352000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f078b170000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f078b021000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f078b006000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f078ae12000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f078b5c6000)
gprofiler/resources/java/musl/libasyncProfiler.so
	linux-vdso.so.1 (0x00007ffcfdb7f000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007ffa12677000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007ffa1265c000)
	libc.musl-x86_64.so.1 => not found
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ffa1250d000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ffa1231b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007ffa128a4000)
```

After:
```
root@2fffbd329bd9:/app# for f in  gprofiler/resources/java/*/libasyncProfiler.so ; do echo $f; ldd $f; done
gprofiler/resources/java/glibc/libasyncProfiler.so
	linux-vdso.so.1 (0x00007ffdc8f15000)
[...]

PR #304, opened 2022-02-23
Build async-profiler on CentOS 7

Jongy:
My trigger for that are the lousy mirrors of CentOS 6, which make our CI flaky.
In any case, it's a hassle to continue using CentOS 6 in any way.

This commit fixes the build to be CentOS 7 based. It's old enough glibc to run on
old kernels, and I verified that the only relevant symbol which required glibc>2.12
now gets versioned to something older as well.

See the fix for AP here: https://github.com/Granulate/async-profiler/commit/0d04078e0586793e852beb11ab285185c447e986

I tested the before & after with https://gist.github.com/Jongy/56f33f4b8a3e04b41d1afb85b47121a1:
before - 
```
glibc version: 2.14
glibcxx version: 3.4.9
```
after -
```
glibc version: 2.4
glibcxx version: 3.4.9
```

and I expect that the executable test on `centos:6` will also cover me here :sweat_smile:

PR #453, opened 2022-08-30
java: Fix async-profiler compatibility on CentOS 6

Jongy:
This PR fixes compatibility with CentOS 6. When we switched to build on [CentOS 7](https://github.com/Granulate/gprofiler/pull/304) we didn't pass the `-static-libstdc++` flag, and at the point when [we fixed that](https://github.com/Granulate/gprofiler/pull/339), we didn't test CentOS 6 support, which broke because `-static-libstdc++` pulls in some symbols which require newer glibc.

This PR avoids `-static-libstdc++` on glibc; it's required for musl-based because musl-based JVMs aren't always dependent on libstdc++; and we've never seen that with glibc.

Additionally, this PR adds a build test for (in)existence of libstdc++, and for final dependency resolving compatibility on CentOS 6.


Jongy:
Ready @adi-benz

scripts/async_profiler_build_shared.sh
```diff
@@ -5,12 +5,27 @@
...
 
 git clone --depth 1 -b "$VERSION" https://github.com/Granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
 source "$1"
 make all
 
 # add a version file to the build directory
 echo -n "$VERSION" > build/async-profiler-version
+
+# build verifications regarding libstdc++.
+if ldd /bin/ls | grep -q musl ; then
```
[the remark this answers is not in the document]
Jongy: Nope, inside an `if` it doesn't take effect.
```
$ bash -e -o pipefail -c "if false ; then echo false ; else echo true ; fi"
true
```

scripts/async_profiler_build_shared.sh
```diff
@@ -5,12 +5,27 @@
...
+GIT_REV="008a8cf435bedce8eec3ef36fdc67e742b070566"
 
 git clone --depth 1 -b "$VERSION" https://github.com/Granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
 source "$1"
 make all
 
 # add a version file to the build directory
 echo -n "$VERSION" > build/async-profiler-version
+
+# build verifications regarding libstdc++.
```
adi-benz, line 18: Can u copy the comment from the PR description?
> avoids -static-libstdc++ on glibc; it's required for musl-based because musl-based JVMs aren't always dependent on libstdc++; and we've never seen that with glibc.

scripts/async_profiler_build_shared.sh
```diff
@@ -5,12 +5,27 @@
...
+if ldd /bin/ls 2>&1 | grep -q musl ; then
```
adi-benz, line 19: Why do we need those build time tests? The CI runs on centos 6 and thus verifies that we don't use new symbols

And we have a test in `test_java.py` that makes sure that on musl we don't load libstd++

adi-benz (changes requested):

scripts/async_profiler_build_shared.sh
Jongy: Kk will add

scripts/async_profiler_build_shared.sh
Jongy: (CentOS 6 test only tests gProfiler, not async-profiler loading on CentOS 6)

adi-benz (approved):

PR #455, opened 2022-09-01
Use mutex function from granulate-utils & test preconditions

Jongy:
* Use `try_acquire_mutex()` instead of our clone.
* Add tests for some of the common precondition checks.


netaneld122 (approved):
Looks good

PR #459, opened 2022-09-06
dotnet: Support Aarch64

Jongy:
It just worked :shrug: so why not.


Jongy:
Tested profiling of `tests/containers/dotnet/Dockerfile` app with gProfiler exe & binary from this branch - both work.

Jongy:
The same manifest hash `749439ff7a431ab4bc38d43cea453dff9ae1ed89a707c318b5082f9b2b25fa22` has an image for both x86 and aarch64.

netaneld122 (approved):

PR #461, opened 2022-09-06
tests: app metadata: Add dotnet

Jongy:
Straightforward

### DIFF
--- a/tests/test_app_metadata.py
+++ b/tests/test_app_metadata.py
@@ -81,6 +81,16 @@ from tests.utils import run_gprofiler_in_container_for_one_session
                 "libc": "glibc",
             },
         ),
+        (
+            True,
+            "dotnet",
+            "dotnet-trace",
+            {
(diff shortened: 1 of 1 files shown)


netaneld122 (approved):

netaneld122 (changes requested):
Looks like this needs to be fixed https://github.com/Granulate/gprofiler/runs/8218001478?check_suite_focus=true#step:7:414

Jongy:
> Looks like this needs to be fixed https://github.com/Granulate/gprofiler/runs/8218001478?check_suite_focus=true#step:7:414

This looks like just flakiness. It's time that we start tracking flaky tests so that we can prioritize fixing them (here - the bpf program we use to find offsets appears to be flaky :shrug:)
Created https://github.com/Granulate/gprofiler/issues/462

netaneld122 (approved):

Jongy:
Flaky stuff

PR #473, opened 2022-09-15
Use read_proc_file helper from granulate-utils

d3dave:
Java, Python, and Ruby profilers rely on searching /proc/PID/maps the same way, so refactored that to a function and replaced read_proc_file with the function from granulate-utils.

### DIFF
--- a/gprofiler/utils/process.py
+++ b/gprofiler/utils/process.py
@@ -6,13 +6,14 @@
 import os
 import re
 from functools import lru_cache
+from typing import Match, Optional
 
 from granulate_utils.linux.process import process_exe, read_proc_file
 from psutil import Process
 
 
-def search_proc_maps(process: Process, pattern: str):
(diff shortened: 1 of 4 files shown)


gprofiler/utils/process.py
```diff
@@ -6,30 +6,17 @@
...
-def read_proc_file(process: Process, file: str) -> str:
-    try:
-        data = Path(f"/proc/{process.pid}/{file}").read_text()
-    except FileNotFoundError as e:
-        raise NoSuchProcess(process.pid) from e
-    else:
-        # ensures we read the right file (i.e PID was not reused)
-        ensure_running(process)
-    return data
+def search_proc_maps(process: Process, pattern: str):
```
Jongy, line 14: type annotation

Jongy (changes requested):
.

Jongy (approved):

Issue #382, opened 2022-06-18
Windows support


Jongy:
This issue lists the features / tasks we need to implement towards supporting Windows with an MVP version.

Java
* [ ] [JFR support](https://github.com/Granulate/gprofiler/issues/496) in `JavaProfiler` (Like `PythonProfiler ` which can select either eBPF or py-spy, we want `JavaProfiler` to be able to work with JFR and to allow the user to select; and on windows, automatically select JFR because async-profiler is not supported).

Python
  * [ ] Build our py-spy for Windows - this should probably be the same as Linux
General
   * [ ] Identify processes to profile in a Windows-compatible way (currently we use `grep`).
   * [ ] Make Linux-only code conditional (e.g checking for host PID NS, copying files via `/proc/pid/root`, etc)

perf
* [ ] `perf` substitute?

Build
* [ ] Package windows executable (for Linux we distribute a binary executable and a container image; for Windows windows we should probably start with an executable, and possibly later add an installer. I suppose py2exe would do?)
* [ ] Build script for Windows - we have `Dockerfile` and `pyi.Dockerfile` which build the dependencies like py-spy, async-profiler, etc; For Windows dependencies (currently py-spy) we'll need a build script that runs prior the packaging itself to download & build the dependencies.

Jongy:
#442 implemented all tasks under the "build" and "python" sections.

Java JFR has https://github.com/Granulate/gprofiler/issues/496, and "perf" (or system-wide substitute) is https://github.com/Granulate/gprofiler/issues/593.

PR #475, opened 2022-09-16
Refactoring of PythonEbpfProfiler and PythonEbpfError

teoluwaf:
## Description
Split the implementation of PythonEbpfProfiler and PythonEbpfError to its own file python_ebpf.py. Also added the implementation of is_windows to gprofiler/__init__.py as PythonEbpfProfiler conditionally imports packages if it is run on Linux.

## Related Issue
[See PR](https://github.com/Granulate/gprofiler/pull/442)


gprofiler/profilers/python_ebpf.py
```diff
@@ -0,0 +1,251 @@
...
+        random_prefix,
+        resource_path,
+        run_process,
+        start_process,
+        wait_event,
+        wait_for_file_by_prefix,
+)
+
+from gprofiler import is_windows
+if not is_windows():
```
Jongy, line 29: Instead of handling windows in this file - it is better if we refrain from importing it, PyPerf will not run on Windows as it's ebpf. Let's remove `is_windows` code from here and instead focus on avoiding to import the file

gprofiler/profilers/python.py
```diff
@@ -40,6 +40,7 @@
 from gprofiler.metadata.system_metadata import get_arch
 from gprofiler.profilers.profiler_base import ProfilerBase, ProfilerInterface, SpawningProcessProfilerBase
 from gprofiler.profilers.registry import ProfilerArgument, register_profiler
+from gprofiler.profilers.python_ebpf import PythonEbpfProfiler, PythonEbpfError
```
Jongy, line 43: So as I suggested in https://github.com/Granulate/gprofiler/pull/475/files#r973784312, let's refrain from importing it if Windows.

And then, to ensure the name `PythonEbpfProfiler` isn't used, we will:
1. Add logic in `PythonProfiler.__init__` that like the `if get_arch() != "x86_64"` will ensure we're on Linux, otherwise log a warning `PyPerf is supported only on Linux, falling back to py-spy`.
2. This way, `_create_ebpf_profiler` is never called.
3. `_create_ebpf_profiler` itself can be defined in the class `PythonEbpfProfiler` under `if is_windows()`, i.e:
```
class PythonEbpfProfiler(...):
    if is_windows():
    def _create_ebpf_profiler(
            self,
            frequency: int,
            duration: int,
            stop_event: Event,
            storage_dir: str,
            profile_spawned_processes: bool,
            add_versions: bool,
            user_stacks_pages: Optional[int],
        ) -> Optional[PythonEbpfProfiler]:
            ....
    ....
```
its code & definition never runs so it's okay `PythonEbpfProfiler` isn't defined
5. All other branches containing references to names imported from `python_ebpf` are under `if self._ebpf_profiler is not None` which will never be true, so we're protected :+1: 

After these changes, this PR is ready to go.

Jongy (commented):
2 comments

gprofiler/__init__.py
```diff
@@ -2,4 +2,12 @@
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
+import sys
+from functools import lru_cache
+
 __version__ = "1.6.0"
+WINDOWS_PLATFORM_NAME = "win32"
```
Jongy, line 9: Let's move this code to `gprofiler/platform.py`? Then you can still import it early enough.

gprofiler/profilers/python_ebpf.py
teoluwaf: Done

gprofiler/profilers/python.py
teoluwaf: Made requested changes. Please review.

gprofiler/__init__.py
teoluwaf: Done

Jongy:
@teoluwaf , you can run `./lint.sh` to check the `isort` `black` etc before the CI catches them :)

Jongy:
@teoluwaf can you please fix conflicts and I'll run the CI again?

gprofiler/profilers/python.py
```diff
@@ -566,7 +329,7 @@ def __init__(
                 logger.warning("PyPerf is supported only on x86_64, falling back to py-spy")
             python_mode = "pyspy"
 
-        if python_mode in ("auto", "pyperf"):
+        if python_mode in ("auto", "pyperf") and not is_windows():
```
Jongy, line 332: Here you implicitly avoid pyperf even if it is explicitly requested. Don't do it implicitly - you can add a `is_windows()` check to the `if` on line 564 - if we're not x86_64 OR we're Windows, then fallback to py-spy; and if pyperf was requested explicitly in that case, log `logger.warning("PyPerf is supported only on Linux x86_64, falling back to py-spy")`

gprofiler/profilers/python_ebpf.py
```diff
@@ -0,0 +1,256 @@
+import glob
```
Jongy, line 1: Please add the project's license here :)

gprofiler/profilers/python.py
Jongy: Only point 1 remains as commented [here](https://github.com/Granulate/gprofiler/pull/475#discussion_r976991116)

Jongy (changes requested):
2 final comments and it's good to go

Jongy:
```
ImportError while loading conftest '/home/runner/work/gprofiler/gprofiler/tests/conftest.py'.
tests/conftest.py:23: in <module>
    from gprofiler.metadata.application_identifiers import get_java_app_id, get_python_app_id, set_enrichment_options
gprofiler/metadata/application_identifiers.py:18: in <module>
    from gprofiler.profilers.java import jattach_path
gprofiler/profilers/__init__.py:6: in <module>
    from gprofiler.profilers.python import PythonProfiler
gprofiler/profilers/python.py:44: in <module>
    from gprofiler.profilers.python_ebpf import PythonEbpfProfiler, PythonEbpfError
gprofiler/profilers/python_ebpf.py:[19](https://github.com/Granulate/gprofiler/actions/runs/3100841093/jobs/5022305059#step:7:20): in <module>
    from gprofiler.profilers.python import PythonMetadata, _add_versions_to_stacks
E   ImportError: cannot import name 'PythonMetadata' from partially initialized module 'gprofiler.profilers.python' (most likely due to a circular import) (/home/runner/work/gprofiler/gprofiler/gprofiler/profilers/python.py)
```

This can be fixed by importing `from gprofiler.profilers import python` and then using the name `python.PythonMetadata`.

Jongy:
Related: https://github.com/Granulate/gprofiler/issues/382

gprofiler/profilers/python_ebpf.py
teoluwaf: Done

Jongy (approved):
I verified the copies themselves (i.e that code in `python_ebpf.py` is the exact copy of what was removed from `python.py`).

Issue #462, opened 2022-09-07
Flaky tests


Jongy:
An issue to track flaky tests, and mark them with X X X like "upvotes".

* `tests/test_executable.py::test_executable[False-python-pyperf]` & `tests/test_executable.py::test_executable[True-python-pyperf]` https://github.com/Granulate/gprofiler/runs/8218001478?check_suite_focus=true#step:7:414 X

PR #477, opened 2022-09-17
tests: Break up application_docker_images()

Jongy:
More lines-of-code for the same task, but the image generators are much more clear now. This code has become a hassle to work with, and now it's simple again :)

As said I'll do [here](https://github.com/Granulate/gprofiler/pull/440#discussion_r945267173).


tests/type_utils.py
```diff
@@ -2,11 +2,16 @@
...
 
 T = TypeVar("T")
 
 
 def cast_away_optional(arg: Optional[T]) -> T:
     assert arg is not None
     return arg
+
+
+def assert_cast(typ: Type[T], arg: Any) -> T:
```
Jongy, line 15: I added it but after a refactor I won't need it. Still useful, so keeping it

Jongy, line 15: I added it but after a refactor I won't need it. Still useful, so keeping it

Jongy, line 15: I added it but after a refactor I won't need it. Still useful, so keeping it

Jongy:
Documented the 2 flaky issues in https://github.com/Granulate/gprofiler/issues/462

tests/conftest.py
```diff
@@ -184,58 +183,73 @@ def gprofiler_docker_image(docker_client: DockerClient) -> Iterable[Image]:
...
+    base_path = CONTAINERS_DIRECTORY / runtime
+    return docker_client.images.build(path=str(base_path), rm=True, dockerfile=str(base_path / dockerfile), **kwargs)[0]
+
+
+def image_name(runtime: str, image_tag: str) -> str:
+    return runtime + ("_" + image_tag if image_tag else "")
+
+
 @fixture(scope="session")
 def application_docker_images(docker_client: DockerClient) -> Mapping[str, Image]:
```
netaneld122, line 198: Maybe not in this PR, but perhaps it might be better to build the images lazily rather than building them all in advance, because right now even if you run one specific test it will build all images.

netaneld122 (approved):

tests/conftest.py
Jongy: https://github.com/Granulate/gprofiler/issues/480 :(

PR #483, opened 2022-09-20
Changed granulate-utils version and code rearrangement accordingly

grariel:
## Description
Moved some methods to granulate-utils.
See: https://github.com/Granulate/granulate-utils/pull/73

## How Has This Been Tested?
Existing automatic tests only.


Jongy (approved):
@michelhe - trusting your CR in granulate-utils that the functions match. Can merge if CI passes

gprofiler/profilers/perf.py
```diff
@@ -266,44 +267,24 @@ def add_exe_metadata(self, process: Process, metadata: Dict[str, Any]) -> None:
...
         return metadata
 
 
 class NodePerfMetadata(PerfMetadata):
     def relevant_for_process(self, process: Process) -> bool:
-        return is_process_basename_matching(process, r"^node$")
+        try:
+            return is_node_process(process)
+        except MissingExePath:
+            return False
```
Jongy, line 287: While this `MissingExePath` change is logical, I don't understand why it didn't occur before... Do you?

grariel: @michelhe is to blame:
https://github.com/Granulate/granulate-utils/pull/73#discussion_r974358851

Jongy: No, that's about golang, not about the nodejs one..

michelhe: כל הזמן מאשימים אותי ורוצים להרוג אותי

image

grariel: You're right.

grariel: It seems as if the code is literally equal. Can't seem to find the reason.

michelhe (approved):

gprofiler/profilers/perf.py
michelhe: It looks necessary, i'm fine with it

michelhe (changes requested):
Ahh tests failed NVM

gprofiler/profilers/perf.py
Jongy: Yeah I guess... still weird

Jongy: I actually liked more the fix done in https://github.com/Granulate/gprofiler/pull/451 (avoids calling the function for kernel threads). The same, logically, needs to happen in the callsite for `relevant_for_process`, because we don't collect metadata for kernel threads.

Jongy:
@grariel please finish it

Jongy:
> > @grariel please finish it
> 
> It needs to be prioritized...

It will be. As a directive, we are not generating more code dups here. So if you move something to utils, we take the effort to update the profiler to use it as well

gprofiler/profilers/perf.py
michelhe: @Jongy Can you clarify what needs to be done here? we are doing positive matching (the function in utils does `r"^node"` so how we'll get kernel threads here)

Jongy: Resolving this, talked f2f

grariel: The exception handling isn't required now.
@Jongy I think the kernel threads check isn't required here.

If you're fine with it, feel free to merge/let me merge.

Jongy: I'm okay with it now, although I don't understand why it's not required. We still use the `is_process_basename_matching` function on kernel threads and it should raise. Anyway.. we'll deal when we see it.

Jongy (approved):

Jongy:
I'll merge it now

Issue #480, opened 2022-09-18
tests: Build images on demand


Jongy:
Currently `application_docker_images` attempts to build all images first, and then tests run.

If some image build fails - I want only the tests that require it to get ERROR, and others should pass okay. Easiest & simplest way to implement would be to build lazily

PR #486, opened 2022-09-21
tests: Build images lazily

Jongy:
Closes: #480

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -247,7 +247,7 @@ def application_image_tag() -> str:
     return ""
 
 
-@fixture(scope="session")  # session so it caches each image.
+@fixture
 def application_docker_image(
     docker_client: DockerClient,
     application_docker_image_configs: Mapping[str, Dict[str, Any]],
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,7 +6,7 @@ import os
 import stat
 import subprocess
 from contextlib import _GeneratorContextManager, contextmanager
-from functools import partial
+from functools import lru_cache, partial
 from pathlib import Path
 from typing import Any, Callable, Dict, Generator, Iterable, Iterator, List, Mapping, Optional, cast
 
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -233,7 +234,12 @@ def application_docker_images(docker_client: DockerClient) -> Mapping[str, Image
         for tag, args in tags_listing.items():
             name = image_name(runtime, tag)
             assert name not in images
-            images[name] = _build_image(docker_client, runtime, **args)
+
+            @lru_cache(maxsize=1)
+            def image_builder() -> Image:
+                return _build_image(docker_client, runtime, **args)
```
Jongy, line 240: Eh this won't work because `runtime` and `args` get re--bound

netaneld122 (dismissed):

tests/conftest.py
```diff
@@ -183,6 +183,7 @@ def gprofiler_docker_image(docker_client: DockerClient) -> Iterable[Image]:
     yield docker_client.images.get("gprofiler")
 
 
+@lru_cache(maxsize=1024)
```
Jongy, line 186: I implemented a cache this way, another option would be to use `fixture(scope="session") on `application_docker_image` but I prefer to have an explicit cache

netaneld122 (dismissed):

Jongy:
Bleh dicts are not cachable, I could use https://stackoverflow.com/a/53394430/797390 but I'll think of something else

Jongy:
Meh, I'll use `scope="session"`.

netaneld122 (dismissed):

Jongy:
```
ScopeMismatch: You tried to access the 'function' scoped fixture 'runtime' with a 'session' scoped request object, involved factories
```

kill me

Jongy:
```
In [26]: %timeit -n 100 docker_client.images.build(path=".", rm=True, dockerfile="Dockerfile")[0]                                                                                                          
40.1 ms ± 1.26 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

This is from building `tests/containers/java`. So Docker does enough caching for us.

Jongy:
This is same as 
image

netaneld122 (approved):

Jongy:
This already passed as https://github.com/Granulate/gprofiler/pull/486/commits/5a26241562ea73586bb7ed1a962d1f7277869387

PR #495, opened 2022-09-26
Import windows specific profilers only

teoluwaf:
Import only supported profilers on Windows

## Description
Conditionally import windows supported profilers

## Related Issue
[See PR](https://github.com/Granulate/gprofiler/pull/442)

### DIFF
--- a/gprofiler/profilers/__init__.py
+++ b/gprofiler/profilers/__init__.py
@@ -1,16 +1,17 @@
 # NOTE: Make sure to import any new process profilers to load it
-from gprofiler.profilers.dotnet import DotnetProfiler
-from gprofiler.profilers.java import JavaProfiler
+from gprofiler.platform import is_windows
 from gprofiler.profilers.perf import SystemProfiler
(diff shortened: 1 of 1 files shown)


gprofiler/profilers/__init__.py
```diff
@@ -1,16 +1,17 @@
 # NOTE: Make sure to import any new process profilers to load it
-from gprofiler.profilers.dotnet import DotnetProfiler
-from gprofiler.profilers.java import JavaProfiler
+from gprofiler.platform import is_windows
 from gprofiler.profilers.perf import SystemProfiler
```
Jongy, line 3: What about the system profiler itself? It cannot be used in Windows either. AFAICT it won't be used if we pass `--no-perf` - we can make it the default in Windows (just like the default would be `--no-dotnet`, `--no-php` etc because these are not supported.

teoluwaf: Done. Moved SystemProfiler import to non-Windows group

Jongy (approved):

PR #498, opened 2022-09-26
Log HTTP proxy upon connection error

Jongy:
Continues https://github.com/Granulate/gprofiler/pull/369 to provide more information for the users upon a connection error.

We use the same logic as `urllib` uses to extract the proxy.


Jongy:
This is how it looks:
* No proxy given, `--server-host` directed at unreachable address: `ERROR: gprofiler: Failed to connect to server. It might be blocked by your security rules / firewall, or you might require a proxy to access it from your environment? Proxy used: none. Error: HTTPSConnectionPool(host='10.0.2.4', port=443): ...`
* Proxy given (but unreachable) - `ERROR: gprofiler: Failed to connect to server. It might be blocked by your security rules / firewall, or you might require a proxy to access it from your environment? Proxy used: 'http://10.0.2.4/aaa'. Error: HTTPSConnectionPool(host='profiler.granulate.io', port=443): ...`

netaneld122 (approved):

Issue #481, opened 2022-09-18
tests: Lift requirement on java/dotnet on the host tester


Jongy:
For `dotnet_command_line` and `java_command_line` we require `dotnet` and `java` respectively to be installed on the host.

Instead of requiring them - we should copy the artifacts from the build application containers. The host tester should not require dotnet/java, or rather, should require them only for the tests that truly use them and not as a fixture that's always called. (alternatively we can make this fixture lazy)

PR #508, opened 2022-09-27
tests: Turn command_line fixture lazy

Jongy:
dotnet_command_line & java_command_line are 2 fixtures that require java/dotnet to be installed, and if they're not installed, the fixture fails. By being eager - this fails the entire test suite. Now they're lazy so the lack of java/dotnet will fail only relevant tests.

Additionally, I made the fixture cached - to avoid re-compilation of the artifacts. This will be improved in the future by avoiding the build at all and copying the artifacts from the matching container.

Related: #481

### DIFF
--- a/tests/test_java.py
+++ b/tests/test_java.py
@@ -219,7 +219,7 @@ def test_java_safemode_build_number_check(
     "in_container,java_args,check_app_exited",
     [
         (False, [], False),  # default
-        (False, ["-XX:ErrorFile=/tmp/my_custom_error_file.log"], False),  # custom error file
+        (False, ("-XX:ErrorFile=/tmp/my_custom_error_file.log",), False),  # custom error file
         (True, [], False),  # containerized (other params are ignored)
     ],
 )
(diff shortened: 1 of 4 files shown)


tests/test_java.py
```diff
@@ -219,7 +219,7 @@ def test_java_safemode_build_number_check(
     "in_container,java_args,check_app_exited",
     [
         (False, [], False),  # default
-        (False, ["-XX:ErrorFile=/tmp/my_custom_error_file.log"], False),  # custom error file
+        (False, ("-XX:ErrorFile=/tmp/my_custom_error_file.log",), False),  # custom error file
         (True, [], False),  # containerized (other params are ignored)
```
netaneld122, line 223: Please modify the java_args to be a tuple everywhere (it fails the tests)

netaneld122 (changes requested):

tests/test_java.py
Jongy: Missed it. Anyway I'm not sure if that's the only cause, let's see what's failing next

Jongy:
Flaky failure: https://github.com/Granulate/gprofiler/issues/512

Issue #492, opened 2022-09-26
Different image tags in script and test fixtures


kwisniewski98:
Tag of gprofiler image defaults to `granulate/gprofiler` in [copy_resource_from_image.sh](https://github.com/Granulate/gprofiler/blob/master/scripts/copy_resources_from_image.sh#L8), but tests expects it to be [gprofiler](https://github.com/Granulate/gprofiler/blob/master/tests/conftest.py#L183). It's just minor inconvenience, that can be overcame providing argument to `copy_resource_from_image.sh` script, but I think it would be nice to have it standardized.

It might be also good idea to provide default value for tag in image building scripts.

Jongy:
The reasoning is -
* `copy_resources_from_image.sh` is meant to be used during development process, e.g to get the latest builds of the resources - so from the latest image (`granulate/gprofiler:latest`).
* `gprofiler_docker_image` is used during the CI - we build the image but don't tag it as `granulate/gprofiler:latest` because it's just a CI image, it's not going to be published. That's built [here](https://github.com/Granulate/gprofiler/blob/4bcb3aedae1c286d3058a0dc0f559cf88fe5e940/.github/workflows/build-container-and-test.yml#L45).

In which flow did you encounter it? When running tests locally, the fixture expected the image but it wasn't found?

> It might be also good idea to provide default value for tag in image building scripts.

Yes. If your flow was the tests flow - I suppose we could swap the defaults - change the build scripts to build `-t gprofiler` by default, and have the fixture use it, and during publish we will pass `-t granulate/gprofiler:latest`.

kwisniewski98:
> In which flow did you encounter it? When running tests locally, the fixture expected the image but it wasn't found?

Yes, that was the case

If there is a reason behind this, I think we can close this issue

Jongy:
We will change the default of `build_x86_64_container.sh` to tag the image as `gprofiler` - this way the tests can copy from it.

During deploy we'll pass the deploy tag.

PR #511, opened 2022-09-28
scripts: build_x86_64_container.sh: Tag image to the name expected by tests

Jongy:
This is done by default now, to avoid confusion when running tests. Instead of requiring the user to pass this manually.

Closes: #492


.github/workflows/build-container-and-test.yml
```diff
@@ -42,7 +42,7 @@ jobs:
     # the tests need the gprofiler image built (from Dockerfile). I run it separately here, because "docker build" prints the build logs
     # more nicely. the tests will then be able to use the built image.
     - name: Build gProfiler image
-      run: ./scripts/build_x86_64_container.sh -t gprofiler
```
Jongy, line 45: Tests expect this name in the `gprofiler_docker_image` fixture

netaneld122 (approved):

PR #515, opened 2022-09-29
windows: Utils

teoluwaf:
Windows Python profiler split PR

## Description
This split PR contains changes to the metadata and utils packages. It also contains workarounds to the lack of availability of windows specific packages imported by the utils package on Linux environments. Without these workarounds, the linter fails with missing library stub errors.

## Related Issue
See [PR](https://github.com/Granulate/gprofiler/pull/442/) for complete details


gprofiler/metadata/application_identifiers.py
```diff
@@ -15,7 +15,11 @@
 from gprofiler.exceptions import CalledProcessError
 from gprofiler.log import get_logger_adapter
 from gprofiler.metadata.enrichment import EnrichmentOptions
-from gprofiler.profilers.java import jattach_path
+from gprofiler.platform import is_windows
```
Jongy, line 18: Suggestion - move relevant Java code to `gprofiler/metadata/application_identifier_java.py`, and import it & add to `_IDENTIFIERS_MAP` iff Windows?

gprofiler/metadata/system_metadata.py
```diff
@@ -142,6 +147,10 @@ def get_cpu_info() -> Tuple[str, str]:
     Parse /proc/cpuinfo to get model name & flags.
     """
     try:
+        if is_windows():
```
Jongy, line 150: This function returns flags in a format specific for Linux, so I don't want to add Windows yet which creates a different format. I suggest we don't call it in Linux at all (use unknown value for both)

gprofiler/metadata/system_metadata.py
```diff
@@ -195,14 +204,22 @@ class SystemInfo:
...
-        spawn_uptime_ms = time.clock_gettime(clock)
-    except OSError as error:
-        if error.errno != errno.EINVAL:
-            raise
-        spawn_uptime_ms = time.clock_gettime(time.CLOCK_MONOTONIC)
+    if is_windows():
+        hostname = platform.node()
+        distribution = platform.system(), platform.release(), platform.version()
+        libc_tuple = platform.libc_ver()
+        mac_address = ""  # Use netifaces pip package
```
Jongy, line 211: Let's add this `netifaces`? For mac address & IP

gprofiler/utils/__init__.py
```diff
@@ -378,6 +407,10 @@ def grab_gprofiler_mutex() -> bool:
     on filesystem structure (as happens with file-based locks).
     In order to see who's holding the lock now, you can run "sudo netstat -xp | grep gprofiler".
     """
+    # Bypass due to lack of Windows support
```
Jongy, line 410: Instead of adding it here - I think it's nicer if we add in `main.py` before the call to it:
```
if is_linux() and not grab_gprofiler_mutex():
```

(I think it's also time to add a helper `is_linux` which `sys.platform == linux`, in some cases it's easier to read than `not is_windows()`)

gprofiler/utils/__init__.py
```diff
@@ -402,6 +435,9 @@ def atomically_symlink(target: str, link_node: str) -> None:
     """
     tmp_path = link_node + ".tmp"
     os.symlink(target, tmp_path)
+    # Windows does not support renaming existing symlink
```
Jongy, line 438: This breaks the "atomically" in this function.

I suggest instead - in the call site of this function, `if is_windows()` then remove the target name with `remote_path(.., missing_ok=True)` and then call `os.symlink`.

gprofiler/utils/process.py
```diff
@@ -11,16 +11,22 @@
...
     return re.search(pattern, read_proc_file(process, "maps").decode(), re.MULTILINE)
 
 
 def process_comm(process: Process) -> str:
-    status = read_proc_file(process, "status").decode()
-    name_line = status.splitlines()[0]
-    assert name_line.startswith("Name:\t")
-    return name_line.split("\t", 1)[1]
+    if is_windows():
+        # TODO: Check if process is running
```
Jongy, line 23: What happens if the process is not running? Will it raise `NoSuchProcess`?

Jongy (changes requested):
Done reviewing

gprofiler/utils/process.py
teoluwaf: Yes, a NoSuchProcess exception is raised. I included that comment because prior to some previous commits (from another author) there used to be a check to see if the process was still running.

teoluwaf (commented):
Addressed the comments. Please review.

gprofiler/metadata/system_metadata.py
teoluwaf: Done

Jongy:
You can merge from master post #519 to solve the lint failures

gprofiler/metadata/system_metadata.py
```diff
@@ -18,8 +17,14 @@
 from granulate_utils.linux.ns import run_in_ns
 
 from gprofiler.log import get_logger_adapter
+from gprofiler.platform import is_linux, is_windows
 from gprofiler.utils import is_pyinstaller, run_process
 
+if is_linux():
+    import fcntl
+else:
+    import netifaces  # type: ignore
```
Jongy, line 26: Instead of `type: ignore` on the import, prefer to add the ignore in `mypy.ini`. This is cleaner & lets us see all used libraries with missing typing information, in one spot

gprofiler/utils/__init__.py
```diff
@@ -267,15 +282,28 @@ def run_process(
...
+        return [
+            Process(pid=p.ProcessId)
+            for p in w.Win32_Process()
+            if match in p.Name.lower() and p.ProcessId != os.getpid()
+        ]
+    else:
+        pattern = re.compile(match)
+        procs = []
+        for process in psutil.process_iter():
+            # Added to prevent AttributeError on processes that have pid == 0
```
Jongy, line 297: Can this edit be removed now that we have a separate branch for windows?

gprofiler/utils/__init__.py
```diff
@@ -267,15 +282,28 @@ def run_process(
...
-    procs = []
-    for process in psutil.process_iter():
-        try:
-            if pattern.match(process_exe(process)):
-                procs.append(process)
-        except psutil.NoSuchProcess:  # process might have died meanwhile
-            continue
-    return procs
+    if is_windows():
+        pythoncom.CoInitialize()
```
Jongy, line 286: Is it okay to call it multiple times? (i.e, I suppose it needs to be initialized only once)

gprofiler/utils/__init__.py
```diff
@@ -267,15 +282,28 @@ def run_process(
...
+        w = wmi.WMI()
+        return [
+            Process(pid=p.ProcessId)
+            for p in w.Win32_Process()
```
Jongy, line 290: Why using this library to iterate and not psutil?

gprofiler/utils/process.py
Jongy: I see. In any case, raising `NoSuchProcess` is the intended behavior and that'll be handled by caller code. I suppose that could be documented. Can you remove this comment, and add a docstring to this function, something like
```
"""
:raises NoSuchProcess: if the process isn't running
"""
```

gprofiler/metadata/system_metadata.py
```diff
@@ -194,15 +205,39 @@ class SystemInfo:
     spawn_uptime_ms: float
 
 
+def get_windows_network_details() -> Tuple[str, str]:
```
Jongy, line 208: Definition can be placed under `if is_windows()`. If needed only on Windows - strive to place under `if is_windows()`

gprofiler/metadata/system_metadata.py
```diff
@@ -194,15 +205,39 @@ class SystemInfo:
...
-    clock = getattr(time, "CLOCK_BOOTTIME", time.CLOCK_MONOTONIC)
-    try:
-        spawn_uptime_ms = time.clock_gettime(clock)
-    except OSError as error:
-        if error.errno != errno.EINVAL:
-            raise
-        spawn_uptime_ms = time.clock_gettime(time.CLOCK_MONOTONIC)
+    if is_windows():
+        hostname = platform.node()
+        distribution = platform.system(), platform.release(), platform.version()
```
Jongy, line 228: What does it return on Windows?

gprofiler/metadata/system_metadata.py
```diff
@@ -194,15 +205,39 @@ class SystemInfo:
...
+    MAC_KEY = -1000
+    for iface in netifaces.interfaces():
+        mac_address = None
+        for key, addresses in netifaces.ifaddresses(iface).items():
+            for address in addresses:
+                if key == MAC_KEY:
+                    mac_address = address[ADDR_KEY]
+                    continue
+                if re.match(IPV4_PATTERN, address[ADDR_KEY]) is not None and address[ADDR_KEY] != LOOPBACK:
+                    return cast(Tuple[str, str], (mac_address, address[ADDR_KEY]))
```
Jongy, line 221: If you expect `mac_address` to be set by this point, it's nicer to add an explicit `assert mac_address is not None`. You can also assert that both are really `str`s. Generally I prefer `assert`s to `cast`s because the former prove their correctness in runtime. Use `cast` only when it's truly very obvious from the code.

Here you can use a utility `assert_cast`, so you'll have instead
```suggestion
                    return (assert_cast(str, mac_address), assert_cast(str, address[ADDR_KEY]))
```

gprofiler/metadata/system_metadata.py
```diff
@@ -98,6 +103,8 @@ def get_mac_address() -> str:
     """
     Gets the MAC address of the first non-loopback interface.
     """
+    if is_windows():
```
Jongy, line 106: Can be removed now?

gprofiler/utils/__init__.py
```diff
@@ -32,6 +32,12 @@
 from granulate_utils.linux.process import process_exe
 from psutil import Process
 
+from gprofiler.platform import is_linux, is_windows
+
+if is_windows():
+    import pythoncom  # type: ignore
```
Jongy, line 38: Same as https://github.com/Granulate/gprofiler/pull/515/files#r990703524

gprofiler/metadata/system_metadata.py
```diff
@@ -194,15 +205,39 @@ class SystemInfo:
...
+    ADDR_KEY = "addr"
+    MAC_KEY = -1000
+    for iface in netifaces.interfaces():
+        mac_address = None
+        for key, addresses in netifaces.ifaddresses(iface).items():
+            for address in addresses:
+                if key == MAC_KEY:
+                    mac_address = address[ADDR_KEY]
+                    continue
+                if re.match(IPV4_PATTERN, address[ADDR_KEY]) is not None and address[ADDR_KEY] != LOOPBACK:
```
Jongy, line 220: Instead of the homemade RE pattern for IPv4, you can use `ipaddress.IPv4Address` and see if it raises.

gprofiler/metadata/application_identifiers.py
```diff
@@ -277,64 +281,6 @@ def get_app_id(self, process: Process) -> Optional[str]:
         return None
 
 
-class _JavaJarApplicationIdentifier(_ApplicationIdentifier):
```
Jongy, line 280: I validated the copy.

gprofiler/metadata/system_metadata.py
Jongy: Yeah I understand, that's what I meant - the function, when runs on Linux, returns data in very specific format, and the Windows values returned by your implementation will probably be very different and will mix it up... So I prefer we just use it at all on Windows, i.e `return UNKNOWN_VALUE, UNKNOWN_VALUE` if Windows.

I now notice that my comment was "we don't call it in Linux at all", I meant Windows, sorry :sweat_smile:

Jongy (changes requested):
Finished next CR round. You should also merge from master to solve existing conflicts

gprofiler/metadata/system_metadata.py
teoluwaf: platform.node() returns the hostname of your machine
platform.system() returns "Windows"
platform.release() returns '10'
platform.version() returns '10.0.19044'

gprofiler/metadata/system_metadata.py
teoluwaf: I have tried adding the missing packages to mypy.ini but still got lint errors. This was the only approach I found that worked to get past the lint errors.

This is what I added to mypy.ini file:

[netifaces]
ignore_missing_imports = True

gprofiler/metadata/system_metadata.py
teoluwaf: Done

gprofiler/metadata/system_metadata.py
teoluwaf: Done

gprofiler/utils/__init__.py
teoluwaf: See response above

gprofiler/utils/__init__.py
teoluwaf: Yes. I have never had a problem calling this multiple times and if I remember correctly, it is needed to call this function before creating an instance of wmi.WMI

gprofiler/utils/__init__.py
teoluwaf: I noticed psutil wasn't including all python processes running on Windows. This is why I found an alternative

gprofiler/utils/process.py
teoluwaf: Actually, process_comm does not raise an exception. However, the caller code raises CalledProcessError if the process is not running as follows:

`            except CalledProcessError as e:
                if (
                    b"Error: Failed to get process executable name. Check that the process is running.\n" in e.stderr
                    and not is_process_running(process)
                ):
                    logger.debug(f"Profiled process {process.pid} exited before py-spy could start")
                    return ProfileData(
                        self._profiling_error_stack("error", comm, "process exited before py-spy started"),
                        appid,
                        app_metadata,
                    )
                raise`

gprofiler/metadata/system_metadata.py
teoluwaf: Done

gprofiler/utils/__init__.py
teoluwaf: Done

gprofiler/metadata/system_metadata.py
teoluwaf: Done

gprofiler/metadata/system_metadata.py
Jongy: It should be
```
[mypy-netifaces.*]
ignore_missing_imports = True
```
(following the others)

gprofiler/utils/__init__.py
Jongy: Replied there

gprofiler/utils/__init__.py
Jongy: Okay, let's document it in the code ("psutil doesn't include all processes ...")

Jongy: If the implementation is completely separate, it is more readable to read the code if structured like this:
```
if is_windows():
    def pgrep_exe(...):
          windows one
else:
    def pgrep_exe(...):
          linux one
```

gprofiler/utils/process.py
Jongy: I get `NoSuchProcess` if `process_comm` is called with a `Process` object that's already exited (on Linux). In any case, the comment was not needed

Jongy (changes requested):
Done reviewing again :)

gprofiler/metadata/system_metadata.py
teoluwaf: Done

gprofiler/utils/__init__.py
teoluwaf: Done

gprofiler/utils/__init__.py
teoluwaf: Done

Jongy:
@teoluwaf , I'll do the final review now, please also merge from master to resolve conflicts

exe-requirements.txt
```diff
@@ -11,3 +11,5 @@ pyinstaller==4.10; platform.machine == "aarch64"
 # During startup program terminated with signal SIGSEGV, Segmentation fault.
 # staying with 0.12.1 for the mean time...
 staticx==0.12.1; platform.machine == "x86_64"
+netifaces==0.11.0; sys.platform == "win32"
```
Jongy, line 14: These (netifaces & WMI) should move to `requirements.txt`. Logically, `requirements.txt` is the general requirements file for gProfiler, and `exe-requirements.txt` is the one for requirements that are only for the pyinstaller build. These 2 libraries are part of gProfiler and not of the exe build.

gprofiler/utils/__init__.py
```diff
@@ -32,6 +32,12 @@
 from granulate_utils.linux.process import process_exe
 from psutil import Process
 
+from gprofiler.platform import is_linux, is_windows
+
+if is_windows():
+    import pythoncom
+    import wmi  # type: ignore
```
Jongy, line 39: type ignore can be moved to mypy.ini as well

Jongy (changes requested):
Done reviewing, 2 small comments, fix them & merge master and we're good to go

exe-requirements.txt
teoluwaf: Done

gprofiler/utils/__init__.py
teoluwaf: Done

Jongy (approved):
LGTM. I'll merge after CI passes

Jongy:
Seeing import errors:
> E   AttributeError: partially initialized module 'gprofiler.metadata.application_identifiers' has no attribute '_ApplicationIdentifier' (most likely due to a circular import)

Jongy:
Also, PyInstaller complains about the installations of Windows-specific libraries missing on Linux (this is from the CI):
```
#127 58.80 missing module named wmi - imported by gprofiler.utils (conditional)
#127 58.80 missing module named pythoncom - imported by gprofiler.utils (conditional)
#127 58.80 missing module named netifaces - imported by gprofiler.metadata.system_metadata (conditional)
#127 58.80 missing module named tests - imported by gprofiler.metadata.system_metadata (top-level)
#127 58.80 PyInstaller failed to pack gProfiler code! See lines above. Make sure to check for SyntaxError as this is often the reason.
```

I bet there is some way to tell it "this import isn't really needed", or "needed only on platform == windows.

requirements.txt
```diff
@@ -14,3 +14,4 @@ websocket-client==1.3.1
 netifaces==0.11.0; sys.platform == "win32"
 WMI==1.5.1; sys.platform == "win32"
 ./granulate-utils/
+./tests
```
Jongy, line 17: What is that for?

teoluwaf: assert_cast. I removed it it and added the implementation of assert_cast to gprofiler/metadata/system_metadata.py

Jongy: Ah I see, if needed then ofc move it to `gprofiler/utils` instead of installing the tests package :)

Jongy:
@teoluwaf you can debug the build locally with `./build_x86_64_executable.sh`

Jongy (approved):

Jongy:
I checked that `check_pyinstaller.sh` works as expected, i.e - injecting an error causes the build to fail.

PR #517, opened 2022-10-06
Use snapshot_pid function instead of snapshot_one in tests

mpozniak95:
## Description
Functions snapshot_one_profile and snapshot_one_collapse are replaced by snapshot_pid_profile and snapshot_one_collapse in test_java.py and test_sanity.py tests.

## Related Issue
https://github.com/Granulate/gprofiler/issues/514

## How Has This Been Tested?
Run test_java and test_sanity

## Checklist:
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have updated the relevant documentation.
- [ ] I have added tests for new logic.

Issue #513, opened 2022-09-29
tests: test_from_container_spawned_process flaky failure - app starts "too late"


Jongy:
See test failure of test_from_container_spawned_process[False-ruby-rbspy] [here](https://github.com/Granulate/gprofiler/actions/runs/3142740525/jobs/5110359098).

It seems that `SpawningProcessProfilerBase` didn't catch the new Ruby process. Maybe it started too late (although, the entire idea in this state is to catch Ruby processes starting late...). In any case, we see the Ruby in the `perf` native profile.

Test needs to `assert` that the Ruby profiler worked (check logs as well) and maybe wait more for app start?

Jongy:
Also note, as stated in #514 , this leads to the Ruby process leaking the test (not being killed afterwards)

PR #519, opened 2022-10-06
lint: Upgrade flake8

Jongy:
## Description

Trying to solve https://github.com/Granulate/gprofiler/actions/runs/3173626903/jobs/5220214768.

I searched online for the error and got to this one: https://stackoverflow.com/a/73932581/ which matches what I suggested [here](https://github.com/Granulate/gprofiler/pull/517#issuecomment-1270023442). By upgrading flake8 I now see in its downloaded dependencies that it adds the necessary constraint: `Collecting importlib-metadata<4.3,>=1.1.0`.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,5 @@
 pytest==6.2.4
-flake8==3.9.2
+flake8==5.0.4
 black==22.3.0
 mypy==0.942
 isort==5.8.0


Jongy:
Seems to have solved the problem. I'll wait for tests to pass because this could change other dependencies as well.

Jongy:
Flaky failure: https://github.com/Granulate/gprofiler/issues/513

PR #531, opened 2022-10-13
remove obsolete line-ends

wzur-r3:
do not add line ends when going through the list of `extraArguments`


Jongy:
Hi @wzur-r3 , thanks for the PR :) I'm not very acquainted with Helm YAML syntax so I'll take some time to study what the effect of this change are, and then merge it.

wzur-r3:
Hi @Jongy - it simply removes empty lines generated when this loop is processed. If you looked at any other template directives, even in the same file, you would see then all have `~` to avoid these empty lines

wzur-r3:
@Jongy I'm going to close this PR.

Our company no longer uses Granulate and after 3 months of inactivity one would though that particular issue has been fixed in the main branch.

PR #532, opened 2022-10-13
Add --no-dotnet flag to profiler flags fixture in tests

mpozniak95:
## Description
This PR is adding --no-dotnet flag to profiler_flags fixture to avoid using dotnet profiler while testing other profilers.


Jongy (approved):
2 flaky failures, I will add to the flaky tests issue later.

Issue #503, opened 2022-09-27
python: app metadata: NoSuchProcess causing ERROR logs


Jongy:
Flaky [test failure](https://github.com/Granulate/gprofiler/actions/runs/3131599146/jobs/5083709691) stemming from:
```
[2022-09-26 23:57:39,114] ERROR: gprofiler.metadata.application_metadata: Exception while collecting metadata in PythonMetadata!
Traceback (most recent call last):
  File "gprofiler/metadata/application_metadata.py", line 67, in get_metadata
  File "gprofiler/profilers/python.py", line 143, in make_application_metadata
  File "granulate_utils/linux/elf.py", line 37, in get_elf_id
  File "granulate_utils/linux/elf.py", line 18, in get_elf_buildid
FileNotFoundError: [Errno 2] No such file or directory: '/proc/3128/exe'
[2022-09-26 23:57:39,116] DEBUG: gprofiler.profilers.profiler_base: PySpyProfiler: process went down during profiling 3128 (lsb_release)
Traceback (most recent call last):
  File "granulate_utils/linux/process.py", line 104, in _translate_errors
  File "granulate_utils/linux/process.py", line 69, in read_proc_file
FileNotFoundError: [Errno 2] No such file or directory: '/proc/3128/status'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "gprofiler/profilers/profiler_base.py", line 119, in _wait_for_profiles
  File "concurrent/futures/_base.py", line 425, in result
  File "concurrent/futures/_base.py", line 384, in __get_result
  File "concurrent/futures/thread.py", line 56, in run
  File "gprofiler/profilers/python.py", line 203, in _profile_process
  File "gprofiler/utils/process.py", line 20, in process_comm
  File "granulate_utils/linux/process.py", line 70, in read_proc_file
  File "contextlib.py", line 99, in __exit__
  File "granulate_utils/linux/process.py", line 114, in _translate_errors
psutil.NoSuchProcess: psutil.NoSuchProcess process no longer exists (pid=3128)
```

This should not be an ERROR - this error should be ignored silently as it is okay for processes to go down.

Only the ERROR one is a problem here.

Jongy:
Another example where this happens and fails the test: https://github.com/Granulate/gprofiler/actions/runs/3105234649/jobs/5030894845

Jongy:
[...]

PR #534, opened 2022-10-13
Update granulate-utils submodule to master

mpozniak95:
## Related Issue
https://github.com/Granulate/gprofiler/issues/503


Jongy:
I think it will fail because https://github.com/Granulate/gprofiler/pull/483 wasn't merged yet and it has some fixes after updating granulate-utils. We will get to this PR after that one is merged

Jongy:
https://github.com/Granulate/gprofiler/pull/483 ended up updating granulate-utils to a newer commit, which includes your changes from https://github.com/Granulate/granulate-utils/pull/87, so we don't need it anymore :) Closing this one.

Issue #538, opened 2022-10-14
tests: nodejs: flaky attach-maps


Jongy:
See 2 test failures here: https://github.com/Granulate/gprofiler/actions/runs/3243777862/jobs/5320303422. They are flaky.

In both cases, the error is:
```
[2022-10-13 17:29:46,323] WARNING: gprofiler.profilers.node: Could not create debug symbols for pid 22987. Reason: HTTPConnectionPool(host='127.0.0.1', port=9229): Max retries exceeded with url: /json/list (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f1691fd6370>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

At least in one of them, I see this:
```
--------------------------- Captured stdout teardown ---------------------------
stdout: 
stderr: Debugger listening on ws://127.0.0.1:9229/19543cbe-cfe6-44e3-9a74-991b667f88e5
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.
```

which tells us that the debug port was indeed opened and we did manage to connect to it... So this failure isn't obvious to me. Is it because we wait too little time for the debugger to start?

mpozniak95:
I think this can be closed right @Jongy ?

Jongy:
Yes, what I described here is resolved by https://github.com/Granulate/gprofiler/pull/561/files.

Issue #523, opened 2022-10-10
python: pyperf: get_stack_offset randomly fails


Jongy:
[Flaky test failures](https://github.com/Granulate/gprofiler/issues/462) stemming from `get_stack_offset` occasionally failing:
```
[2022-10-06 12:22:46,079] DEBUG: gprofiler.utils: (['/tmp/_MEIg4dLPw/gprofiler/resources/python/pyperf/get_stack_offset']) stderr: 'stack not found, searched for 16384 bytes into task_struct\n'
[2022-10-06 12:22:46,079] DEBUG: gprofiler.profilers.python: eBPF profiler error: Command '['/tmp/_MEIg4dLPw/gprofiler/resources/python/pyperf/get_stack_offset']' returned non-zero exit status 1. 
stdout: b''
stderr: b'stack not found, searched for 16384 bytes into task_struct\n'
```
(excerpt from [this failure](https://github.com/Granulate/gprofiler/actions/runs/3196892055/jobs/5219800478)).

This might be only some issue in the test, but might also be a real problem occuring in the wild.

Jongy:
When it happens - it does seem to happen multiple times for the same test run (i.e the same problem reproduces on the same kernel / state).

Jongy:
Another failure example - https://github.com/Granulate/gprofiler/actions/runs/3220091322/jobs/5266664858

PR #540, opened 2022-10-14
Avoid using const multiple times for one variable in nodejs profiler

mpozniak95:
## Description
Instead of assigning `process.mainModule.require("fs")` to a const, now we call `readlinkSync("/proc/self/ns/pid")` directly.

## Motivation and Context
This PR should fix problems with: 
`gprofiler.profilers.node: Could not create debug symbols for pid Reason: SyntaxError: Identifier 'fs' has already been declared` 
`Traceback (most recent call last):
  File "gprofiler/profilers/node.py", line 129, in _execute_js_command
KeyError: 'value'`

### DIFF
--- a/gprofiler/profilers/node.py
+++ b/gprofiler/profilers/node.py
@@ -145,7 +145,7 @@ def _change_dso_state(sock: WebSocket, module_path: str, action: str) -> None:
 
 
 def _validate_ns_node(sock: WebSocket, expected_ns_link_name: str) -> None:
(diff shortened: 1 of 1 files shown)


Jongy (approved):
Looks good. What is the KeyError, though? Do we get that when the code we eval raised an error? Let's handle that on our end. Can be a separate PR.

Jongy:
Flaky failures: https://github.com/Granulate/gprofiler/issues/523 and https://github.com/Granulate/gprofiler/issues/538

Issue #521, opened 2022-10-09
tests: nodejs: Full matrix tests for supported NodeJS versons in attach-maps mode


Jongy:
Basically like https://github.com/Granulate/gprofiler/pull/478 but for the new `attach-maps` mode we added in #451 (matrix of - all versions supported by it X libc type)

Jongy:
App only inside container is enough for these tests (we test the NodeJS version support specifically).

PR #543, opened 2022-10-18
Add matrix tests nodejs

mpozniak95:
## Description
This PR adds matrix tests for nodejs profiler (for different nodejs versions)

## Related Issue
https://github.com/Granulate/gprofiler/issues/521


tests/conftest.py
```diff
@@ -240,6 +240,20 @@ def application_docker_image_configs() -> Mapping[str, Dict[str, Any]]:
         "nodejs": {
             "": dict(buildargs={"NODE_RUNTIME_FLAGS": "--perf-prof --interpreted-frames-native-stack"}),
             "without-flags": dict(buildargs={"NODE_RUNTIME_FLAGS": ""}),
+            "10-alpine": dict(dockerfile="matrix.Dockerfile", buildargs={"NODE_IMAGE_TAG": "10.24.1-alpine"}),
```
Jongy, line 243: Instead of `10-alpine` and `10-slim`, better names are `10-musl` and `10-glibc` - that's what we're testing after all, and `slim` / `alpine` images are only the implemntation detail. Let's change for all

tests/containers/nodejs/matrix.Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG NODE_IMAGE_TAG
```
Jongy, line 1: I suppose we could use the same Dockerfile? This can be the only version we maintain (named `Dockerfile`), and we just need to add the `node@sha256:59531d2835edd5161c8f9512f9e095b1836f7a1fcb0ab73e005ec46047384911` as a build arg to the `""` and `"without-flags"` images.

tests/test_node.py
```diff
@@ -67,3 +67,46 @@ def test_nodejs_attach_maps_from_container(
...
+    output_directory: Path,
+    output_collapsed: Path,
+    runtime_specific_args: List[str],
+    profiler_flags: List[str],
+    application_image_tag: str,
+    ) -> None:
+    _ = application_pid
+    node_version = application_image_tag.split("-")[0]
+    if node_version in ["14", "15", "16"]:
+        pytest.xfail("This test fails with these nodejs versions")
```
Jongy, line 107: I created issue https://github.com/Granulate/gprofiler/issues/550 for it, you can refer to it in this xfail message here.

tests/test_node.py
```diff
@@ -67,3 +67,46 @@ def test_nodejs_attach_maps_from_container(
...
+    collapsed_text = run_gprofiler_in_container_for_one_session(
```
Jongy, line 108: For simpler & faster tests - we don't need a full run of the gProfiler container here, let's initialize `SystemProfiler` and use it here, i.e like `test_nodejs_attach_maps`.

Jongy:
@mpozniak95 , about [the failure](https://github.com/Granulate/gprofiler/actions/runs/3288931948/jobs/5420231530), it seems like the flakiness of attach-maps as described [here](https://github.com/Granulate/gprofiler/issues/538). A connection could not be made post starting the debugger.

Jongy (changes requested):
Done reviewing :)

mpozniak95:
Addressed comments :)

Jongy (dismissed):
LGTM, let's see the tests

mpozniak95:
@Jongy it works on my env:
tests/test_node.py::test_nodejs_matrix[nodejs-attach-maps-12-musl-True] PASSED

Jongy:
I re-ran the failing ones, let's see

Jongy (approved):

Jongy (approved):

PR #544, opened 2022-10-18
Update perf version to v6.0g1

### LINKED ISSUE #493: perf: understand/eliminate [unknown] frames
What are `[unknown]` frames (when neither the symbol name nor DSO name are present). Is JITted code the reason?

IlayRosenberg:
## Description

Update our perf version to include the latest version of the linux sources. This should resolve #493 thanks to the fix in https://github.com/Granulate/linux/commit/363afa3aef24f5e08df6a539f5dc3aae4cddcc1a

## Motivation and Context

This was mainly done in order to fix #493 

## How Has This Been Tested?

I ran a small program locally which reproduces the `[unknown]` frames in the graph by spawning threads until the tid becomes large enough to be alphabetically smaller (for example "100" < "70"). This reproduces the issue which is just incorrect thread initialization order in perf script. I then verified that with the mentioned fix the issue does not reproduce.

## Checklist:

### DIFF
--- a/scripts/perf_build.sh
+++ b/scripts/perf_build.sh
@@ -6,7 +6,7 @@
 set -euo pipefail
 
 # downloading the zip because the git is very large (this is also very large, but still smaller)
-curl -SL https://codeload.github.com/Granulate/linux/zip/37e33e68d0a02f31f46e089d4a33d9877ed79768 -o linux.zip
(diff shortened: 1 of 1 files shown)


Jongy:
I will review the perf merge, then approve and merge this

IlayRosenberg:
@Jongy I manually tested again using my local program and a perf version I built from `v6.0g1` and it looks good

Jongy (approved):

Issue #524, opened 2022-10-10
"Upload profiling data" mode


Jongy:
We'd like to add a new operation mode to gProfiler, where it can take the output of external profilers and convert it to the protocol accepted by the server & upload the profiles. This is helpful as an intermediate step before incorporating a new profiler into gProfiler - we will be able to run any external profiler, and use gProfiler to upload the profiles and have them visible in the Performance Studio UI.

As an MVP, what I have in mind is adding a subcommand to `gprofiler` - for example, `gprofiler upload-profile`. This command will accept a file path as input (and additionally, other parameters like `--token` and `--service-name`, those that are still relevant).

The input file will be in the stackcollapsed format, without the container name & app metadata as first frames (so gProfiler needs to specify that in the profile header send as the first line of the profile). As if gProfiler was run with `--disable-container-names` and `--disable-application-metadata` - you can run the profiler this way and see the generated profile, then prepare the necessary conversions on the input file to get it like the server would expect.
Additionally, allow the fille will begin with a comment line `#` like our profiles, and might contain the metadata keys that we send to the server, in the same format as we use, such as `hostname` or `instance_type`. If any keys are present in the input file, they will be used in the profile that gProfiler submits to the server. If keys such as `hostname` are not present, gProfiler will fill them in (this way, the UI has everything it needs to operate).

Implementation wise - it makes sense to avoid instantiating the `GProfiler` class - you need only the `APIClient` class. Some refactors may be required to achieve that. Also, preferably - allow pasing only the relevant arguments (so e.g all profiler parameters are irrelevant; `--profile-spawned-processes` is irrelevant; parameters about logging are not irrelevant. For that we will need to break arguments into groups maybe, and add only the relevant groups for the argparse subcommand `upload-profile`).

Tip - use `--curlify-requests` which shows the final profile being uploaded & parameters sent by gProfiler.

Sample metadata line (you can get it by running `gprofiler -o /tmp` and you'll have it in the `last_profile.col` file.
```
[...]

PR #546, opened 2022-10-19
"Upload profiling data" mode

pfilipko1:
New subcommand: upload-file

## Description
Addition of a new subcommnad that enables uploading external .col files to Granulate Performance Studio.

## Related Issue
https://github.com/Granulate/gprofiler/issues/524

## Motivation and Context
It adds new feature, that enables

## How Has This Been Tested?
It was tested manually, using sample .col data without metadata and container names in such scenarios: 

- right file, right path
- wrong file, right path
- wrong path

## Checklist:


gprofiler/main.py
```diff
@@ -463,6 +508,13 @@ def parse_cmd_args() -> configargparse.Namespace:
         default=False,
         help="Whether to upload the profiling results to the server",
     )
+    parser.add_argument(
+        "--upload-collapsed-file",
+        action="store",
+        type=str,
+        default="",
```
Jongy, line 515: Don't set `default` - it will be `None` which makes more sense

README.md
```diff
@@ -111,6 +111,10 @@ gProfiler can be run in a continuous mode, profiling periodically, using the `--
 Note that when using `--continuous` with `--output-dir`, a new file will be created during *each* sampling interval.
 Aggregations are only available when uploading to the Granulate Performance Studio.
 
+### Uploading profiling data from another source
+gProfiler can be used to upload external .col file to Granulate Performance Studio by using `--upload-collapsed-file <path-to-.col>` flag.
+In case of gProfiler run volume mapping flag must be added. Example: `docker run --name granulate-gprofiler -v <path-to.col>:/home/outputdir/ --pid=host --userns=host --privileged  gprofiler:latest --token=<token> --service-name="<service>" --upload-collapsed-file /home/outputdir/<name>.col -u`
```
Jongy, line 116: UX wise, I think `--upload-collapsed-file` should be enough (so `-u` needn't be given if that one's given).

Alternatively - enforce having `-u` given if `--upload-collapsed-file` is given

gprofiler/main.py
```diff
@@ -382,6 +382,51 @@ def run_continuous(self) -> None:
...
+    metrics = NoopSystemMetricsMonitor().get_metrics()
+    # TODO:container names man, remember them
+    merged_result, total_samples = concatenate_from_external_file(
+        args.upload_collapsed_file,
+        None,
+        enrichment_options,
+        metadata,
+        metrics,
+    )
+    local_end_time = local_start_time + datetime.timedelta(seconds=(time.monotonic() - monotonic_start_time))
```
Jongy, line 404: I think these 2 should also be accepted in the file - so file creator can specify the timeframe. Let's define it as 2 fields in the first line of the collapsed (2 UTC timestamps) and try to load them from there. If not given - use a NULL timestamp (i.e UTC timestamp of the epoch)

Jongy:
Not done reviewing

gprofiler/main.py
pfilipko1: Done, info in the README.md

gprofiler/merge.py
```diff
@@ -332,6 +332,38 @@ def _enrich_and_finalize_stack(
...
+    """
+    total_samples = 0
+    lines = []
+    start_time = None
+    end_time = None
+    # TODO: container names and application metadata
+    with open(collapsed_file_path) as file:
+        for index, line in enumerate(file):
+            if index == 0:
+                data = json.loads(line[2:])
```
Jongy, line 351: You should enrich the metadata here with the result of `get_current_metadata`. Values in `data` take precedence, so we need to deep merge `data` onto the resulting dict from `get_current_metadata`.

gprofiler/main.py
```diff
@@ -463,6 +503,13 @@ def parse_cmd_args() -> configargparse.Namespace:
         default=False,
         help="Whether to upload the profiling results to the server",
     )
+    parser.add_argument(
```
Jongy, line 506: Did you try adding it as a subcommand? I think it makes more sense UX wise, and also allows us to structure the code better. With the current way this is written, the preconditions for gProfiler run are still checked in the upload mode, despite there's no need for that (checking for root, checking for init PID NS, ...).

What I had in mind is - `./gprofiler upload-file` will be a subcommand accepting only the relevant parameters - let's begin with `token`, `service-name` and later if needed we can add others such as `server-upload-timeout`. It will reject other parameters because they are truly not used. `./gprofiler` will run the default behavior (i.e what we currently have) and accept all arguments.
See [this](https://stackoverflow.com/questions/6365601/default-sub-command-or-handling-no-sub-command-with-argparse) SO question with an example of how to do that with `argparse`, and then there's a way to define the argument `--token` once, and have it available both for the default parser (`./gprofiler`) and for the upload subparser (`./gprofiler upload-file`)

gprofiler/main.py
```diff
@@ -824,29 +871,31 @@ def main() -> None:
         )
 
         set_enrichment_options(enrichment_options)
-
-        gprofiler = GProfiler(
```
Jongy, line 828: If we change it to be a subcommand, this diff will be eliminated because we won't need to change anything here

Jongy (changes requested):
Left some comments, we still need to change the design IMO, mostly with argparse

gprofiler/merge.py
```diff
@@ -332,6 +332,38 @@ def _enrich_and_finalize_stack(
...
 
+def concatenate_from_external_file(
+    collapsed_file_path: str,
+    container_names_client: Optional[ContainerNamesClient],
+) -> Tuple[str, str, str, int]:
+    """
+    Concatenate all stacks from all stack mappings in process_profiles.
+    Add "profile metadata" and metrics as the first line of the resulting collapsed file.
+    """
+    total_samples = 0
```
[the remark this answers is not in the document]
Jongy: It's not - I created https://github.com/Granulate/gprofiler/pull/556 to remove it, you can merge it to your branch meanwhile, or wait for it to be merged

README.md
```diff
@@ -111,6 +111,11 @@ gProfiler can be run in a continuous mode, profiling periodically, using the `--
 Note that when using `--continuous` with `--output-dir`, a new file will be created during *each* sampling interval.
 Aggregations are only available when uploading to the Granulate Performance Studio.
 
+### Uploading profiling data from another source
+gProfiler can be used to upload external .col file to Granulate Performance Studio by using `--upload-collapsed-file <path-to-.col>` flag.
```
Jongy, line 115: `--upload-collapsed-file` that's old, needs to be updated

gprofiler/merge.py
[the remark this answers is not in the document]
Jongy: It can be removed now

gprofiler/merge.py
```diff
@@ -332,6 +333,46 @@ def _enrich_and_finalize_stack(
...
+    end_time = None
+
+    # TODO: container names and application metadata
+    with open(collapsed_file_path) as file:
+        for index, line in enumerate(file):
+            if index == 0:
+                read_metadata = json.loads(line[2:])
+                metadata = _merge_metadata(read_metadata, obtained_metadata)
+                try:
+                    start_time = datetime.datetime.strptime(metadata["start_time"], "%Y-%m-%dT%H:%M:%S.%f")
```
Jongy, line 367: Let's move parsing to a utility function `parse_iso8601_timestamp`? Place it next to `get_iso8601_format_time`.

gprofiler/merge.py
```diff
@@ -332,6 +333,46 @@ def _enrich_and_finalize_stack(
...
+    total_samples = 0
+    lines = []
+    start_time = None
+    end_time = None
+
+    # TODO: container names and application metadata
+    with open(collapsed_file_path) as file:
+        for index, line in enumerate(file):
+            if index == 0:
+                read_metadata = json.loads(line[2:])
```
Jongy, line 364: Check if line begins with comment (i.e `#`, and then the json itself can be immediately after it `#{...}` or with a space `# {}`. I suggest you just `json.loads(line[1:])` and it'll work for both cases

gprofiler/merge.py
```diff
@@ -332,6 +333,46 @@ def _enrich_and_finalize_stack(
...
+                metadata = _merge_metadata(read_metadata, obtained_metadata)
+                try:
+                    start_time = datetime.datetime.strptime(metadata["start_time"], "%Y-%m-%dT%H:%M:%S.%f")
+                    end_time = datetime.datetime.strptime(metadata["end_time"], "%Y-%m-%dT%H:%M:%S.%f")
+                except KeyError:
+                    pass
+            lines.append(line.rstrip())
+
+    return start_time, end_time, "\n".join(lines), total_samples
```
Jongy, line 373: total_samples can be removed

gprofiler/main.py
```diff
@@ -815,6 +890,11 @@ def main() -> None:
         if client is not None and remote_logs_handler is not None:
             remote_logs_handler.init_api_client(client)
 
+        if hasattr(args, "func"):
+            if args.subcommands == "upload-file":
```
Jongy, line 894: "else" what? It can't happen right? So instead
```suggestion
            assert args.subcommands == "upload-file"
```

gprofiler/merge.py
```diff
@@ -332,6 +333,46 @@ def _enrich_and_finalize_stack(
     return f"{enrich_data.application_prefix}{enrich_data.container_prefix}{stack} {count}"
 
 
+def _merge_metadata(primary, secondary):
```
Jongy, line 336: Typing please :) `Dict[str, Any]` suffices

Jongy, line 336: This is fairly generic. Let's move to utils and keep with generic name `merge_dicts`? Also names like `source` and `dest` are more clear IMO, I liked them [here](https://stackoverflow.com/a/20666342/797390)

gprofiler/merge.py
Jongy, line 373: Metadata is not returned?

gprofiler/main.py
```diff
@@ -604,13 +679,13 @@ def parse_cmd_args() -> configargparse.Namespace:
     args.perf_inject = args.nodejs_mode == "perf"
     args.perf_node_attach = args.nodejs_mode == "attach-maps"
 
-    if args.upload_results:
+    if args.upload_results or args.file_path:
```
Jongy, line 682: In one site you use `args.upload_results or args.subcommands == "upload-file"` and in this site you use `args.upload_results or args.file_path`. Let's be consistent. You can create a helper function that gets the `args` object and returns true/false if we should be uploading.

Alternatively - and possible simpler - is to add after the `parse_args` call:
```
if args.subcommand == "upload-file":
   args.upload_results = True
```
then we don't get the 2 branches of the code.

gprofiler/main.py
```diff
@@ -463,18 +522,34 @@ def parse_cmd_args() -> configargparse.Namespace:
...
-        "--curlify-requests", help="Log cURL commands for HTTP requests (used for debugging)", action="store_true"
+
+    subparsers = parser.add_subparsers(dest="subcommands")
+    upload_file = subparsers.add_parser("upload-file")
+    upload_file.add_argument(
+        "--file-path",
+        action="store",
+        type=str,
+        default=None,
+        help="Path for the collapsed file to be",
```
Jongy, line 533:
```suggestion
        help="Path for the collapsed file to be uploaded",
```

gprofiler/main.py
```diff
@@ -463,18 +522,34 @@ def parse_cmd_args() -> configargparse.Namespace:
...
-    parser.add_argument(
-        "--curlify-requests", help="Log cURL commands for HTTP requests (used for debugging)", action="store_true"
+
+    subparsers = parser.add_subparsers(dest="subcommands")
+    upload_file = subparsers.add_parser("upload-file")
+    upload_file.add_argument(
+        "--file-path",
+        action="store",
+        type=str,
+        default=None,
```
Jongy, line 532: The default of `default` is None, I think :) No need to specify

gprofiler/main.py
```diff
@@ -463,18 +522,34 @@ def parse_cmd_args() -> configargparse.Namespace:
...
-    parser.add_argument("--token", dest="server_token", help="Server token")
-    parser.add_argument("--service-name", help="Service name")
-    parser.add_argument(
-        "--curlify-requests", help="Log cURL commands for HTTP requests (used for debugging)", action="store_true"
+
+    subparsers = parser.add_subparsers(dest="subcommands")
+    upload_file = subparsers.add_parser("upload-file")
+    upload_file.add_argument(
+        "--file-path",
+        action="store",
```
Jongy, line 530: That's the default for `action`, no need to specify?

gprofiler/main.py
```diff
@@ -382,6 +373,74 @@ def run_continuous(self) -> None:
...
+        logger.error("Upload of profile to server timed out.")
+    except APIError as e:
+        logger.error(f"Error occurred sending profile to server: {e}")
+    except RequestException:
+        logger.exception("Error occurred sending profile to server")
+    else:
+        logger.info("Successfully uploaded profiling data to the server")
+
+
+def send_collapsed_file_only(args, client):
```
Jongy, line 409: Typing please

gprofiler/main.py
```diff
@@ -382,6 +373,74 @@ def run_continuous(self) -> None:
...
+    spawn_time = time.time()
+    gpid = ""
+    metrics = NoopSystemMetricsMonitor().get_metrics()
+    if args.collect_metadata:
+        _static_metadata = get_static_metadata(spawn_time=spawn_time, run_args=args.__dict__)
+    metadata = (
+        get_current_metadata(cast(Metadata, _static_metadata))
+        if args.collect_metadata
+        else {"hostname": get_hostname()}
+    )
```
Jongy, line 419:
```suggestion
    if args.collect_metadata:
        metadata = get_current_metadata(get_static_metadata(spawn_time=spawn_time, run_args=args.__dict__)))
    else:
        metadata = {"hostname": get_hostname()}
```
simpler

gprofiler/main.py
```diff
@@ -382,6 +373,74 @@ def run_continuous(self) -> None:
                     break
 
 
+def _submit_profile_logged(
+    client: APIClient,
+    start_time: datetime.datetime,
+    end_time: datetime.datetime,
+    profile: str,
+    total_samples: int,
```
Jongy, line 381: Don't need `total_samples`

gprofiler/main.py
```diff
@@ -313,26 +313,17 @@ def _snapshot(self) -> None:
...
-                    local_start_time,
-                    local_end_time,
-                    merged_result,
-                    total_samples,
-                    self._profile_api_version,
-                    self._spawn_time,
-                    metrics,
-                    self._gpid,
-                )
-                self._gpid = response_dict.get("gpid", "")
```
Jongy, line 327: `self._gpid` is saved across sessions - you can return it from `_submit_profile_logged` and write it back to `self._gpid` here.

gprofiler/main.py
```diff
@@ -382,6 +382,46 @@ def run_continuous(self) -> None:
...
+    # TODO:container names, application metadata, remember them
+    start_time, end_time, merged_result, total_samples = concatenate_from_external_file(
+        args.upload_collapsed_file,
+        None,
+    )
+    local_end_time = local_start_time + datetime.timedelta(seconds=(time.monotonic() - monotonic_start_time))
+    if start_time is not None and end_time is not None:
+        local_start_time = datetime.datetime.strptime(start_time, "%Y-%m-%d %H:%M:%S.%f")
+        local_end_time = datetime.datetime.strptime(end_time, "%Y-%m-%d %H:%M:%S.%f")
+    try:
```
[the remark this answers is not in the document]
Jongy: See https://github.com/Granulate/gprofiler/pull/546/files#r1008565473

gprofiler/merge.py
Jongy: Or actually, it should be embedded back into the first profile line. It's currently not done

Jongy:
If I run with no args, I get:
```
root@40f149145ccb:/app# python3 -m gprofiler
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/app/gprofiler/__main__.py", line 8, in <module>
    main()
  File "/app/gprofiler/main.py", line 799, in main
    args = parse_cmd_args()
  File "/app/gprofiler/main.py", line 682, in parse_cmd_args
    if args.upload_results or args.file_path:
AttributeError: 'Namespace' object has no attribute 'file_path'
```

Let's ensure the "standard" behavior is kept in that case, i.e the same error message that we used to get before `upload-file`.

gprofiler/main.py
```diff
@@ -463,18 +522,34 @@ def parse_cmd_args() -> configargparse.Namespace:
...
+    subparsers = parser.add_subparsers(dest="subcommands")
+    upload_file = subparsers.add_parser("upload-file")
+    upload_file.add_argument(
+        "--file-path",
+        action="store",
+        type=str,
+        default=None,
+        help="Path for the collapsed file to be",
     )
+    for subparser in [parser, upload_file]:
```
Jongy, line 535: Cool. Liked this solution :)

gprofiler/main.py
```diff
@@ -382,6 +373,74 @@ def run_continuous(self) -> None:
...
+    )
+    # TODO:container names, application metadata
+    local_start_time, local_end_time, merged_result, total_samples = concatenate_from_external_file(
+        args.file_path,
+        metadata,
+    )
+
+    if local_start_time is None and local_end_time is None:
+        assert (
+            local_start_time is None and local_end_time is None
```
Jongy, line 428: Assert is the same condition as the `if`, so it will never be triggered. I suppose the `if` is meant to be `local_start_time is None or local_end_time is None`? (notice the `and` -> `or` diff)

gprofiler/main.py
Jongy: Good now :)

Jongy (changes requested):
Finished reviewing, got some comments, overall looks good - that's the way this API & UX should be.

gprofiler/main.py
pfilipko1: Are we sure there will be no other subcommands in the future?

Jongy: Once we add them, the assert will trigger and we will remove it :)

Jongy:
Many lint errors, please fix

Jongy:
Also, the main repo's master appears a few commits ahead from your fork's master, you should pull the main fork master before merging "master" (otherwise you're not as up to date as could be)

README.md
```diff
@@ -118,6 +118,11 @@ gProfiler can be run in a continuous mode, profiling periodically, using the `--
 Note that when using `--continuous` with `--output-dir`, a new file will be created during *each* sampling interval.
 Aggregations are only available when uploading to the Granulate Performance Studio.
 
+### Uploading profiling data from another source
+gProfiler can be used to upload external .col file to Granulate Performance Studio by using `upload-file` subcommand.
+In case of gProfiler run volume mapping flag must be added. Example: `docker run --name granulate-gprofiler -v <path-to-.col>:<path-to-.col> --pid=host --userns=host --privileged  gprofiler:latest upload-file  --token=<token> --service-name="<service>" --file-path <path-to-.col>`
```
Jongy, line 123: Mention that it's only for docker mode (maybe you meant "in case of docker run, ..."?)

gprofiler/utils/__init__.py
```diff
@@ -357,6 +357,10 @@ def get_iso8601_format_time(time: datetime.datetime) -> str:
     return time.replace(microsecond=0).isoformat()
 
 
+def parse_iso8601_timestamp(time: str):
+    return datetime.datetime.strptime(time, "%Y-%m-%dT%H:%M:%S.%f")
```
Jongy, line 361:
```suggestion
    return datetime.datetime.fromisoformat(time)
```

gprofiler/merge.py
```diff
@@ -332,6 +333,41 @@ def _enrich_and_finalize_stack(
...
+            if index == 0:
+                read_metadata = json.loads(line[1:])
+                metadata = merge_dicts(read_metadata, obtained_metadata)
+                try:
+                    start_time = parse_iso8601_timestamp(metadata["start_time"])
+                    end_time = parse_iso8601_timestamp(metadata["end_time"])
+                except KeyError:
+                    pass
+                try:
+                    metadata["run_arguments"]["func"] = str(metadata["run_arguments"]["func"])
```
Jongy, line 361: I suppose you're casing it here to replace the "function" object with a string?

In any case, we don't need it, so you can just ensure `func` is popped out from `run_arguments` (we pass `args.__dict__` to `GProfiler` class, you can pass a copy of the dict with `func` popped). Please explain in comment why it's done, because this code isn't clear

gprofiler/merge.py
```diff
@@ -332,6 +333,41 @@ def _enrich_and_finalize_stack(
...
+    """
+
+    lines = []
+    start_time = None
+    end_time = None
+
+    # TODO: container names and application metadata
+    with open(collapsed_file_path) as file:
+        for index, line in enumerate(file):
+            if index == 0:
```
Jongy, line 352: Ensure the line starts with `#` as the format expects. (can be `assert line.startswith(...)`)

gprofiler/main.py
```diff
@@ -465,18 +520,32 @@ def parse_cmd_args() -> configargparse.Namespace:
...
-        help="Timeout for upload requests to the server in seconds (default: %(default)s)",
-    )
-    parser.add_argument("--token", dest="server_token", help="Server token")
-    parser.add_argument("--service-name", help="Service name")
-    parser.add_argument(
-        "--curlify-requests", help="Log cURL commands for HTTP requests (used for debugging)", action="store_true"
+
+    subparsers = parser.add_subparsers(dest="subcommand")
+    upload_file = subparsers.add_parser("upload-file")
+    upload_file.add_argument(
```
Jongy, line 526: Make it a required argument?

gprofiler/main.py
```diff
@@ -817,6 +889,11 @@ def main() -> None:
         if client is not None and remote_logs_handler is not None:
             remote_logs_handler.init_api_client(client)
 
+        if hasattr(args, "func"):
```
Jongy, line 892: This runs after `verify_preconditions()`, so we still require root & init PID NS. Let's ensure these are not requried, I think best would be to refrain from calling `verify_preconditions()` if we're in `upload-file` mode.

Jongy (changes requested):
I managed to try it, works fine :) I have no more UX comments - `./gprofiler` behaves good when passing `--help`, `upload-file --help` and no arguments at all. I also managed to upload a file. We can close the existing comments and we're done

gprofiler/merge.py
pfilipko1: You are right, the purpose was to omit json errors when function object was encountered. I'll pop it from run arguments then.

gprofiler/main.py
```diff
@@ -384,6 +375,69 @@ def run_continuous(self) -> None:
...
+        )
+    except Timeout:
+        logger.error("Upload of profile to server timed out.")
+    except APIError as e:
+        logger.error(f"Error occurred sending profile to server: {e}")
+    except RequestException:
+        logger.exception("Error occurred sending profile to server")
+    else:
+        logger.info("Successfully uploaded profiling data to the server")
+    return response_dict.get("gpid", "")
```
Jongy, line 406: If an exception was raised - `response_dict` is not assigned and this will result in an exception.

I think this code can be placed in `else` clause of the `try`, and at the end of the function, as a fallback, you can use the default value of `""`.

gprofiler/main.py
```diff
@@ -730,7 +802,8 @@ def init_pid_file(pid_file: str) -> None:
 
 def main() -> None:
     args = parse_cmd_args()
-    verify_preconditions(args)
+    if args.subcommand is None:
```
Jongy, line 805:
```suggestion
    if args.subcommand != "upload-file":
```

Let's be conservative, and be explicit for that case.

Jongy (approved):

Issue #497, opened 2022-09-26
nodejs: perf.map attach: Allow 2 running processes in the same net namespace


Jongy:
If port is already opened it won't open. We can have one of two solutions for having multiple node processes in same network namespace:

1.  Close debugger after generating perf maps and starting it each time we want to do any operation on debugger. In this method no two debuggers will be open at the same time. This will also be useful when someone started container with i. e. `--network=host`. (I've started implementing that)
2.  We can change debug port, but that can only be done inside process (we need to change `process.debugPort`). This means we need to start it on port 9229 anyway, then change `debugPort` and restart debugger.

In both methods we have to start it at 9229 port and I didn't find any solution for that.

Answering your other questions:
If module is already loaded, CDP request will respond with False when loading module
I'll add this sanity check

_Originally posted by @kwisniewski98 in https://github.com/Granulate/gprofiler/pull/451#discussion_r979972958_

kwisniewski98:
I'm thinking it should be done this way:

1. Open debugger and connect to it, like it is done now.
2. Do what should be done with `replMode: true`, this way changes should stay even when debugger is closed.
3. Close debugger port executing `process._debugEnd()` via `Runtime.evaluate`

Alternately it can be done like that:

1. Open debugger and connect to it, like it is done now.
2. Change debugger port via executing `process.debugPort=<some_number>` via `Runtime.evaluate`. This way we can leave debugger port open.

I'm worried that second way will cause problems when we restart gProfiler, since we won't know at which port debugger is running.

PR #547, opened 2022-10-20
nodejs: Allow running two processes in the same net namespace

mpozniak95:
## Description
This PR adds closing debugger after his job is done.

## Related Issue
https://github.com/Granulate/gprofiler/issues/497

## How Has This Been Tested?
Run two nodejs applications, then run gprofiler. Previously such situation resulted in `"Could not clean up debug symbols for pid Y. Reason: Wrong pid, expected Y, actual X"`


Jongy:
Good solution.

Can you add a test that runs 2 node apps and proves both are profiled? You can convert the `test_nodejs_attach_maps` test (in its container mode) to use `application_factory`, then you can use it like:
```
with application_factory() as pid1:
  with application_factory() as pid2:
    # profile here
```

I think that'll work. Then you use `profiler.snapshot()` and run the standard `assert_collapsed` on profiles of both PIDs.

Jongy (approved):
LGTM, I'll merge if tests pass

Jongy:
Very weird. `FAILED tests/test_node.py::test_nodejs_matrix[nodejs-attach-maps-12-musl-True]` on 4 but passed on one. Does it pass locally for you?

Jongy:
> Very weird. `FAILED tests/test_node.py::test_nodejs_matrix[nodejs-attach-maps-12-musl-True]` on 4 but passed on one. Does it pass locally for you?

Huh I thought I wrote it in #543 :sweat_smile:

tests/test_node.py
```diff
@@ -67,3 +68,33 @@ def test_nodejs_attach_maps_from_container(
...
+                6,
+                Event(),
+                str(tmp_path),
+                False,
+                perf_mode="fp",
+                perf_inject=False,
+                perf_dwarf_stack_size=0,
+                perf_node_attach=True,
+            ) as profiler:
+                process1_collapsed = snapshot_pid_collapsed(profiler, pid1)
```
Jongy, line 97: Here, instead of invoking the profiler twice, you could run `results = profiler.snapshot()` once, and then get `results[pid1]` and `results[pid2]` from the results dict

Jongy:
Conflicts exist - merge master?

mpozniak95:
Yep, done

Jongy (dismissed):

Jongy:
Let's merge master & remove the `test_preconditions.py` change and see if now things pass :)

mpozniak95:
All tests passed :)

Jongy (approved):

PR #554, opened 2022-10-23
Updated granulate-utils version and used is_kernel_process function

grariel:
The purpose of this PR is to avoid checking if kernel threads are golang/node processes, hence adding https://github.com/Granulate/granulate-utils/pull/95.

## Description
Used `is_kernel_process` function.

## How Has This Been Tested?
Existing automatic tests only.

### DIFF
--- a/gprofiler/utils/__init__.py
+++ b/gprofiler/utils/__init__.py
@@ -300,7 +300,6 @@ else:
         procs = []
         for process in psutil.process_iter():
             try:
-                # kernel threads should be child of process with pid 2
                 if not is_kernel_thread(process) and pattern.match(process_exe(process)):
                     procs.append(process)
             except psutil.NoSuchProcess:  # process might have died meanwhile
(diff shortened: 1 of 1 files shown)


gprofiler/utils/__init__.py
```diff
@@ -301,7 +301,7 @@ def pgrep_exe(match: str) -> List[Process]:
         for process in psutil.process_iter():
             try:
                 # kernel threads should be child of process with pid 2
```
Jongy, line 303: Comment is not related here anymore. Maybe move to granulate-utils

Jongy:
@grariel - please check it by running gProfiler in high frequency (e.g 2k) and seeing that on master, the `MissingExePath` is triggered in the golang version check flow; and on your branch, it doesn't trigger anymore.

Jongy (approved):
Merge after performing the requested test

PR #561, opened 2022-10-26
Fix flaky NodeJS tests

mpozniak95:
## Description
This PR:
- increases timeout to creating connection with debugger socket in NodeJS profiler.
- adds sleep in fibonacci.js application to lower CPU usage

## Related Issue
https://github.com/Granulate/gprofiler/issues/538

## Motivation and Context
NodeJS Tests are flaky.

### DIFF
--- a/gprofiler/profilers/node.py
+++ b/gprofiler/profilers/node.py
@@ -159,7 +159,7 @@ def _validate_pid(expected_pid: int, sock: WebSocket) -> None:
 
 def create_debugger_socket(nspid: int, ns_link_name: str) -> WebSocket:
     debugger_url = _get_debugger_url()
-    sock = create_connection(debugger_url)
+    sock = create_connection(url=debugger_url, timeout=15.0)
     sock.settimeout(10)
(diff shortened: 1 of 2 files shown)


gprofiler/profilers/node.py
```diff
@@ -61,6 +62,7 @@ def _get_dest_inside_container(musl: bool, node_version: str) -> str:
 def _start_debugger(pid: int) -> None:
     # for windows: in shell node -e "process._debugProcess(PID)"
     os.kill(pid, signal.SIGUSR1)
+    sleep(2)
```
Jongy, line 65: Instead of a sleep here, I think it makes more sense to re-try the connection if it fails early

mpozniak95:
Run CI/CD 3 times, there was one fail in `tests/test_sanity.py::test_from_container[True-nodejs-perf]` but I think it is not connected with issue that this PR is resolving. Created new issue here: https://github.com/Granulate/gprofiler/issues/562

Jongy (approved):

Issue #529, opened 2022-10-12
Limit gProfiler memory usage


Jongy:
The easiest way to place a hard limit on the memory used by gProfiler, as gProfiler is a privileged app, is to use cgroups. We have such API in granulate-utils - see [this file](https://github.com/Granulate/granulate-utils/blob/master/granulate_utils/linux/cgroups/memory_cgroup.py). We can use it to place gProfiler in a cgroup & set its limit to e.g 1GB to ensure we don't grow too much, e.g due to https://github.com/Granulate/gprofiler/issues/528

Jongy:
The same can be applied on CPU limits

Creatone:
Hello @Jongy! I am interested in working on this issue. Can I be assigned to it?

Jongy:
Hi @Creatone ! You're welcome to try it :)

Let's define an MVP:
* Only cgroups v1 support (cgroups v2 will be done later).
* gProfiler will check if it's already in a cgroup (this check is already done in [move_to_cgroup](https://github.com/Granulate/granulate-utils/blob/6e6c028b8abde4beb811d5929a335fd5bcd33449/granulate_utils/linux/cgroups/base_cgroup.py#L58) for "predefined" ones, which are Docker/k8s/ECS. So we can just use the API of `move_to_cgroup` and handle the raised exception if we're already in a cgroup).
* The default limits should be as [set in the k8s DaemonSet](https://github.com/Granulate/gprofiler/blob/59f85a9a22a4a9218f307332244c5f638f909998/deploy/k8s/gprofiler.yaml#L35).
* Allow passing `--limit-memory` and `--limit-cpu` which override the defaults. Units are bytes for memory (int) and cores (float, 1 == 1 core) for CPU. If `-1` is given it implies no limit, but you can still move to the cgroup, as this can help us do measurements of utilization, as explained in https://github.com/Granulate/gprofiler/issues/227. After this issue (#529), I want `--log-usage` to work for exe mode as well (i.e I expect this issue to resolve #227 as well).
* Allow passing `--no-cgroups` to avoid doing ANY cgroups changes (moving / changing limits etc). It's always helpful to have a kill switch.
* Try moving into cgroups & limiting as early as possible, but after the logger is initialized, so that you can log errors.
* If any error happens during cgroups handling, swallow it and log (i.e don't block gProfiler).
[...]

Issue #227, opened 2021-11-27
--log-usage support in exe mode


Jongy:
`--log-usage` lets us track the CPU & memory usage of gProfiler. It does so by reading cgroups files for the cgroup gProfiler runs in. See https://github.com/Granulate/gprofiler/blob/89fcf8e271706029bedd486b2b35727e92058086/gprofiler/usage_loggers.py for how it's currently done.

This works in containers, as we can just use the cgroup files of the container.
In "executable" mode, however, we don't run in separate cgroups, so measuring the existing cgroups will not yield the correct result.

If we move ourselves into a child cgroup of the current one, we can then track that new one.

Note: we sometimes run in semi-containers, such as Databricks / Dataflow / Fargate, where we're an executable but we're on low privileges so I don't know if we can create new cgroups. We need to handle this case as well (that is, do not fail hard if we fail to move cgroups)

Resources:
* [cgroups memory controller](https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt) and specifically `usage_in_bytes`.
* [cgroups cpuacct controller](https://www.kernel.org/doc/Documentation/cgroup-v1/cpuacct.txt)
* Moving cgroups - new process in cgrouup with [cgexec](https://linux.die.net/man/1/cgexec)
* Moving cgroups - [existing process in cgroup](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/resource_management_guide/sec-moving_a_process_to_a_control_group) - this is probably what we should use; not `cgclassify` but "manually" writing our PID to the `tasks` file of a new cgroup.
* [Creating cgroups](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/resource_management_guide/sec-creating_cgroups)

Cgroups are per-thread actually. So we should move our thread to a new cgroup as early as possible (before other threads/processes are spawned), and make sure it's really the only thread in the process once we do so.

Jongy:
https://github.com/Granulate/granulate-utils/pull/56 lets us create & move between cgroups

PR #564, opened 2022-10-27
Limit gProfiler memory & CPU usage and --log-usage support in exe mode.

Creatone:
## Description

TBA / WIP

## Related Issue

#529 #227

## Motivation and Context

TBA / WIP

## How Has This Been Tested?

TBA / WIP

## Checklist:

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -614,7 +614,7 @@ def parse_cmd_args() -> configargparse.Namespace:
         default=DEFAULT_MEMORY_LIMIT,
         dest="memory_limit",
         type=int,
-        help="Limit on the memory used by gProfiler. Units are bytes and the default is %DEFAULT_MEMORY_LIMIT."
+        help=f"Limit on the memory used by gProfiler. Units are bytes and the default is '{DEFAULT_MEMORY_LIMIT}'."
     )
 
     parser.add_argument(
@@ -622,7 +622,7 @@ def parse_cmd_args() -> configargparse.Namespace:
         default=DEFAULT_CPU_LIMIT,
         dest="cpu_limit",
         type=float,
-        help="Limit on the cpu used by gProfiler. Units are cores and the default is %DEFAULT_CPU_LIMIT."
+        help=f"Limit on the cpu used by gProfiler. Units are cores and the default is '{DEFAULT_CPU_LIMIT}'."
     )
 
     parser.add_argument(
@@ -790,7 +790,7 @@ def main() -> None:
     # TODO(Creatone): Check the containerized scenario.
     if args.cgroups_changes and get_run_mode() not in ("k8s", "container"):
         logger.info(f"Trying to set resource limits, cpu='{args.cpu_limit}' "
-                    f"cores and memory='{args.memory_limit >> 20}' MB.")
+                    f"cores and memory='{args.memory_limit >> 20:.2f}' MB.")
         try:
             set_limits(args.cpu_limit, args.memory_limit)
         except Exception:
--- a/gprofiler/usage_loggers.py
+++ b/gprofiler/usage_loggers.py
@@ -14,7 +14,7 @@ import psutil
 from granulate_utils.linux.cgroups.cgroup import find_v1_hierarchies, find_v2_hierarchy
 
 
-# TODO(Creatone): Move it to granulate-utils.
+# TODO(Creatone): Move it to granulate-utils. Consider change.
 def _obtain_cgroup_controller_path(cgroup: str, controller: str) -> str:
     cgroup_v1_hierarchies = find_v1_hierarchies()
     if len(cgroup_v1_hierarchies) != 1:
--- a/tests/test_cgroups.py
+++ b/tests/test_cgroups.py
@@ -1,0 +1,59 @@
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+import os
+import subprocess
+from pathlib import Path
+from subprocess import Popen
+from typing import List
+
+import pytest
+from docker import DockerClient
+from docker.models.images import Image
+
+from tests.utils import run_privileged_container, _print_process_output
+
+
+def test_cgroup_limit_container(
+        docker_client: DockerClient,
+        gprofiler_docker_image: Image,
+        output_directory: Path,
+) -> None:
+    logs = run_privileged_container(docker_client, gprofiler_docker_image,
+                                    command=['-v', '--limit-cpu', '0.5', '--limit-memory', '1048576', '-o',
+                                             str(output_directory)])
+
+    limit_log = "Trying to set resource limits, cpu='0.5' cores and memory='1024.00' MB."
+
+    assert limit_log not in logs
+
+
+def test_cgroup_limit_privileged_executable(
+    gprofiler_exe: Path,
+    output_directory: Path,
+) -> None:
+    os.mkdir(output_directory)
+
+    command = (
+        ['sudo', str(gprofiler_exe), '-v', '--limit-cpu', '0.5',
+         '--limit-memory', str((1 << 30)), '-o', str(output_directory), "-d", "5",
+         "--no-java", "--no-python", "--no-php", "--no-ruby", "--no-nodejs", "--no-dotnet"]
+    )
+
+    popen = Popen(command, stdout=subprocess.PIPE)
+    assert popen.wait() == 0
+    stdout, _ = popen.communicate()
+    logs = stdout.decode("utf-8").splitlines()
+    limit_log = "Trying to set resource limits, cpu='0.5' cores and memory='1024.00' MB."
+
+    present = False
+    for line in logs:
+        if limit_log in line:
+            present = True
+    assert present
+
+
+# Not implemented yet.
+def test_cgroup_try_limit_no_privileged_executable():
+    assert False


gprofiler/main.py
```diff
@@ -602,6 +605,30 @@ def parse_cmd_args() -> configargparse.Namespace:
...
+        default=0.5,  # 500m, same as in the k8s DaemonSet
+        dest="cpu_limit",
+        type=float,
+        help="Limit on the cpu used by gProfiler."
+    )
+
+    parser.add_argument(
+        "--no-cgroups",
+        action="store_true",
+        dest="disable_cgroups",
```
Jongy, line 627:
```suggestion
        dest="cgroups_changes",
```
Code is more clear if it works on the "positive" case and not the "negative" case. The argument can then be `store_false`.

gprofiler/main.py
```diff
@@ -602,6 +605,30 @@ def parse_cmd_args() -> configargparse.Namespace:
...
 
+    parser.add_argument(
+        "--limit-memory",
+        default=(1 << 30),  # 1Gi, same as in the k8s DaemonSet
+        dest="memory_limit",
+        type=int,
+        help="Limit on the memory used by gProfiler."
+    )
+
+    parser.add_argument(
```
Jongy, line 616: How do I specific no CPU limit? i.e `--limit-cpu none` should be a possible parameter, not necessarily this way, but you should be able to limit ONLY the memory or CPU.

gprofiler/main.py
```diff
@@ -723,6 +744,45 @@ def init_pid_file(pid_file: str) -> None:
...
 
+# Set limits and return path of the cgroup.
+def set_limits(cpu: float, memory: int) -> str:
+    cgroups = {}
+    logger.debug("Check if cgroup version is supported.")
+    try:
+        cgroups["cpu"] = CpuCgroup()
+        cgroups["memory"] = MemoryCgroup()
+    except UnsupportedCGroupV2:
+        logger.error("cgroup v2 is not supported by gProfiler, cpu and memory limits wouldn't be set.")
```
Jongy, line 755: Created ticket https://github.com/Granulate/gprofiler/issues/567 for it.

Jongy, line 755: Use `debug` log here. It's not an error, it's a yet unsupported "failsafe" mechanism, the user needn't see it as an error.

gprofiler/main.py
```diff
@@ -723,6 +744,45 @@ def init_pid_file(pid_file: str) -> None:
     Path(pid_file).write_text(str(os.getpid()))
 
 
+# Set limits and return path of the cgroup.
+def set_limits(cpu: float, memory: int) -> str:
+    cgroups = {}
```
Jongy, line 749: Any reason to use this dict? Why not 2 variables `cpu_cgroup` and `memory_cgroup`?

gprofiler/main.py
```diff
@@ -723,6 +744,45 @@ def init_pid_file(pid_file: str) -> None:
...
+    logger.debug("Check if cgroup version is supported.")
+    try:
+        cgroups["cpu"] = CpuCgroup()
+        cgroups["memory"] = MemoryCgroup()
+    except UnsupportedCGroupV2:
+        logger.error("cgroup v2 is not supported by gProfiler, cpu and memory limits wouldn't be set.")
+        return
+
+    logger.debug("Prepare gProfiler cpu cgroup.")
```
Jongy, line 758: Too verbose IMO, we don't need this log

gprofiler/main.py
```diff
@@ -723,6 +744,45 @@ def init_pid_file(pid_file: str) -> None:
...
+    try:
+        cgroups["cpu"].move_to_cgroup("gprofiler", os.getpid())
+    except AlreadyInCgroup:
+        logger.warning("gProfiler have already a cpu group.")
+
+    logger.debug("Set cpu limit in the cgroup.")
+    cgroups["cpu"].set_cpu_limit_cores(cpu)
```
Jongy, line 765:
```suggestion
    try:
        cgroups["cpu"].move_to_cgroup("gprofiler", os.getpid())
    except AlreadyInCgroup:
        logger.warning("gProfiler already has a cpu group")
    else:
        cgroups["cpu"].set_cpu_limit_cores(cpu)
```

also removed the "set cpu limit" log, I think it's too verbose. You're already logging once that you are about to set resource limits - no need to do so again. Log only in the error / skip case - failed to set limits OR decided not to set limits due to already in cgroup.

Jongy: Please make the same change for the memory part below.

gprofiler/main.py
```diff
@@ -723,6 +744,45 @@ def init_pid_file(pid_file: str) -> None:
     Path(pid_file).write_text(str(os.getpid()))
 
 
+# Set limits and return path of the cgroup.
+def set_limits(cpu: float, memory: int) -> str:
+    cgroups = {}
+    logger.debug("Check if cgroup version is supported.")
```
Jongy, line 750: This log is not needed IMO. As I suggested in https://github.com/Granulate/gprofiler/pull/564/files#r1007993619, log only the negative case.

gprofiler/main.py
```diff
@@ -738,10 +798,20 @@ def main() -> None:
         remote_logs_handler,
     )
 
+    # check if there is no kill switch for managing cgroups
```
Jongy, line 801: Comment is redundant. You're just checking the CLI argument here

gprofiler/usage_loggers.py
```diff
@@ -11,7 +11,17 @@
...
+
+
+# TODO(Creatone): Move it to granulate-utils.
+def _obtain_cgroup_controller_path(cgroup: str, controller: str) -> str:
+    cgroup_v1_hierarchies = find_v1_hierarchies()
+    if len(cgroup_v1_hierarchies) != 1:
+        assert controller in cgroup_v1_hierarchies
+        return f"{cgroup_v1_hierarchies[controller]}{cgroup}"
+    else:
+        return f"{find_v2_hierarchy()}/{controller}{cgroup}"
```
Jongy, line 24: I don't think this will work now - because cgroups v2 files are different. For example - there is no `cpuacct.usage` file, but `cpu.stat` file which contains a `usage_usec` field.

I created a [ticket](https://github.com/Granulate/gprofiler/issues/567) for cgroups v2 support. Until that's done, I suggest you raise an exception here if v2 is in use.

Also I see that granulate-utils does this check: `len(get_cgroups(os.getpid())) == 1` (in `_verify_preconditions`). It's subtly different from what you did here (checks the controllers available for this processes instead of the mounted controllers). I think it'll produce the same results, but let's be consistent.

gprofiler/main.py
```diff
@@ -738,10 +798,20 @@ def main() -> None:
...
+    # TODO(Creatone): Check the containerized scenario.
+    cgroup = "/" # assume we run in the root cgroup (when containerized, that's our view)
+    if not args.disable_cgroups and get_run_mode() not in ("k8s", "container"):
+        logger.info(f"Trying to set resource limits, cpu='{args.cpu_limit}' and memory='{args.memory_limit}'.")
```
Jongy, line 805:
```suggestion
        logger.info(f"Trying to set resource limits, cpu='{args.cpu_limit}' cores and memory='{args.memory_limit}' bytes.")
```

Or maybe

```suggestion
        logger.info(f"Trying to set resource limits, cpu='{args.cpu_limit}' cores and memory='{args.memory_limit / 1024 / 1024}' MB.")
```

gprofiler/main.py
```diff
@@ -723,6 +744,45 @@ def init_pid_file(pid_file: str) -> None:
...
+    except AlreadyInCgroup:
+        logger.warning("gProfiler have already a memory group.")
+
+    logger.debug("Set memory limit in the cgroup.")
+    cgroups["memory"].set_limit_in_bytes(memory)
+
+    return cgroups['cpu'].cgroup
+
+
+def setup_usage_logger(log_usage: bool, cgroup: str) -> UsageLoggerInterface:
```
Jongy, line 779: Unused :) (and I also prefer keeping it as `CgroupsUsageLogger(logger, cgroup) if args.log_usage else NoopUsageLogger()`)

gprofiler/main.py
```diff
@@ -723,6 +744,45 @@ def init_pid_file(pid_file: str) -> None:
...
+    logger.debug("Prepare gProfiler memory cgroup.")
+    try:
+        cgroups["memory"].move_to_cgroup("gprofiler", os.getpid())
+    except AlreadyInCgroup:
+        logger.warning("gProfiler have already a memory group.")
+
+    logger.debug("Set memory limit in the cgroup.")
+    cgroups["memory"].set_limit_in_bytes(memory)
+
+    return cgroups['cpu'].cgroup
```
Jongy, line 776: This API is weird. This function possibly moved the CPU cgroup, then possibly moves the memory cgroup. Moving the memory cgroup can fail, in which case you don't return the moved CPU cgroup, and the caller expects it to remain `/`.

Instead of returning a value, I think it's better to resolve the actual cgroup AFTER this function completes running.

Creatone:
@Jongy To be honest I added "[WIP]" to the title to show you that I'm working on that. There will be plenty of force pushes there. Many things will look differently. I wasn't guessing that you'll review that already. :v:

Jongy (changes requested):
Finished reviewing.

Let's test:
* [ ] cgroup limitting when a container - gprofiler should not move / make changes.
* [ ] cgroup limitting when a privileged executable - gprofiler should move & make changes
* [ ] cgroup limitting when a non-privileged executable - gprofiler should try to move, fail, and continue running okay (this can be tested by running as root but with dropped capabilities.. happens in AWS Fargate, Databricks and a few other platforms)
I have a few more ideas, will add

Jongy:
> 

Oh sorry :sweat_smiley: I saw that it's not a draft so I started. Nevermind, no problem

Jongy:
By the way - we eventually use squash-and-merge so there is no need to force push, you can just add commits on top of it. But up to you

Creatone:
@Jongy Ahhhh it is my mistake, nevertheless, the code will be better reviewed :)

Creatone:
@Jongy Could you check the tests from c4be935ad69994033730037d1a8d484c988a7502?
It is a good way to check your conditions?

Jongy:
> @Jongy Could you check the tests from [c4be935](https://github.com/Granulate/gprofiler/commit/c4be935ad69994033730037d1a8d484c988a7502)? It is a good way to check your conditions?

The tests check if gProfiler behaved as predicted, but I prefer if we add true "end-to-end" tests, i.e - the test will actually check the cgroup limit imposed on the gProfiler process and see that it matches what we expected. You can get the gProfiler PID, get its cgroup, verify that it actually moved to a new cgroup named `gprofiler`, and that the limits of that cgroup are what we expect. You can use the same APIs from granulate-utils to read the actual limits set on the cgroup.

Issue #550, opened 2022-10-21
tests: nodejs: Fix matrix test for NodeJS 14, 15, 16


Jongy:
Test is added in https://github.com/Granulate/gprofiler/pull/543 and we found out the mechanism fails for versions 14, 15, 16.

This most likely means the attach-maps feature fails for these versions. @mpozniak95 will check.

PR #566, opened 2022-10-28
Fix using attach_maps on NodeJS 14+

mpozniak95:
## Description
This PR removes experimental replMode (it is not needed in our case because we are not declaring any variables, and it seems to not work on NodeJS 14+). Also this PR adds closing socket after closing debugger.

## Related Issue
https://github.com/Granulate/gprofiler/issues/550

## How Has This Been Tested?
Run test_nodejs_matrix

### DIFF
--- a/gprofiler/profilers/node.py
+++ b/gprofiler/profilers/node.py
@@ -39,7 +39,7 @@ class NodeDebuggerProcessUndefined(Exception):
     pass
 
 
-class ExpectedResult():
+class ExpectedResult:
     STRING = 1
     INTEGER = 2
     BOOLEAN = 3
--- a/gprofiler/profilers/node.py
+++ b/gprofiler/profilers/node.py
@@ -129,7 +129,7 @@ def _execute_js_command(sock: WebSocket, command: str) -> Any:
         "id": 1,
         "method": "Runtime.evaluate",
         "params": {
-            "expression": f"process === undefined ? 'process undefined' : {command}",
+            "expression": f"typeof(process) === 'undefined' ? 'process undefined' : {command}",
         },
     }
     sock.send(json.dumps(cdp_request))
@@ -139,7 +139,10 @@ def _execute_js_command(sock: WebSocket, command: str) -> Any:
     except json.JSONDecodeError:
         raise NodeDebuggerUnexpectedResponse(message) from None
     try:
-        if isinstance(message["result"]["result"]["value"], str) and "process undefined" in message["result"]["result"]["value"]:
+        if (
+            isinstance(message["result"]["result"]["value"], str)
+            and message["result"]["result"]["value"] == "process undefined"
+        ):
             raise NodeDebuggerProcessUndefined(message)
         else:
             return message["result"]["result"]["value"]
@@ -149,12 +152,12 @@ def _execute_js_command(sock: WebSocket, command: str) -> Any:
 
 def _change_dso_state(sock: WebSocket, module_path: str, action: str) -> None:
     assert action in ("start", "stop"), "_change_dso_state supports only start and stop actions"
-    command = f'process.mainModule === undefined ? "process undefined" : process.mainModule.require("{os.path.join(module_path, "linux-perf.js")}").{action}()'
+    command = f'typeof(process.mainModule) === "undefined" ? "process undefined" : process.mainModule.require("{os.path.join(module_path, "linux-perf.js")}").{action}()'
     cdp_request = {
         "id": 1,
         "method": "Runtime.evaluate",
         "params": {
-            "expression": f'process === undefined ? "process undefined" : {command}',
+            "expression": f'typeof(process) === "undefined" ? "process undefined" : {command}',
         },
     }
     _send_socket_request(sock, cdp_request)
@@ -166,7 +169,7 @@ def _close_debugger(sock: WebSocket) -> None:
         "id": 1,
         "method": "Runtime.evaluate",
         "params": {
-            "expression": "process === undefined ? 'process undefined' : process._debugEnd()",
+            "expression": "typeof(process) === 'undefined' ? 'process undefined' : process._debugEnd()",
         },
     }
     sock.send(json.dumps(cdp_request))
@@ -176,7 +179,7 @@ def _close_debugger(sock: WebSocket) -> None:
         sock.close()
     else:
         try:
-            if "process undefined" in message["result"]["result"]["value"]:
+            if message["result"]["result"]["value"] == "process undefined":
                 raise NodeDebuggerProcessUndefined(message)
             else:
                 raise NodeDebuggerUnexpectedResponse(message)
--- a/gprofiler/profilers/node.py
+++ b/gprofiler/profilers/node.py
@@ -162,7 +162,6 @@ def _change_dso_state(sock: WebSocket, module_path: str, action: str) -> None:
     _evaluate_js_command(sock, command, ExpectedResult.BOOLEAN)
 
 
-@retry(NodeDebuggerProcessUndefined, 5, 0.5)
 def _close_debugger(sock: WebSocket) -> None:
     command = "process._debugEnd()"
     _evaluate_js_command(sock, command, ExpectedResult.NONE)
--- a/gprofiler/profilers/node.py
+++ b/gprofiler/profilers/node.py
@@ -10,7 +10,7 @@ import stat
 from functools import lru_cache
 from pathlib import Path
 from threading import Event
-from typing import Any, Dict, List, cast
+from typing import Any, List, cast
 
 import psutil
 import requests
--- a/gprofiler/profilers/node.py
+++ b/gprofiler/profilers/node.py
@@ -39,6 +39,13 @@ class NodeDebuggerProcessUndefined(Exception):
     pass
 
 
+class ExpectedResult():
+    STRING = 1
+    INTEGER = 2
+    BOOLEAN = 3
+    NONE = 4
+
+
 def _get_node_major_version(process: psutil.Process) -> str:
     node_version = get_exe_version(process, Event(), 3)
     # i. e. v16.3.2 -> 16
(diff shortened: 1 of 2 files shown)


Jongy:
Interesting. What effect does `replMode` have when given to NodeJS 10-12?

Jongy:
Closing the socket is just something forgotten earlier, I suppose. And good to do explicitly, because implicit close as the object is GCed can happen anytime, and this introduces inconsistent behavior indeed...

mpozniak95:
image
According to: https://chromedevtools.github.io/devtools-protocol/v8/Runtime/#method-evaluate

Jongy:
I see this failure in the logs:
```
gprofiler.profilers.node.NodeDebuggerUnexpectedResponse: {'id': 1, 'result': {'result': {'type': 'object', 'subtype': 'error', 'className': 'ReferenceError', 'description': 'ReferenceError: process is not defined', 'objectId': '{"injectedScriptId":1,"id":1}'}, 'exceptionDetails': {'exceptionId': 1, 'text': 'Uncaught', 'lineNumber': 0, 'columnNumber': 0, 'scriptId': '11', 'exception': {'type': 'object', 'subtype': 'error', 'className': 'ReferenceError', 'description': 'ReferenceError: process is not defined', 'objectId': '{"injectedScriptId":1,"id":2}'}}}}
```

It's clearly a flaky problem, but I don't understand how a ReferenceError can be flaky o.O

Jongy:
```
gprofiler.profilers.node.NodeDebuggerUnexpectedResponse: {'id': 1, 'result': {'result': {'type': 'object', 'subtype': 'error', 'className': 'TypeError', 'description': "TypeError: Cannot read properties of undefined (reading 'require')\n    at <anonymous>:1:20\n    at createWriteErrorHandler (node:internal/console/constructor:354:33)\n    at Object.value (node:internal/console/constructor:235:18)\n    at initializeGlobalConsole (node:internal/console/constructor:689:33)\n    at patchProcessObject (node:internal/bootstrap/pre_execution:136:3)\n    at prepareMainThreadExecution (node:internal/bootstrap/pre_execution:38:3)\n    at node:internal/main/run_main_module:9:1", 'objectId': '-608984767610883370.1.1'}, 'exceptionDetails': {'exceptionId': 1, 'text': 'Uncaught', 'lineNumber': 0, 'columnNumber': 19, 'scriptId': '101', 'exception': {'type': 'object', 'subtype': 'error', 'className': 'TypeError', 'description': "TypeError: Cannot read properties of undefined (reading 'require')\n    at <anonymous>:1:20\n    at createWriteErrorHandler (node:internal/console/constructor:354:33)\n    at Object.value (node:internal/console/constructor:235:18)\n    at initializeGlobalConsole (node:internal/console/constructor:689:33)\n    at patchProcessObject (node:internal/bootstrap/pre_execution:136:3)\n    at prepareMainThreadExecution (node:internal/bootstrap/pre_execution:38:3)\n    at node:internal/main/run_main_module:9:1", 'objectId': '-608984767610883370.1.2'}}}}
```

Again flakiness, and this time a different error.

 I'm thinking, maybe without the repl mode, we might land at unexpected JS contexts, which don't have `process` or `require` defined. So to solve the flakiness we can add some retries until we land in a "good" mode - for example, run the code under `if (typeof process !== "undefined")`. If the `if` check fails, you report back some value that lets gProfiler know it should try again in 0.5 seconds. If the check passes, we can continue to the require.

gprofiler/profilers/node.py
```diff
@@ -156,6 +155,7 @@ def _close_debugger(sock: WebSocket) -> None:
     }
     sock.send(json.dumps(cdp_request))
     sock.recv()
+    sock.close()
```
Jongy, line 158: There are 2 sites that use `create_debugger_socket` and both are not protecting from exceptions with regards to closing the socket. Let's change the sites `with create_debugger_socket(...)` and ensure that the object's context manager actually closes it after the `with`. This ensures no leaks even if exceptions occur.

Jongy (changes requested):
2 suggestions

mpozniak95:
WIP version, just wanted to see if CI/CD will pass with these changes. I will merge `_execute_js_command` and `_send_socket_request` functions, because I feel that we have a lot of duplicated code now.

gprofiler/profilers/node.py
```diff
@@ -127,39 +139,54 @@ def _execute_js_command(sock: WebSocket, command: str) -> Any:
     except json.JSONDecodeError:
         raise NodeDebuggerUnexpectedResponse(message) from None
     try:
-        return message["result"]["result"]["value"]
+        if isinstance(message["result"]["result"]["value"], str) and "process undefined" in message["result"]["result"]["value"]:
```
Jongy, line 142:
```suggestion
        if isinstance(message["result"]["result"]["value"], str) and "process undefined" == message["result"]["result"]["value"]:
```

Shouldn't it be equal?

mpozniak95:
CI/CD green :)

gprofiler/profilers/node.py
mpozniak95: I will add it in separate PR.

mpozniak95: Is it ok :D ?

Jongy: Yes

Jongy: Did you open an issue for it, or you remember to fix?

gprofiler/profilers/node.py
```diff
@@ -94,80 +105,79 @@ def _get_debugger_url() -> str:
...
-
-
-def _execute_js_command(sock: WebSocket, command: str) -> Any:
+@retry(NodeDebuggerProcessUndefined, 5, 0.5)
+def _evaluate_js_command(sock: WebSocket, command: str, expected_result: int) -> Any:
+    # Check if process or process.mainModule in command, and if it is, check if it is defined in js context
+    if "process.mainModule" in command:
+        command = f'typeof(process.mainModule) === "undefined" ? "process undefined" : {command}'
+    if "process" in command:
+        command = f'typeof(process) === "undefined" ? "process undefined" : {command}'
```
Jongy, line 114: So for `process.mainModule` we first check if `process` is defined and THEN check if `process.mainModule` is defined. Ay?

gprofiler/profilers/node.py
```diff
@@ -35,6 +35,17 @@ class NodeDebuggerUnexpectedResponse(Exception):
     pass
 
 
+class NodeDebuggerProcessUndefined(Exception):
+    pass
+
+
+class ExpectedResult:
```
Jongy, line 42: Can be an `enum`, e.g like `class JavaSafemodeOptions(str, Enum)` in `java.py`. If you pass around instances of the enum instead of `int`, you get type checking for free. After all, you don't want others to pass e.g `7777` as the `expected_resutl` value

Jongy, line 42: I think `ResultType` is a better name for it.

Code like
```
        if expected_result == ExpectedResult.NONE:
            return message
        else:
            raise NodeDebuggerUnexpectedResponse(message) from None
```

which is confusing (what was expected and what was received?) now becomes:
```
        if expected_result == ResultType.NONE:
            return message
        else:
            raise NodeDebuggerUnexpectedResponse(message) from None
```
which is far more clear :)

gprofiler/profilers/node.py
mpozniak95: https://github.com/Granulate/gprofiler/issues/576 -> just opened it

gprofiler/profilers/node.py
```diff
@@ -94,80 +105,79 @@ def _get_debugger_url() -> str:
...
+    if (
+        "result" not in message.keys()
+        or "result" not in message["result"].keys()
+        or "type" not in message["result"]["result"].keys()
+    ):
+        raise NodeDebuggerUnexpectedResponse(message) from None
+    if expected_result == ExpectedResult.BOOLEAN:
+        if message["result"]["result"]["type"] != "boolean":
+            raise NodeDebuggerUnexpectedResponse(message) from None
+        else:
```
Jongy, line 148: Why does `boolean` behave different from others? (returning the entire message and not just the value?)

gprofiler/profilers/node.py
Jongy: Excellent

gprofiler/profilers/node.py
mpozniak95: Exactly

gprofiler/profilers/node.py
mpozniak95: Previously we didn't return anything, but actually that's true, it should return value too. I will change that

gprofiler/profilers/node.py
```diff
@@ -94,80 +106,77 @@ def _get_debugger_url() -> str:
...
-            "replMode": True,
         },
     }
     sock.send(json.dumps(cdp_request))
     message = sock.recv()
     try:
         message = json.loads(message)
     except json.JSONDecodeError:
-        raise NodeDebuggerUnexpectedResponse(message) from None
+        if expected_result == ResultType.NONE:
```
Jongy, line 128: Do you expect a specific value if `ResultType.NONE`?

Such code is not very clear - IDK why `ResultType.NONE` would yield `JSONDecodeError`, I think it's nicer if you check for the value you expct for `None` and handle it. Otherwise, there could be calls with ResultType.NONE`, which actually result in unexpected, non-JSON output, and you'll ignore it here

gprofiler/profilers/node.py
```diff
@@ -94,80 +106,77 @@ def _get_debugger_url() -> str:
...
+        ):
+            raise NodeDebuggerProcessUndefined(message)
     except KeyError:
         raise NodeDebuggerUnexpectedResponse(message) from None
+    if (
+        "result" not in message.keys()
+        or "result" not in message["result"].keys()
+        or "type" not in message["result"]["result"].keys()
+    ):
+        raise NodeDebuggerUnexpectedResponse(message) from None
```
Jongy, line 145: In this one (and the next 4) you don't need `from None` because we're not in exception context (i.e not in `except` block)

gprofiler/profilers/node.py
```diff
@@ -94,80 +106,77 @@ def _get_debugger_url() -> str:
...
+    if expected_result == ResultType.BOOLEAN:
+        if message["result"]["result"]["type"] != "boolean":
+            raise NodeDebuggerUnexpectedResponse(message) from None
+    elif expected_result == ResultType.INTEGER:
+        if message["result"]["result"]["type"] != "number":
+            raise NodeDebuggerUnexpectedResponse(message) from None
+    elif expected_result == ResultType.STRING:
+        if message["result"]["result"]["type"] != "string":
+            raise NodeDebuggerUnexpectedResponse(message) from None
```
Jongy, line 154: You can use your enum values :)
```suggestion
    # check if result is not boolean/number/string as expected
    if expected_result.value != message["result"]["result"]["type"]:
        raise NodeDebuggerUnexpectedResponse(message) from None
```

Jongy (changes requested):
Done reviewing

gprofiler/profilers/node.py
mpozniak95: When evaluating _debugEnd() JS debugger returns empty message == '' thats why it yields `JSONDecodeError`. I can add check `if len(message) == 0 and expected_result == ResultType.NONE`

mpozniak95:
I addressed the comments :)

gprofiler/profilers/node.py
Jongy: Yes, that's better

gprofiler/profilers/node.py
Jongy: I suggested that the entire set of `if`s can be removed, but that can be done in the next PR

Jongy:
The 2 failures are [flakiness in installing pip](https://github.com/Granulate/gprofiler/actions/runs/3392263761/jobs/5638588307) and [get_stack_offset failure](https://github.com/Granulate/gprofiler/actions/runs/3392263749/jobs/5638503357) (https://github.com/Granulate/gprofiler/issues/523).

Merging it

PR #568, opened 2022-10-28
ci: Fix black not failing the CI & shellcheck all shell files

Jongy:
Fixes #335 which broke it :/

### DIFF
--- a/shell_lint.sh
+++ b/shell_lint.sh
@@ -8,4 +8,5 @@ set -euo pipefail
 SHELLCHECK_VERSION=v0.8.0
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
-docker run --rm -t -v "$SCRIPT_DIR":"$SCRIPT_DIR" "koalaman/shellcheck:$SHELLCHECK_VERSION" -x $SCRIPT_DIR/scripts/*.sh
+# shellcheck disable=SC2046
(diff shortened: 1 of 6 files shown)


Jongy:
I wonder why shellcheck didn't complain on it

Jongy:
Okay, we're only shellchecking the `scripts/` directory. I'll fix that as well.

tests/test.sh
```diff
@@ -16,7 +16,7 @@ fi
 # require resources.
 # TODO split them to 2 pytest files
 PERF_RESOURCE="$SCRIPT_DIR/../gprofiler/resources/perf"
-if [ ! -f "$PERF_RESOURCE" ] && [[ "$@" != *"--executable"* ]]; then
+if [ ! -f "$PERF_RESOURCE" ] && [[ "$*" != *"--executable"* ]]; then
```
Jongy, line 19: See https://www.shellcheck.net/wiki/SC2199

shell_lint.sh
```diff
@@ -8,4 +8,5 @@ set -euo pipefail
 SHELLCHECK_VERSION=v0.8.0
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
-docker run --rm -t -v "$SCRIPT_DIR":"$SCRIPT_DIR" "koalaman/shellcheck:$SHELLCHECK_VERSION" -x $SCRIPT_DIR/scripts/*.sh
+# shellcheck disable=SC2046
+docker run --rm -t -v "$SCRIPT_DIR":"$SCRIPT_DIR" "koalaman/shellcheck:$SHELLCHECK_VERSION" -x $(git ls-tree -r HEAD --name-only | grep '.sh$' | sed "s,.*,$SCRIPT_DIR/&,")
```
Jongy, line 12: I use `git ls-tree` and not just `find | grep .sh` because I have many untracked `.sh` files that I don't want to remove :)

PR #571, opened 2022-11-02
Introduce diagnostics mode

Jongy:
Introduces a `--diagnostics` flag. This causes gProfiler to log lots of local diagnostics information about the system - making it easier for us to debug the results of gProfiler, in deployments where this flag was used and the log is provided.


liorgorb (approved):
LGTM

liorgorb (dismissed):

liorgorb (approved):

PR #575, opened 2022-11-03
rbspy 0.12.1

Jongy:


Jongy:
Merged this into https://github.com/Granulate/gprofiler/pull/575, I'll do them together because upgrading either requires upgrading rust, and after rust is upgraded, the one non upgraded breaks.

Issue #576, opened 2022-11-03
nodejs: attach-maps: Use 'with' statement when creating socket connection


mpozniak95:
As mentioned in comment in: https://github.com/Granulate/gprofiler/pull/566

We should add protection from exceptions with regards to closing the socket. Let's change the sites with create_debugger_socket(...) and ensure that the object's context manager actually closes it after the with. This ensures no leaks even if exceptions occur.

PR #577, opened 2022-11-04
Use with statement when creating Node.js debugger socket

mpozniak95:
## Description
This PR adds protection from exceptions with regards to closing debugger socket.

## Related Issue
 https://github.com/Granulate/gprofiler/issues/576

### DIFF
--- a/gprofiler/profilers/node.py
+++ b/gprofiler/profilers/node.py
@@ -207,12 +207,12 @@ def _copy_module_into_process_ns(process: psutil.Process, musl: bool, version: s
 
 
 def _generate_perf_map(module_path: str, nspid: int, ns_link_name: str, pid: int) -> None:
-    with create_debugger_socket(nspid, ns_link_name) as sock:
+    with create_debugger_socket(nspid, ns_link_name, pid) as sock:
         _change_dso_state(sock, module_path, "start", pid)
 
 
 def _clean_up(module_path: str, nspid: int, ns_link_name: str, pid: int) -> None:
-    with create_debugger_socket(nspid, ns_link_name) as sock:
+    with create_debugger_socket(nspid, ns_link_name, pid) as sock:
         try:
             _change_dso_state(sock, module_path, "stop", pid)
         finally:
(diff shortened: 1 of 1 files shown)


gprofiler/profilers/node.py
```diff
@@ -180,13 +174,17 @@ def _validate_pid(expected_pid: int, sock: WebSocket) -> None:
...
-    return sock
+    try:
+        debugger_url = _get_debugger_url()
+        sock = create_connection(url=debugger_url, timeout=15.0)
+        sock.settimeout(10)
+        _validate_ns_node(sock, ns_link_name)
+        _validate_pid(nspid, sock)
+        yield sock
+    finally:
+        _close_debugger(sock)
```
Jongy, line 187: If `_get_debugger_url` raised - this code runs and access the unbound `sock`. In such cases you need to set it to `None` (preferably outside of the `try` block) and in the `finally` you check `if sock is not None`.

Jongy, line 187: I realized that we never close the `sock` object itself, i.e release the connection. Let's add it here as well, explicitly (or in `_close_debugger`). You can change `_close_debugger` to be `try ... finally: sock.close()`.

gprofiler/profilers/node.py
```diff
@@ -143,15 +144,8 @@ def _evaluate_js_command(sock: WebSocket, command: str, expected_result: ResultT
...
-    if expected_result == ResultType.BOOLEAN:
-        if expected_result.value != message["result"]["result"]["type"]:
-            raise NodeDebuggerUnexpectedResponse(message)
-    elif expected_result == ResultType.NUMBER:
-        if expected_result.value != message["result"]["result"]["type"]:
-            raise NodeDebuggerUnexpectedResponse(message)
-    elif expected_result == ResultType.STRING:
-        if expected_result.value != message["result"]["result"]["type"]:
-            raise NodeDebuggerUnexpectedResponse(message)
+    if expected_result.value != message["result"]["result"]["type"]:
```
Jongy, line 147: Good :)

Jongy (changes requested):
Done reviewing :)

Jongy (approved):

Issue #560, opened 2022-10-25
nodejs: attach-maps: Add success logs 


Jongy:
We should have some logs upon successful attaching to Node processes (in the DSO enable & DSO disable, per PID)

Jongy:
Closed by: https://github.com/Granulate/gprofiler/pull/578

Issue #562, opened 2022-10-26
nodejs: perf: test_from_container failed - fibonacci application wasn't profiled


mpozniak95:
Sporadic fail in `tests/test_sanity.py::test_from_container[True-nodejs-perf]`:
```
function_name = 'fibonacci'
collapsed = Counter({'0;;swapper;secondary_startup_64_no_verify_[k];start_secondary_[k];cpu_startup_entry_[k];do_idle_[k];default_...unknown];[/lib/x86_64-linux-gnu/ld-2.24.so];[/lib/x86_64-linux-gnu/ld-2.24.so];[/lib/x86_64-linux-gnu/ld-2.24.so]': 1})

    def assert_function_in_collapsed(function_name: str, collapsed: StackToSampleCount) -> None:
        print(f"collapsed: {collapsed}")
>       assert is_function_in_collapsed(function_name, collapsed), f"function {function_name!r} missing in collapsed data!"
E       AssertionError: function 'fibonacci' missing in collapsed data!

tests/utils.py:162: AssertionError
```
Looks like fibonacci application didn't start. Full logs here:
https://github.com/Granulate/gprofiler/actions/runs/3327993341/jobs/5504674315

PR #578, opened 2022-11-04
Add logging for changing DSO state and creating debugger socket

mpozniak95:
## Related Issue
https://github.com/Granulate/gprofiler/issues/560

### DIFF
--- a/gprofiler/profilers/node.py
+++ b/gprofiler/profilers/node.py
@@ -155,9 +155,10 @@ def _evaluate_js_command(sock: WebSocket, command: str, expected_result: ResultT
     return message["result"]["result"]["value"]
 
 
-def _change_dso_state(sock: WebSocket, module_path: str, action: str) -> None:
+def _change_dso_state(sock: WebSocket, module_path: str, action: str, pid: int) -> None:
     assert action in ("start", "stop"), "_change_dso_state supports only start and stop actions"
     command = f'process.mainModule.require("{os.path.join(module_path, "linux-perf.js")}").{action}()'
+    logger.debug("Changing DSO state", action=action, pid=pid, module_path=module_path)
     _evaluate_js_command(sock, command, ResultType.BOOLEAN)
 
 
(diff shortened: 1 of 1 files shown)


gprofiler/profilers/node.py
```diff
@@ -186,6 +186,7 @@ def create_debugger_socket(nspid: int, ns_link_name: str) -> WebSocket:
     sock.settimeout(10)
     _validate_ns_node(sock, ns_link_name)
     _validate_pid(nspid, sock)
+    logger.info(f"Created debugger socket for pid: {nspid} (nodejs)")
```
Jongy, line 189: No need to add `(nodejs)` - we log the module name which owns the logger, so this will look like `INFO: gprofiler.profilers.node: Created debugger socket for pid...` anyway.

Jongy, line 189: Prefer to pass arguments as extras instead of embedding them in the message itself. For example, as done here https://github.com/Granulate/gprofiler/blob/c95af658df7723ce0a9e74e67c94e1e6224c16ac/gprofiler/profilers/java.py#L247

gprofiler/profilers/node.py
```diff
@@ -205,13 +206,15 @@ def _copy_module_into_process_ns(process: psutil.Process, musl: bool, version: s
 def _generate_perf_map(module_path: str, nspid: int, ns_link_name: str) -> None:
     sock = create_debugger_socket(nspid, ns_link_name)
     _change_dso_state(sock, module_path, "start")
+    logger.info(f"Changed DSO state to start for pid: {nspid} (nodejs)")
```
Jongy, line 209: For consistency - you can move this log message to `_change_dso_state` and log the `action` parameter, so the message is just `Changing DSO state` and the PID & action are 2 extras

gprofiler/profilers/node.py
Jongy, line 189: Prefer to log the host PID and not NS PID. The NSPID is not unique and can (and will) repeat across containers, so by reading the log you don't truly know which process it's referring to. Please log the PID itself (if not available in the function, you can pass it from a higher-level function which does have it)

gprofiler/profilers/node.py
Jongy, line 209: All these log messages can be `logger.debug`, they're truly only for debug purposes

gprofiler/profilers/node.py
Jongy: It's helpful to log the NSPID as well since you have it here. I'm thinking of what could be useful when we debug failures here, and the NSPID could help as well, but PID itself is required

gprofiler/profilers/node.py
Jongy, line 209: I think that logging the resolved `module_path` can also help (i.e pass it as an extra to the `debug(..)` call)

Jongy (changes requested):
Done reviewing

gprofiler/profilers/node.py
mpozniak95: I added host PID to NSPID in that line, but in _changed_dso I log only host PID

Jongy:
Flaky: https://github.com/Granulate/gprofiler/issues/562 and https://github.com/Granulate/gprofiler/issues/584

Jongy (approved):

Issue #432, opened 2022-08-06
Embed DSO name in native frames


Jongy:
We'd like native frames to contain the DSO name. So when I see a function like `write`, I know that it belongs to `libc.so`. I think that it can be enabled with a flag to `perf script` (or rather - it's already printed by `perf script` but we discard the information) and then we just need to fix gProfiler to collect it as well.

Collecting the basename of the DSO is enough.

This is mainly desired for `perf`, also nice to add in PyPerf and async-profiler. It's already a feature of async-profiler - see [this commit](https://github.com/jvm-profiling-tools/async-profiler/commit/d8228a1f), we can just enable this flag when the flag is enabled in gProfiler.

Jongy:
In PyPerf, the relevant code to look at & change is [this](https://github.com/Granulate/bcc/blob/257430e792575b131441c776ea44e9e80bad6396/examples/cpp/pyperf/PyPerfNativeStackTrace.cc#L76).

Jongy:
See https://github.com/Granulate/gprofiler/pull/491#issuecomment-1258187962 for the definition we implemented

PR #491, opened 2022-09-26
Introduce and test reporting DSO names in perf and java-ap stack profiles.

marcin-ol:
We'd like native frames to contain the DSO name. 

## Description
- Async-profiler was modified to change format for embedding DSO names into parentheses (`()`) notation,
- For `perf` the merging procedure was modified to expose DSO name.

## Related Issue
#432 

## How Has This Been Tested?
Testcases have been added into:
- test_java,
- test_perf.

## Checklist:
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have updated the relevant documentation.
- [x] I have added tests for new logic.

PR #579, opened 2022-11-04
Report dso from pyperf

marcin-ol:
Report dso name for native frames generated by PyPerf.

## Description
Include PyPerf version recently updated with the implementation of DSO name reporting.  
Feature is not enabled by default - has to be enabled with run-time flag `--insert-dso-name`.

## Related Issue
#432 

## How Has This Been Tested?
Test has been defined python test suite.

## Screenshots
--

## Checklist:


gprofiler/profilers/python_ebpf.py
```diff
@@ -178,6 +179,8 @@ def start(self) -> None:
             str(self._SYMBOLS_MAP_SIZE),
             # Duration is irrelevant here, we want to run continuously.
         ] + self._offset_args()
+        if self._insert_dso_name:
+            cmd.extend(["--do-insert-dso-name"])
```
Jongy, line 183:
```suggestion
            cmd.extend(["--insert-dso-name"])
```
More consistent :)

Jongy (commented):
Test looks good.

Jongy:
BCC PR https://github.com/Granulate/bcc/pull/38

Jongy:
Let's update tag & hash per https://github.com/Granulate/bcc/pull/38#issuecomment-1329861435

Jongy:
@marcin-ol - I got around to play with it and found a bug in PyPerf now :/ but it has an easy fix.

See this stack - 
```
(more python frames here ...);yaml.loader.SafeLoader.__init__ (/home/jong/.local/lib/python3.10/site-packages/yaml/loader.py:33 [PyYAML==6.0])_[p];PyObject_SelfIter_[pn];_PyEval_EvalFrameDefault (/usr/bin/python3.10)_[pn];PyObject_SelfIter_[pn];_PyEval_EvalFrameDefault (/usr/bin/python3.10)_[pn];PyObject_SelfIter_[pn];_PyEval_EvalFrameDefault (/usr/bin/python3.10)_[pn];PyCell_New_[pn];PyFrozenSet_New_[pn];PyEval_EvalCode (/usr/bin/python3.10)_[pn];_PyArg_ParseTuple_SizeT_[pn];_PyEval_EvalFrameDefault (/usr/bin/python3.10)_[pn];_PyFunction_Vectorcall (/usr/bin/python3.10)_[pn];_PyEval_EvalFrameDefault (/usr/bin/python3.10)_[pn];_PyFunction_Vectorcall (/usr/bin/python3.10)_[pn];_PyEval_EvalFrameDefault (/usr/bin/python3.10)_[pn];_PyObject_MakeTpCall (/usr/bin/python3.10)_[pn];_PyStack_AsDict_[pn];_PyObject_FastCallDictTstate (/usr/bin/python3.10)_[pn];_PyEval_EvalFrameDefault (/usr/bin/python3.10)_[pn];_PyObject_GetMethod (/usr/bin/python3.10)_[pn];PyObject_GetAttr (/usr/bin/python3.10)_[pn];_PyObject_LookupAttr_[pn] 3
```

These are generated from this Python code:
```

In [2]: while 1: 
   ...:     import yaml 
   ...:     yaml.safe_load("") 
   ...:      
   ...:         
```

We see many `_PyEval_EvalFrameDefault` after the last Python frame. This is not expected - every Python frame has exactly one `_PyEval_EvalFrameDefault`. What PyPerf is trying to do is, to strip all native frames BELOW the last `_PyEval_EvalFrameDefault`, so that logically you see Python frames, followed by the last `_PyEval_EvalFrameDefault` which belongs to the last Python frame, and then the native functions that it calls.

The responsible code is:
```
    if (memcmp(buf, "_PyEval_EvalFrameDefault",
                sizeof("_PyEval_EvalFrameDefault")) == 0 ||
        memcmp(buf, "PyEval_EvalFrameEx", sizeof("PyEval_EvalFrameEx")) == 0)
        {
      break;
    }
```

I missed that in the review - we now need to check only the function address, not the entire frame which includes the DSO. We can also make it a startswith check.

Jongy:
Additionally - many native frames in my profile don't have the DSO. For example `_PyObject_LookupAttr_[pn]`. Can you check a local profile on your box? I don't expect this many failures.

Maybe it's a matter of symbols which are on the binary (`/usr/bin/python3.10` in this case?) vs symbols from `python-dbg` debug symbols package?

Jongy:
And a final note - not related to this PR but to https://github.com/Granulate/gprofiler/pull/491 actually - see this:
image

see the functions with `(inlined)`? (this is from the DWARF profile of PyPerf itself).

I know that perf writes `(inlined)` for inlined functions, but Is `(inlined)` ALL that perf gives us? Maybe it's just our sample regex failing here? Becuase `(inlined)` is certainly not the filename :sweat_smile:

marcin-ol:
Proposed fix for repeating `_PyEval_EvalFrameDefault`: 
https://github.com/Granulate/bcc/pull/39

marcin-ol:
> Additionally - many native frames in my profile don't have the DSO. For example `_PyObject_LookupAttr_[pn]`. Can you check a local profile on your box? I don't expect this many failures.
> 
> Maybe it's a matter of symbols which are on the binary (`/usr/bin/python3.10` in this case?) vs symbols from `python-dbg` debug symbols package?

- Do you have any specific example that seems to be lacking many of DSO names in PyPerf stacks?
- I do not see it very often with the recent commits of bcc.
- I did some research to identify what might be happening:
  - It's possible that dynamic symbols might be missing from profile,
  - Missing debug symbols in the system might be another option. But does python bundle debug symbols anyway?
- It would be great to have a specific example to verify those behaviors.

Jongy:
> * Do you have any specific example that seems to be lacking many of DSO names in PyPerf stacks?

I ran the same snippet as I wrote [here](https://github.com/Granulate/gprofiler/pull/579#issuecomment-1329923547) on my Ubuntu 22.04 Python 10.

> * I do not see it very often with the recent commits of bcc.

You tried upgrading the BCC we base off for PyPerf? I'm open to that as well! It's time we do it. We're based on BCC from `Mon Mar 8 09:40:17 2021 +0900`.

> * I did some research to identify what might be happening:
>   
>   * It's possible that dynamic symbols might be missing from profile,
>   * Missing debug symbols in the system might be another option. But does python bundle debug symbols anyway?

Python has some symbols in the binary itself, and extra symbols (those not needed for linking C extensions) in a separate file.

I ran these commands on my box:
```
$ readelf -Ws /usr/bin/python3.10| grep _PyObject_LookupAttr_
$ readelf -Ws /usr/bin/python3.10 | grep _PyEval_EvalFrameDefault
   521: 0000000000144fb0 42323 FUNC    GLOBAL DEFAULT   16 _PyEval_EvalFrameDefault
```

`_PyObject_LookupAttr_` (which didn't have a DSO) doesn't have a symbol. `_PyEval_EvalFrameDefault` which did have a DSO, DOES have a symbol. However, in both cases, the name is resolved, so BCC does know how to find the debug symbols installed on my system, but doesn't know how to link them back to the owning DSO. Maybe that was fixed in new BCC version as well.

> * It would be great to have a specific example to verify those behaviors.

Suggested above :) You can try it on an `ubuntu:22.04` container where you install the system Python 3.

marcin-ol:
Proposed fix for missing DSO names in native frames:
https://github.com/Granulate/bcc/pull/40

Jongy:
> However, in both cases, the name is resolved, so BCC does know how to find the debug symbols installed on my system, but doesn't know how to link them back to the owning DSO. Maybe that was fixed in new BCC version as well.

Combining this (false) assumption with https://github.com/Granulate/bcc/pull/40#pullrequestreview-1204392872, I think we can understand that BCC fails to symbolicate when debug symbols are required, while libunwind succeeds. Might be worth to check how to make BCC use debug symbols.

Anyway @marcin-ol , what else do we need to merge this PR besides https://github.com/Granulate/bcc/pull/40?

Jongy:
I merged https://github.com/Granulate/bcc/pull/40 and tagged, we can update the ref here and merge :)

marcin-ol:
> > However, in both cases, the name is resolved, so BCC does know how to find the debug symbols installed on my system, but doesn't know how to link them back to the owning DSO. Maybe that was fixed in new BCC version as well.
> 
> Combining this (false) assumption with [Granulate/bcc#40 (review)](https://github.com/Granulate/bcc/pull/40#pullrequestreview-1204392872), I think we can understand that BCC fails to symbolicate when debug symbols are required, while libunwind succeeds. Might be worth to check how to make BCC use debug symbols.
> 
> Anyway @marcin-ol , what else do we need to merge this PR besides [Granulate/bcc#40](https://github.com/Granulate/bcc/pull/40)?

There's nothing left besides a successful run of CI/CD with these changes.

Jongy (approved):

Jongy:
I will merge soon

PR #586, opened 2022-11-08
Remove passthrough_exception parameter in run_in_ns()

Jongy:
Post https://github.com/Granulate/granulate-utils/pull/97


Jongy:
I'll let the full tests run

Issue #504, opened 2022-09-27
appids: Implement for nodejs


Jongy:
Basic appids for NodeJS would be - find the single `*.js` argument passed to the process (where a Node process is `pgrep_exe("node")`), and the appid could be `appid: nodejs: myfile.js`.

Jongy:
When you add it - make sure to update `assert_app_id`.

Jongy:
Logically, the appid needs to be inserted here: https://github.com/Granulate/gprofiler/blob/d1c341260813cf15e6e10ef6e89cb7603ad54f64/gprofiler/profilers/perf.py#L260

Current code calls `_get_metadata` which checks for relevant Golang/NodeJS processes. At the point where it decides a process is e.g NodeJS, it needs to get the appid as well.
For Golang, we don't want an appid right now (so we can remain with `None`).

Note on NodeJS - I suggested to find the single `*.js` file, technically there might be multiple, at least with `node -r myfile.js my2ndfile.js`. We can find all `*.js` arguments, and ignore those preceded by a `-r` argument?

Jongy:
You can get inspiration from https://github.com/Granulate/gprofiler/blob/master/tests/test_appids.py (and also, make sure to add a test for it :))

PR #591, opened 2022-11-10
nodejs: Implement getting appid

mpozniak95:
## Related Issue
https://github.com/Granulate/gprofiler/issues/504

### DIFF
--- a/gprofiler/metadata/application_identifiers.py
+++ b/gprofiler/metadata/application_identifiers.py
@@ -25,6 +25,7 @@ if is_linux():
 _logger = get_logger_adapter(__name__)
 
 _PYTHON_BIN_RE = re.compile(r"^python([23](\.\d{1,2})?)?$")
+_NODE_BIN_RE = re.compile(r"^node$|^nodejs$")
 
 
 # Python does string interning so just initializing str() as a sentinel is not enough.
@@ -50,6 +51,10 @@ def _is_python_m_proc(process: Process) -> bool:
 
 def _is_python_bin(bin_name: str) -> bool:
     return _PYTHON_BIN_RE.match(os.path.basename(bin_name)) is not None
+
+
+def _is_node_bin(bin_name: str) -> bool:
+    return _NODE_BIN_RE.match(os.path.basename(bin_name)) is not None
 
 
 def _get_cli_arg_by_name(
@@ -273,6 +278,17 @@ class _PythonModuleApplicationIdentifier(_ApplicationIdentifier):
         return None
 
 
+class _NodeModuleApplicationIdentifier(_ApplicationIdentifier):
+    def get_app_id(self, process: Process) -> Optional[str]:
+        if not _is_node_bin(_get_cli_arg_by_index(process.cmdline(), 0)):
+            return None
+
+        arg_1 = _get_cli_arg_by_index(process.cmdline(), 1)
+        if arg_1.endswith(".js"):
+            return f"node: {arg_1}"
+        return None
+
+
 # Please note that the order matter, because the FIRST matching identifier will be used.
 # so when adding new identifiers pay attention to the order, unless aggregate_all is used.
 _IDENTIFIERS_MAP: Dict[str, List[_ApplicationIdentifier]] = {
@@ -287,6 +303,7 @@ _IDENTIFIERS_MAP: Dict[str, List[_ApplicationIdentifier]] = {
 }
 
 if is_linux():
+    _IDENTIFIERS_MAP["node"] = [_NodeModuleApplicationIdentifier()]
     _IDENTIFIERS_MAP["java"] = [_JavaJarApplicationIdentifier()]
     _IDENTIFIERS_MAP["java_spark"] = _IDENTIFIERS_MAP["java"] + [_JavaSparkApplicationIdentifier()]
 
@@ -334,3 +351,7 @@ def get_python_app_id(process: Process) -> Optional[str]:
 
 def get_java_app_id(process: Process, should_collect_spark_app_name: bool = False) -> Optional[str]:
     return get_app_id(process, "java_spark" if should_collect_spark_app_name else "java", aggregate_all=True)
+
+
+def get_node_app_id(process: Process) -> Optional[str]:
+    return get_app_id(process, "node")
--- a/gprofiler/metadata/application_identifiers.py
+++ b/gprofiler/metadata/application_identifiers.py
@@ -25,7 +25,6 @@ if is_linux():
 _logger = get_logger_adapter(__name__)
 
 _PYTHON_BIN_RE = re.compile(r"^python([23](\.\d{1,2})?)?$")
-_NODE_BIN_RE = re.compile(r"^node$|^nodejs$")
 
 
 # Python does string interning so just initializing str() as a sentinel is not enough.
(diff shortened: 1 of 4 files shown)


gprofiler/metadata/application_identifiers.py
```diff
@@ -273,6 +278,17 @@ def get_app_id(self, process: Process) -> Optional[str]:
         return None
 
 
+class _NodeModuleApplicationIdentifier(_ApplicationIdentifier):
+    def get_app_id(self, process: Process) -> Optional[str]:
+        if not _is_node_bin(_get_cli_arg_by_index(process.cmdline(), 0)):
+            return None
+
+        arg_1 = _get_cli_arg_by_index(process.cmdline(), 1)
```
Jongy, line 286: The command line can be:
* `node myapp.js`
* `node --some-flag myapp.js`
* `node -r myrequire.js --some-flag myapp.js`

So you need a more sophisticated way to find the arg than simply getting the first one. Please also write tests for the cases I suggested

gprofiler/metadata/application_identifiers.py
```diff
@@ -287,6 +303,7 @@ def get_app_id(self, process: Process) -> Optional[str]:
 }
 
 if is_linux():
+    _IDENTIFIERS_MAP["node"] = [_NodeModuleApplicationIdentifier()]
```
Jongy, line 306: Can be under the generic one (not only under `is_linux()`, that is, next to the python and others)

gprofiler/metadata/application_identifiers.py
```diff
@@ -273,6 +278,17 @@ def get_app_id(self, process: Process) -> Optional[str]:
         return None
 
 
+class _NodeModuleApplicationIdentifier(_ApplicationIdentifier):
+    def get_app_id(self, process: Process) -> Optional[str]:
+        if not _is_node_bin(_get_cli_arg_by_index(process.cmdline(), 0)):
```
Jongy, line 283: If you call it only if node app, you don't need to do this check here. In any case we have `is_node_process` for it so you don't need the new code here

tests/test_appids.py
```diff
@@ -186,3 +186,9 @@ def test_python() -> None:
         process_with_cmdline(["python2.7", "/path/to/mod.py"])
     )
     assert f"python: mod.py ({PROCESS_CWD}/mod.py)" == get_python_app_id(process_with_cmdline(["python2.7", "mod.py"]))
+
+
+def test_node() -> None:
+    assert "node: myapp.js" == get_node_app_id(process_with_cmdline(["node", "myapp.js"]))
```
Jongy, line 192: You also need to update `assert_app_id` to handle node as well

Jongy (changes requested):
Done reviewing :)

gprofiler/profilers/perf.py
```diff
@@ -257,15 +268,22 @@ def snapshot(self) -> ProcessToProfileData:
...
             for k, v in merge.merge_global_perfs(
                 self._perf_fp.wait_and_script() if self._perf_fp is not None else None,
                 self._perf_dwarf.wait_and_script() if self._perf_dwarf is not None else None,
                 self._insert_dso_name,
             ).items()
         }
 
+    def _generate_profile_data(self, stacks: StackToSampleCount, pid: int) -> ProfileData:
+        metadata = self._get_metadata(pid)
+        if type(metadata) is dict and "node_version" in metadata.keys():
```
Jongy, line 281:
```suggestion
        if metadata is not None and "node_version" in metadata:
```
1. The check `type(metadata) is dict` tells the reader "metadata might be a dict and might be other things" while in reality "metadata" is a dict or `None`.
2. No need for `.keys()` - `x in dict` does that.

gprofiler/profilers/perf.py
```diff
@@ -243,6 +244,16 @@ def _get_metadata(self, pid: int) -> Optional[AppMetadata]:
             pass
         return None
 
+    def _get_appid(self, pid: int) -> Optional[str]:
+        if pid in (0, -1):
```
Jongy, line 248: Don't need this check here, it's to remove "sentinel" values returned by `perf`, but `_get_metadata` already removed them, so you don't need to check again.

Jongy: (this function won't be called if `pid in (0, -1)`)

gprofiler/metadata/application_identifiers.py
```diff
@@ -273,6 +273,21 @@ def get_app_id(self, process: Process) -> Optional[str]:
         return None
 
 
+class _NodeModuleApplicationIdentifier(_ApplicationIdentifier):
+    def get_app_id(self, process: Process) -> Optional[str]:
+        skip = False
```
Jongy, line 278: `skip_next`

gprofiler/metadata/application_identifiers.py
```diff
@@ -273,6 +273,21 @@ def get_app_id(self, process: Process) -> Optional[str]:
...
+        for arg in process.cmdline()[1:]:
+            if skip:
+                skip = False
+                continue
+            if arg in ["--require", "-r"]:
```
Jongy, line 283: `node --require=b.js r.js` is also a legit commandline, let's also check `if arg.startswith("--require=")`

gprofiler/metadata/application_identifiers.py
```diff
@@ -273,6 +273,21 @@ def get_app_id(self, process: Process) -> Optional[str]:
...
+                skip = True
+                continue
+            if arg.endswith(".js"):
+                return f"nodejs: {arg}"
```
Jongy, line 287: Let's use `_append_file_to_proc_wd` here like used in others, e.g `_PythonModuleApplicationIdentifier`. The idea is that the final appid will be `nodejs: {file_arg} ({abs path of file arg})`.

tests/test_appids.py
```diff
@@ -186,3 +186,13 @@ def test_python() -> None:
         process_with_cmdline(["python2.7", "/path/to/mod.py"])
     )
     assert f"python: mod.py ({PROCESS_CWD}/mod.py)" == get_python_app_id(process_with_cmdline(["python2.7", "mod.py"]))
+
+
+def test_node_appid() -> None:
```
Jongy, line 191: Let's add one test for the `node --require=b.js r.js` case

gprofiler/metadata/application_identifiers.py
Jongy: (See tests use of `process_with_cmdline` and `PROCESS_CWD`)

Jongy (changes requested):
Done reviewing :)

tests/conftest.py
```diff
@@ -482,6 +487,7 @@ def assert_app_id(application_pid: int, runtime: str, in_container: bool) -> Gen
     desired_name_and_getter = {
         "java": (get_java_app_id, "java: Fibonacci.jar"),
         "python": (get_python_app_id, "python: lister.py (/app/lister.py)"),
+        "node": (get_node_app_id, "nodejs: fibonacci.js"),
```
Jongy, line 490: This needs to be fixed post the `_append_file_to_proc_wd` change

gprofiler/metadata/application_identifiers.py
Jongy: I don't see this change committed

Jongy (changes requested):
Done reviewing

Jongy:
Lint fails

Jongy:
Looks good, I'll wait for tests to pass

PR #594, opened 2022-11-13
Bugfix/proc basename argv0

guybortnikov:
## Description
Use newer version of `granulate-utils`, that fixes a bug on process matching, also verify the bug was fixed with one more test

## How Has This Been Tested?
Added automatic tests


Jongy (approved):
Will merge after tests pass

PR #599, opened 2022-11-14
Bugfix/multiple-containers-being-run-in-tests

guybortnikov:
## Description
Small bugfix to fix a race-condition where 2 containers are being run from the same test (see `test_preconditions:test_mutex_taken_twice`)

### DIFF
--- a/tests/test_preconditions.py
+++ b/tests/test_preconditions.py
@@ -3,7 +3,7 @@
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
 from pathlib import Path
-from typing import Optional
+from typing import Optional, List
 
 import pytest
 from docker import DockerClient
@@ -11,7 +11,7 @@ from docker.errors import ContainerError
 from docker.models.containers import Container
 from docker.models.images import Image
 
-from tests.utils import start_gprofiler_in_container_for_one_session, wait_for_container
+from tests.utils import start_gprofiler_in_container_for_one_session, wait_for_container, wait_for_log
 
 
 def start_gprofiler(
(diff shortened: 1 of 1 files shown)


tests/test_preconditions.py
```diff
@@ -20,35 +21,58 @@ def start_gprofiler(
...
-        ["-d", "1"],
+        ["-d", "1"] + extra_profiler_args,
         privileged=privileged,
         user=user,
         pid_mode=pid_mode,
     )
 
 
+@contextmanager
+def run_gprofiler_container(
```
Jongy, line 41: This feels like `run_gprofiler_in_container_for_one_session` without the waiting part, no? Do you want to try merging them?

tests/test_preconditions.py
```diff
@@ -20,35 +21,58 @@ def start_gprofiler(
...
     """
-    gprofiler1 = start_gprofiler(docker_client, gprofiler_docker_image)
-    gprofiler2 = start_gprofiler(docker_client, gprofiler_docker_image)
-
-    # exits without an error
-    assert wait_for_container(gprofiler2) == (
-        "Could not acquire gProfiler's lock. Is it already running?"
-        " Try 'sudo netstat -xp | grep gprofiler' to see which process holds the lock.\n"
-    )
+    # Run the first one continuously
```
Jongy, line 67: This is the bug then, right? The race when the first one already exits

tests/test_preconditions.py
```diff
@@ -20,35 +21,58 @@ def start_gprofiler(
...
+) -> Container:
+    container: Container = None
+    try:
+        container = start_gprofiler(
+            docker_client, gprofiler_docker_image, privileged, user, pid_mode, extra_profiler_args
+        )
+        yield container
+    finally:
+        if container is not None:
+            container.stop()  # Fine to be run also if already stopped
```
Jongy, line 57: Leaving the "remove" for a future PR?

tests/test_preconditions.py
guybortnikov: 1. the first one might have exited before 2 have run
2. the first one didn't acquire the mutex and the second one will acquire it before (with the current solution we're waiting for log that is being logged only after the mutex successfully acquired)

tests/test_preconditions.py
guybortnikov: yes, not related, part of https://github.com/Granulate/gprofiler/issues/600

tests/test_preconditions.py
guybortnikov: No reason, only needed here, all the other tests are waiting for the container to exit, this is only needed when running in continuously

Jongy: Ok

Jongy (approved):

Issue #505, opened 2022-09-27
appids: Implement for Ruby


Jongy:
Basic appids for Ruby would be - find the single `*.rb` argument passed to the process (where a Rode process is as `RbSpyProfiler._select_processes_to_profile()`), and the appid could be `appid: ruby: myfile.rb`.

Jongy:
When you add it - make sure to update `assert_app_id`.

Jongy:
Needs to be added here: https://github.com/Granulate/gprofiler/blob/d1c341260813cf15e6e10ef6e89cb7603ad54f64/gprofiler/profilers/ruby.py#L110

PR #601, opened 2022-11-15
ruby: Implement getting appid

mpozniak95:
## Related Issue
https://github.com/Granulate/gprofiler/issues/505

### DIFF
--- a/gprofiler/metadata/application_identifiers.py
+++ b/gprofiler/metadata/application_identifiers.py
@@ -290,6 +290,21 @@ class _NodeModuleApplicationIdentifier(_ApplicationIdentifier):
         return None
 
 
+class _RubyModuleApplicationIdentifier(_ApplicationIdentifier):
+    def get_app_id(self, process: Process) -> Optional[str]:
+        skip_next = False
+        for arg in process.cmdline():
+            if skip_next:
+                skip_next = False
+                continue
+            if arg == "-r":
+                skip_next = True
+                continue
+            if arg.endswith(".rb"):
+                return f"ruby: {arg} ({_append_file_to_proc_wd(process, arg)})"
+        return None
+
+
 # Please note that the order matter, because the FIRST matching identifier will be used.
 # so when adding new identifiers pay attention to the order, unless aggregate_all is used.
 _IDENTIFIERS_MAP: Dict[str, List[_ApplicationIdentifier]] = {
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -487,7 +488,8 @@ def assert_app_id(application_pid: int, runtime: str, in_container: bool) -> Gen
     desired_name_and_getter = {
         "java": (get_java_app_id, "java: Fibonacci.jar"),
         "python": (get_python_app_id, "python: lister.py (/app/lister.py)"),
-        "node": (get_node_app_id, "nodejs: fibonacci.js (/app/fibonacci.js"),
```
Jongy, line 490: How did we merge it this way o.O I mean how did the tests pass

gprofiler/metadata/application_identifiers.py
```diff
@@ -290,6 +290,21 @@ def get_app_id(self, process: Process) -> Optional[str]:
...
 
 
+class _RubyModuleApplicationIdentifier(_ApplicationIdentifier):
+    def get_app_id(self, process: Process) -> Optional[str]:
+        skip_next = False
+        for arg in process.cmdline():
+            if skip_next:
+                skip_next = False
+                continue
+            if arg == "-r":
```
Jongy, line 300: I read that these 2 are equal:
```
ruby -r require.rb app.rb
ruby -rrequire.rb app.rb
```

so you need to check if anything follows the `-r` and if so, don't skip next and only skip the current argument.

Jongy (changes requested):
One small comment :)

tests/conftest.py
mpozniak95: So this was never tested :(

Jongy: Ohh.

gprofiler/metadata/application_identifiers.py
```diff
@@ -290,6 +290,23 @@ def get_app_id(self, process: Process) -> Optional[str]:
...
 
 
+class _RubyModuleApplicationIdentifier(_ApplicationIdentifier):
+    def get_app_id(self, process: Process) -> Optional[str]:
+        skip_next = False
+        for arg in process.cmdline():
+            if skip_next:
+                skip_next = False
+                continue
+            if re.compile(r"-r.+").match(arg):
```
Jongy, line 300: Simpler - if arg startswith "-r". Covers both cases so we won't need the ==.

gprofiler/metadata/application_identifiers.py
```diff
@@ -290,6 +290,22 @@ def get_app_id(self, process: Process) -> Optional[str]:
...
 
+class _RubyModuleApplicationIdentifier(_ApplicationIdentifier):
+    def get_app_id(self, process: Process) -> Optional[str]:
+        skip_next = False
+        for arg in process.cmdline():
+            if skip_next:
+                skip_next = False
+                continue
+            if arg.startswith("-r"):
+                if len(arg) <= 2:
```
Jongy, line 301: Could be `== 2` because it's always at least 2 due to the startswith, but meh

Jongy (approved):

Issue #604, opened 2022-11-17
nodejs: attach-maps: Build on older Ubuntu version


Jongy:
We use `ubuntu:20.04`, see
https://github.com/Granulate/gprofiler/blob/914e05081bf90f96059f1a3b5995dec71b334b7c/Dockerfile#L26

This means the lib is compiled against new libstdc++ and glibc and will not necessary load on older systems (e.g on a system/container based on Ubuntu 16.04).

Logically, this is the same problem as we have with async-profiler - we build it on CentOS 7 and do many "shenanigans" to have it load on systems as old as CentOS 6. NodeJS is more modern so I don't think we need to support *that* old system. I suppose getting the build to work on 14.04 is good enough.

We can also add a test that verifies it, or at the very least, test manually on an Ubuntu 14.04 container where NodeJS was installed. If we try to `attach-maps` such container now, it will fail with an error like `Error: /lib64/libstdc++.so.6: version CXXABI_1.3.9 not found (required by /tmp/gprofiler_tmp/node_module/20eb88a/glibc/10/build/Release/linux-perf.node)`.

Jongy:
Let's also make sure it supports CentOS 7 (glibc 2.17 libstdc++ 3.4.19)

PR #597, opened 2022-11-14
Use modern Python for gProfiler exe

pfilipko1:
Previously, gprofiler binary is being built using python 3.6, it is now updated to 3.10

## Description
Addition of openssl and python building in the centos 7 container

## Related Issue
https://github.com/Granulate/gprofiler/issues/583

## Motivation and Context
It will allow to use modern python functionalities

## How Has This Been Tested?
Manually

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] I have updated the relevant documentation.
- [ ] I have added tests for new logic.

PR #610, opened 2022-11-21
nodejs: attach-maps: Build nodejs resources on centos7 image

mpozniak95:
## Description
This PR changes base image for building nodejs resources from ubuntu20 to centos7. This change is needed because profiling nodejs processes on centos7 resulted in: `Error: /lib64/libstdc++.so.6: version CXXABI_1.3.9 not found (required by /tmp/gprofiler_tmp/node_module/20eb88a/glibc/10/build/Release/linux-perf.node)`. 
This PR also changes image of 12-glibc nodejs_matrix_test to centos7.

## Related Issue
https://github.com/Granulate/gprofiler/issues/604

## How Has This Been Tested?
pytest -v -k "node"


Dockerfile
```diff
@@ -163,12 +163,14 @@ COPY scripts/build_node_package.sh .
...
 # node-package-builder-glibc
-FROM ubuntu${NODE_PACKAGE_BUILDER_GLIBC} AS node-package-builder-glibc
+FROM centos/devtoolset-7-toolchain-centos7${NODE_PACKAGE_BUILDER_GLIBC} AS node-package-builder-glibc
+USER 0
 WORKDIR /tmp
 COPY scripts/node_builder_glibc_env.sh .
 RUN ./node_builder_glibc_env.sh
 COPY scripts/build_node_package.sh .
 RUN ./build_node_package.sh
+USER 1001
```
Jongy, line 173: Not required because the next command is `FROM` so we switch image.

pyi.Dockerfile
```diff
@@ -111,12 +111,14 @@ COPY scripts/build_node_package.sh .
...
 # node-package-builder-glibc
-FROM ubuntu${NODE_PACKAGE_BUILDER_GLIBC} AS node-package-builder-glibc
+FROM centos/devtoolset-7-toolchain-centos7${NODE_PACKAGE_BUILDER_GLIBC} AS node-package-builder-glibc
+USER 0
 WORKDIR /tmp
 COPY scripts/node_builder_glibc_env.sh .
 RUN ./node_builder_glibc_env.sh
 COPY scripts/build_node_package.sh .
 RUN ./build_node_package.sh
+USER 1001
```
Jongy, line 121: Not needed, as said earlier

tests/containers/nodejs/Dockerfile
```diff
@@ -9,5 +11,6 @@ RUN mkdir /app
 ADD fibonacci.js /app
 
 ENV NODE_RUNTIME_FLAGS ${NODE_RUNTIME_FLAGS}
+USER 1001
```
Jongy, line 14: Why this change?

tests/conftest.py
```diff
@@ -248,29 +248,31 @@ def application_docker_image_configs() -> Mapping[str, Dict[str, Any]]:
             "": dict(
                 buildargs={
                     "NODE_RUNTIME_FLAGS": "--perf-prof --interpreted-frames-native-stack",
+                    "NODE_IMAGE": "node",
                     "NODE_IMAGE_TAG": "node@sha256:59531d2835edd5161c8f9512f9e095b1836f7a1fcb0ab73e005ec46047384911",
```
Jongy, line 252: This means we have `node:node@sha256:59531d2835edd5161c8f9512f9e095b1836f7a1fcb0ab73e005ec46047384911` as the image, is that the way it works? Weird o.O

tests/containers/nodejs/Dockerfile
```diff
@@ -1,5 +1,7 @@
 ARG NODE_IMAGE_TAG
-FROM node:${NODE_IMAGE_TAG}
+ARG NODE_IMAGE
```
Jongy, line 2:
```suggestion
ARG NODE_IMAGE=node
```
then we don't need to add `NODE_IMAGE` everywhere :)

Dockerfile
mpozniak95: That's what I thought too but linter failed: https://github.com/Granulate/gprofiler/actions/runs/3514434057/jobs/5888441878

tests/conftest.py
mpozniak95: Oh, didn't notice that. I will change it

tests/containers/nodejs/Dockerfile
mpozniak95: I wasn't able to create /app directory on centos image. Can we f.e. switch dir to /tmp/app instead of /app?

scripts/node_builder_glibc_env.sh
```diff
@@ -5,8 +5,15 @@
...
-export DEBIAN_FRONTEND=noninteractive
-apt update -y && apt install -y --no-install-recommends curl g++ python3 make gcc git ca-certificates npm
-# apt has node v10 by default, so we need to add newer version to run node-gyp
-curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
-apt install -y --no-install-recommends nodejs
+yum update -y && yum install -y curl python3 make git ca-certificates npm
+# yum has node v10 by default, so we need to add newer version to run node-gyp
+curl -fsSL https://rpm.nodesource.com/setup_16.x | bash -
+yum install -y nodejs
+# node-gyp needs python 3.7+, so we need to install it
```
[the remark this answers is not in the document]
mpozniak95: On my env python installation takes around 3 minutes. Entire node_builder_glibc_env.sh script takes around 4.5 minutes

tests/conftest.py
```diff
@@ -248,29 +248,31 @@ def application_docker_image_configs() -> Mapping[str, Dict[str, Any]]:
...
-            "14-musl": dict(buildargs={"NODE_IMAGE_TAG": "14-alpine"}),
-            "15-glibc": dict(buildargs={"NODE_IMAGE_TAG": "15-slim"}),
-            "15-musl": dict(buildargs={"NODE_IMAGE_TAG": "15-alpine"}),
-            "16-glibc": dict(buildargs={"NODE_IMAGE_TAG": "16-slim"}),
-            "16-musl": dict(buildargs={"NODE_IMAGE_TAG": "16-alpine"}),
+            "10-glibc": dict(buildargs={"NODE_IMAGE": "node", "NODE_IMAGE_TAG": "10-slim"}),
+            "10-musl": dict(buildargs={"NODE_IMAGE": "node", "NODE_IMAGE_TAG": "10.24.1-alpine"}),
+            "11-glibc": dict(buildargs={"NODE_IMAGE": "node", "NODE_IMAGE_TAG": "11-slim"}),
+            "11-musl": dict(buildargs={"NODE_IMAGE": "node", "NODE_IMAGE_TAG": "11-alpine"}),
+            "12-glibc": dict(buildargs={"NODE_IMAGE": "centos/nodejs-12-centos7", "NODE_IMAGE_TAG": "12"}),
```
Jongy, line 264: The reason for this test is - to ensure that the nodejs module is successfully loaded on glibc 2.17. I suggest we also add something in the test that "proves" this assumption - e.g in case the base image is changed in the future, and the new image doesn't have this glibc version. For example, a similar check is done here - https://github.com/Granulate/gprofiler/blob/5b82e9bd58ff412f9ed834cd1a423bb9888b2117/tests/test_java.py#L162 the test wants to run on a musl-based image, and the test actually verifies it in runtime. You can check in the test in the 12-glibc case, that `ldd --version` in the container gives 2.17

scripts/node_builder_glibc_env.sh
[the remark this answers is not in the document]
Jongy: Nice. Not too much. Even with that - we can eliminate this as well with https://github.com/Granulate/gprofiler/pull/597 because that PR sets up python 3.10 on CentOS 7, we'll be able to base off that image them. We can do that later

Dockerfile
```diff
@@ -23,8 +23,8 @@ ARG BURN_BUILDER_GOLANG=@sha256:f7d3519759ba6988a2b73b5874b17c5958ac7d0aa48a8b1d
 ARG GPROFILER_BUILDER_UBUNTU=@sha256:cf31af331f38d1d7158470e095b132acd126a7180a54f263d386da88eb681d93
 # node-package-builder-musl alpine
 ARG NODE_PACKAGE_BUILDER_MUSL=@sha256:69704ef328d05a9f806b6b8502915e6a0a4faa4d72018dc42343f511490daf8a
-# node-package-builder-glibc - ubuntu:20.04
-ARG NODE_PACKAGE_BUILDER_GLIBC=@sha256:cf31af331f38d1d7158470e095b132acd126a7180a54f263d386da88eb681d93
+# node-package-builder-glibc - centos7 with gcc 7
```
Jongy, line 26: "centos7 with gcc 7" - let's replace it with the image:tag you got the hash from?

Jongy: (I suppose `centos/devtoolset-7-toolchain-centos7:latest`?)

pyi.Dockerfile
Jongy: Let's comment here as well

tests/containers/nodejs/Dockerfile
Jongy: Yes we can, no problem

Jongy: We can leave this here as well

Jongy: Or was it needed for hadolint and that's it?

tests/utils.py
```diff
@@ -168,6 +168,18 @@ def assert_function_in_collapsed(function_name: str, collapsed: StackToSampleCou
...
 
+def assert_ldd_version_container(container: Container, version: str) -> None:
+    pattern = r"^ldd \(.*\) (\d*.\d*)"
+    command = "ldd --version"
+    exec_output = container.exec_run(command).output.decode("utf-8")
+    search_result = re.search(pattern, exec_output)
+    if search_result:
+        version_in_container = search_result.group(1)
+    else:
+        version_in_container = False
```
Jongy, line 179: Assigning a string or a boolean is not okay (I think mypy should complain on it o.O). It does work because, well, `False` will always be different from the version.

It's better to assign `None` if you couldn't find what you need, the comparison will fail as well

tests/utils.py
```diff
@@ -168,6 +168,18 @@ def assert_function_in_collapsed(function_name: str, collapsed: StackToSampleCou
     assert is_function_in_collapsed(function_name, collapsed), f"function {function_name!r} missing in collapsed data!"
 
 
+def assert_ldd_version_container(container: Container, version: str) -> None:
+    pattern = r"^ldd \(.*\) (\d*.\d*)"
+    command = "ldd --version"
+    exec_output = container.exec_run(command).output.decode("utf-8")
+    search_result = re.search(pattern, exec_output)
```
Jongy, line 175: Just embed the pattern here?
```suggestion
    search_result = re.search(r"^ldd \(.*\) (\d*.\d*)", exec_output)
```

tests/utils.py
```diff
@@ -168,6 +168,18 @@ def assert_function_in_collapsed(function_name: str, collapsed: StackToSampleCou
     assert is_function_in_collapsed(function_name, collapsed), f"function {function_name!r} missing in collapsed data!"
 
 
+def assert_ldd_version_container(container: Container, version: str) -> None:
+    pattern = r"^ldd \(.*\) (\d*.\d*)"
+    command = "ldd --version"
+    exec_output = container.exec_run(command).output.decode("utf-8")
```
Jongy, line 174:
```suggestion
    exec_output = container.exec_run("ldd --version").output.decode("utf-8")
```

No need for extra variables in this case IMO, it makes the code unnecessarily longer

tests/containers/nodejs/Dockerfile
mpozniak95: Actually I think that linter doesn't scan this file, I removed this line

mpozniak95:
https://github.com/Granulate/gprofiler/issues/562 in CICD -> https://github.com/Granulate/gprofiler/actions/runs/3523260293/jobs/5907558800

tests/utils.py
```diff
@@ -168,6 +168,16 @@ def assert_function_in_collapsed(function_name: str, collapsed: StackToSampleCou
...
 
 
+def assert_ldd_version_container(container: Container, version: str) -> None:
+    exec_output = container.exec_run("ldd --version").output.decode("utf-8")
+    search_result = re.search(r"^ldd \(.*\) (\d*.\d*)", exec_output)
+    if search_result:
+        version_in_container = search_result.group(1)
+    else:
+        version_in_container = None
+    assert version_in_container == version, f"ldd version in container: {version_in_container}, expected {version}"
```
Jongy, line 178: Tip - in pytest, you don't need to add the `, f"ldd version ..."`, pytest does it for you, it will print the rhs & lhs values nicely.

Jongy (approved):

Jongy:
@mpozniak95 - in a separate PR, can you please update `build_aarch64_...` scripts with the new base image hash for the node glibc builder?

PR #611, opened 2022-11-25
Block building glibc on aarch64, update nodejs glibc builder image in aarch64 scripts

mpozniak95:
Update base images in aarch64 scripts to be in line with https://github.com/Granulate/gprofiler/pull/610

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -167,13 +167,9 @@ FROM centos/devtoolset-7-toolchain-centos7${NODE_PACKAGE_BUILDER_GLIBC} AS node-
 USER 0
 WORKDIR /tmp
 COPY scripts/node_builder_glibc_env.sh .
-RUN if [ "$(uname -m)" != "aarch64" ]; then \
-      ./node_builder_glibc_env.sh; \
-    fi
+RUN ./node_builder_glibc_env.sh
 COPY scripts/build_node_package.sh .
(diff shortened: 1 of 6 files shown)


Jongy:
Oof, it seems that this image doesn't support Aarch64 - when running it on a Graviton server:
```
$ docker run -it centos/devtoolset-7-toolchain-centos7 bash
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
exec /usr/bin/container-entrypoint: exec format error
```

What we can do -
* Check if there's another base image of CentOS 7 that comes with devtoolset and has Aarch64
* For Aarch64, use `centos:7` and install devtoolset ourselves (only on CentOS 7, on Aarch64 we can remain with what we got)

In any case, after #597 , we will be able to base off the image which produces the gProfiler exe, as it will have a functioning Python 3.10. If this fix grows complex, we can disable the support for Aarch64 now (after all it's still experimental, we'll first use the x86_64 one) and bring it back after #597.

Jongy:
Build passes for both exe & container of Aarch64 :) I will merge after tests pass.

mpozniak95:
Great, thanks for checking it on aarch64 ! :)

Jongy (approved):

Issue #528, opened 2022-10-12
perf: memory leak


Jongy:
Occasionally I'm seeing perf processes whose memory continues growing forever. Here's one example for `/proc/pid/status` of such a perf - it shows that anonymous memory takes up most of the space (700mb). This perf process is at least 2 weeks old.
```
ame:	perf
Umask:	0022
State:	S (sleeping)
Tgid:	3966
Ngid:	0
Pid:	3966
PPid:	3825
TracerPid:	0
Uid:	0	0	0	0
Gid:	0	0	0	0
FDSize:	64
Groups:	0 1001 
NStgid:	3966
NSpid:	3966
NSpgid:	3966
NSsid:	3665
VmPeak:	  732316 kB
VmSize:	  732208 kB
VmLck:	       0 kB
VmPin:	    5136 kB
VmHWM:	  707368 kB
VmRSS:	  707368 kB
RssAnon:	  702548 kB
RssFile:	    4820 kB
RssShmem:	       0 kB
VmData:	  712220 kB
VmStk:	     132 kB
VmExe:	   11512 kB
VmLib:	       4 kB
VmPTE:	    1452 kB
VmSwap:	       0 kB
HugetlbPages:	       0 kB
CoreDumping:	0
THP_enabled:	1
Threads:	2
SigQ:	0/15540
SigPnd:	0000000000000000
ShdPnd:	0000000000000000
SigBlk:	0000000008000000
SigIgn:	0000000000000004
SigCgt:	0000000180014c02
CapInh:	0000000000000000
CapPrm:	0000003fffffffff
CapEff:	0000003fffffffff
CapBnd:	0000003fffffffff
CapAmb:	0000000000000000
NoNewPrivs:	0
Seccomp:	0
Speculation_Store_Bypass:	thread vulnerable
Cpus_allowed:	3
Cpus_allowed_list:	0-1
Mems_allowed:	00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000001
Mems_allowed_list:	0
voluntary_ctxt_switches:	130270
nonvoluntary_ctxt_switches:	1686839
```

[...]

PR #613, opened 2022-11-28
perf: Restart perf every 60 minutes if perf used memory exceeds 512 MB

mpozniak95:
## Description
Restart perf once per 60 minutes to avoid memory leak from issue https://github.com/Granulate/gprofiler/issues/528.

## Related Issue
https://github.com/Granulate/gprofiler/issues/528

## How Has This Been Tested?
Run gprofiler in continuous mode manually.

### DIFF
--- a/gprofiler/profilers/perf.py
+++ b/gprofiler/profilers/perf.py
@@ -5,6 +5,7 @@
 
 import os
 import signal
+import time
 from pathlib import Path
 from subprocess import Popen
 from threading import Event
@@ -37,6 +38,7 @@ DEFAULT_PERF_DWARF_STACK_SIZE = 8192
 class PerfProcess:
     _dump_timeout_s = 5
     _poll_timeout_s = 5
+    _restart_after_s = 3600
     # default number of pages used by "perf record" when perf_event_mlock_kb=516
     # we use double for dwarf.
     _mmap_sizes = {"fp": 129, "dwarf": 257}
@@ -50,6 +52,7 @@ class PerfProcess:
         inject_jit: bool,
         extra_args: List[str],
     ):
+        self.start_time = None
         self._frequency = frequency
         self._stop_event = stop_event
         self._output_path = output_path
@@ -77,11 +80,17 @@ class PerfProcess:
             str(self._mmap_sizes[self._type]),
         ] + self._extra_args
 
+    def check_if_restart(self) -> None:
+        if time.time() - self.start_time >= self._restart_after_s:
+            self.stop()
+            self.start()
+
     def start(self) -> None:
         logger.info(f"Starting perf ({self._type} mode)")
         process = start_process(self._get_perf_cmd(), via_staticx=False)
         try:
             wait_event(self._poll_timeout_s, self._stop_event, lambda: os.path.exists(self._output_path))
+            self.start_time = time.time()
         except TimeoutError:
             process.kill()
             assert process.stdout is not None and process.stderr is not None
@@ -267,7 +276,7 @@ class SystemProfiler(ProfilerBase):
         for perf in self._perfs:
             perf.switch_output()
 
-        return {
+        data = {
             k: self._generate_profile_data(v, k)
             for k, v in merge.merge_global_perfs(
                 self._perf_fp.wait_and_script() if self._perf_fp is not None else None,
@@ -275,6 +284,11 @@ class SystemProfiler(ProfilerBase):
                 self._insert_dso_name,
             ).items()
         }
+
+        for perf in self._perfs:
+            perf.check_if_restart()
+
+        return data
 
     def _generate_profile_data(self, stacks: StackToSampleCount, pid: int) -> ProfileData:
         metadata = self._get_metadata(pid)
--- a/gprofiler/profilers/perf.py
+++ b/gprofiler/profilers/perf.py
@@ -84,7 +84,7 @@ class PerfProcess:
     def check_if_restart(self) -> None:
         if (
             time.monotonic() - self.start_time >= self._restart_after_s
-            and Process(self._process.pid).memory_info().rss >= self._perf_memory_usage_treshhold # type: ignore
+            and Process(self._process.pid).memory_info().rss >= self._perf_memory_usage_treshhold  # type: ignore
         ):
             self.stop()
             self.start()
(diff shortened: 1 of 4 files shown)


Jongy:
It's a fair approach I think.

However instead of "just restarting" every 60 minutes, I suggest we restart only if memory exceedds some threshold (defined at 512mb to begin with). So every 60 minutes, if perf memory usage is > 512mb, restart it. You can retrieve process memory utilization with psutil.

gprofiler/profilers/perf.py
```diff
@@ -77,11 +80,17 @@ def _get_perf_cmd(self) -> List[str]:
             str(self._mmap_sizes[self._type]),
         ] + self._extra_args
 
+    def check_if_restart(self) -> None:
+        if time.time() - self.start_time >= self._restart_after_s:
```
Jongy, line 84: Use `time.monotinic` - `time.time` can be affected by clock changes

gprofiler/profilers/perf.py
```diff
@@ -37,6 +38,8 @@
 class PerfProcess:
     _dump_timeout_s = 5
     _poll_timeout_s = 5
+    _restart_after_s = 3600
+    _perf_memory_usage_treshhold = 536870912  # 512Mb
```
Jongy, line 42:
```suggestion
    _perf_memory_usage_treshhold = 512 * 1024 * 1024
```

then, we don't need the comment :)

gprofiler/profilers/perf.py
```diff
@@ -50,6 +53,7 @@ def __init__(
         inject_jit: bool,
         extra_args: List[str],
     ):
+        self.start_time = 0.0
```
Jongy, line 56:
```suggestion
        self._start_time = 0.0
```
it's private

gprofiler/profilers/perf.py
```diff
@@ -77,11 +81,20 @@ def _get_perf_cmd(self) -> List[str]:
             str(self._mmap_sizes[self._type]),
         ] + self._extra_args
 
+    def check_if_restart(self) -> None:
```
Jongy, line 84:
```suggestion
    def check_if_needs_restart(self) -> None:
```
Let's add a docstring explaining the reasoning (what is "needs restart")

gprofiler/profilers/perf.py
```diff
@@ -77,11 +81,20 @@ def _get_perf_cmd(self) -> List[str]:
...
+        if (
+            time.monotonic() - self.start_time >= self._restart_after_s
+            and Process(self._process.pid).memory_info().rss >= self._perf_memory_usage_treshhold  # type: ignore
```
Jongy, line 87: Is `rss` enough? As far as I understand, some of the data that perf uses is actually "shared". I'm not sure if it will account in `rss` of psutil.

You can test that with a simple python program that reads many files from e.g `/usr` (after all - this is what perf does, it reads tons of files to extract symbols) and see if it causes its `rss` to increase.

gprofiler/profilers/perf.py
```diff
@@ -276,6 +289,11 @@ def snapshot(self) -> ProcessToProfileData:
             ).items()
         }
 
+        for perf in self._perfs:
```
Jongy, line 292: It's always good to guard over features (especially very "active" features like this one) with a feature flag. Let's add one here? It can be enabled by default, and we add a flag `--perf-no-restart` that disables it.

Jongy (changes requested):
Done reviewing

gprofiler/profilers/perf.py
mpozniak95: Ok, I will check that

mpozniak95: Added debug prints to check rss and shared values:
```
Type: fp
Rss memory: 4829184
Shared memory: 4186112
Type: dwarf
Rss memory: 8355840
Shared memory: 7704576
Type: fp
Rss memory: 5242880
Shared memory: 4546560
Type: dwarf
Rss memory: 11493376
Shared memory: 10788864
```

> RSS does not include memory that is swapped out. It does include memory from shared libraries as long as the pages from those libraries are actually in memory. It does include all stack and heap memory.

Checking if RSS exceeds threshold should be enough

gprofiler/profilers/perf.py
```diff
@@ -77,11 +81,21 @@ def _get_perf_cmd(self) -> List[str]:
             str(self._mmap_sizes[self._type]),
         ] + self._extra_args
 
+    def check_if_needs_restart(self) -> None:
+        """Checks if perf used memory exceeds threshold, and if it does, restarts perf"""
+        if (
+            time.monotonic() - self._start_time >= self._restart_after_s
+            and Process(self._process.pid).memory_info().rss >= self._perf_memory_usage_threshold  # type: ignore
```
Jongy, line 88: What's the `type: ignore` for? psutil?

tests/test_perf.py
```diff
@@ -43,6 +43,7 @@ def make_system_profiler(tmp_path: Path, perf_mode: str, insert_dso_name: bool)
         perf_inject=False,
         perf_dwarf_stack_size=DEFAULT_PERF_DWARF_STACK_SIZE,
         perf_node_attach=False,
+        perf_restart=False,
```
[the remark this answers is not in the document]
Jongy: Here is a good site (and keep the others `False`)

Jongy (changes requested):
Done reviewing

Jongy (changes requested):
Done reviewing

gprofiler/profilers/perf.py
mpozniak95: `gprofiler/profilers/perf.py:88: error: Item "None" of "Optional[Popen[Any]]" has no attribute "pid"`
same issue as in line 134. self._process is initialized with None value

Jongy: Okay. So in such cases - this function expects `self._process` to be non-None, and using `type: ignore` weakens the static linting.

Instead, what is preferable, is to provide additional help to the static linter so that it's able to do its job. In this case, you can do that by adding `assert self._process is not None` at the beginning of the function - it's indeed a bug if the function is called while this field is not initialized.

The only times it's far to use `type: ignore` is when a package doesn't have type hints, or particular methods/classes etc don't have type hints. You DON'T use `type: ignore` to solve implicit cases like here (something is `None` but we don't expect it to be `None`, ...)

Jongy:
One more comment :)

Jongy:
Excellent :)

Jongy (approved):

PR #614, opened 2022-11-29
Fix aarch64 centos image issue

mpozniak95:
This PR fixes problem with building gProfiler on aarch64 architecture.


Jongy:
Apparently - https://github.com/Granulate/gprofiler/pull/611 worked for me on Aarch64 because I pulled the image manually, and Docker was able to find the image with the hash spec despite the incorrect image name for the hash.

Tested master on a clean Docker env, and it fails. Tested this PR and it passes.

PR #618, opened 2022-11-30
tests: align pytest version across requirements

guybortnikov:
## Description
Align `pytest` across pip requirements files


guybortnikov:
> Is this file even needed? If there was a mismatch but CI & build pass, is it even used? One of the 2 is not required I presume


I was afraid.. now removed from `dev-requirements`, let's see CI passes

Jongy:
I don't think it'll pass because the tests don't install dev-requirements.txt. For all I care you can merge both into dev-requirements.txt and install it in `test.sh`.

test-requirements.txt
```diff
@@ -1,4 +1,3 @@
-pytest==6.2.4
 PyYAML==6.0
 docker==5.0.0
 six==1.16.0
```
Jongy, line 3: Let's just remove this file and move the other things to `dev-requirements.txt`. It's not needed & confusing

guybortnikov: done

dev-requirements.txt
```diff
@@ -1,8 +1,11 @@
-pytest==7.0.1
+pytest==7.0.1 # pytest types are needed for linting (dev process)
```
Jongy, line 1: Comment irrelevant anymore?

Jongy (approved):
Can merge once tests pass

Issue #612, opened 2022-11-27
perf: hangs in wait_and_script()


Jongy:
I encountered a hanging gProfiler 1.7.0. Ran `py-spy dump` and got this stack trace:
```
Thread 0x7F768F77C740 (idle): "MainThread"
    wait (threading.py:295)
    result (_base.py:427)
    _snapshot (main.py:278)
    run_continuous (main.py:369)
    main (main.py:841)
    <module> (pyi_build.py:13)
Thread 0x7F76848B4700 (idle): "Thread-6"
    wait (threading.py:299)
    wait (threading.py:551)
    _continuously_poll_memory (system_metrics.py:76)
    run (threading.py:864)
    _bootstrap_inner (threading.py:916)
    _bootstrap (threading.py:884)
Thread 0x7F767F7BE700 (idle): "Process Events Listener"
    select (selectors.py:445)
    _listener_loop (proc_events.py:129)
    _proc_events_listener (proc_events.py:198)
    run (threading.py:864)
    _bootstrap_inner (threading.py:916)
    _bootstrap (threading.py:884)
Thread 0x7F767FFBF700 (idle): "ThreadPoolExecutor-0_0"
    wait (threading.py:295)
    get (queue.py:164)
    _worker (thread.py:67)
    run (threading.py:864)
    _bootstrap_inner (threading.py:916)
    _bootstrap (threading.py:884)
Thread 0x7F767EFBD700 (idle): "ThreadPoolExecutor-0_1"
    wait (threading.py:295)
    get (queue.py:164)
    _worker (thread.py:67)
    run (threading.py:864)
    _bootstrap_inner (threading.py:916)
    _bootstrap (threading.py:884)
Thread 0x7F767E7BC700 (idle): "ThreadPoolExecutor-0_2"
    wait (threading.py:295)
    get (queue.py:164)
    _worker (thread.py:67)
    run (threading.py:864)
    _bootstrap_inner (threading.py:916)
    _bootstrap (threading.py:884)
Thread 0x7F767DFBB700 (idle): "ThreadPoolExecutor-0_3"
    wait (threading.py:295)
    get (queue.py:164)
    _worker (thread.py:67)
    run (threading.py:864)
    _bootstrap_inner (threading.py:916)
    _bootstrap (threading.py:884)
Thread 0x7F767D7BA700 (active): "ThreadPoolExecutor-0_4"
    wait_and_script (perf.py:109)
    snapshot (perf.py:242)
    run (thread.py:56)
[...]

PR #619, opened 2022-11-30
perf: Set stdout and stderr streams of perf process to non blocking

mpozniak95:
## Description
This PR sets stdout and stderr streams of perf process to non blocking.

## Related Issue
https://github.com/Granulate/gprofiler/issues/612

## How Has This Been Tested?
Added `raise exception()` to try block in `wait_and_script()` -> on master - gProfiler stucks on reading from stdout and stderr (like in related issue), after these changes it reads the stream and continues

### DIFF
--- a/gprofiler/profilers/perf.py
+++ b/gprofiler/profilers/perf.py
@@ -89,8 +89,8 @@ class PerfProcess:
             raise
         else:
             self._process = process
-            os.set_blocking(self._process.stdout.fileno(), False)
-            os.set_blocking(self._process.stderr.fileno(), False)
+            os.set_blocking(self._process.stdout.fileno(), False)  # type: ignore
+            os.set_blocking(self._process.stderr.fileno(), False)  # type: ignore
             logger.info(f"Started perf ({self._type} mode)")
 
     def stop(self) -> None:
--- a/gprofiler/profilers/perf.py
+++ b/gprofiler/profilers/perf.py
@@ -89,6 +89,8 @@ class PerfProcess:
             raise
         else:
             self._process = process
+            os.set_blocking(self._process.stdout.fileno(), False)
+            os.set_blocking(self._process.stderr.fileno(), False)
             logger.info(f"Started perf ({self._type} mode)")
 
     def stop(self) -> None:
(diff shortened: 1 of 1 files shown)


gprofiler/profilers/perf.py
```diff
@@ -107,6 +107,8 @@ def wait_and_script(self) -> str:
             perf_data = wait_for_file_by_prefix(f"{self._output_path}.", self._dump_timeout_s, self._stop_event)
         except Exception:
             assert self._process is not None and self._process.stdout is not None and self._process.stderr is not None
+            os.set_blocking(self._process.stdout.fileno(), False)
```
Jongy, line 110: Let's move these 2 to the `start()` method, not here. Can do that in the `else` branch in `start()`.

gprofiler/profilers/perf.py
```diff
@@ -107,6 +107,8 @@ def wait_and_script(self) -> str:
...
+            os.set_blocking(self._process.stderr.fileno(), False)
             logger.critical(
                 f"perf failed to dump output. stdout {self._process.stdout.read()!r}"
```
Jongy, line 113: What happens to this `read()` when it's nonblocking? We get back an empty string?

mpozniak95: We get None

Jongy: kk

Jongy:
I managed to reproduce with this diff:
```
diff --git a/gprofiler/profilers/perf.py b/gprofiler/profilers/perf.py
index 990e077e..a385d8b1 100644
--- a/gprofiler/profilers/perf.py
+++ b/gprofiler/profilers/perf.py
@@ -264,8 +264,8 @@ class SystemProfiler(ProfilerBase):
         if self._stop_event.wait(self._duration):
             raise StopEventSetException
 
-        for perf in self._perfs:
-            perf.switch_output()
+        # for perf in self._perfs:
+        #     perf.switch_output()
 
         return {
             k: self._generate_profile_data(v, k)
```

we don't send a signal, so perf doesn't write the output file, then we time-out in `wait_for_file_by_prefix` and eventually I get the same hang in
```
Thread 0x7F11117FA700 (active): "ThreadPoolExecutor-0_4"
    wait_and_script (gprofiler/profilers/perf.py:111)
    snapshot (gprofiler/profilers/perf.py:273)
    run (concurrent/futures/thread.py:57)
    _worker (concurrent/futures/thread.py:80)
    run (threading.py:870)
    _bootstrap_inner (threading.py:932)
    _bootstrap (threading.py:890)
```

On your branch - the same diff doesn't hang and just keeps erroring. Assuming the problem is interleaved, it will repair itself I bet.

Jongy (approved):

Issue #616, opened 2022-11-29
pyperf: Upgrade bcc


Jongy:
As suggested in https://github.com/Granulate/gprofiler/pull/579#issuecomment-1331134830, we haven't upgraded bcc in a while (or rather, never since we incorporated it). It's a live project that probably received bugifxes and performance improvements. We have proper tests in place for pyperf, so we should give it a try.

marcin-ol:
One of good reasons for upgrading bcc - a commit in the upstream bcc: 
[bcc/syms: Fix shared lib module offset <-> global addr conversion](https://github.com/iovisor/bcc/commit/4b3cbf44ba9eb9d5a7d11c69a7238497889ce2fc)

Issue #582, opened 2022-11-06
Consolidate image & executable build into exe-in-image


Jongy:
We have 2 Dockerfiles (`Dockerfile` and `Dockerfile.pyi`) which result in an image and in an exe. Instead of maintaining 2 builds, we can remove the image build, and instead the image Dockerfile would become:
```
FROM scratch # or an almost-empty image
COPY build/.../gprofiler /gprofiler
ENTRYPOINT /gprofiler
```

Upsides:
1. Only one build to maintain
2. Smaller image size (100mb ~ as opposed to 300-400 now)

Downsides:
1. Image build is faster for local testing & execution, building the exe takes longer so this makes development cycles a bit longer. We can support local execution with `python -m gprofiler` (haven't tried that in a long while).

Jongy:
I think we'll go for this after https://github.com/Granulate/gprofiler/pull/597 is merged and we're using a newer Python in the exe, I don't want to force all releases to use the Python 3.6 we have in the exe which has some bugs that we encounter in flaky tests.

Jongy:
Additional elaboration:
1. Ideally, the base image should be empty (since staticx-ed binaries have no dependencies). If possible - the best would be an EMPTY image with the `gprofiler` binary in `/`. In Dokcer that's `FROM scratch` as the base of the Dockerfile.
2. Personally, I use the container build when debugging/prototyping because it's faster to build than the pyinstaller+staticx. Once we remove the standard container build, I won't have that option anymore, and so, I'd like a replacement. One option is to be able to build the exe while skipping the staticx - this is considerably faster, while not enough for all profilers (PyPerf is not built statically at the moment, but I can live with that and use the full build when I want to test PyPerf stuff). Any other idea on how to remain with fast (less than 5 seconds) time for the build are acceptable :pray:
3. `ENV GPROFILER_IN_CONTAINER=1` should be kept.
4. The `build_..._container.sh` scripts should be simple now and not contain any reference to other base images. In other words, only `build_..._executable.sh` should contain references to the base images `CENTOS_VERSION` etc.

Jongy:
More notes:
* @mpozniak95 correctly noted that gProfiler uses shell - `/bin/sh` - for running commands, e.g in `pgrep_maps`. So we decided to go with `alpine` as a base image, not `scratch`.
[...]

PR #622, opened 2022-12-05
Build latest version of bcc

marcin-ol:
Upgrade bcc in project.

## Description
Bcc fork has been upgraded with changes from upstream bcc.

## Related Issue
https://github.com/Granulate/gprofiler/issues/616

## Motivation and Context
[bcc/syms: Fix shared lib module offset <-> global addr conversion](https://github.com/iovisor/bcc/commit/4b3cbf44ba9eb9d5a7d11c69a7238497889ce2fc)

## How Has This Been Tested?
Regular battery of tests.

## Checklist:


Jongy:
Nice.

Why did you upgrade LLVM to 12? Is this a requirement now?

Jongy:
1. On BCC version - on https://github.com/iovisor/bcc/issues/3881 it's claimed that 0.24.0 supports LLVM 7, I see that [0.24.0 is from Jan 14](https://github.com/iovisor/bcc/releases/tag/v0.24.0) which is a bit old, so that's why we want the latest? (which should be 0.25.0 and NOT the master branch, btw). If we go with 0.24.0, does it get built?
2. Considering that we want to upgrade to 0.25.0 and it truly requires upgrading LLVM - then in the container build, we build PyPerf on `ubuntu:20.04` and for the exe build we build over CentOS. I don't recall why this discrepancy exists. We can very likely upgrade both to Ubuntu 20.04 or even 22.04.
3. I saw https://github.com/Granulate/bcc/pull/41 - it shows the entire diff. Did you merge the 0.25.0 tag into master? I wonder if there's a way to see the "diff" or "conflicts" from the merge, since I'm not going to review the entire diff haha. Can you post the conflicts you had on that PR?
4. As for testing PyPerf itself - PyPerf is an executable that's built with BCC, and in runtime PyPerf builds the eBPF program, and this step is most likely failing after you've upgraded PyPerf. So the container/exe might build but PyPerf will fail to start. In order to debug it, I suggest you set up an environment to build & debug PyPerf, this will make iterations faster.. You can follow my instructions [here](https://github.com/Granulate/gprofiler/blob/master/CONTRIBUTING.md#debugging-the-build) to "stop" the build after PyPerf is built, then run that container as privileged and observe the errors PyPerf gets, then you can edit the code and just re-build PyPerf immediately and re-try it. You can post here the errors you get when PyPerf starts (if you get any) and I'll help :)

marcin-ol:
> 1. On BCC version - on [BCC does not support LLVLM 7 or earlier anymore with the master branch iovisor/bcc#3881](https://github.com/iovisor/bcc/issues/3881) it's claimed that 0.24.0 supports LLVM 7, I see that [0.24.0 is from Jan 14](https://github.com/iovisor/bcc/releases/tag/v0.24.0) which is a bit old, so that's why we want the latest? (which should be 0.25.0 and NOT the master branch, btw). If we go with 0.24.0, does it get built?
- Yes, that’s the reason. Fixes directly mentioning the symbol resolution process landed in 0.25.
- Yeah, I realized a bit later I should’ve referenced tag 0.25. I intend to restart this merge properly.
> 2. Considering that we want to upgrade to 0.25.0 and it truly requires upgrading LLVM - then in the container build, we build PyPerf on `ubuntu:20.04` and for the exe build we build over CentOS. I don't recall why this discrepancy exists. We can very likely upgrade both to Ubuntu 20.04 or even 22.04.
- We link PyPerf statically, right? So if possible I would like to rework the Dockerfiles to use common process for PyPerf, with Ubuntu 20.04.
> 3. I saw [Upgrade bcc reference bcc#41](https://github.com/Granulate/bcc/pull/41) - it shows the entire diff. Did you merge the 0.25.0 tag into master? I wonder if there's a way to see the "diff" or "conflicts" from the merge, since I'm not going to review the entire diff haha. Can you post the conflicts you had on that PR?
- I did a rebase actually; I will retry this upgrade with a merge.
> 4. As for testing PyPerf itself - PyPerf is an executable that's built with BCC, and in runtime PyPerf builds the eBPF program, and this step is most likely failing after you've upgraded PyPerf. So the container/exe might build but PyPerf will fail to start. In order to debug it, I suggest you set up an environment to build & debug PyPerf, this will make iterations faster.. You can follow my instructions [here](https://github.com/Granulate/gprofiler/blob/master/CONTRIBUTING.md#debugging-the-build) to "stop" the build after PyPerf is built, then run that container as privileged and observe the errors PyPerf gets, then you can edit the code and just re-build PyPerf immediately and re-try it. You can post here the errors you get when PyPerf starts (if you get any) and I'll help :)
- Oh, I’m working in a similar way. I noticed a build up to PyPerf takes mere seconds, and PyPerf needs <2mins to build completely.   
  Then I run resulting gprofiler container and change its entrypoint to PyPerf.
- I will post the errors next time, when I redo the merge.

marcin-ol:
PR mentioned previously (https://github.com/Granulate/gprofiler/pull/622#issuecomment-1342345155) was opened only as a reference.
The valid PR for bcc is: https://github.com/Granulate/bcc/pull/41  
For now only a container build of gProfiler builds a working PyPerf. Executable build will be updated after verifying the container build.

marcin-ol:
Building a PyPerf from this branch results in a valid binary, but the following warning is produced while starting PyPerf:
```
sh: 1: modprobe: not found
Unable to find kernel headers. Try rebuilding kernel with CONFIG_IKHEADERS=m (module) or installing the kernel development package for your running kernel version.
Ready to profile
```

Jongy:
> * Yes, that’s the reason. Fixes directly mentioning the symbol resolution process landed in 0.25.
> * Yeah, I realized a bit later I should’ve referenced tag 0.25. I intend to restart this merge properly.

Kk.

> * We link PyPerf statically, right? So if possible I would like to rework the Dockerfiles to use common process for PyPerf, with Ubuntu 20.04.

It's linked dynamically but gProfiler exe uses `staticx` to push inside all dependencies of gProfiler. I think the reason to build PyPerf on CentOS 7 in the exe build, is because it's dynamic and with `staticx` it will have all of its dependencies brought from the fs of CentOS 7. If we build on another image, the dependencies of PyPerf won't interop with the dependencies of e.g the Python interpreter itself (to begin with: there will be a different ld.so of a different version).

Also, see https://github.com/Granulate/gprofiler/issues/582 which is why I'm not sure if I want to spend efforts into it now, I think we'll end up staying with just one eventually.

> * I did a rebase actually; I will retry this upgrade with a merge.

Perhaps with a rebase it'd be easier (we can use `git range-diff` to view the differences). In any case I'd like to see the diff somehow.

> * Oh, I’m working in a similar way. I noticed a build up to PyPerf takes mere seconds, and PyPerf needs <2mins to build completely.
>   Then I run resulting gprofiler container and change its entrypoint to PyPerf.

Oh okay, that's a fine method as well :)

Jongy:
> Building a PyPerf from this branch results in a valid binary, but the following warning is produced while starting PyPerf:
> 
> ```
> sh: 1: modprobe: not found
> Unable to find kernel headers. Try rebuilding kernel with CONFIG_IKHEADERS=m (module) or installing the kernel development package for your running kernel version.
> Ready to profile
> ```

I think this is about https://github.com/Granulate/bcc/commit/3fd97d89efb79553a20e16653c2538b14674f6c6.

By design, PyPerf tries to get kernel headers, to make them available for the eBPF program compilation. I think this failing modprobe is for the module "kheaders" or "ikheaders" which exposes the kernel headers, PyPerf tries to load it to be able to use kernel headers (see https://elixir.bootlin.com/linux/v6.1/source/kernel/kheaders.c).

However- we don't need it, PyPerf strives to be header-less so it can runon systems w/o kernel headers installed / without the ikheaders module. It's possible that the new BCC version attempts to load this module. We need to disable it, as you can see e.g in https://github.com/Granulate/bcc/commit/3fd97d89efb79553a20e16653c2538b14674f6c6#diff-47c39b83cd86c86e601398274e965e302e009105f9b2a2eecc6619fb93cf816f I did some tricks to prevent BCC from trying to use kernel headers.

Jongy:
@marcin-ol given the plan in https://github.com/Granulate/gprofiler/issues/582, I think the more interesting case to solve is the build in container mode - 
* Either we get it built in CentOS 7
* Or we get it build in a new system such as Ubuntu 22.04, but get the resulting binary and its dependencies pulled into the staticx build, in a way that they interop with the binaries on the building image.

By the way, the reason we build on CentOS 7 is to be able to run on old systems such as CentOS 6. If I build a *static* executable on my Ubuntu 20.04 and try to run it on CentOS 6, I get:
```
[centos@ip-xxx ~]$ ./a
FATAL: kernel too old
Aborted
```

because glibc, at some point, dropped support for old kernels (not sure from which version)

marcin-ol:
Minor __new__ warning observed when starting PyPerf:
```
Failed to iterate over ELF symbols: /proc/946/root/usr/bin/python3.8 (deleted)
[   946] Setup new python failed
Failed to iterate over ELF symbols: /proc/1263/root/usr/bin/python3.8 (deleted)
[  1263] Setup new python failed

$ sudo xargs -0n1 </proc/1263/cmdline
/usr/bin/python3
/usr/share/unattended-upgrades/unattended-upgrade-shutdown
--wait-for-signal
$ sudo xargs -0n1 </proc/946/cmdline
/usr/bin/python3
/usr/bin/networkd-dispatcher
--run-startup-triggers
```

Jongy:
> Minor **new** warning observed when starting PyPerf:
> 
> ```
> Failed to iterate over ELF symbols: /proc/946/root/usr/bin/python3.8 (deleted)
> [   946] Setup new python failed
> Failed to iterate over ELF symbols: /proc/1263/root/usr/bin/python3.8 (deleted)
> [  1263] Setup new python failed
> 
> $ sudo xargs -0n1 </proc/1263/cmdline
> /usr/bin/python3
> /usr/share/unattended-upgrades/unattended-upgrade-shutdown
> --wait-for-signal
> $ sudo xargs -0n1 </proc/946/cmdline
> /usr/bin/python3
> /usr/bin/networkd-dispatcher
> --run-startup-triggers
> ```

These are expected - BCC doesn't handle files that were removed, it happens probably because the Python was upgraded (e.g `apt upgrade python3`) and there are still files running from the old binary (see the ` (deleted)` suffix).

marcin-ol:
Changed the way that PyPerf is being built and used in gProfiler:
- Now bcc is built using almost the same instructions as container buid.
- PyPerf is packed using staticx to ensure all libraries it needs are available.
- Launching PyPerf under gProfiler was modified to treat PyPerf as regular binary, not as part of staticx context.
- Finally we also need a modified staticx bootloader, hence we build patched staticx. Otherwise dump signal `USR2` doesn't reach PyPerf executable but kills staticx bootloader instead.

gprofiler/profilers/python_ebpf.py
```diff
@@ -157,7 +157,7 @@ def test(self) -> None:
             "--duration",
             "1",
         ] + self._offset_args()
-        process = start_process(cmd, via_staticx=True)
+        process = start_process(cmd, via_staticx=False)
```
Jongy, line 160: This is the last site using `via_staticx=True`, let's make it a default parameter with `False` value and remove the `via_staticx=False` everywhere.
I do suggest we keep support for `via_staticx` in case we need it in the future.

Jongy:
> PyPerf is packed using staticx to ensure all libraries it needs are available.

I didn't understand why this step is required. In `list_needed_libs.sh` we pack all dependencies of all binaries under `gprofiler/resources` into the "staticx context" of gProfiler itself - so it should include everything that PyPerf requires, just as it has included the relevant libraries in the past. If we can do it that way there will be much less diff.

Jongy:
To review the changes between current BCC and BCC based on v0.25.0, I can think of 2 ways:
1. If rebase was used - I can use `git range-diff` to compare the Granulate-added commits.
2. If merge was used - I suppose I can just review the diff under `examples/cpp/...`.

Reviewing the diff of rebase would be easier, I think. I see that you went with merge though. Would it be possible to re-implement as a rebase without too much hassle?

gprofiler/profilers/python_ebpf.py
marcin-ol: 👍 
Will do.

marcin-ol:
> > PyPerf is packed using staticx to ensure all libraries it needs are available.
> 
> I didn't understand why this step is required. In `list_needed_libs.sh` we pack all dependencies of all binaries under `gprofiler/resources` into the "staticx context" of gProfiler itself - so it should include everything that PyPerf requires, just as it has included the relevant libraries in the past. If we can do it that way there will be much less diff.

This has changed as PyPerf is now built in a separate stage, based on Ubuntu.  It wouldn't build on Centos 7.9, as it needs recent llvm (>10).
Due to this pyperf needs its own set of libraries, coming from Ubuntu and won't load properly with Centos runtime libs (actually it segfaulted).

Jongy:
@marcin-ol can you please merge master and resolve conflicts?

Jongy:
My plan to review the diff is:
```
gd 711f0302 new_head > bcc_diff_new
gd cf6fe1c3 old_head > bcc_diff_old
```

diff the diffs.

pyi.Dockerfile
```diff
@@ -111,33 +111,72 @@ RUN ./node_builder_musl_env.sh
...
+    fi && \
+    ./libunwind_build.sh
+
+WORKDIR /bcc
+COPY ./scripts/pyperf_build.sh .
+RUN ./pyperf_build.sh
 
-# bcc & gprofiler
+# build staticx dedicated for PyPerf and process PyPerf binary;
+# apply patch to ensure staticx bootloader propagates dump signal to actual PyPerf binary
```
Jongy, line 161: Remind me - we chose this approach to have all different PyPerf dependency libraries packed into the final binary? What was the reason we decided not to make use of the existing `./scripts/list_needed_libs.sh` which packs relevant libraries into the staticx context of gProfiler itself?

pyi.Dockerfile
```diff
@@ -343,10 +331,9 @@ COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # hadolint ignore=SC2046,SC2086
 RUN set -e; \
     if [ $(uname -m) != "aarch64" ]; then \
-        source scl_source enable devtoolset-8 llvm-toolset-7 ; \
```
Jongy, line 346: Not needed here anymore - because we needed devtoolset for BCC and now it's a separate layer so we don't require it anymore?

pyi.Dockerfile
```diff
@@ -111,33 +111,72 @@ RUN ./node_builder_musl_env.sh
...
 
-# bcc & gprofiler
+# build staticx dedicated for PyPerf and process PyPerf binary;
+# apply patch to ensure staticx bootloader propagates dump signal to actual PyPerf binary
+COPY scripts/staticx_for_pyperf_patch.diff staticx_for_pyperf_patch.diff
+# hadolint ignore=DL3003
+RUN if [ "$(uname -m)" = "aarch64" ]; then \
+      exit 0; \
+    fi && \
+    git clone -b v0.13.8 https://github.com/JonathonReinhart/staticx.git && \
```
Jongy, line 167: Any particular reason you upgraded to v0.13.8? Just for the sake of a new software, or was there a problem with v0.13.6?

pyi.Dockerfile
```diff
@@ -111,33 +111,72 @@ RUN ./node_builder_musl_env.sh
...
+WORKDIR /bcc
+COPY ./scripts/pyperf_build.sh .
+RUN ./pyperf_build.sh
 
-# bcc & gprofiler
+# build staticx dedicated for PyPerf and process PyPerf binary;
+# apply patch to ensure staticx bootloader propagates dump signal to actual PyPerf binary
+COPY scripts/staticx_for_pyperf_patch.diff staticx_for_pyperf_patch.diff
+# hadolint ignore=DL3003
+RUN if [ "$(uname -m)" = "aarch64" ]; then \
```
Jongy, line 164: Move the build after libunwind build? staticx needn't be rebuilt if pyperf is rebuilt, and currently it will be since it follows the pyperf build in the dockerfile.

As a rule of thumb - build & env stuff first, then copying in sources and building the actual software.

pyi.Dockerfile
```diff
@@ -111,33 +111,72 @@ RUN ./node_builder_musl_env.sh
...
+      git \
+      patchelf scons
 
 COPY --from=perf-builder /bpftool /bpftool
 
 COPY scripts/bcc_helpers_build.sh .
 RUN ./bcc_helpers_build.sh
 
+# build bcc
+# TODO: copied from the main Dockerfile... but modified a lot. we'd want to share it some day.
```
Jongy, line 147: Btw - now that BCC & BCC helpers are built both on Ubuntu, can we share more of their build? Only if easy.

scripts/pyperf_build.sh
```diff
@@ -5,7 +5,7 @@
 #
 set -euo pipefail
 
-git clone --depth 1 -b v1.2.6 https://github.com/Granulate/bcc.git && cd bcc && git reset --hard 6a995642d12287f26cbb97edd3d29e62a5fde337
+git clone --depth 1 -b upgrade-bcc-reference https://github.com/marcin-ol/bcc.git && cd bcc && git reset --hard 863d545
```
Jongy, line 8: `863d545` is still the commit from Dec 2022, I recall you said you upgraded it? Or there was no need to update BCC again?

Jongy (changes requested):
I finished reviewing. I am revieiwng the BCC diff itself now.

pyi.Dockerfile
marcin-ol: 👍 
That's a good idea, I guess it should work.
I'll look into it.

pyi.Dockerfile
marcin-ol: That's because bcc is now built with Ubuntu, borrowing process from gProfiler container build.  

Could no longer build newer version in Centos 7, due to lack of recent builds of LLVM.

pyi.Dockerfile
marcin-ol: Will look into it.

pyi.Dockerfile
marcin-ol: No, I just chose the recent one.
But to speed up the build I will change back to v0.13.6. Nothing changed in bootloader v0.13.2..v0.13.8.
👍 Will fix.

pyi.Dockerfile
marcin-ol: That's right. We no longer need those toolsets around for staticx to compile gprofiler.

scripts/pyperf_build.sh
marcin-ol: Our reference is still bcc 0.25.
We might upgrade to 0.27, but I believe it doesn't bring us anything immediately useful besides perhaps this:
- [Implement symbol lookup in zipped libraries. #4440](https://github.com/iovisor/bcc/pull/4440)

I guess we needn't upgrade further atm. 
But I might try to check the upgrade anyway.

Jongy:
The failures appear to be real? (in PyPerf tests)

pyi.Dockerfile
Jongy: Thanks. Let's add it in a comment in this file? (we're using staticx in a separate env due to ubntuu vs centos ...)

scripts/pyperf_build.sh
Jongy: Not a must. Can be done a separate PR.

Jongy:
On my machine (Ubuntu 22.04):
```
[2023-04-23 17:10:57,919] DEBUG: gprofiler.utils: Running command (command=['/tmp/_MEIp27VnO/gprofiler/resources/python/pyperf/PyPerf', '--fs-offset', '5416', '--stack-offset', '32', '--output', '/tmp/gprofiler_tmp/tmpkbec5y9m/pyperf.PAeDBJZOqEHzPMAi.col', '-F', '1', '--duration', '1'])
[2023-04-23 17:10:58,024] DEBUG: gprofiler.profilers.python: eBPF profiler error: Command ['/tmp/_MEIp27VnO/gprofiler/resources/python/pyperf/PyPerf', '--fs-offset', '5416', '--stack-offset', '32', '--output', '/tmp/gprofiler_tmp/tmpkbec5y9m/pyperf.PAeDBJZOqEHzPMAi.col', '-F', '1', '--duration', '1'] died with <Signals.SIGSEGV: 11>.
stdout: 
stderr: 
[2023-04-23 17:10:58,025] INFO: gprofiler.profilers.python: Python eBPF profiler initialization failed
```
(built from master)

pyi.Dockerfile
Jongy: I mean between the 2 dockerfiles - both now use the same base - `FROM ubuntu${PYPERF_BUILDER_UBUNTU} AS bcc-builder-base` in Dockerfile and `FROM ubuntu${PYPERF_BUILDER_UBUNTU} AS bcc-build` in pyi.Dockerfile, so they can be shared with eg `bcc_env.sh` and `bcc_build.sh` scripts like the rest of the artifacts that we build.

pyi.Dockerfile
marcin-ol: 👍

pyi.Dockerfile
marcin-ol: 👍

scripts/pyperf_build.sh
```diff
@@ -17,5 +32,15 @@ fi
...
-cmake -DPYTHON_CMD=python3 -DINSTALL_CPP_EXAMPLES=y -DCMAKE_INSTALL_PREFIX=/bcc/root ..
+cmake -DPYTHON_CMD=python3 -DINSTALL_CPP_EXAMPLES=y -DCMAKE_INSTALL_PREFIX=/bcc/root -DBCC_CLOADER_KERNEL_HEADERLESS=1 ..
 make -C examples/cpp/pyperf -j -l VERBOSE=1 install
+# leave build directory
+cd ..
+# leave bcc repository
+cd ..
+
+# We're using staticx to build a distribution-independent binary of PyPerf because PyPerf
+# can only build with latest llvm (>10), which cannot be obtained on CentOS.
```
Jongy, line 43:
```suggestion
# can only build with latest llvm (>10), which cannot be obtained on CentOS 7 (which is the base for the executable environment).
```

scripts/pyperf_build.sh
```diff
@@ -5,7 +5,22 @@
...
+    with_staticx="$1"
+else
+    echo "Unexpected argument: $1"
+    exit 1
+fi
+
+if [ "$(uname -m)" = "aarch64" ]; then
+    exit 0;
+fi
+git clone --depth 1 -b upgrade-bcc-to-0.27 https://github.com/marcin-ol/bcc.git && cd bcc && git reset --hard e1657a1f
```
Jongy, line 23: Reminder to udpate before merge

scripts/pyperf_env.sh
```diff
@@ -0,0 +1,72 @@
...
+    libz-dev \
+    liblzma-dev \
+    ca-certificates \
+    git \
+    patchelf scons
+
+if [ -n "$with_staticx" ]; then
+    apt-get install -y --no-install-recommends \
+        python3-pip ca-certificates git \
+        patchelf scons
```
Jongy, line 44: These are all already installed by the previous installer, so remove from either location?

scripts/pyperf_env.sh
```diff
@@ -0,0 +1,72 @@
...
+    rm -rf staticx
+fi
+
+./bcc_helpers_build.sh
+
+# cd /tmp
+# if [ "$(uname -m)" = "aarch64" ]; then
+#     exit 0
+# fi
+# ./libunwind_build.sh
```
Jongy, line 69: I see you put it outside in the dockerfile, so remove here?

Jongy (changes requested):
Done reviewing. Also tested locally, all looks well :)

I finished reviewing the BCC side as well, and left a comment: https://github.com/Granulate/bcc/pull/44#pullrequestreview-1406743442

scripts/pyperf_env.sh
marcin-ol: 👍

scripts/pyperf_env.sh
marcin-ol: 👍

scripts/pyperf_build.sh
marcin-ol: 👍

Jongy (approved):
At last :)

PR #624, opened 2022-12-07
Enable building glibc on aarch64

mpozniak95:
## Description
Enable building glibc resources for aarch64

## How it has been tested?
Built aarch64 executable via build_aarch64_executable.sh script. Then run gprofiler to profile Fibonacci application run on centos7


scripts/build_aarch64_container.sh
```diff
@@ -17,8 +17,8 @@ GOLANG_VERSION=@sha256:f7d3519759ba6988a2b73b5874b17c5958ac7d0aa48a8b1d84d66ef25
 ALPINE_VERSION=@sha256:b06a5cf61b2956088722c4f1b9a6f71dfe95f0b1fe285d44195452b8a1627de7
 # mcr.microsoft.com/dotnet/sdk:6.0-focal
 DOTNET_BUILDER=@sha256:749439ff7a431ab4bc38d43cea453dff9ae1ed89a707c318b5082f9b2b25fa22
-# node-package-builder-glibc - centos7:latest
-NODE_PACKAGE_BUILDER_GLIBC=@sha256:65a4aad1156d8a0679537cb78519a17eb7142e05a968b26a5361153006224fdc
+# node-package-builder-glibc - centos8
```
Jongy, line 20: `centos8:latest`?

If we build on CentOS 8, does it load on CentOS 7?

pyi.Dockerfile
```diff
@@ -13,7 +13,7 @@ ARG GPROFILER_BUILDER
 ARG PYPERF_BUILDER_UBUNTU
 ARG DOTNET_BUILDER
 ARG NODE_PACKAGE_BUILDER_MUSL
-ARG NODE_PACKAGE_BUILDER_GLIBC
+ARG NODE_PACKAGE_BUILDER_GLIBC=build-prepare
```
Jongy, line 16: Instead of specifying a default value here, I'd pass this value in the Aarch64 scripts (so that the x86_64 scripts don't have any interaction with it).

Jongy: I know it's different from the `Dockerfile` where the defaults are for x86_64, I think we should remove the defaults there as well :sweat_smile: and keep it consistent.

scripts/build_aarch64_container.sh
mpozniak95: It loads because we are building with: `-static-libstdc++`

scripts/build_node_package.sh
```diff
@@ -26,6 +20,10 @@ export NAN_PATH
 sed -i 's/node \-e \\"require('\''nan'\'')\\"/echo $NAN_PATH/g' binding.gyp
 rm -rf nan.tar.gz
 mkdir $BUILD_TARGET_DIR
+if [ "$(uname -m)" = "aarch64" ] && [ "$(< /etc/os-release grep VERSION | head -n1 | cut -d \" -f2)" = "8" ]; then
```
Jongy, line 23:
```suggestion
if [ "$(uname -m)" = "aarch64" ] && [ "$(grep VERSION /etc/os-release | head -n1 | cut -d \" -f2)" = "8" ]; then
```
More readable :) Also, whenever you're writing parsing code, it's nicer if you explain in comment what is the expected input you're going to parse, and what is the expected result, and why are you parsing. So - an example line of "VERSION", and the reason that we're trying to extract the CentOS version, to check if 8.

scripts/build_node_package.sh
```diff
@@ -26,6 +20,10 @@ export NAN_PATH
...
+    export LDFLAGS="-static-libstdc++"
+    export CPPFLAGS="-static-libstdc++"
```
Jongy, line 25: Do we need both?

scripts/build_node_package.sh
Jongy: Btw in one site we use `grep -q "CentOS Linux 8" /etc/os-release` and here we do something else, be consistent :sweat_smile:

scripts/build_aarch64_container.sh
mpozniak95: I am not sure if this is centos8:latest though. I took image that is already used in build_aarch64_executable as base for gprofiler_builder. I'll change it to centos8:latest and check if it will still work.

mpozniak95: Changed to centos8:latest and it works.

scripts/build_node_package.sh
mpozniak95: I checked and only LDFLAGS matter

scripts/node_builder_glibc_env.sh
```diff
@@ -4,20 +4,24 @@
...
 yum -y install openssl-devel bzip2-devel
-curl -L https://www.python.org/ftp/python/3.7.9/Python-3.7.9.tgz -o Python-3.7.9.tgz
-tar xzf Python-3.7.9.tgz
-cd Python-3.7.9
-./configure --enable-optimizations
-make altinstall
-cd ..
\ No newline at end of file
+
+if [ "$(uname -m)" != "aarch64" ]; then
```
Jongy, line 16: In the x86_64 exe build we're also building Py 3.7, so we can use the same Python instead of installing? Not very important anyway.. maybe for a next PR or for when we finally get rid of the container build.

scripts/build_node_package.sh
```diff
@@ -26,6 +20,9 @@ export NAN_PATH
 sed -i 's/node \-e \\"require('\''nan'\'')\\"/echo $NAN_PATH/g' binding.gyp
 rm -rf nan.tar.gz
 mkdir $BUILD_TARGET_DIR
+if grep -q "CentOS Linux 8" /etc/os-release; then
```
Jongy, line 23: Add explaining comment why

scripts/build_aarch64_container.sh
```diff
@@ -17,8 +17,8 @@ GOLANG_VERSION=@sha256:f7d3519759ba6988a2b73b5874b17c5958ac7d0aa48a8b1d84d66ef25
 ALPINE_VERSION=@sha256:b06a5cf61b2956088722c4f1b9a6f71dfe95f0b1fe285d44195452b8a1627de7
 # mcr.microsoft.com/dotnet/sdk:6.0-focal
 DOTNET_BUILDER=@sha256:749439ff7a431ab4bc38d43cea453dff9ae1ed89a707c318b5082f9b2b25fa22
-# node-package-builder-glibc - centos7:latest
-NODE_PACKAGE_BUILDER_GLIBC=@sha256:65a4aad1156d8a0679537cb78519a17eb7142e05a968b26a5361153006224fdc
+# node-package-builder-glibc - centos8:latest
```
Jongy, line 20: Image is probably `centos:8` not `centos8:latest`? :)

Jongy: Also, how come the hash didn't change if we changed the image? o.O

It's truly `centos:8`: https://hub.docker.com/layers/library/centos/8/images/sha256-65a4aad1156d8a0679537cb78519a17eb7142e05a968b26a5361153006224fdc?context=explore

Dockerfile
```diff
@@ -161,8 +161,16 @@ COPY scripts/build_node_package.sh .
...
 # node-package-builder-glibc
-FROM centos${NODE_PACKAGE_BUILDER_GLIBC} AS node-package-builder-glibc
+FROM ${NODE_PACKAGE_BUILDER_GLIBC} AS node-package-builder-glibc
 USER 0
+RUN if grep -q "CentOS Linux 8" /etc/os-release ; then \
+        sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*; \
+        sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*; \
+        yum install -y dnf-plugins-core; \
+        dnf config-manager --set-enabled powertools; \
+        yum clean all; \
```
Jongy, line 171: This part repeats, can we move to a shared script? i.e `scripts/fix_centos8.sh`

scripts/node_builder_glibc_env.sh
mpozniak95: centos/devtoolset-7-toolchain-centos7:latest image doesn't have python3 installed by default. We could use centos8 image and use static libstdc++ as well for x86_64, but I am not sure if it is the best way since centos 8 EOL.

Jongy: Wait.... also for x86_64 we don't need it anymore. We are building Python 3.7 for x86_64, but we have 3.8 in the container build and 3.10 in the exe build (since https://github.com/Granulate/gprofiler/pull/597), so why do we need to build 3.7?

scripts/fix_centos8.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*
+sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
+yum install -y dnf-plugins-core
+dnf config-manager --set-enabled powertools
+yum clean all
```
Jongy, line 10: Add a newline to all files

scripts/node_builder_glibc_env.sh
mpozniak95: Added a comment to file node_builder_glibc_env.sh with a summary of when we use what image and python

scripts/fix_centos8.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) Granulate. All rights reserved.
+# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+#
+
```
Jongy, line 6: Add `set -e` otherwise some commands here can fail.

The original code of 
```
RUN if grep -q "CentOS Linux 8" /etc/os-release ; then \
        sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*; \
        sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*; \
        yum install -y dnf-plugins-core; \
        dnf config-manager --set-enabled powertools; \
        yum clean all; \
```
was broken as well, because it didn't use `&&` or `set -e`, and that's buggy.

Always use `set -e` in your scrpts :)

pyi.Dockerfile
```diff
@@ -269,6 +255,18 @@ RUN if grep -q "CentOS Linux 8" /etc/os-release ; then \
     fi
 RUN python3 -m pip install --no-cache-dir -r exe-requirements.txt
 
+FROM ${NODE_PACKAGE_BUILDER_GLIBC} as node-package-builder-glibc
```
Jongy, line 258: This can happen earlier - it only requires the installation of Python, not the installation of our `requirements.txt` etc.

I suppose it can be placed installed after
```
# we want the latest pip
# hadolint ignore=DL3013
RUN python3 -m pip install --no-cache-dir --upgrade pip
```

Strive to place things as earlier as possible in Dockerfiles, to avoid unnecessary coupling. In the current state, if I modify e.g the `requirements.txt`, then this thing will be rebuilt unnecessarily

scripts/node_builder_glibc_env.sh
```diff
@@ -4,20 +4,29 @@
...
-yum update -y && yum install -y curl python3 make git ca-certificates npm
+# x86_64 container: Centos7 -> python3.7 built by code from node_builder_glibc_env.sh
+# x86_64 exe: Centos7 -> python3.7 built by code from node_builder_glibc_env.sh
+# aarch64 container: Centos8 -> python3.6.8 (default python from centos8 image) -> building with --static-libstdc++
+# aarch64 exe: Centos8 -> python3.10 based from build-prepare stage where it is built -> building with --static-libstc++
+
+yum update -y && yum install -y curl make git ca-certificates npm
 # yum has node v10 by default, so we need to add newer version to run node-gyp
 curl -fsSL https://rpm.nodesource.com/setup_16.x | bash -
+yum remove -y nodejs npm
```
Jongy, line 16: Why is it needed? Especially due tonext line re-installing `nodejs`?

scripts/node_builder_glibc_env.sh
Jongy: Good. Now that I have that in front of me - do we need to build 3.7.9 in ANY case?
* x86_64 container - is centos 7, can install python 3.6.8 like aarch64
* x86_64 exe - is centos 7, has the python 3.10 we built
* aarch64 container - is centos 8, use the python 3.6.8
* aarch64 exe - is centos 8, use the 3.10 we built

Jongy (changes requested):
Done revieiwng. Finally everything is clear, good documentation :) left some comments

scripts/node_builder_glibc_env.sh
mpozniak95: Yep, I deleted code that built 3.7.9. One note -> in x86_64 exe we use now centos7 with python 3.6.8 (python 10 we are building on centos8 so we would need to use static libstdc++)

scripts/node_builder_glibc_env.sh
mpozniak95: There was some mismatch between nodejs version and npm version, but I guess we don't have to install it separately as nodejs package includes npm. I deleted this line

mpozniak95:
I addressed the comments :)

Issue #600, opened 2022-11-14
Cleanup of gprofiler containers being run during test suite


guybortnikov:
During the run of the test suite, many gProfiler container instances are being spawned for testing purposes, these containers aren't being cleaned (after being stopped), leaving a big amount of garbage leftover on the machine.
For example, even when running only a very small amount of tests there a lot of gprofiler containers left:
```
CONTAINER ID   IMAGE          COMMAND                  CREATED       STATUS                     PORTS     NAMES
1b8f29fa799d   ed9ec930e37b   "python3 -m gprofile…"   2 hours ago   Exited (1) 2 hours ago               lucid_goodall
352f78029a3e   ed9ec930e37b   "python3 -m gprofile…"   2 hours ago   Exited (1) 2 hours ago               heuristic_burnell
c4165a26f7be   ed9ec930e37b   "python3 -m gprofile…"   2 hours ago   Exited (1) 2 hours ago               boring_meninsky
2e76c031d5ab   ed9ec930e37b   "python3 -m gprofile…"   2 hours ago   Exited (0) 2 hours ago               inspiring_williams
32d43d419adc   ed9ec930e37b   "python3 -m gprofile…"   2 hours ago   Exited (0) 2 hours ago               hungry_mendel
a056c47d001e   ed9ec930e37b   "python3 -m gprofile…"   2 hours ago   Exited (0) 2 hours ago               kind_solomon
50129a2f3006   ed9ec930e37b   "python3 -m gprofile…"   2 hours ago   Exited (0) 2 hours ago               suspicious_black
78cbf63a517a   ed9ec930e37b   "python3 -m gprofile…"   2 hours ago   Exited (143) 2 hours ago             competent_volhard
7c05de21e257   ed9ec930e37b   "python3 -m gprofile…"   5 hours ago   Exited (0) 5 hours ago               brave_lalande
c37242b123ae   ed9ec930e37b   "python3 -m gprofile…"   5 hours ago   Exited (0) 5 hours ago               silly_elbakyan
75176e5c4558   ed9ec930e37b   "python3 -m gprofile…"   5 hours ago   Exited (0) 5 hours ago               musing_lehmann
c708d65f77e7   ed9ec930e37b   "python3 -m gprofile…"   5 hours ago   Exited (0) 5 hours ago               unruffled_swanson
82e90b144551   ed9ec930e37b   "python3 -m gprofile…"   5 hours ago   Exited (0) 5 hours ago               beautiful_villani
[...]

PR #621, opened 2022-12-04
Drop support for Python 3.6, 3.7

Jongy:
After https://github.com/Granulate/granulate-utils/pull/107 is merged.

Since https://github.com/Granulate/gprofiler/pull/597 we no longer need any 3.6, 3.7 support. Time to move forward

PR #627, opened 2022-12-08
exited container cleanup

pfilipko1:
## Checklist:


Jongy:
Closes: https://github.com/Granulate/gprofiler/issues/600

Jongy:
Waiting on https://github.com/Granulate/gprofiler/pull/621 so CI is fixed.

tests/utils.py
```diff
@@ -242,7 +243,7 @@ def start_gprofiler_in_container_for_one_session(
     volumes = {
         str(output_directory): {"bind": inner_output_directory, "mode": "rw"},
     }
-    args = ["-v", "-d", "3", "-o", inner_output_directory] + runtime_specific_args + profiler_flags
+    args = ["--rm", "-v", "-d", "3", "-o", inner_output_directory] + runtime_specific_args + profiler_flags
```
Jongy, line 246: @pfilipko1 this argument is passed to gProfiler, not to the underlying Docker

tests/conftest.py
```diff
@@ -202,22 +222,27 @@ def application_process(
 
 
 @fixture(scope="session")
-def docker_client() -> DockerClient:
-    return docker.from_env()
+def docker_client(tests_id: str) -> Generator[Tuple[DockerClient, str], None, None]:
```
Jongy, line 225: Changing this fixture is a great clutter in the code, now all code needs to use `docker_client[0]` which is not convenient. This is too many changes in the tests. It makes no sense to change all sites + require all sites inn the future to use a tuple and extract the value.

Options I can think of:
* Extend the `DockerClient` class to one which implements the functionality (e.g accepts the label in its c'tor and adds the label for all created containers)
* Apply the change only in the `start_container` method, if it's the only method calling `container.run`. Need to find a way to pass the value to it without modifying ALL tests

pfilipko1: True, I was thinking about saving the test_id to a local file and reading it there :)

But I like the idea about extended class better.

tests/conftest.py
```diff
@@ -203,7 +205,13 @@ def application_process(
 
 @fixture(scope="session")
 def docker_client() -> DockerClient:
-    return docker.from_env()
+    tests_id = "".join(random.choice(string.ascii_uppercase) for _ in range(10))
```
Jongy, line 208: Now it looks good!

Now, 2 final changes I'd like - 
1. Before removing the containers, please note down if any exist. If so, raise an exception AFTER removing them.
2. Let's make the fixture `scope="function"`, this will cause the fixture to error tests which fail to remove their containers, and error them only (unlike `scope="session"` which will mark the session as failed, I bet).

Jongy, line 208: Can use `secrets.token_hex(5)` instead, shorter :)

tests/conftest.py
```diff
@@ -203,7 +205,13 @@ def application_process(
...
+    docker_client = docker.from_env()
+    setattr(docker_client, "test_id", tests_id)
+    yield docker_client
+    docker_client.containers.prune(filters={"label": tests_id})
+    if len(docker_client.containers.list(filters={"label": tests_id})) > 0:
+        raise Exception("Container has not been properly pruned")
```
Jongy, line 214: Always give the problematic data when raising the exception, otherwise one cannot debug the problem based on it.
So store the problematic containers in a list, and include `repr(..)` of that list in the `Exception`.

Jongy (changes requested):
.

tests/conftest.py
pfilipko1: wouldn't this fail always in that case - because the exited containers would exist before pruning them and that is the reason we prune them?

Jongy (changes requested):
Done reviewing :)

tests/utils.py
```diff
@@ -341,13 +342,18 @@ def _application_docker_container(
     application_docker_capabilities: List[str],
     application_docker_command: Optional[List[str]] = None,
 ) -> Container:
-    container: Container = docker_client.containers.run(
+    if application_docker_command is None:
+        application_docker_command = []
```
Jongy, line 346: Is it the same behavior for `docker_client.containers.run` - when you pass `command=None` and `command=[]`?

If unsure I suggest you refrain from changing the behavior (and e.g make `command` a `Optional[List[str]]` in `start_container`)

tests/utils.py
```diff
@@ -341,13 +342,18 @@ def _application_docker_container(
...
+    container: Container = start_container(
+        docker_client,
         application_docker_image,
-        detach=True,
```
Jongy, line 346: Why removed `detach=True`?

Jongy: Ah because `start_container` does it

tests/utils.py
```diff
@@ -341,13 +342,18 @@ def _application_docker_container(
...
+        application_docker_command,
+        volumes=None,
+        privileged=False,
+        pid_mode="host",
         user="5555:6666",
         mounts=application_docker_mounts,
         cap_add=application_docker_capabilities,
-        command=application_docker_command,
```
Jongy, line 350: Style-wise, I prefer we remain with kwargs instead of positional arguments.

tests/conftest.py
```diff
@@ -220,7 +221,21 @@ def application_process(
 
 @fixture(scope="session")
 def docker_client() -> DockerClient:
-    return docker.from_env()
+    tests_id = secrets.token_hex(5)
+    docker_client = docker.from_env()
+    setattr(docker_client, "test_id", tests_id)
+    yield docker_client
+
+    exited_ids = []  # type: List[str]
```
Jongy, line 229:
```suggestion
    exited_ids: Listr[str] = []
```
If explicit typing is needed, use Python 3 types.

tests/conftest.py
```diff
@@ -220,7 +221,21 @@ def application_process(
 
 @fixture(scope="session")
 def docker_client() -> DockerClient:
-    return docker.from_env()
+    tests_id = secrets.token_hex(5)
+    docker_client = docker.from_env()
+    setattr(docker_client, "test_id", tests_id)
```
Jongy, line 226:
```suggestion
    setattr(docker_client, "_gprofiler_test_id", tests_id)
```

As you're patching a global object - refrain from patching possibly used attributes. The `_gprofiler_` prefix ensures it's not used.

tests/test_executable.py
```diff
@@ -129,6 +129,7 @@ def test_executable_not_privileged(
     exit_code, output = application_docker_container.exec_run(cmd=command, privileged=False, user="root:root")
 
     _no_errors(output.decode())
+    print(output.decode())
```
Jongy, line 132: Can swap this line with the previous (so we get the output print, before calling `_no_errors`)

tests/utils.py
```diff
@@ -341,13 +342,19 @@ def _application_docker_container(
...
+    container: Container = start_container(
+        docker_client,
         application_docker_image,
-        detach=True,
-        user="5555:6666",
-        mounts=application_docker_mounts,
-        cap_add=application_docker_capabilities,
-        command=application_docker_command,
+        application_docker_command,
+        volumes=None,
```
Jongy, line 354: That's the default value, don't pass it?

tests/utils.py
```diff
@@ -341,13 +342,19 @@ def _application_docker_container(
...
+        privileged=True,
+        pid_mode="host",
```
Jongy, line 356: Why are you passing these 2? Previously we didn't run as privileged/pid_host when `containers.run` was used.

Jongy (changes requested):

tests/conftest.py
```diff
@@ -221,7 +222,21 @@ def application_process(
...
+    exited_ids: List[str] = []
+    exited_container_list = docker_client.containers.list(filters={"status": "exited", "label": tests_id})
+    for container in exited_container_list:
+        exited_ids.append(container.id)
+
+    pruned_ids = docker_client.containers.prune(filters={"label": tests_id}).get("ContainersDeleted", [])
+
+    for exited_id in exited_ids:
+        if exited_id not in pruned_ids:
+            raise Exception(f"Container with id {exited_id} has not been properly pruned")
```
Jongy, line 239: If we have multiple "non pruned" containers, we only get an exception for one. Lets create the full set of elements that appear in exited but not in pruned, and if that's a list longer than 0, raise an exception with them?

Jongy (changes requested):

Jongy:
Looks good. I will take a final look tmrw and merge.

Jongy (approved):

Issue #525, opened 2022-10-10
Use gLogger for remote logging


Jongy:
gLogger was introduced in https://github.com/Granulate/granulate-utils/pull/62. Before we make the full transition, let's support both the old infra and the new one, with a flag like `--glogger`. We can also enable it on 5% (e.g `if random.random() > 0.95: use glogger`) to verify stability.

PR #628, opened 2022-12-11
Use glogger as remote logging infra

Jongy:
## Description

Closes: https://github.com/Granulate/gprofiler/issues/525

I didn't support this feature in A/B mode as suggested in the ticket, it's much changes than I expected and it'll be hard to maintain both infras in the code.

Some changes I did as part of it:
* state `run_id` and `cycle_id` are no longer added in logging level, then omitted in local logger - I add them in glogger level.

TODOs:
* [x] Call the handler.close() method of glogger - is it needed to ensure all logs are sent back? Or does glogger have some kind of finalization?

## How Has This Been Tested?

* [x] `NO_SERVER_LOG_KEY` doesn't send the message to the server.
* [x] `NO_SERVER_EXTRA_KEY` doesn't send `extra`s to the server.
* [x] `NO_SERVER_LOG_KEY` and `NO_SERVER_EXTRA_KEY` are not sent to server / printed locally.
* [x] plain messages are logged locally
* [x] plain messages marked with `NO_SERVER_LOG_KEY` are logged locally
* [x] messages with extras are logged locally with the extras.
* [x] messages with extras marked with `NO_SERVER_EXTRA_KEY` are logged locally with the extras.
* [x] messages with traceback logged locally with traceback
* [x] messages with traceback logged remotely with traceback
* [x] `cycle_id` passes to the backend properly
* [x] changing `--glogger-server` works

### DIFF
--- a/gprofiler/log.py
+++ b/gprofiler/log.py
@@ -58,10 +58,12 @@ class RemoteLogsHandler(BatchRequestsHandler):
         )
 
     def emit(self, record: LogRecord) -> None:
-        if self.get_extra_fields(record).pop(NO_SERVER_LOG_KEY, False):
+        extra = self.get_extra_fields(record)
+
+        if extra.pop(NO_SERVER_LOG_KEY, False):
             return
 
-        if self.get_extra_fields(record).pop(NO_SERVER_EXTRA_KEY, False):
+        if extra.pop(NO_SERVER_EXTRA_KEY, False):
             record.extra = {}
 
         return super().emit(record)
--- a/gprofiler/log.py
+++ b/gprofiler/log.py
@@ -23,6 +23,7 @@ DEFAULT_API_SERVER_ADDRESS = "https://api.granulate.io"
 
 NO_SERVER_LOG_KEY = "no_server_log"
 NO_SERVER_EXTRA_KEY = "no_extra_to_server"
+CYCLE_ID_KEY = "cycle_id"
 LOGGER_NAME_RE = re.compile(r"gprofiler(?:\..+)?")
 
 
@@ -37,8 +38,8 @@ class GProfilerExtraAdapter(ExtraAdapter):
         extra = super().get_extra(**kwargs)
         # here we add fields which change during the lifetime of gProfiler.
         # fields that do not change go in RemoteLogsHandler.get_metadata().
-        assert "cycle_id" not in extra
-        return {**extra, "cycle_id": get_state().cycle_id}
+        assert CYCLE_ID_KEY not in extra
+        return {**extra, CYCLE_ID_KEY: get_state().cycle_id}
 
 
 class RemoteLogsHandler(BatchRequestsHandler):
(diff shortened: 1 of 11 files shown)


gprofiler/log.py
```diff
@@ -89,101 +41,49 @@ class RemoteLogsHandler(logging.Handler):
...
+
+        return super().emit(record)
+
+    def get_metadata(self) -> Dict[str, str]:
+        from gprofiler.metadata.system_metadata import get_hostname_or_none
+
+        state = get_state()
+        hostname = get_hostname_or_none()
+
+        metadata = {
```
Jongy, line 66: I kept the same set of metadatas as we had in the previous version. Will add more on demand.

gprofiler/log.py
```diff
@@ -89,101 +41,49 @@ class RemoteLogsHandler(logging.Handler):
...
+            "run_id": state.run_id,
+            "cycle_id": state.cycle_id,
+            "gprofiler_version": __version__,
+            "service_name": self._service_name,
         }
```
d3dave, line 71: are you sure this is supposed to be part of metadata and not extra? the metadata is shared across all sent logs at time of sending, so its possible for a batch to contain logs from separate runs/cycles.

Jongy: You're right. Good catch

Jongy: Using `get_extra`.

gprofiler/log.py
```diff
@@ -89,101 +50,50 @@ class RemoteLogsHandler(logging.Handler):
...
-        return {
-            "message": record.message,
-            "level": record.levelname,
-            "timestamp": formatted_timestamp,
-            "extra": extra,
-            "logger_name": record.name,
-            "exception": exception_traceback,
-            RUN_ID_KEY: run_id,
-            CYCLE_ID_KEY: cycle_id,
+        if self.get_extra_fields(record).pop(NO_SERVER_EXTRA_KEY, False):
```
d3dave, line 64: calling `get_extra_fields` twice

d3dave:
reviewed all. just that one comment.

d3dave (dismissed):

Jongy:
Waiting on https://github.com/Granulate/gprofiler/pull/621 so CI is fixed.

Jongy:
I need to fix the tests (some tests referenced the `extra`s via the old way)

d3dave (approved):

Jongy:
@d3dave - about this note:

>  Call the handler.close() method of glogger - is it needed to ensure all logs are sent back? Or does glogger have some kind of finalization?

How does glogger handle it? It seems that messages are sent by some mechanism, before the process goes down.

d3dave:
> How does glogger handle it? It seems that messages are sent by some mechanism, before the process goes down.

`logging` registers an `atexit` handler that closes all stray handlers. You only need to do it if you need to flush before process exit.

Jongy:
Made one final change - ready to merge now

d3dave (approved):

PR #634, opened 2022-12-16
python: Log PyPerf stdout+stderr after a restart

Jongy:
If PyPerf fails, we restart it, but we log only the exit code and not the output, so we cannot know WHY, if the reason is present in its outputs.. For example, here:
```
[12:56:24] PyPerf dead/not responding, killing it
[12:56:24] Python eBPF profiler failed, restarting PyPerf... (exit_code=-6)
[12:56:24] Starting profiling of Python processes with PyPerf
```

It probably it an `assert` but I cannot know. With this PR, we'll know :)


michelhe (approved):

gprofiler/profilers/python_ebpf.py
```diff
@@ -256,9 +256,14 @@ def _terminate(self) -> Optional[int]:
             self.process.terminate()  # okay to call even if process is already dead
             code = self.process.wait()
             self.process = None
+
```
michelhe, line 259: Can comment on the line but change above to use `self.is_running` instead of `        if self.process is not None:
`

Jongy: Bleh, mypy then whines, it doesn't follow the validty of the assert through functions :/

michelhe (approved):

PR #638, opened 2022-12-18
utils: Nicer process start/exit logs

Jongy:


michelhe (dismissed):
Failing tests

Jongy:
Flakiness, I think. Merged master

gprofiler/utils/__init__.py
```diff
@@ -118,11 +118,11 @@ def wrapper() -> Any:
 def start_process(
     cmd: Union[str, List[str]], via_staticx: bool, term_on_parent_death: bool = True, **kwargs: Any
 ) -> Popen:
-    cmd_text = " ".join(cmd) if isinstance(cmd, list) else cmd
-    logger.debug(f"Running command: ({cmd_text})")
     if isinstance(cmd, str):
         cmd = [cmd]
 
+    logger.debug("Running command", command=cmd)
```
d3dave, line 124: Are lists in extras valid? Shouldn't it be a string value?

Jongy: I was about to say that I'm certain it does support, but realized that I mix it with the metadata, and it seems that lists are not supported. I'll fix. thx

Jongy: I think repr makes sense

Jongy: Actually - it is supported, I looked at the wrong logs view. Supported fine :)

Jongy:
It seems that tests are truly failing

Jongy:
Now should pass

d3dave (approved):

PR #644, opened 2022-12-22
Update granulate-utils

Jongy:
Includes the change from https://github.com/Granulate/granulate-utils/pull/113 about Zing version parsing.


Jongy:
Merged master, but there's no actual change so all tests will pass again.

YoniKF (approved):

PR #655, opened 2023-01-08
java: async-profiler: Default async-profiler safemode of PROBE_SP

Jongy:
PROBE_SP was proven useful to get more accurate traces. This also removes reference from JAVA_STATE which is now named diffrently and has a different value, and we no longer want.

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -129,7 +129,8 @@ JAVA_SAFEMODE_DEFAULT_OPTIONS = [
     JavaSafemodeOptions.HSERR.value,
 ]
 
-JAVA_ASYNC_PROFILER_DEFAULT_SAFEMODE = 64  # StackRecovery.JAVA_STATE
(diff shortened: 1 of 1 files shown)


gprofiler/profilers/java.py
```diff
@@ -608,8 +609,8 @@ def read_output(self) -> Optional[str]:
             "--java-async-profiler-safemode",
             dest="java_async_profiler_safemode",
             default=JAVA_ASYNC_PROFILER_DEFAULT_SAFEMODE,
-            type=integer_range(0, 128),
-            metavar="[0-127]",
+            type=integer_range(0, 0x200),
+            metavar="[0-512]",
```
YoniKF, line 613:
```suggestion
            metavar="[0-511]",
```

YoniKF (changes requested):

gprofiler/profilers/java.py
Jongy: That's true

YoniKF (approved):

PR #656, opened 2023-01-13
Improve app metadata robustness

Jongy:
* Don't fail if getting exe/execfn fails
* Handle ZombieProcess errors gracefully

The trigger for this was some system where we're able to profile Java processes, but unable to read execfns (unable to access processes' `/proc/pid/mem`).
```
[2023-01-11 16:40:12,511] ERROR: gprofiler.metadata.application_metadata: Exception while collecting metadata in JavaMetadata!
  File "granulate_utils/linux/process.py", line 106, in _translate_errors
  File "granulate_utils/linux/process.py", line 98, in _read_process_memory
PermissionError: [Errno 13] Permission denied: '/proc/620/mem'

During handling of the above exception, another exception occurred:

  File "gprofiler/metadata/application_metadata.py", line 52, in get_metadata
  File "gprofiler/profilers/java.py", line 263, in make_application_metadata
  File "gprofiler/metadata/application_metadata.py", line 68, in make_application_metadata
  File "granulate_utils/linux/process.py", line 78, in read_process_execfn
  File "granulate_utils/linux/process.py", line 100, in _read_process_memory
  File "contextlib.py", line 99, in __exit__
  File "granulate_utils/linux/process.py", line 111, in _translate_errors
psutil.AccessDenied: psutil.AccessDenied (pid=620)
```

in such cases, I prefer that we get the rest of the metadata, and a clear error for `execfn`.

### DIFF
--- a/gprofiler/metadata/application_metadata.py
+++ b/gprofiler/metadata/application_metadata.py
@@ -7,8 +7,8 @@ import functools
 from threading import Event, Lock
 from typing import Any, Dict, Optional
 
-from granulate_utils.linux.process import is_process_running, read_process_execfn
-from psutil import NoSuchProcess, Process
+from granulate_utils.linux.process import is_process_running, process_exe, read_process_execfn
+from psutil import NoSuchProcess, Process, ZombieProcess
 
 from gprofiler.log import get_logger_adapter
 from gprofiler.metadata.versions import get_exe_version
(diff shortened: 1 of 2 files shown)


gprofiler/metadata/application_metadata.py
```diff
@@ -50,19 +50,41 @@ def get_metadata(self, process: Process) -> Optional[Dict]:
...
+            exe = f"error: {e.__class__.__name__}"
+        md["exe"] = exe
+
+        try:
+            execfn = "error: not supported on Windows" if is_windows() else read_process_execfn(process)
+        except (NoSuchProcess, ZombieProcess):
+            raise  # let caller handle
+        except Exception as e:
+            logger.exception("Exception while reading process execfn", pid=process.pid)
+            execfn = f"error: {e.__class__.__name__}"
```
netaneld122, line 87: Do we usually leave errors where the values should be instead of setting to None? It feels like a bad practice that may lead to other code parsing the error as if it's a real exe/execfn

Jongy: I want the error to be clear, as the type of error itself might be part of the logic. Since the error value is not a legitimate value and can be clearly distinguished, I think it's okay.
Alternatively, I can use `None` and add another `exe_error` and `execfn_error` fields. Makes more sense?

netaneld122: If we have full control over the parsing side then it's ok I guess, I mainly wanted to bring some attention to this 👍

netaneld122 (approved):
LGTM

PR #660, opened 2023-01-23
Add gProfiler flags retrieval

uriariel:
Add jvm flags retrieval for gProfiler

## Description
Allow gProfiler to get jvm flags for the profiled java process.

## Checklist:


gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+    process: Process,
+) -> List[Tuple[Optional[str], Optional[str], Optional[str], str, Optional[List[str]]]]:
+    try:
+        output = run_process([jattach_path(), str(process.pid), "jcmd", "VM.flags -all"]).stdout.decode()
+    except CalledProcessError as e:
+        logging.warning(f"Couldn't get jvm flags for process {process.pid}: {e.stderr}")
+        return []
+
+    vm_flags = []
+    vm_flags_pattern = re.compile(r"(\S+)\s+(\S+)\s+(:)?= (\S*)\s+{(.+?)}(?:\s*{(.*)})*")
```
Jongy, line 325: Document some example lines of what you're trying to catch here. Always do so when embedding regexes in code.

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+
+    for line in output.splitlines():
+        match = vm_flags_pattern.search(line)
+        if match:
+            flag_type = match.group(1)
+            flag_name = match.group(2)
+            flag_is_non_default_source_only_jdk_8 = match.group(3) == ":"
```
Jongy, line 332: Document this part (jdk 8 vs 11 behavior)

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+            flag_value = match.group(4)
+            flag_kind = match.group(5).split()
```
Jongy, line 334: Why split? document

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
 
 
+@functools.lru_cache(maxsize=1024)
+def get_jvm_flags(
+    process: Process,
+) -> List[Tuple[Optional[str], Optional[str], Optional[str], str, Optional[List[str]]]]:
+    return [
+        (flag_name, flag_type, flag_value, flag_source, flag_kind)
+        for flag_name, flag_type, flag_value, flag_source, flag_kind in get_jvm_flags_raw(process)
+        if filter_jvm_flags(flag_name, flag_type, flag_value, flag_source, flag_kind) is True
```
Jongy, line 290: Let's create a namedtuple for them instead of passing around bunches of variables

Jongy, line 290:
```suggestion
        (flag_name, flag_type, flag_value, flag_source, flag_kind)
        for flag_name, flag_type, flag_value, flag_source, flag_kind in get_jvm_flags_raw(process)
        if filter_jvm_flags(flag_name, flag_type, flag_value, flag_source, flag_kind)
```

No `is True`

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+    _flag_name: Optional[str],
+    flag_type: Optional[str],
+    _flag_value: Optional[str],
+    flag_source: str,
+    flag_kind: Optional[List[str]],
+) -> bool:
+    if flag_source == "default":
+        return False
+
+    if flag_type and flag_type not in ["bool", "int", "uint", "intx", "uintx", "uint64_t", "size_t", "double"]:
```
Jongy, line 304: Why would `flag_type` be None?  It doesn't appear to be the case in your parsing

gprofiler/profilers/java.py
```diff
@@ -258,7 +259,7 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
         # that loads other libs.
         libjvm_elfid = get_mapped_dso_elf_id(process, "/libjvm")
 
-        metadata = {"java_version": version, "libjvm_elfid": libjvm_elfid}
+        metadata = {"java_version": version, "libjvm_elfid": libjvm_elfid, "jvm_flags": get_jvm_flags(process)}
```
Jongy, line 262: I'm not sure this can handle nested dicts. Please check with the backend team if that's supported. If not, let's make it a JSON

tests/test_app_metadata.py
```diff
@@ -55,6 +55,18 @@
                 "OpenJDK Runtime Environment (build 1.8.0_275-b01)\n"
                 "OpenJDK 64-Bit Server VM (build 25.275-b01, mixed mode)",
                 "libjvm_elfid": "buildid:0542486ff00153ca0bcf9f2daea9a36c428d6cde",
+                "jvm_flags": [
```
YoniKF, line 58: Let's add a test for JDK 9+, afaict the output of `jcmd VM.flags -all` is a bit different there.
In addition, consider adding unit tests for the parsing alone.

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+        return False
+
+    return True
+
+
+def get_jvm_flags_raw(
+    process: Process,
+) -> List[Tuple[Optional[str], Optional[str], Optional[str], str, Optional[List[str]]]]:
+    try:
+        output = run_process([jattach_path(), str(process.pid), "jcmd", "VM.flags -all"]).stdout.decode()
```
YoniKF, line 319: I see that when loading async-profiler we run jattach with a timeout. We probably want the same defense here.

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+) -> bool:
+    if flag_source == "default":
+        return False
+
+    if flag_type and flag_type not in ["bool", "int", "uint", "intx", "uintx", "uint64_t", "size_t", "double"]:
+        return False
+
+    if flag_kind and any(
+        (not_supported_kind in flag_kind) for not_supported_kind in ["notproduct", "pd", "experimental"]
+    ):
```
YoniKF, line 309: Use set intersecation.

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+    flag_kind: Optional[List[str]],
+) -> bool:
+    if flag_source == "default":
+        return False
+
+    if flag_type and flag_type not in ["bool", "int", "uint", "intx", "uintx", "uint64_t", "size_t", "double"]:
+        return False
+
+    if flag_kind and any(
+        (not_supported_kind in flag_kind) for not_supported_kind in ["notproduct", "pd", "experimental"]
```
YoniKF, line 308: We should probably filter out `develop` flags (or does `jcmd VM.flags -all` filter them out itself?).

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+
+    for line in output.splitlines():
+        match = vm_flags_pattern.search(line)
+        if match:
+            flag_type = match.group(1)
+            flag_name = match.group(2)
+            flag_is_non_default_source_only_jdk_8 = match.group(3) == ":"
+            flag_value = match.group(4)
+            flag_kind = match.group(5).split()
+            flag_source_jdk_9 = match.group(6)
```
YoniKF, line 335: Looks like [the terminology in OpenJDK is "flag origin"](https://github.com/openjdk/jdk/blob/544c16e0bdd4335b2624158fd1f6521984aa5079/src/hotspot/share/runtime/flags/jvmFlag.hpp#L36), let's rename accordingly.

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
     return Path(resource_path("java/async-profiler-version")).read_text()
 
 
+@functools.lru_cache(maxsize=1024)
+def get_jvm_flags(
+    process: Process,
+) -> List[Tuple[Optional[str], Optional[str], Optional[str], str, Optional[List[str]]]]:
```
YoniKF, line 286: I'm not sure about the types here: Why are the flag name, type, and value optional? Even the kind should always have some value afaict from the code.

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+            flag_source_jdk_9 = match.group(6)
+
+            if flag_is_non_default_source_only_jdk_8:
+                flag_source = "non-default"
+            elif flag_source_jdk_9:
+                flag_source = flag_source_jdk_9
+            else:
+                flag_source = "default"
+
+            vm_flags += [(flag_name, flag_type, flag_value, flag_source, flag_kind)]
```
YoniKF, line 344: Use `list.append`.

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+
+
+def get_jvm_flags_raw(
+    process: Process,
+) -> List[Tuple[Optional[str], Optional[str], Optional[str], str, Optional[List[str]]]]:
+    try:
+        output = run_process([jattach_path(), str(process.pid), "jcmd", "VM.flags -all"]).stdout.decode()
+    except CalledProcessError as e:
+        logging.warning(f"Couldn't get jvm flags for process {process.pid}: {e.stderr}")
+        return []
```
YoniKF, line 322: I think that we should return `None` to distinguish between the cases of failure to retrieve flags and all flags filtered out.

YoniKF (changes requested):

gprofiler/profilers/java.py
uriariel: All of them are regex match groups they could be None, Ill change it to also handle this comment https://github.com/Granulate/gprofiler/pull/660#discussion_r1084952682 and Ill add to the list only flags that were parsed correctly

gprofiler/profilers/java.py
```diff
@@ -279,6 +280,72 @@ def get_ap_version() -> str:
...
+
```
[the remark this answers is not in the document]
uriariel: since its a regex nothing happens, all the match groups will just be None, Ill change it like I replied on https://github.com/Granulate/gprofiler/pull/660#discussion_r1084946779

gprofiler/profilers/java.py
uriariel: `flag_type` was a regex match so it could be None, but Im changing it now as I wrote in https://github.com/Granulate/gprofiler/pull/660#discussion_r1085408196

gprofiler/profilers/java.py
uriariel: double checked with the backend team, its OK

tests/conftest.py
```diff
@@ -235,6 +235,12 @@ def application_docker_image_configs() -> Mapping[str, Dict[str, Any]]:
         },
         "java": {
             "": {},
+            "hotspot-jdk-8": {},  # add for clarity when testing with multiple JDKs
+            "hotspot-jdk-11": dict(
+                buildargs={
+                    "JAVA_BASE_IMAGE": "openjdk@sha256:e81b7f317654b0f26d3993e014b04bcb29250339b11b9de41e130feecd4cd43c"
```
YoniKF, line 241:
```suggestion
                    "JAVA_BASE_IMAGE": "openjdk/11-jdk"
```
or similar does not work?

gprofiler/profilers/java.py
```diff
@@ -259,11 +284,107 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+    def get_jvm_flags_raw(self, process: Process) -> Optional[List[JvmFlag]]:
+        try:
+            output = self.jattach_jcmd_runner.run(process, "VM.flags -all")
+        except CalledProcessError as e:
+            logging.warning(f"Couldn't get jvm flags for process {process.pid}: {e.stderr}")
+            return None
+
+        vm_flags = []
+        # The output of VM.flags -all format on jdk 8:
```
YoniKF, line 330: A docstring will be clearer here.

gprofiler/profilers/java.py
```diff
@@ -259,11 +284,107 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+                flag_type = match.group("flag_type")
+
+                if flag_is_non_default_origin_only_jdk_8:
+                    flag_origin = "non-default"
+                elif flag_origin_jdk_9:
+                    flag_origin = flag_origin_jdk_9
+                else:
+                    flag_origin = "default"
+
+                if flag_type and flag_name and flag_value and flag_kind:
```
YoniKF, line 375: Isn't there a way to make the `search` method return a match object only when all non-optional groups were matched? I hoped this is the default behavior.

gprofiler/profilers/java.py
```diff
@@ -197,6 +198,26 @@ def __str__(self) -> str:
...
+        self.jattach_timeout = jattach_timeout
+
+    def run(self, process: Process, cmd: str) -> str:
+        return run_process(
+            [jattach_path(), str(process.pid), "jcmd", cmd], stop_event=self.stop_event, timeout=self.jattach_timeout
+        ).stdout.decode()
+
+
+@dataclasses.dataclass
+class JvmFlag:
```
YoniKF, line 213: Any reason for the `flag_` prefix on all attributes? If not, let's remove it.

gprofiler/profilers/java.py
```diff
@@ -259,11 +284,107 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+            "uint",
+            "intx",
+            "uintx",
+            "uint64_t",
+            "size_t",
+            "double",
+        ]:
+            return False
+
+        if set(flag.flag_kind).intersection({"notproduct", "pd", "experimental", "develop"}):
```
YoniKF, line 317: Don't filter out `experimental` flags.

gprofiler/profilers/java.py
```diff
@@ -259,11 +284,107 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+        # "lp64_product", "rw", "pd", "JVMCI", "C1", "C2", "ARCH"
+
+        # possible flag types: "bool", "int", "uint", "intx", "uintx", "uint64_t", "size_t", "double", "ccstr",
+        # "ccstrlist"
+
+        # possible flag origins: default, non-default, command line, environment, config file, management, ergonomic,
+        # attach, internal, jimage
+
+        vm_flags_pattern = re.compile(
+            r"(?P<flag_type>\S+)\s+(?P<flag_name>\S+)\s+(?P<flag_is_non_default_origin_only_jdk_8>:)?= (?P<flag_value>\S*)\s+{(?P<flag_kind>.+?)}(?:\s*{(?P<flag_origin_jdk_9>.*)})*"  # noqa: E501
```
YoniKF, line 351:
```suggestion
            r"(?P<flag_type>\S+)\s+(?P<flag_name>\S+)\s+(?P<flag_is_non_default_origin_only_jdk_8>:)?= (?P<flag_value>\S*)\s+{(?P<flag_kind>.+?)}(?:\s*{(?P<flag_origin_jdk_9>.*)})?"  # noqa: E501
```
Doesn't the origin appear at most once?

tests/test_app_metadata.py
YoniKF: Are we adding unit tests for the parsing?

YoniKF (changes requested):

gprofiler/profilers/java.py
```diff
@@ -259,11 +284,107 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
         metadata.update(super().make_application_metadata(process))
         return metadata
 
+    @functools.lru_cache(maxsize=1024)
+    def get_jvm_flags(self, process: Process) -> Optional[List[tuple]]:
+        jvm_raw_flags = self.get_jvm_flags_raw(process)
+        if jvm_raw_flags is None:
+            return None
+
+        return [dataclasses.astuple(flag) for flag in jvm_raw_flags if self.filter_jvm_flags(flag)]
```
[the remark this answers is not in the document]
uriariel: since we send it as a json it might cause a problem with the DB, as json dicts don't guarantee order whereas json list do.
https://stackoverflow.com/questions/7214293/is-the-order-of-elements-in-a-json-list-preserved

gprofiler/profilers/java.py
uriariel: none that I know of

gprofiler/profilers/java.py
```diff
@@ -259,11 +285,109 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+
+    @staticmethod
+    def filter_jvm_flags(flag: JvmFlag) -> bool:
```
Jongy, line 302:
```suggestion
    def filter_jvm_flag(flag: JvmFlag) -> bool:
```

It filters a single flag.

gprofiler/profilers/java.py
```diff
@@ -259,11 +285,109 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+            "uint",
+            "intx",
+            "uintx",
+            "uint64_t",
+            "size_t",
+            "double",
+        ]:
+            return False
+
+        if set(flag.kind).intersection({"notproduct", "pd", "develop"}):
```
Jongy, line 318:
```suggestion
        if flag.kind not in ("notproduct", "pd", "develop"):
```
? isn't it a single string?

Jongy: Or kind is a list of strings, let's name it `kinds` not `kind`?

gprofiler/profilers/java.py
```diff
@@ -259,11 +285,109 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+        "ccstrlist"
+
+        possible flag origins: default, non-default, command line, environment, config file, management, ergonomic,
+        attach, internal, jimage
+        """
+
+        try:
+            output = self.jattach_jcmd_runner.run(process, "VM.flags -all")
+        except CalledProcessError as e:
+            logging.warning(f"Couldn't get jvm flags for process {process.pid}: {e.stderr}")
```
Jongy, line 349:
```suggestion
            logging.warning("Couldn't get jvm flags for process", pid=process.pid, stderr=e.stderr)
```
Prefer extras to formatting into messages

tests/conftest.py
Jongy: The hash is when we want a precise build (i.e matching buildid, etc). IDK if needed in this case

gprofiler/profilers/java.py
```diff
@@ -259,11 +285,109 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+                flag_type = match.group("flag_type")
+
+                if flag_is_non_default_origin_only_jdk_8:
+                    flag_origin = "non-default"
+                elif flag_origin_jdk_9:
+                    flag_origin = flag_origin_jdk_9
+                else:
+                    flag_origin = "default"
+
+                if flag_type and flag_name and flag_value and flag_kind:
```
Jongy, line 378: Pls `x is not None` instead of `x`

gprofiler/profilers/java.py
YoniKF: Kind can be both `product` and `C2`, for example.

gprofiler/profilers/java.py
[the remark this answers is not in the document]
YoniKF: We just need to make sure to dump to JSON with `sort_keys=True`. Can you check with the backend team that they load the JSON sorted?

Another thing: Maybe we should return a dict of (flag name) -> (flag attributes) instead of a list of (flag name and attributes).

gprofiler/profilers/java.py
YoniKF: Are you sure about it?
``` python
In [14]: re.search(r"(?P<flag_type>\S+)\s+(?P<flag_name>\S+)\s+(?P<flag_is_non_default_origin_only_jdk_8>:)?= (?P<flag_value>\S*)\s+{(?P<flag_kind>.+?)}(?:\s*{(?P<flag_origin_jdk_9>.*)})?", "t n = v {k} {o}") is None
Out[14]: False

In [15]: re.search(r"(?P<flag_type>\S+)\s+(?P<flag_name>\S+)\s+(?P<flag_is_non_default_origin_only_jdk_8>:)?= (?P<flag_value>\S*)\s+{(?P<flag_kind>.+?)}(?:\s*{(?P<flag_origin_jdk_9>.*)})?", "t n = v {k}") is None
Out[15]: False

In [16]: re.search(r"(?P<flag_type>\S+)\s+(?P<flag_name>\S+)\s+(?P<flag_is_non_default_origin_only_jdk_8>:)?= (?P<flag_value>\S*)\s+{(?P<flag_kind>.+?)}(?:\s*{(?P<flag_origin_jdk_9>.*)})?", "t n =  {k}") is None
Out[16]: False

In [17]: re.search(r"(?P<flag_type>\S+)\s+(?P<flag_name>\S+)\s+(?P<flag_is_non_default_origin_only_jdk_8>:)?= (?P<flag_value>\S*)\s+{(?P<flag_kind>.+?)}(?:\s*{(?P<flag_origin_jdk_9>.*)})?", "t n =  ") is None
Out[17]: True
```
The issue is only with `t n =  {k}` where the value is matched to `""`, but fixing `(?P<flag_value>\S*)` to `(?P<flag_value>\S+)` solves that:
``` python
In [22]: re.search(r"(?P<flag_type>\S+)\s+(?P<flag_name>\S+)\s+(?P<flag_is_non_default_origin_only_jdk_8>:)?= (?P<flag_value>\S+)\s+{(?P<flag_kind>.+?)}(?:\s*{(?P<flag_origin_jdk_9>.*)})?", "t n = v {k} {o}") is None
Out[22]: False

In [23]: re.search(r"(?P<flag_type>\S+)\s+(?P<flag_name>\S+)\s+(?P<flag_is_non_default_origin_only_jdk_8>:)?= (?P<flag_value>\S+)\s+{(?P<flag_kind>.+?)}(?:\s*{(?P<flag_origin_jdk_9>.*)})?", "t n = v {k}") is None
Out[23]: False

In [24]: re.search(r"(?P<flag_type>\S+)\s+(?P<flag_name>\S+)\s+(?P<flag_is_non_default_origin_only_jdk_8>:)?= (?P<flag_value>\S+)\s+{(?P<flag_kind>.+?)}(?:\s*{(?P<flag_origin_jdk_9>.*)})?", "t n = {k}") is None
Out[24]: True

In [25]: re.search(r"(?P<flag_type>\S+)\s+(?P<flag_name>\S+)\s+(?P<flag_is_non_default_origin_only_jdk_8>:)?= (?P<flag_value>\S+)\s+{(?P<flag_kind>.+?)}(?:\s*{(?P<flag_origin_jdk_9>.*)})?", "t n = v") is None
Out[25]: True
```

YoniKF (changes requested):
Open comments.

gprofiler/profilers/java.py
uriariel: opposite of @YoniKF's comment, I consider storing flag.kind as `set`

gprofiler/profilers/java.py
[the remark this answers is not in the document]
uriariel: I think its OK both ways, discussed it briefly with @Jongy f2f, but lets rethink this, what format do we prefer?

gprofiler/profilers/java.py
```diff
@@ -197,6 +204,50 @@ def __str__(self) -> str:
...
+            flag_origin = flag_origin_jdk_9
+        else:
+            flag_origin = "default"
+
+        return cls(
+            name=match.group("flag_name"),
+            type=match.group("flag_type"),
+            value=match.group("flag_value"),
+            origin=flag_origin,
+            kind=flag_kind,
```
YoniKF, line 247:
```suggestion
            kind=sorted(flag_kind),
```
Prevent accidental cardinality increase.

YoniKF (changes requested):

tests/test_app_metadata.py
uriariel: no we are not adding unitests

gprofiler/profilers/java.py
uriariel: "" is indeed a legal value

YoniKF: Discussion continues in https://github.com/Granulate/granulate-utils/pull/122.

gprofiler/metadata/application_identifiers.py
```diff
@@ -306,25 +307,33 @@ def get_app_id(self, process: Process) -> Optional[str]:
...
-    _IDENTIFIERS_MAP["java"] = [_JavaJarApplicationIdentifier()]
-    _IDENTIFIERS_MAP["java_spark"] = _IDENTIFIERS_MAP["java"] + [_JavaSparkApplicationIdentifier()]
+    @classmethod
+    def init_java(cls, jattach_jcmd_runner: JattachJcmdRunner) -> None:
+        if is_linux():
+            cls.identifiers_map["java"] = [_JavaJarApplicationIdentifier(jattach_jcmd_runner)]
+            cls.identifiers_map["java_spark"] = cls.identifiers_map["java"] + [_JavaSparkApplicationIdentifier()]
 
 
 def set_enrichment_options(enrichment_options: EnrichmentOptions) -> None:
```
Jongy, line 339: Is it still used at all?

gprofiler/profilers/java.py
YoniKF: @uriariel Please note that you used `%s` formatting instead of extra args like suggested here.

YoniKF (commented):
A few comments, after https://github.com/Granulate/granulate-utils/pull/122 is merged and the current PR is updated I'll review again from b3f355ae13baa384c9e8f7b4ad738c17a023de9d.

gprofiler/profilers/java.py
uriariel: extra args should be defined in the logger format, don't they?

YoniKF: @Jongy Can you clarify? AFAICT any extra argument other than some built-ins which have special meaning (like `exc_info`) are fine.

gprofiler/profilers/java.py
```diff
@@ -135,6 +138,11 @@ class JavaSafemodeOptions(str, Enum):
 SUPPORTED_AP_MODES = ["cpu", "itimer", "alloc"]
 
 
+class JavaFlagRetrievalOptions(str, Enum):
+    ALL = "all"
+    DEFAULT = "default"
```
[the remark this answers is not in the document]
uriariel: https://github.com/Granulate/gprofiler/blob/552af301d49a7d2f899fb6268bdd76ac68c9d047/gprofiler/profilers/java.py#L760

tests/test_app_metadata.py
```diff
@@ -124,5 +197,11 @@ def test_app_metadata(
     # stack begins with index
     idx = int(stack.split(";")[0])
 
+    if runtime == "java":
+        # dont check the jvm flags, they have different values on different machines
```
Jongy, line 201: You *do* check them after all, so what you mean is "don't check the JVM flags in direct comparison, as they might change a bit across machines due to ergonomics"

tests/test_java.py
```diff
@@ -718,3 +720,284 @@ def test_java_frames_include_no_semicolons(
...
+def test_get_default_jvm_flags(
+    tmp_path: Path,
+    application_pid: int,
+    assert_collapsed: AssertInCollapsed,
+    expected_flags: List[Dict[str, Union[str, List[str]]]],
+) -> None:
+    with make_java_profiler(
+        frequency=99,
+        storage_dir=str(tmp_path),
+        java_async_profiler_mode="cpu",
```
Jongy, line 888: No reason to override parameters to `make_java_profiler`, you don't profile with it anyway

tests/test_java.py
```diff
@@ -718,3 +720,284 @@ def test_java_frames_include_no_semicolons(
...
+                },
+                {"name": "UseG1GC", "type": "bool", "value": "true", "origin": "ergonomic", "kind": ["product"]},
+            ],
+        ),
+    ],
+)
+def test_get_default_jvm_flags(
+    tmp_path: Path,
+    application_pid: int,
+    assert_collapsed: AssertInCollapsed,
```
Jongy, line 882: Unused? remove

tests/conftest.py
```diff
@@ -235,6 +238,8 @@ def application_docker_image_configs() -> Mapping[str, Dict[str, Any]]:
         },
         "java": {
             "": {},
+            "hotspot-jdk-8": {},  # add for clarity when testing with multiple JDKs
```
Jongy, line 241: To make sure it doesn't break if the default image is changed, you can pass here `"JAVA_BASE_IMAGE": "openjdk:8u275"`

tests/test_java.py
```diff
@@ -718,3 +720,284 @@ def test_java_frames_include_no_semicolons(
...
+    [
+        (
+            "-XX:SelfDestructTimer=5",
+            "-XX:+PrintCodeCache",
+        ),
+    ],
+)
+@pytest.mark.parametrize("in_container", [True])
+@pytest.mark.parametrize(
+    "application_image_tag,expected_flags",
```
Jongy, line 906: If the `expected_flags` is constant across different images, don't parametrize it alongisde the image. Parametrize it once OR keep it a const in the test.

Jongy: I see there are slight differences between JDK 8 and JDK 11, so up to you if u want to keep it here or reduce the boilerplate somehow

tests/test_java.py
```diff
@@ -718,3 +720,284 @@ def test_java_frames_include_no_semicolons(
...
+    application_docker_image: Image,
+    assert_collapsed: AssertInCollapsed,
+    java_cli_flags: str,
+    java_env_flags: str,
+    expected_flags: List[Dict[str, Union[str, List[str]]]],
+    application_pid: int,
+    application_image_tag: str,
+) -> None:
+    """
+    Tests that we can profile a Java app that runs with non-java "comm", by reading the argv0 instead.
```
Jongy, line 979: Fix docstring to reflect what this test actually tests - 
1. Tests retrieving specific flags only
2. Tests retrival from env & commandline sources and reporting the correct source

tests/test_java.py
```diff
@@ -718,3 +720,284 @@ def test_java_frames_include_no_semicolons(
...
+    """
+    with make_java_profiler(
+        frequency=99,
```
Jongy, line 982: Don't override fields u don't need to

tests/test_java.py
```diff
@@ -718,3 +720,284 @@ def test_java_frames_include_no_semicolons(
...
+        # container and not from the manully started one
+        with _application_docker_container(
+            docker_client,
+            application_docker_image,
+            [],
+            [],
+            application_docker_command=[
+                "bash",
+                "-c",
+                f"export JAVA_TOOL_OPTIONS={java_env_flags}" ";" f"java {java_cli_flags} -jar Fibonacci.jar",
```
Jongy, line 997: Fix the string concatenation here to be a single string

Jongy, line 997:
```suggestion
                f"export JAVA_TOOL_OPTIONS={java_env_flags}" ";" f"exec java {java_cli_flags} -jar Fibonacci.jar",
```

if you do that you won't need the `find_application_pid`. You can keep the exporting fo it to a function, though.

tests/utils.py
```diff
@@ -351,3 +355,28 @@ def _application_docker_container(
...
+        process.cmdline()[0].endswith("sh")
+        and process.cmdline()[1] == "-c"
+        and len(process.children(recursive=False)) == 1
+    ):
+        pid = process.children(recursive=False)[0].pid
+
+    return pid
+
+
+def assert_jvm_flags_equal(actual_jvm_flags: Optional[List], expected_jvm_flags: List) -> None:
```
Jongy, line 373: Why accepting `Optional` then asserting?

Jongy, line 373: What's the idea? Why is it not a strict `==` comparison? because you don't want to compare the `value` due to ergonomics? Document that next to `pop`ing the items.

tests/utils.py
```diff
@@ -351,3 +355,28 @@ def _application_docker_container(
...
+    assert actual_jvm_flags is not None, "actual_jvm_flags is None"
+
+    assert len(actual_jvm_flags) == len(expected_jvm_flags), f"{actual_jvm_flags} != {expected_jvm_flags}"
+
+    for actual_flag_dict, expected_flag_dict in zip(actual_jvm_flags, expected_jvm_flags):
+        actual_flag_dict.pop("value")
+        expected_flag_dict.pop("value")
+
+        assert actual_flag_dict == expected_flag_dict, f"{actual_flag_dict} != {expected_flag_dict}"
```
Jongy, line 382:
```suggestion
        assert actual_flag_dict == expected_flag_dict
```

didn't we agree that pytest does the pretty print for you? :sweat_smile:

tests/test_java.py
```diff
@@ -718,3 +720,284 @@ def test_java_frames_include_no_semicolons(
...
+                    "name": "UseCompressedClassPointers",
+                    "type": "bool",
+                    "value": "true",
+                    "origin": "ergonomic",
+                    "kind": ["lp64_product"],
+                },
+                {
+                    "name": "UseCompressedOops",
+                    "type": "bool",
+                    "value": "true",
```
Jongy, line 870: Why do you specify the `value` everywhere if you gonna `pop` it before comparing?

Instead - and because we should probably test the collection of `value` - how about you don't store the `value` for flags which might change due to ergonomics, and when `value` is not present, `assert_jvm_flags_equal` will not test it, but when present it will test?

gprofiler/profilers/java.py
```diff
@@ -768,6 +872,19 @@ def _init_java_safemode(self, java_safemode: str) -> None:
                 f" --java-safemode={JavaSafemodeOptions.JAVA_EXTENDED_VERSION_CHECKS}"
             )
 
+    def _init_java_flag_retrieval(self, java_jvm_flags_to_retrieve: str) -> Union[JavaFlagRetrievalOptions, List[str]]:
+        if java_jvm_flags_to_retrieve is None or java_jvm_flags_to_retrieve == "":
```
Jongy, line 876: `is None`? Type is defined as `: str`.

gprofiler/profilers/java.py
```diff
@@ -661,6 +750,15 @@ def read_output(self) -> Optional[str]:
...
             help="Disable collection of async-profiler meminfo at the end of each cycle (collected by default)",
         ),
+        ProfilerArgument(
+            "--java-jvm-flags-to-retrieve",
+            dest="java_jvm_flags_to_retrieve",
+            type=str,
+            nargs="?",
+            default=JavaFlagRetrievalOptions.DEFAULT.value,
+            help="Comma-separated list of JVM flags to retrieve from the JVM process, 'all' to retrieve all flags, "
+            "or 'default' to retrieve only non-default jvm flags. Defaults to '%(default)s'",
```
Jongy, line 760: Maybe "'non-default' to retrieve only 'non-default' JVM flags"? Makes more sense to me. Also fix `JavaFlagRetrievalOptions`.

gprofiler/profilers/java.py
Jongy: I see, it's a collection. Closing

gprofiler/profilers/java.py
```diff
@@ -259,11 +288,71 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+            logger.error("Flags not found in jvm flags: %s", not_found_requested_flag_names)
+
+            return requested_flags
+
+        logging.error("Unrecognized jvm_flags_to_retrieve value: %s", self.jvm_flags_to_retrieve)
+
+        return None
+
+    @staticmethod
+    def filter_jvm_flag(flag: JvmFlag) -> bool:
```
Jongy, line 334: Document the decisions in this function.

"default" we skip because - we don't want (?)

"ccstrlist" we skip because not implemented

"notproduct" etc we skip because it's debugging flags (?)

gprofiler/profilers/java.py
```diff
@@ -259,11 +288,71 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
         # that loads other libs.
         libjvm_elfid = get_mapped_dso_elf_id(process, "/libjvm")
 
-        metadata = {"java_version": version, "libjvm_elfid": libjvm_elfid}
+        metadata = {
+            "java_version": version,
+            "libjvm_elfid": libjvm_elfid,
+            "jvm_flags": self.get_jvm_flags_serialized(process),
```
Jongy, line 294: Please collect this in try/except to refrain from distrupting the collection of other metadata items (such as `java_version`). Log upon error and use `None` for `jvm_flags`.

gprofiler/profilers/java.py
Jongy: Okay, I read the code and I see that the "default" here means gProfiler's default, which goes beyond the non-default JVM flags (skips nonproduct flags, etc).
Partial documentation is the worst case - this misleads the reader. Either make it complete, or remove tiny bits and just write `or 'default' for the defaults filtering settings for flags; see filter_jvm_flag(). Defaults to ...`.

gprofiler/profilers/java.py
```diff
@@ -259,11 +288,71 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+                if flag_name not in [flag.name for flag in requested_flags]
+            ]
+
+            logger.error("Flags not found in jvm flags: %s", not_found_requested_flag_names)
+
+            return requested_flags
+
+        logging.error("Unrecognized jvm_flags_to_retrieve value: %s", self.jvm_flags_to_retrieve)
+
+        return None
```
Jongy, line 331: Nicer in an `else` clause for the `if` - it is clearer that this is the fallback case.

Anyway, can it happen? Can't you change the last `elif` to be `else` followed with an `assert isinstance`?

gprofiler/profilers/java.py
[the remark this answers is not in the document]
Jongy: Please drop a reference here - either to the `--java-jvm-flags-to-retrieve` argument help section, or just a short description here of what these 2 mean

gprofiler/metadata/application_identifiers_java.py
```diff
@@ -11,16 +11,19 @@
...
 
 _logger = get_logger_adapter(__name__)
 
 
 class _JavaJarApplicationIdentifier(_ApplicationIdentifier):
+    def __init__(self, jattach_jcmd_runner: JattachJcmdRunner):
+        self.jattach_jcmd_runner = jattach_jcmd_runner
+
     def get_app_id(self, process: Process) -> Optional[str]:
+        assert self.jattach_jcmd_runner is not None
```
Jongy, line 24: Is this assert needed? Your constructor assigns a non-optional `JattachJcmdRunner` to this field.

gprofiler/metadata/application_identifiers_java.py
```diff
@@ -11,16 +11,19 @@
...
 from gprofiler.metadata.base_application_identifier import _ApplicationIdentifier
-from gprofiler.profilers.java import jattach_path
-from gprofiler.utils import run_process
+from gprofiler.profilers.java import JattachJcmdRunner
 
 _logger = get_logger_adapter(__name__)
 
 
 class _JavaJarApplicationIdentifier(_ApplicationIdentifier):
+    def __init__(self, jattach_jcmd_runner: JattachJcmdRunner):
```
Jongy, line 20: Missing call to `super()` c'tor

gprofiler/profilers/java.py
Jongy: @YoniKF - the convention in gProfiler is to pass extras this way. Non-logging kwargs are converted to extras - it is handled by `ExtraAdapter` in glogger. Please follow it and pass extras that way (and refrain from using `%` formatting anyway lol, resort to `.format()` but not `%` please)

gprofiler/metadata/application_identifiers_java.py
```diff
@@ -40,7 +43,7 @@ def get_app_id(self, process: Process) -> Optional[str]:
             if java_command:
                 return f"java: {java_command}{' (' + ' '.join(java_args) + ')' if java_args else ''}"
         except CalledProcessError as e:
-            _logger.warning(f"Couldn't get Java properties for process {process.pid}: {e.stderr}")
+            _logger.warning("Couldn't get Java properties for process %d: %s", process.pid, e.stderr)
```
Jongy, line 46: Can fix it to newstyle extra formatting, i.e `pid=process.pid, jattach_stderr=e.stderr`

gprofiler/profilers/java.py
```diff
@@ -259,11 +288,71 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+
+        if jvm_raw_flags is None:
+            return None
+
+        elif self.jvm_flags_to_retrieve == JavaFlagRetrievalOptions.ALL:
+            return jvm_raw_flags
+        elif self.jvm_flags_to_retrieve == JavaFlagRetrievalOptions.DEFAULT:
+            return [flag for flag in jvm_raw_flags if self.filter_jvm_flag(flag)]
+        elif isinstance(self.jvm_flags_to_retrieve, list):
+            requested_flags = [flag for flag in jvm_raw_flags if flag.name in self.jvm_flags_to_retrieve]
```
Jongy, line 318: Nicer with `filter` builtin

gprofiler/profilers/java.py
```diff
@@ -259,11 +288,71 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+                flag_name
+                for flag_name in self.jvm_flags_to_retrieve
+                if flag_name not in [flag.name for flag in requested_flags]
+            ]
+
+            logger.error("Flags not found in jvm flags: %s", not_found_requested_flag_names)
+
+            return requested_flags
+
+        logging.error("Unrecognized jvm_flags_to_retrieve value: %s", self.jvm_flags_to_retrieve)
```
Jongy, line 329: Please `warning` - user configuration errors are warning by convention. `error` is for things that were configured okay but do not behave okay.

Jongy: Also remove `%s`

gprofiler/profilers/java.py
```diff
@@ -259,11 +288,71 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+            return [flag for flag in jvm_raw_flags if self.filter_jvm_flag(flag)]
+        elif isinstance(self.jvm_flags_to_retrieve, list):
+            requested_flags = [flag for flag in jvm_raw_flags if flag.name in self.jvm_flags_to_retrieve]
+            not_found_requested_flag_names = [
+                flag_name
+                for flag_name in self.jvm_flags_to_retrieve
+                if flag_name not in [flag.name for flag in requested_flags]
+            ]
+
+            logger.error("Flags not found in jvm flags: %s", not_found_requested_flag_names)
```
Jongy, line 325: Also `warning`

Jongy, line 325: Don't log it if the list is empty?

gprofiler/profilers/java.py
```diff
@@ -259,11 +288,71 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+        if jvm_raw_flags is None:
+            return None
+
+        elif self.jvm_flags_to_retrieve == JavaFlagRetrievalOptions.ALL:
+            return jvm_raw_flags
+        elif self.jvm_flags_to_retrieve == JavaFlagRetrievalOptions.DEFAULT:
+            return [flag for flag in jvm_raw_flags if self.filter_jvm_flag(flag)]
+        elif isinstance(self.jvm_flags_to_retrieve, list):
+            requested_flags = [flag for flag in jvm_raw_flags if flag.name in self.jvm_flags_to_retrieve]
+            not_found_requested_flag_names = [
```
Jongy, line 319:
```suggestion
            not_found_flags = [
```

Shorter naming please. The context is clear

gprofiler/profilers/java.py
```diff
@@ -259,11 +288,71 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+
+        return [flag.to_dict() for flag in sorted(jvm_flags, key=lambda flag: flag.name)]
+
+    def get_jvm_flags(self, process: Process) -> Optional[List[JvmFlag]]:
+        jvm_raw_flags: Optional[List[JvmFlag]] = self.get_jvm_flags_raw(process)
+
+        if jvm_raw_flags is None:
+            return None
+
+        elif self.jvm_flags_to_retrieve == JavaFlagRetrievalOptions.ALL:
```
Jongy, line 313: General - `collect` is a more appropriate term here and used in other parts of the codebase. Let's rename this field to `collected_jvm_flags` and `JavaFlagRetrievalOptions` to `JavaFlagsCollectionOptions`

gprofiler/profilers/java.py
```diff
@@ -730,8 +829,13 @@ def __init__(
         self._pid_to_java_version: Dict[int, Optional[str]] = {}
         self._kernel_messages_provider = get_kernel_messages_provider()
         self._enabled_proc_events_java = False
+        self._java_jvm_flags_to_retrieve = self._init_java_flag_retrieval(java_jvm_flags_to_retrieve)
```
Jongy, line 832: Per https://github.com/Granulate/gprofiler/pull/660/files#r1096566312
```suggestion
        self._collected_jvm_flags = self._init_java_flags_collection(java_collect_jvm_flags_to)
```

Also rename to `--java-collect-jvm-flags=all`, `--java-collect-jvm-flags=default`.

gprofiler/profilers/java.py
```diff
@@ -730,8 +829,13 @@ def __init__(
...
+        self._jattach_jcmd_runner = JattachJcmdRunner(
+            stop_event=self._stop_event, jattach_timeout=self._jattach_timeout
+        )
         self._ap_timeout = self._duration + self._AP_EXTRA_TIMEOUT_S
-        self._metadata = JavaMetadata(self._stop_event)
+        application_identifiers.ApplicationIdentifiers.init_java(self._jattach_jcmd_runner)
```
Jongy, line 837: This has the (hidden?) assumption that java appids are used only from the `JavaProfiler`, i.e if the `JavaProfiler` was not initialized, it's okay that we don't have java appids. I checked now and it seems fine, and I guess, in any case we will get a `KeyError`...

Jongy (changes requested):
Done reviewing

gprofiler/profilers/java.py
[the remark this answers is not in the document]
YoniKF: `default is to retrieve only non-default jvm flags` is not accurate though, we perform some additional filtering.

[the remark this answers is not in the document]
Jongy: Yeah yeah I saw that later, please refer to the function in the help message or explain thoroughly what it does or don't explain at all. Don't document partially

tests/conftest.py
uriariel: but we don't specify version or hash in openjdk11, so I prefer either to pin them both to one version or leave them both "unpinned"

tests/test_java.py
uriariel: totally different, since JDK11 flags contains "origin" whilst JDK8 only specify "`default"` or `"non-default"`

gprofiler/metadata/application_identifiers_java.py
uriariel: but `_ApplicationIdentifier` is an abstract class that doesn't define an `__init__`, why should `_JavaJarApplicationIdentifier` call `super.__init__()`?

gprofiler/metadata/application_identifiers_java.py
uriariel: one can pass `None` somehow, I its OK to assert

tests/test_java.py
uriariel: we agreed that its not that important anyway, I would prefer leave it as is, as it checks default flag collection

gprofiler/profilers/java.py
uriariel: I want to support `None` as default for example when called from `make_java_profiler`, although I pass default value explicitly, if you want Ill remove it

gprofiler/profilers/java.py
uriariel: yes we will, it was the same with `set_enrichment_options` wasn't it?

gprofiler/profilers/java.py
```diff
@@ -259,11 +287,77 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+            return jvm_raw_flags
+        elif self.jvm_flags_to_collect == JavaFlagCollectionOptions.DEFAULT:
+            return [flag for flag in jvm_raw_flags if self.filter_jvm_flag(flag)]
+        elif isinstance(self.jvm_flags_to_collect, list):
+            requested_flags = [flag for flag in jvm_raw_flags if flag.name in self.jvm_flags_to_collect]
+            not_found_flags = [
+                flag_name
+                for flag_name in self.jvm_flags_to_collect
+                if flag_name not in [flag.name for flag in requested_flags]
+            ]
```
YoniKF, line 322:
```suggestion
            not_found_flags = set(self.jvm_flags_to_collect) - {flag.name for flag in requested_flags}
```

gprofiler/profilers/java.py
```diff
@@ -259,11 +287,77 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+
+            if not_found_flags:
+                logger.error("Flags not found in jvm flags:", not_found_flags=not_found_flags)
+
+            return requested_flags
+        else:
+            logger.error("Unrecognized jvm_flags_to_collect value:", jvm_flags_to_collect=self.jvm_flags_to_collect)
+            return None
```
YoniKF, line 330: As before, if getting here is a programming error and not an actual flow, either remove (and depend on type checking) or convert to an assertion.

gprofiler/profilers/java.py
```diff
@@ -768,6 +877,19 @@ def _init_java_safemode(self, java_safemode: str) -> None:
...
+            return JavaFlagCollectionOptions.DEFAULT
+        if any(
+            [
+                java_jvm_flags_to_collect == java_flag_collection_option.value
+                for java_flag_collection_option in JavaFlagCollectionOptions
+            ]
+        ):
+            return JavaFlagCollectionOptions(java_jvm_flags_to_collect)
+        else:
+            return java_jvm_flags_to_collect.split(",")
```
YoniKF, line 891: Is there nothing built-in argparse to do that for you?

gprofiler/profilers/java.py
```diff
@@ -768,6 +877,19 @@ def _init_java_safemode(self, java_safemode: str) -> None:
...
 
+    def _init_java_flag_collection(self, java_jvm_flags_to_collect: str) -> Union[JavaFlagCollectionOptions, List[str]]:
+        if java_jvm_flags_to_collect is None or java_jvm_flags_to_collect == "":
+            return JavaFlagCollectionOptions.DEFAULT
+        if any(
+            [
+                java_jvm_flags_to_collect == java_flag_collection_option.value
+                for java_flag_collection_option in JavaFlagCollectionOptions
+            ]
+        ):
```
YoniKF, line 888:
```suggestion
        if java_jvm_flags_to_collect in (java_flag_collection_option.value for java_flag_collection_option in JavaFlagCollectionOptions):
```
or keep the logic but drop the intermediate list.

Or, it might be more Pythonic to just try `JavaFlagCollectionOptions(java_jvm_flags_to_collect)` first and on `ValueError` do the split.

YoniKF (changes requested):

gprofiler/profilers/java.py
uriariel: thats how we do it for safemode options.[
](https://github.com/Granulate/gprofiler/blob/b07653ccba6b3536a6d94c222204e74f7b1e1a5f/gprofiler/profilers/java.py#L865)

gprofiler/profilers/java.py
```diff
@@ -259,11 +287,77 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+        if flag.type in ["ccstr", "ccstrlist"]:
+            return False
+
+        if set(flag.kind).intersection({"notproduct", "pd", "develop"}):
+            return False
+
+        return True
+
+    @functools.lru_cache(maxsize=1024)
+    def get_jvm_flags_raw(self, process: Process) -> Optional[List[JvmFlag]]:
```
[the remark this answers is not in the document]
uriariel: IIUC the only JVM flags that could change during execution time are `manageable` flags. Ill add `manageable` to the filter.

gprofiler/profilers/java.py
YoniKF: Still using an intermediate list.

gprofiler/profilers/java.py
uriariel: why we don't try except also for ```get_java_version```?

gprofiler/metadata/application_identifiers_java.py
Jongy: It's not abstract, it's an empty class with no c'tor definition, hence the default one.

If you don't leave a call to the c'tor here, then at some point in the future, a c'tor might be added to the parent class, and this new class will misbehave because you don't call the c'tor. It's the best practice to call it anyway.

gprofiler/metadata/application_identifiers_java.py
Jongy: You could tell that on every parameter to every function...

We don't add asserts in this case - it clutters the code. You can add asserts on public-facing APIs, or where mypy lacks info so an assert helps, or when it helps for code readability. Here's it's none of the cases, please remove

gprofiler/profilers/java.py
Jongy: It's a good point, we should. You can fix it in this PR like I did [here](https://github.com/Granulate/gprofiler/pull/656/files), or keep for a later PR. For new features that we add - we add them more robust to begin with

gprofiler/profilers/java.py
```diff
@@ -259,11 +287,86 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+            requested_flags = [flag for flag in jvm_raw_flags if flag.name in self.jvm_flags_to_collect]
+            not_found_flags = set(self.jvm_flags_to_collect) - {flag.name for flag in requested_flags}
+
+            # log if the set is not empty
+            if not_found_flags:
+                logger.error("Flags not found in jvm flags:", not_found_flags=not_found_flags)
+
+            return requested_flags
+        else:
+            assert False, f"Unrecognized jvm_flags_to_collect value: {self.jvm_flags_to_collect}"
```
Jongy, line 328: Uhh less nicer that way, instead you should remove the condition in the last `if`, and `assert` that condition itself.

gprofiler/profilers/java.py
```diff
@@ -661,6 +764,16 @@ def read_output(self) -> Optional[str]:
             default=True,
             help="Disable collection of async-profiler meminfo at the end of each cycle (collected by default)",
         ),
+        ProfilerArgument(
+            "--java-jvm-flags-to-collect",
+            dest="java_jvm_flags_to_collect",
+            type=str,
+            nargs="?",
+            default=JavaFlagCollectionOptions.DEFAULT.value,
```
Jongy, line 772: Add an option to disable collection altogether. `--java-collect-jvm-flags=none`, `--java-collect-jvm-flags=default` and so on. Please rename the flag to `--java-collect-jvm-flag`, this is imperative unlike `flags-to-collect`.

gprofiler/profilers/java.py
Jongy: That wasn't called in `JavaProfiler` though, but in some generic code.

Anyway, we will get an explicit error, so I'm fine with it

gprofiler/profilers/java.py
Jongy: And then rename the `to_collect` name everywhere, to the new name

tests/utils.py
Jongy: k

Jongy: k

tests/conftest.py
Jongy: Fair. okay

tests/utils.py
```diff
@@ -351,3 +355,32 @@ def _application_docker_container(
...
+
+    return pid
+
+
+def assert_jvm_flags_equal(actual_jvm_flags: Optional[List], expected_jvm_flags: List) -> None:
+    """
+    When comparing JVM flags, we want to ignore the "value" field
+    as it might change in the CI due to ergonomic set flags
+    """
+    assert actual_jvm_flags is not None, f"{actual_jvm_flags} != {expected_jvm_flags}"
```
Jongy, line 378: Do you need the assert message? For this one and for the next one?

gprofiler/profilers/java.py
Jongy: No, user configuration is WARNING and errors in profiler are ERRORs. that's the current convention

gprofiler/profilers/java.py
Jongy: https://github.com/Granulate/gprofiler/pull/660#discussion_r1103751968

gprofiler/profilers/java.py
Jongy: > Also rename to --java-collect-jvm-flags=all, --java-collect-jvm-flags=default.

@uriariel you didn't do this one and now I see [I commented it again](https://github.com/Granulate/gprofiler/pull/660#discussion_r1103750687). Please make sure you go over ALL review comments before re-requesting review.

gprofiler/profilers/java.py
uriariel: So there's no point in catching exceptions and returning null in inner methods.

tests/utils.py
uriariel: yes, that way its clearer when it fails

gprofiler/profilers/java.py
Jongy: This code is confusing for the reader.

If the API allows None - then you can check for None and specify it in the APi.
If the API doesn't allow None - don't check for it, assert it. Or don't do anything and trust mypy. But at most, assert.

assert is read differently than ifs

gprofiler/profilers/java.py
```diff
@@ -273,11 +302,91 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
         # that loads other libs.
         libjvm_elfid = get_mapped_dso_elf_id(process, "/libjvm")
 
-        metadata = {"java_version": version, "libjvm_elfid": libjvm_elfid}
+        try:
+            jvm_flags = self.get_jvm_flags_serialized(process)
+        except CalledProcessError as e:
+            jvm_flags = [{"error": e.stderr}]
```
Jongy, line 308: I thought it's nicer to return an error string like I did for the `exe` and `execfn` fields. No need for it to be a list.

Jongy, line 308: `e.stderr` can have high cardinality, let's just use the same `type(e).__name__` & log the exception? (i.e just remove this except clause and fallthrough the default one). We wil be able to check the logs for exact details of the failure.

gprofiler/profilers/java.py
Jongy: Wdym?

gprofiler/profilers/java.py
uriariel: Fixed already removed the check for None

gprofiler/profilers/java.py
Jongy: Ah with respect to this method. Yes, you are right. Letting the exception raise, catching it here and marking `error:` is preferred.

In a separate PR we can fix `java_version` and `libjvm_elfid` to also not fail us

gprofiler/profilers/java.py
```diff
@@ -681,6 +790,16 @@ def read_output(self) -> Optional[str]:
             default=True,
             help="Disable collection of async-profiler meminfo at the end of each cycle (collected by default)",
         ),
+        ProfilerArgument(
+            "--java-collect-jvm-flags",
+            dest="java_collect_jvm_flags",
+            type=str,
+            nargs="?",
+            default=JavaFlagCollectionOptions.DEFAULT.value,
+            help="Comma-separated list of JVM flags to collect from the JVM process, 'all' to collect all flags,"
```
Jongy, line 799: Document `'none'` option as well?

gprofiler/profilers/java.py
```diff
@@ -746,8 +866,15 @@ def __init__(
         self._pid_to_java_version: Dict[int, Optional[str]] = {}
         self._kernel_messages_provider = get_kernel_messages_provider()
         self._enabled_proc_events_java = False
+        self._java_collect_jvm_flags = self._init_java_flag_collection(java_collect_jvm_flags)
```
Jongy, line 869: `_java` prefix is not needed (needed for the args because it's the same namespace with other args, but this object is `JavaProfiler`, note that no other attr has `_java_` prefix

gprofiler/profilers/java.py
```diff
@@ -783,6 +910,16 @@ def _init_java_safemode(self, java_safemode: str) -> None:
                 f" --java-safemode={JavaSafemodeOptions.JAVA_EXTENDED_VERSION_CHECKS}"
             )
 
+    def _init_java_flag_collection(self, java_collect_jvm_flags: str) -> Union[JavaFlagCollectionOptions, List[str]]:
+        if java_collect_jvm_flags == "":
```
Jongy, line 914: Why is this a legit value? UX wise this seems like "none" to me - I passed a list and it was empty. It should have the same behavior as `none`.

gprofiler/profilers/java.py
uriariel: The exception is already logged in the ibber function that rethrows the exception, I'll add another log here for the general case.

btw e.stderr includes pid or something?

gprofiler/profilers/java.py
```diff
@@ -273,11 +302,91 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+            return False
+
+        return True
+
+    @functools.lru_cache(maxsize=1024)
+    def get_jvm_flags_raw(self, process: Process) -> List[JvmFlag]:
+        try:
+            output = self.jattach_jcmd_runner.run(process, "VM.flags -all")
+        except CalledProcessError as error:
+            logger.warning("Couldn't get jvm flags for process", pid=process.pid, stderr=error.stderr)
```
Jongy, line 385: As I suggested in `make_application_metadata`, we can remove this clause and just let it raise, no need for the specific log here. We log the stderr in any case in `run_process`.

gprofiler/profilers/java.py
```diff
@@ -273,11 +302,91 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+        if flag.origin == "default":
+            return False
+
+        if set(flag.kind).intersection({"notproduct", "pd", "develop"}):
+            return False
+
+        return True
+
+    @functools.lru_cache(maxsize=1024)
+    def get_jvm_flags_raw(self, process: Process) -> List[JvmFlag]:
```
Jongy, line 381: What's the name btw? Why `raw`? It gets the parsed flags. You can just call it `get_jvm_flags`

gprofiler/profilers/java.py
uriariel: Just thought it would be nicer, so if we parse it in the backend or ui we won't need to check the type and rather assume jvm_flags is a list

gprofiler/profilers/java.py
Jongy: Ah I see, `_unfiltered`?

gprofiler/profilers/java.py
```diff
@@ -273,11 +302,91 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+        """
+        if flag.origin == "default":
+            return False
+
+        if set(flag.kind).intersection({"notproduct", "pd", "develop"}):
+            return False
+
+        return True
+
+    @functools.lru_cache(maxsize=1024)
```
Jongy, line 380: Nitpick: you can cache the outer function, i.e after filtering, no need to include ALL data in the cache, only the data that passed filtering

gprofiler/profilers/java.py
uriariel: We not only filter in the other get_jvm_flags but also decide which flags to collect according to the argument

gprofiler/profilers/java.py
Jongy: We will still need special handling for the error case - there is only one entry which is a dict of one key == `error`. Doesn't help much. It will be easier to check if it's a string that startswith error lol. Make it a string to be consistent with other sites.

Jongy: Run this code and see the logs emitted by `run_process` - they include the full commandline, so yeah.

gprofiler/profilers/java.py
```diff
@@ -273,11 +302,91 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+        if jvm_flags is None:
+            return None
+
+        return [flag.to_dict() for flag in sorted(jvm_flags, key=lambda flag: flag.name)]
+
+    def get_jvm_flags(self, process: Process) -> Optional[Iterable[JvmFlag]]:
+        if self.java_collect_jvm_flags == JavaFlagCollectionOptions.NONE:
+            return None
+
+        jvm_raw_flags: Iterable[JvmFlag] = filter(self.filter_jvm_flag, self.get_jvm_flags_raw(process))
```
Jongy, line 333: Don't name it `raw` flags like the function `_raw` because it did go through filtering.

Function can be `_unfiltered`

Variable can be `_basic_filters` or sth like that

gprofiler/profilers/java.py
```diff
@@ -273,11 +302,91 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+        elif self.java_collect_jvm_flags == JavaFlagCollectionOptions.DEFAULT:
+            return filter(self.default_collection_filter_jvm_flag, jvm_raw_flags)
+        else:
+            assert isinstance(self.java_collect_jvm_flags, list), f"Unrecognized value: {self.java_collect_jvm_flags}"
+            requested_flags = [flag for flag in jvm_raw_flags if flag.name in self.java_collect_jvm_flags]
+            not_found_flags = set(self.java_collect_jvm_flags) - {flag.name for flag in requested_flags}
+
+            # log if the set is not empty
+            if not_found_flags:
+                logger.warning("Flags not found in jvm flags:", not_found_flags=not_found_flags)
```
Jongy, line 346:
```suggestion
                logger.warning("Requested flags are missing from JVM flags:", missing_flags=not_found_flags)
```

gprofiler/profilers/java.py
```diff
@@ -273,11 +302,91 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+
+        jvm_raw_flags: Iterable[JvmFlag] = filter(self.filter_jvm_flag, self.get_jvm_flags_raw(process))
+
+        if self.java_collect_jvm_flags == JavaFlagCollectionOptions.ALL:
+            return jvm_raw_flags
+        elif self.java_collect_jvm_flags == JavaFlagCollectionOptions.DEFAULT:
+            return filter(self.default_collection_filter_jvm_flag, jvm_raw_flags)
+        else:
+            assert isinstance(self.java_collect_jvm_flags, list), f"Unrecognized value: {self.java_collect_jvm_flags}"
+            requested_flags = [flag for flag in jvm_raw_flags if flag.name in self.java_collect_jvm_flags]
```
Jongy, line 341:
```suggestion
            found_flags = [flag for flag in jvm_raw_flags if flag.name in self.java_collect_jvm_flags]
```

tests/conftest.py
```diff
@@ -566,6 +560,10 @@ def _set_enrichment_options() -> None:
             application_metadata=True,
         )
     )
+
+    ApplicationIdentifiers.init_java(
+        JattachJcmdRunner(stop_event=Event(), jattach_timeout=AsyncProfiledProcess._JATTACH_TIMEOUT)
```
Jongy, line 565:
```suggestion
        JattachJcmdRunner(stop_event=Event(), jattach_timeout=AsyncProfiledProcess._DEFAULT_JATTACH_TIMEOUT)
```

Let's name it that way

tests/utils.py
```diff
@@ -345,6 +349,38 @@ def _application_docker_container(
...
+
+    assert len(actual_jvm_flags) == len(expected_jvm_flags), f"{actual_jvm_flags} != {expected_jvm_flags}"
+
+    for actual_flag_dict, expected_flag_dict in zip(actual_jvm_flags, expected_jvm_flags):
+        actual_flag_value = actual_flag_dict.pop("value")
+        expected_flag_value = expected_flag_dict.pop("value")
+
+        if expected_flag_value is not None:
+            assert actual_flag_value == expected_flag_value, f"{actual_jvm_flags} != {expected_jvm_flags}"
```
Jongy, line 379: I suppose your assert message should show the flag values in it, not the flags themselves. u already popped the values

tests/test_java.py
```diff
@@ -811,3 +813,316 @@ def flush_output_and_stop_async_profiler(self: AsyncProfiledProcess, *args: Any,
...
+            ],
+        ) as container:
+            assert (
+                profiler._metadata.get_jvm_flags_serialized(psutil.Process(container.attrs["State"]["Pid"]))
+                == expected_flags
+            )
+
+
+@pytest.mark.parametrize("in_container", [True])
+@pytest.mark.parametrize("expected_flags", [None])
```
Jongy, line 1120: UX wise let's return an empty list for `none`? Instead of `None`

tests/test_java.py
```diff
@@ -811,3 +813,316 @@ def flush_output_and_stop_async_profiler(self: AsyncProfiledProcess, *args: Any,
...
+    java_cli_flags: str,
+    expected_flags: List[Dict[str, Union[None, str, List[str]]]],
+    application_pid: int,
+    application_image_tag: str,
+    profiler_state: ProfilerState,
+) -> None:
+    """
+    Tests filtering of jvm flags we don't support collecting
+    """
+    with make_java_profiler(profiler_state, java_collect_jvm_flags="MaxHeapFreeRatio, MinHeapFreeRatio") as profiler:
```
Jongy, line 1099:
```suggestion
    with make_java_profiler(profiler_state, java_collect_jvm_flags="MaxHeapFreeRatio,MinHeapFreeRatio") as profiler:
```

Jongy: Errors like this one are why it's best to include a positive case in all negative tests. Your test here could use `caplog` to search for the log message "Requested flag not found" which would catch this error

Jongy: Not a must now, if u want

tests/test_java.py
```diff
@@ -811,3 +813,316 @@ def flush_output_and_stop_async_profiler(self: AsyncProfiledProcess, *args: Any,
...
+    application_docker_image: Image,
+    assert_collapsed: AssertInCollapsed,
+    java_cli_flags: str,
+    expected_flags: List[Dict[str, Union[None, str, List[str]]]],
+    application_pid: int,
+    application_image_tag: str,
+    profiler_state: ProfilerState,
+) -> None:
+    """
+    Tests filtering of jvm flags we don't support collecting
```
Jongy, line 1097: Why don't we support these 2? Because they are managable?

tests/test_java.py
```diff
@@ -811,3 +813,316 @@ def flush_output_and_stop_async_profiler(self: AsyncProfiledProcess, *args: Any,
...
+    application_pid: int,
+    expected_flags: List[Dict[str, Union[None, str, List[str]]]],
+) -> None:
+    with make_java_profiler(profiler_state) as profiler:
+        assert_jvm_flags_equal(
+            profiler._metadata.get_jvm_flags_serialized(psutil.Process(application_pid)), expected_flags
+        )
+
+
+@pytest.mark.parametrize(
```
Jongy, line 978: No test that I saw so far actually checks the collected value - you use `None` everywhere so it's popped and ignored.

This test seems like a good case to test non-None values, as these are hardcoded that you pass.

Jongy (changes requested):
Done reviewing

tests/test_java.py
uriariel: Yes, `manageable` flags can be changed during execution

tests/test_java.py
uriariel: Fixed, however notices that the warning message in `caplog.text` does not include the `extra_kwargs` (the list of not found flags).
Do you have an idea?

gprofiler/profilers/java.py
uriariel: changed to `Missing requested flags:` since `Requested flags are missing from JVM` sounds like all of the requested flags were missing

tests/utils.py
uriariel:
```python
        if expected_flag_value is not None:
            assert (
                actual_flag_value == expected_flag_value
            ), f"{actual_flag_dict|{'value': actual_flag_value}} != {expected_flag_dict|{'value': expected_flag_value}}"
            ```
so that if the value is not the same we will know which flag it belonged to

File: tests/test_java.py (hunk 76 above)
Comment (uriariel) [reply tests/test_java.py:1120]: But its would be nice to tell apart when all flags were filtered vs when none were requested, wouldn't it?

File: gprofiler/profilers/java.py [hunk 80]
```diff
@@ -330,22 +330,22 @@ def get_jvm_flags(self, process: Process) -> Optional[Iterable[JvmFlag]]:
         if self.java_collect_jvm_flags == JavaFlagCollectionOptions.NONE:
             return None
 
-        jvm_raw_flags: Iterable[JvmFlag] = filter(self.filter_jvm_flag, self.get_jvm_flags_raw(process))
+        filtered_jvm_flags = self.get_supported_jvm_flags(process)
```
Jongy, line 333: supported_jvm_flags?

gprofiler/profilers/java.py
Jongy: kk

gprofiler/profilers/java.py
uriariel: supported for collection, `basic_filter` sounds unclear IMO

Jongy: Ok

gprofiler/profilers/java.py
Jongy: ok now

tests/test_java.py
Jongy: "No flags were requested" is not a legit use case. There's no reason to create multiple edge cases where you only need one. Empty list is no flags and that's it

tests/test_java.py
Jongy: Yes, you can use `caplog.records` (search for usage examples in other tests)

Jongy:
2 small comments remaining + merge from master after https://github.com/Granulate/gprofiler/pull/698 is merged, and we can merge this one.

Jongy:
I pushed a fix to the Windows Ci

Jongy:
Merging, will fix the Windows CI later.

Jongy (approved):

Issue #648, opened 2023-01-01
java: async-profiler: remove our semicolon patch


Jongy:
From skimming the code it seems that https://github.com/jvm-profiling-tools/async-profiler/commit/b0a44524bad111a7f18c8309c0ebec56ccfdcf2a now prevents `;` to be outputted in stackcollapses, and it is replaced by `|`.

If that's the case, we can remove [our patch](https://github.com/Granulate/async-profiler/commit/6242654ccf79dcba886daed037aafd434f4eb69e) - the less changes to upstream AP, the better.

When doing this, please add a test that `|` is indeed output where in cases where it should be, instead of `;`. I suppose that in `public static void main(final String[] args)` we'll be able to see `Ljava/lang/String;` vs `Ljava/lang/String|`.

PR #661, opened 2023-01-24
java-ap: remove semicolon patch

marcin-ol:
## Description
Removing local workaround for semicolons in frame names; upstream already handles this.
- Integrate with Granulate/async-profiler#4.
- Add test that verifies the java frames from async profiler include no semicolon `';'`.

## Related Issue
#648 

## Motivation and Context
- the less changes to upstream AP, the better.

## How Has This Been Tested?
Testcase added in `test_java.py` testsuite.

### DIFF
--- a/scripts/async_profiler_build_shared.sh
+++ b/scripts/async_profiler_build_shared.sh
@@ -8,7 +8,7 @@ set -euo pipefail
 VERSION=v2.9g5
 GIT_REV="60cbc128a752c657e8a97300784686e08c8dbe9d"
 
-git clone --depth 1 -b "$VERSION" https://github.com/granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
+git clone --depth 1 -b "$VERSION" https://github.com/Granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
 make all
 
 # add a version file to the build directory
--- a/scripts/async_profiler_build_shared.sh
+++ b/scripts/async_profiler_build_shared.sh
@@ -5,10 +5,10 @@
 #
 set -euo pipefail
 
-VERSION=v2.9g4
-GIT_REV="cde6aead3f8c30ccc79eaa3a44f36ed97b8ebc91"
+VERSION=remove-our-semicolon-patch
+GIT_REV="f9d5b861eb0b0bb2e0a8e2b4ed7bf6ca13a41d63"
 
-git clone --depth 1 -b "$VERSION" https://github.com/Granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
+git clone --depth 1 -b "$VERSION" https://github.com/marcin-ol/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
 make all
 
 # add a version file to the build directory
--- a/scripts/async_profiler_build_shared.sh
+++ b/scripts/async_profiler_build_shared.sh
@@ -5,10 +5,10 @@
 #
 set -euo pipefail
 
-VERSION=remove-our-semicolon-patch
-GIT_REV="f9d5b861eb0b0bb2e0a8e2b4ed7bf6ca13a41d63"
+VERSION=v2.9g5
+GIT_REV="60cbc128a752c657e8a97300784686e08c8dbe9d"
 
-git clone --depth 1 -b "$VERSION" https://github.com/marcin-ol/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
+git clone --depth 1 -b "$VERSION" https://github.com/granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
 make all
 
 # add a version file to the build directory
--- a/tests/test_java.py
+++ b/tests/test_java.py
@@ -705,4 +705,5 @@ def test_java_frames_include_no_semicolons(
     ) as profiler:
         collapsed = snapshot_pid_profile(profiler, application_pid).stacks
         assert is_function_in_collapsed(".main([Ljava/lang/String|)", collapsed)
+        assert not is_function_in_collapsed(";_[j];", collapsed)
         assert not is_pattern_in_collapsed(r"\([^);]+[;][^)]*\)", collapsed)
--- a/tests/test_java.py
+++ b/tests/test_java.py
@@ -715,4 +715,6 @@ def test_java_frames_include_no_semicolons(
         # no semicolon in return type:
         assert not is_function_in_collapsed(";_[j];", collapsed)
         # no semicolon in arguments list signatures:
-        assert not is_pattern_in_collapsed(r"\([^);]+[;][^)]*\)", collapsed)
+        assert not is_pattern_in_collapsed(r"\([^);]+;[^)]*\)", collapsed)
+        # only a pipe can occur within arguments
+        assert is_pattern_in_collapsed(r"\([^);|]+\|[^)]*\)", collapsed)
--- a/tests/test_java.py
+++ b/tests/test_java.py
@@ -693,6 +693,7 @@ def test_meminfo_logged(
         assert "async-profiler memory usage (in bytes)" in caplog.text
 
 
+# test that java frames include no semicolon but use a pipe '|' character instead, as implemented by AP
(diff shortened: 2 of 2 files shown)


scripts/async_profiler_build_shared.sh
```diff
@@ -5,10 +5,10 @@
 #
 set -euo pipefail
 
-VERSION=v2.9g4
-GIT_REV="cde6aead3f8c30ccc79eaa3a44f36ed97b8ebc91"
+VERSION=remove-our-semicolon-patch
```
Jongy, line 8: Let's update it to v2.9g5 60cbc128a752c657e8a97300784686e08c8dbe9d

tests/test_java.py
```diff
@@ -691,3 +691,19 @@ def test_meminfo_logged(
...
+    application_pid: int,
+) -> None:
+    with make_java_profiler(
+        storage_dir=str(tmp_path),
+        duration=3,
+        frequency=999,
+    ) as profiler:
+        collapsed = snapshot_pid_profile(profiler, application_pid).stacks
+        assert is_function_in_collapsed(".main([Ljava/lang/String|)", collapsed)
+        assert not is_function_in_collapsed(";_[j];", collapsed)
```
Jongy, line 708: What is checked here? Why will we ever have an empty function name? :sweat_smile:

tests/test_java.py
```diff
@@ -691,3 +691,19 @@ def test_meminfo_logged(
...
+        assert not is_pattern_in_collapsed(r"\([^);]+[;][^)]*\)", collapsed)
```
Jongy, line 709: What is checked here?

tests/test_java.py
```diff
@@ -691,3 +691,19 @@ def test_meminfo_logged(
...
+    tmp_path: Path,
+    application_pid: int,
+) -> None:
+    with make_java_profiler(
+        storage_dir=str(tmp_path),
+        duration=3,
+        frequency=999,
+    ) as profiler:
+        collapsed = snapshot_pid_profile(profiler, application_pid).stacks
+        assert is_function_in_collapsed(".main([Ljava/lang/String|)", collapsed)
```
Jongy, line 707: Also comment what's checked here (the `|` is used instead of `;` by AP)?

tests/test_java.py
```diff
@@ -691,3 +691,19 @@ def test_meminfo_logged(
     ) as profiler:
         snapshot_pid_profile(profiler, application_pid)
         assert "async-profiler memory usage (in bytes)" in caplog.text
+
+
+@pytest.mark.parametrize("in_container", [True])
+def test_java_frames_include_no_semicolons(
```
Jongy, line 697: Semicolons should be used between frames, but not as part of a frame. Let's also test here that they are used between frames (so check that a function appears after another function with `;` in between?)

Jongy (changes requested):
Don't we need also to set `STYLE_NO_SEMICOLON`?

Jongy:
> Don't we need also to set `STYLE_NO_SEMICOLON`?

Oh, I see that's the default for `dumpCollapsed`, so we don't.

scripts/async_profiler_build_shared.sh
marcin-ol: 👍

tests/test_java.py
marcin-ol: Will do.

tests/test_java.py
marcin-ol: 👍 
Will add a comment.

tests/test_java.py
marcin-ol: It's testing for a object-kind return type of a function.
I've run this test against a non-patched version of async-profiler and this pattern appears in collapsed stacls.

tests/test_java.py
marcin-ol: 👍 
Will explain it in a comment.

scripts/async_profiler_build_shared.sh
```diff
@@ -5,10 +5,10 @@
 #
 set -euo pipefail
 
-VERSION=v2.9g4
-GIT_REV="cde6aead3f8c30ccc79eaa3a44f36ed97b8ebc91"
+VERSION=v2.9g5
+GIT_REV="60cbc128a752c657e8a97300784686e08c8dbe9d"
 
-git clone --depth 1 -b "$VERSION" https://github.com/Granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
+git clone --depth 1 -b "$VERSION" https://github.com/granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
```
Jongy, line 11: Bring back the G :)

tests/test_java.py
Jongy: Ahh because if any function returns an object, it'll be `func()Ljava/lang/String;_[j];`. I see.

Jongy:
Looks good, one small comment, let's see if tests pass

scripts/async_profiler_build_shared.sh
marcin-ol: Right :)

tests/test_java.py
```diff
@@ -691,3 +691,28 @@ def test_meminfo_logged(
...
+        # after each object type. async-profiler replaces them with '|' to have consistent collapsed stacks format.
+        # We make sure semicolons do not appear neither in function arguments list nor in the return type.
+        # make sure pipe character is used where semicolon would be:
+        assert is_function_in_collapsed("Fibonacci.main([Ljava/lang/String|)V_[j];", collapsed)
+        # make sure semicolon is used only to separate function frames:
+        assert is_function_in_collapsed("Fibonacci$1.run()V_[j];java/io/File.list()[Ljava/lang/String|_[j]", collapsed)
+        # no semicolon in return type:
+        assert not is_function_in_collapsed(";_[j];", collapsed)
+        # no semicolon in arguments list signatures:
+        assert not is_pattern_in_collapsed(r"\([^);]+[;][^)]*\)", collapsed)
```
Jongy, line 718:
```suggestion
        assert not is_pattern_in_collapsed(r"\([^);]+;[^)]*\)", collapsed)
```
Simpler for the same purpose?

For a stronger check, let's also test the "positive case", i.e
```
assert is_pattern_in_collapsed(r"\([^)|]+[^)]*\)", collapsed)
```
(or whatever - with the `|` instead of `;`, this proves that the regex is correct; because the regex might not match due to other reasons as well)

Jongy:
```
$ git range-diff upstream/master..cde6aead3f8c30ccc79eaa3a44f36ed97b8ebc91 upstream/master..60cbc128a752c657e8a97300784686e08c8dbe9d
 1:  6242654 <  -:  ------- Replace semicolon in method descriptors to vertical bar, since a semicolon is already used to separate between methods
 2:  35531a4 =  1:  13e835e Remove the "Profiling started" message
 3:  473b925 =  2:  e16fad0 Remove the "Profiling stopped" message
 4:  bacc840 =  3:  0c25728 Compile statically with libstdc++
 5:  f162e85 =  4:  fca73de Build jattach statically
 6:  d2d2c12 =  5:  f8f77fd Build fdtransfer statically
 7:  900bbf0 =  6:  ab6e1de Compile statically with libgcc
 8:  54c85f6 =  7:  ae5daf1 Fix -static-* args passed only in non-MERGE case :(
 9:  6be6afa =  8:  c8c72a1 Close standard streams after forking in fdtransfer
10:  0d3e5a6 =  9:  49080a2 -static-libstdc++ only on musl
11:  bb1f846 = 10:  f4d874c Convert frame name format to 'function (DSO)' for LIB style (#2)
12:  92045e8 = 11:  bb06e23 Add accept timeout as 3rd parameter of fdtransfer
13:  d761b57 = 12:  faaabf8 Use compat-glibc on x86_64 glibc
14:  26d2a66 = 13:  df5df4a dump stats on stop
15:  cde6aea = 14:  60cbc12 Do not try to parse mapped pseudofiles

```

Jongy:
One final comment

tests/test_java.py
marcin-ol: okay, added another pattern.

Jongy:
LGTM :)

Jongy (approved):

Issue #649, opened 2023-01-01
java: async-profiler: test that no output is written to process stdout/stderr


Jongy:
async-profiler writes in the process' stdout/stderr. We have removed those prints - see https://github.com/Granulate/async-profiler/commit/35531a45a197048a450aca38386d144dfa1060bb and https://github.com/Granulate/async-profiler/commit/473b925f4a82bd7bd53b0737e3ead238c38f7a09.

We should have a test that proves that no additional output is written to stdout/stderr during the cycle of a single profiling session, i.e start/stop of AP. stdout & stderr should remain empty, or, should contain the only output that the profiled app is printing. I suggest adding a `System.out.println(...)` and `System.err.println(...)` to our Fibonacci Java app and then compare the output of the process to those 2 prints.

Jongy:
(I want it tested continuously because as we upgrade AP, we might accidentally introduce prints that are added in new versions)

marcin-ol:
Do you mean we want to make sure no code writes explicitly to stdout?

I'm asking because the commits you have mentioned didn't write directly to stdout.
We directed output to a stream **out**, which is an instance of `ofstream` initialized with argument `file=` passed to jattach.
And we always set the `file=` argument.

marcin-ol:
Found out that the test needed only call `flush()` on output stream to reproduce the behavior.

Jongy:
If I run async-profiler from master on a Java process, I get `Profiling started` prints in the stdout of the Java process itself. That's what we are trying to avoid in removing those logs.

PR #664, opened 2023-01-26
Add test for zero stray output from java async-proviler

marcin-ol:
## Description
Make sure that async-profiler generates no output to profiled process stdout/ stderr streams.

## Related Issue
#649 

## Motivation and Context
We don't need any output from async-profiler to appear in profiled process output streams. We don't need it and want to avoid it.  
Recently few commits were merged which removed some print statements reaching process outputs.

## How Has This Been Tested?
Testcase was added to `test_java` testsuite.

### DIFF
--- a/tests/test_java.py
+++ b/tests/test_java.py
@@ -25,6 +25,7 @@ from granulate_utils.java import parse_jvm_version
 from granulate_utils.linux.ns import get_process_nspid
 from granulate_utils.linux.process import is_musl
 from packaging.version import Version
+from psutil import Process
 from pytest import LogCaptureFixture, MonkeyPatch
 
 from gprofiler.profilers.java import (
@@ -80,6 +81,19 @@ class AsyncProfiledProcessForTests(AsyncProfiledProcess):
         self._run_async_profiler(
             self._get_base_cmd() + [f"status,log={self._log_path_process},file={self._output_path_process}"],
         )
+
+    def read_ap_version(self: AsyncProfiledProcess) -> str:
+        # call async-profiler 'version' action to discover its version number
+        # read output from AsyncProfiledProcess, which should be empty
+        output_without_version = cast_away_optional(self.read_output())
+        assert output_without_version == ""
+        self._run_async_profiler(
+            self._get_base_cmd() + [f"version" f"{self._get_ap_output_args()}" f",log={self._log_path_process}"]
+        )
+        # async-profiler emits version number compiled into itself
+        output_with_version = cast_away_optional(self.read_output())
+        version = output_with_version.strip()
+        return version
 
 
 def test_async_profiler_already_running(
@@ -753,13 +767,24 @@ def test_no_stray_output_in_stdout_stderr(
         duration=3,
         frequency=999,
     ) as profiler:
+        # read ap-version from test ap-process to match the proper output against it
+        with AsyncProfiledProcessForTests(
+            process=Process(application_pid),
+            storage_dir=profiler._storage_dir,
+            insert_dso_name=False,
+            stop_event=profiler._stop_event,
+            mode="itimer",
+            ap_safemode=0,
+            ap_args="",
+        ) as ap_proc:
+            ap_version = ap_proc.read_ap_version()
         collapsed = snapshot_pid_collapsed(profiler, application_pid)
         assert_collapsed(collapsed)
     application_docker_container.stop(timeout=3)
     application_docker_container.wait(timeout=3)
     textout = application_docker_container.logs(stdout=True, stderr=False).decode()
     texterr = application_docker_container.logs(stdout=False, stderr=True).decode()
-    # output from Fibonacci should be the only output in stdout
-    assert textout.strip() == "Fibonacci thread starting"
+    # output from Fibonacci and async-profiler version should be the only lines  in stdout
+    assert textout.splitlines() == ["Fibonacci thread starting", ap_version]
     # output from Fibonacci should be the only expected output in stderr
     assert texterr.strip() == "Fibonacci loop starting"
--- a/tests/test_java.py
+++ b/tests/test_java.py
@@ -718,3 +718,42 @@ def test_java_frames_include_no_semicolons(
         assert not is_pattern_in_collapsed(r"\([^);]+;[^)]*\)", collapsed)
         # only a pipe can occur within arguments
         assert is_pattern_in_collapsed(r"\([^);|]+\|[^)]*\)", collapsed)
+
+
+# test that async profiler doesn't print anything to applications stdout, stderr streams
+@pytest.mark.parametrize("in_container", [True])
+def test_no_stray_output_in_stdout_stderr(
+    tmp_path: Path,
+    application_pid: int,
+    application_docker_container: Container,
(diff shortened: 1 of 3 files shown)


tests/test_java.py
```diff
@@ -718,3 +718,42 @@ def test_java_frames_include_no_semicolons(
...
+    with make_java_profiler(
+        storage_dir=str(tmp_path),
+        duration=3,
+        frequency=999,
+    ) as profiler:
+        snapshot_pid_profile(profiler, application_pid)
+        textout = application_docker_container.logs(stdout=True, stderr=False).decode()
+        texterr = application_docker_container.logs(stdout=False, stderr=True).decode()
+        # output from Fibonacci should be the only output in stdout
+        assert textout.splitlines() == ["Fibonacci thread starting"]
```
Jongy, line 757:
```suggestion
        assert textout == "Fibonacci thread starting\n"
```

simpler? Same for stderr

tests/test_java.py
```diff
@@ -718,3 +718,42 @@ def test_java_frames_include_no_semicolons(
...
+    application_pid: int,
+    application_docker_container: Container,
+    monkeypatch: MonkeyPatch,
+) -> None:
+    # save original stop function
+    stop_async_profiler = AsyncProfiledProcess.stop_async_profiler
+
+    # replace async profiler stop routine to trigger flushing standard streams
+    def flush_output_and_stop_async_profiler(self: AsyncProfiledProcess, *args, **kwargs) -> str:
+        # Call 'version' and 'check' actions on async-profiler to make sure writes to stdout, stderr are flushed
```
Jongy, line 736: Why does it cause flushing?

Anyway - a simpler solution: run the container, do a single profile, then stop the container; I suppose Java handles SIGTERM gracefully, so you will get the logs properly.

Jongy (changes requested):
Done reviewing.

Please run a negative test - i.e a test where you change the AP reference to your fork, which removes e.g https://github.com/Granulate/async-profiler/commit/35531a45a197048a450aca38386d144dfa1060bb, so it proves that the test now fails.

tests/test_java.py
marcin-ol: That's true - java applications flush stdout by default (they trigger autoflush by default).  
async-profiler on the other hand is c++, and doesn't flush by default after each write to output stream.

I have observed that `version` action does trigger flush on standard output: ([profiler.cpp#L1548](https://github.com/granulate/async-profiler/blob/master/src/profiler.cpp#L1548)):
```cpp
        case ACTION_VERSION:
            out << PROFILER_VERSION;
            out.flush();
            break;
```
(this code is still there in the upstream, btw).

tests/test_java.py
marcin-ol: Yes.

tests/test_java.py
Jongy: I see. Good point. Please keep the code doing the `version` call and elaborate in a comment on why it's needed (you can also give a github permalink to this code which it the clear explanation why we want to run `version`).

Jongy (changes requested):
Done reviewing

tests/test_java.py
```diff
@@ -718,3 +718,46 @@ def test_java_frames_include_no_semicolons(
...
+        monkeypatch.setattr(AsyncProfiledProcess, "stop_async_profiler", flush_output_and_stop_async_profiler)
+
+    with make_java_profiler(
+        storage_dir=str(tmp_path),
+        duration=3,
+        frequency=999,
+    ) as profiler:
+        collapsed = snapshot_pid_collapsed(profiler, application_pid)
+        assert_collapsed(collapsed)
+        application_docker_container.stop(timeout=3)
```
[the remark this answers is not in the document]
marcin-ol: Thanks, will fix it.

tests/test_java.py
marcin-ol: 👍

marcin-ol:
> Done reviewing.
> 
> Please run a negative test - i.e a test where you change the AP reference to your fork, which removes e.g [Granulate/async-profiler@35531a4](https://github.com/Granulate/async-profiler/commit/35531a45a197048a450aca38386d144dfa1060bb), so it proves that the test now fails.

Added a test commit to async-profiler bringing back the prints that were removed: [marcin-ol/async-profiler@623637b](https://github.com/marcin-ol/async-profiler/commit/623637b593282cc7fe5a07b85d35c00bc6c245d1).
- engaging the test against that commit: [gprofiler@9343ff4](https://github.com/Granulate/gprofiler/pull/664/commits/9343ff484ca0ce3d43c7fc6de83211004e5c1246)
 
I expect the test not to fail - although it should - when flushing is disabled.  
Introduced this commit to back up this claim.

marcin-ol:
> > Done reviewing.
> > Please run a negative test - i.e a test where you change the AP reference to your fork, which removes e.g [Granulate/async-profiler@35531a4](https://github.com/Granulate/async-profiler/commit/35531a45a197048a450aca38386d144dfa1060bb), so it proves that the test now fails.
> 
> Added a test commit to async-profiler bringing back the prints that were removed: [marcin-ol/async-profiler@623637b](https://github.com/marcin-ol/async-profiler/commit/623637b593282cc7fe5a07b85d35c00bc6c245d1).
> 
> * engaging the test against that commit: [gprofiler@9343ff4](https://github.com/Granulate/gprofiler/pull/664/commits/9343ff484ca0ce3d43c7fc6de83211004e5c1246)
> 
> I expect the test not to fail - although it should - when flushing is disabled. Introduced this commit to back up this claim.

Without flushing async-profiler's output, the test fails, even though the profiler emitted additional output:
```
tests/test_java.py::test_no_stray_output_in_stdout_stderr[False-True] PASSED [ 35%]
tests/test_java.py::test_no_stray_output_in_stdout_stderr[True-True] FAILED [ 36%]
```

Jongy:
> Without flushing async-profiler's output, the test fails, even though the profiler emitted additional output:
> 
> ```
> tests/test_java.py::test_no_stray_output_in_stdout_stderr[False-True] PASSED [ 35%]
> tests/test_java.py::test_no_stray_output_in_stdout_stderr[True-True] FAILED [ 36%]
> ```

If I understand it correctly - it looks good - the test fails when we DO flush and passes when we don't.

That's what I meant by a negative test. We can remove the async-profiler build commit change, it looks ready for merge IMO.

marcin-ol:
> > Without flushing async-profiler's output, the test fails, even though the profiler emitted additional output:
> > ```
> > tests/test_java.py::test_no_stray_output_in_stdout_stderr[False-True] PASSED [ 35%]
> > tests/test_java.py::test_no_stray_output_in_stdout_stderr[True-True] FAILED [ 36%]
> > ```
> 
> If I understand it correctly - it looks good - the test fails when we DO flush and passes when we don't.
> 
> That's what I meant by a negative test. We can remove the async-profiler build commit change, it looks ready for merge IMO.

Yeah, that's what should be expected.
I've made flushing nonvoluntary and restored reference to mainstream async-profiler.

Jongy (approved):

Jongy:
Awesome @marcin-ol :)

PR #666, opened 2023-01-29
perf: Cleanups & robustness improvements

Jongy:
* CLean up logging (pass stdout & stderr as extras, use standard "perf ({type} mode)" in all logs)
* Clean up leftover files before starting - leftover files are moved by perf to "xxx.old" which is later identified by our wait_for_file_by_prefix(), which we don't want.
* Refactor check_if_needs_restart() to check_if_needs_memory_restart(), to signify its purpose.
* Check if perf is running every cycle beginning, and re-start it if not. perf might be killed e.g due to OOM, and we had nothing in place to re-start it. This means all future gProfiler cycles would fails in this run :/

@d3dave lmk if you have a better design for this "cleanup".

TODO:
* [ ] add automatic tests for restart if perf is killed
* [ ] add automatic tests for restart based on memory (?)

### DIFF
--- a/gprofiler/profilers/perf.py
+++ b/gprofiler/profilers/perf.py
@@ -26,7 +26,7 @@ from gprofiler.metadata.application_metadata import ApplicationMetadata
 from gprofiler.profilers.node import clean_up_node_maps, generate_map_for_node_processes, get_node_processes
 from gprofiler.profilers.profiler_base import ProfilerBase
 from gprofiler.profilers.registry import ProfilerArgument, register_profiler
-from gprofiler.utils import run_process, start_process, wait_event, wait_for_file_by_prefix
+from gprofiler.utils import remove_path, run_process, start_process, wait_event, wait_for_file_by_prefix
 from gprofiler.utils.perf import perf_path, valid_perf_pid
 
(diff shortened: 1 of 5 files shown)


gprofiler/profilers/perf.py
```diff
@@ -81,39 +85,66 @@ def _get_perf_cmd(self) -> List[str]:
...
+
+    def restart_if_not_running(self) -> None:
+        """
+        Restarts perf if it was stopped for whatever reason.
+        """
+        if not self.is_running():
+            logger.warning(f"{self._log_name} not running (unexpectedly), restarting...")
+            self.restart()
+
+    def check_if_needs_memory_restart(self) -> None:
```
d3dave, line 134: This doesn't just check, it restarts. So `restart_if_rss_exceeded`?

d3dave (commented):
Just name of the function. Everything else lgtm.

gprofiler/profilers/perf.py
Jongy: Hm. Yeah makes sense

d3dave (approved):

d3dave (approved):

Issue #640, opened 2022-12-20
Processes exiting during a session do not get enriched with container name


Jongy:
We enrich container names when merging profilers:
https://github.com/Granulate/gprofiler/blob/c2899341813c5db57b8401356c02c74cc6ec75d5/gprofiler/merge.py#L318-L319

This is too late if the process already exited. With the default profiling duration of 60s, this is common.

* For process profiles (`_profile_process`), we already get the application metadata & appid when the profile begins, so we always get them.
* For perf profiles, we don't have the PIDs that were profiled until we run `perf script`. We can just collect container names for ALL live processes, but then we have a race with PID reuse, not sure how likely is that for a 60s profiling session. If `perf` could output a stronger identifier, e.g PID+start-time, it'd help.

PR #667, opened 2023-01-30
Move getting container name to _profile_process function, expand ProfileData with container_name property

mpozniak95:
## Description
Get container name for profiled process in the same function as we get appid and metadata (_profile_process).

## Related Issue
https://github.com/Granulate/gprofiler/issues/640

### DIFF
--- a/gprofiler/containers_client.py
+++ b/gprofiler/containers_client.py
@@ -82,7 +82,7 @@ class ContainerNamesClient:
                 self._current_container_names.add(container_name)
             return container_name
 
-        self.refresh_container_names_cache()
+        self._refresh_container_names_cache()
         if container_id not in self._container_id_to_name_cache:
             self._container_id_to_name_cache[container_id] = None
             return None
@@ -91,11 +91,8 @@ class ContainerNamesClient:
             self._current_container_names.add(container_name)
         return container_name
 
-    def refresh_container_names_cache(self) -> None:
+    def _refresh_container_names_cache(self) -> None:
         # We re-fetch all of the currently running containers, so in order to keep the cache small we clear it
         self._container_id_to_name_cache.clear()
-        self._pid_to_container_name_cache.clear()
         for container in self._containers_client.list_containers() if self._containers_client is not None else []:
             self._container_id_to_name_cache[container.id] = container.name
-            if container.pid is not None:
-                self._pid_to_container_name_cache[container.pid] = container.name
--- a/gprofiler/containers_client.py
+++ b/gprofiler/containers_client.py
@@ -41,17 +41,27 @@ class ContainerNamesClient:
     def container_names(self) -> List[str]:
         return list(self._current_container_names)
 
-    def get_container_name(self, pid: int) -> str:
+    def get_container_name(self, pid: int, profiling_start_time: float) -> str:
         if self._containers_client is None:
             return ""
 
         if not valid_perf_pid(pid):
             return ""
 
-        if pid in self._pid_to_container_name_cache:
-            name = self._pid_to_container_name_cache[pid]
-            self._current_container_names.add(name)
-            return name
+        try:
+            process_create_time = Process(pid).create_time()
+            process_exists = True
+        except NoSuchProcess:
+            process_create_time = 0
+            process_exists = False
+
+        # If process started after profiling start time then don't take container_name 
+        # from cache to avoid wrong container name that might be caused by PID reuse
+        if not process_exists or process_create_time < profiling_start_time:
+            if pid in self._pid_to_container_name_cache:
+                name = self._pid_to_container_name_cache[pid]
+                self._current_container_names.add(name)
+                return name
 
         container_name: Optional[str] = self._safely_get_process_container_name(pid)
         if container_name is None:
--- a/gprofiler/containers_client.py
+++ b/gprofiler/containers_client.py
@@ -41,27 +41,15 @@ class ContainerNamesClient:
     def container_names(self) -> List[str]:
         return list(self._current_container_names)
 
-    def get_container_name(self, pid: int, profiling_start_time: float) -> str:
+    def get_container_name(self, pid: int) -> str:
         if self._containers_client is None:
             return ""
 
         if not valid_perf_pid(pid):
             return ""
 
-        try:
-            process_create_time = Process(pid).create_time()
-            process_exists = True
-        except NoSuchProcess:
-            process_create_time = 0
-            process_exists = False
-
-        # If process started after profiling start time then don't take container_name
-        # from cache to avoid wrong container name that might be caused by PID reuse
-        if not process_exists or process_create_time < profiling_start_time:
-            if pid in self._pid_to_container_name_cache:
-                name = self._pid_to_container_name_cache[pid]
-                self._current_container_names.add(name)
-                return name
+        if pid in self._pid_to_container_name_cache:
+            return self._pid_to_container_name_cache[pid]
 
         container_name: Optional[str] = self._safely_get_process_container_name(pid)
         if container_name is None:
--- a/gprofiler/containers_client.py
+++ b/gprofiler/containers_client.py
@@ -80,7 +80,7 @@ class ContainerNamesClient:
                 self._current_container_names.add(container_name)
             return container_name
 
-        self._refresh_container_names_cache()
+        self.refresh_container_names_cache()
         if container_id not in self._container_id_to_name_cache:
             self._container_id_to_name_cache[container_id] = None
             return None
@@ -89,8 +89,9 @@ class ContainerNamesClient:
             self._current_container_names.add(container_name)
         return container_name
 
-    def _refresh_container_names_cache(self) -> None:
+    def refresh_container_names_cache(self) -> None:
         # We re-fetch all of the currently running containers, so in order to keep the cache small we clear it
         self._container_id_to_name_cache.clear()
         for container in self._containers_client.list_containers() if self._containers_client is not None else []:
             self._container_id_to_name_cache[container.id] = container.name
+            self._pid_to_container_name_cache[container.pid] = container.name
--- a/gprofiler/containers_client.py
+++ b/gprofiler/containers_client.py
@@ -94,4 +94,5 @@ class ContainerNamesClient:
         self._container_id_to_name_cache.clear()
         for container in self._containers_client.list_containers() if self._containers_client is not None else []:
             self._container_id_to_name_cache[container.id] = container.name
-            self._pid_to_container_name_cache[container.pid] = container.name
+            if container.pid:
+                self._pid_to_container_name_cache[container.pid] = container.name
--- a/gprofiler/containers_client.py
+++ b/gprofiler/containers_client.py
@@ -49,6 +49,7 @@ class ContainerNamesClient:
             return ""
 
         if pid in self._pid_to_container_name_cache:
+            self._current_container_names.add(self._pid_to_container_name_cache[pid])
             return self._pid_to_container_name_cache[pid]
 
         container_name: Optional[str] = self._safely_get_process_container_name(pid)
@@ -92,6 +93,7 @@ class ContainerNamesClient:
     def refresh_container_names_cache(self) -> None:
         # We re-fetch all of the currently running containers, so in order to keep the cache small we clear it
         self._container_id_to_name_cache.clear()
+        self._pid_to_container_name_cache.clear()
         for container in self._containers_client.list_containers() if self._containers_client is not None else []:
             self._container_id_to_name_cache[container.id] = container.name
             if container.pid:
--- a/gprofiler/containers_client.py
+++ b/gprofiler/containers_client.py
@@ -49,8 +49,9 @@ class ContainerNamesClient:
             return ""
 
         if pid in self._pid_to_container_name_cache:
-            self._current_container_names.add(self._pid_to_container_name_cache[pid])
-            return self._pid_to_container_name_cache[pid]
+            name = self._pid_to_container_name_cache[pid]
+            self._current_container_names.add(name)
+            return name
 
         container_name: Optional[str] = self._safely_get_process_container_name(pid)
         if container_name is None:
@@ -96,5 +97,5 @@ class ContainerNamesClient:
         self._pid_to_container_name_cache.clear()
         for container in self._containers_client.list_containers() if self._containers_client is not None else []:
             self._container_id_to_name_cache[container.id] = container.name
-            if container.pid:
+            if container.pid is not None:
                 self._pid_to_container_name_cache[container.pid] = container.name
--- a/gprofiler/containers_client.py
+++ b/gprofiler/containers_client.py
@@ -55,7 +55,7 @@ class ContainerNamesClient:
             process_create_time = 0
             process_exists = False
 
-        # If process started after profiling start time then don't take container_name 
+        # If process started after profiling start time then don't take container_name
         # from cache to avoid wrong container name that might be caused by PID reuse
         if not process_exists or process_create_time < profiling_start_time:
             if pid in self._pid_to_container_name_cache:
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -283,6 +283,7 @@ class GProfiler:
     def _snapshot(self) -> None:
         local_start_time = datetime.datetime.utcnow()
         monotonic_start_time = time.monotonic()
+        self._profiler_state.start_time = monotonic_start_time
         process_profilers_futures = []
         for prof in self.process_profilers:
             prof_future = self._executor.submit(prof.snapshot)
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -281,7 +281,6 @@ class GProfiler:
             prof.stop()
 
     def _snapshot(self) -> None:
-        self._profiler_state.refresh_container_cache()
         local_start_time = datetime.datetime.utcnow()
         monotonic_start_time = time.monotonic()
         process_profilers_futures = []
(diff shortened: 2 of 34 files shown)


gprofiler/profilers/php.py
```diff
@@ -66,9 +67,12 @@ def __init__(
         profile_spawned_processes: bool,
         php_process_filter: str,
         php_mode: str,
+        container_names_client: Optional[ContainerNamesClient],
     ):
         assert php_mode == "phpspy", "PHP profiler should not be initialized, wrong php_mode value given"
-        super().__init__(frequency, duration, stop_event, storage_dir, insert_dso_name, profiling_mode)
+        super().__init__(
+            frequency, duration, stop_event, storage_dir, insert_dso_name, profiling_mode, container_names_client
+        )
```
Jongy, line 75: I don't like it much that we have to make this change in all profilers.

Logically - there are profiler parameters which are truly parameters per profiler and might change per profiler - `frequency`, `duration`. All others (and specifically `stop_event`, `storage_dir`, `containers_name_client`) are actually passed because we cannot pass "general state parameters" in another way.

Let's refactor it (in a different PR) and then base this PR on that refactor.

* One option - profilers can get the `GProfiler` object and access the settings configured on it via accessors `.get_storage_dir()` etc.
* We can create another object `ProfilerState`, create it in `GProfiler` and pass it in the c'tor

My point is, adding a new "state parameter" like `containers_name_client` should not induce diff in all classes of all c'tors.

Let's open another PR for this refactor please

gprofiler/profilers/python.py
```diff
@@ -215,6 +224,10 @@ def _profile_process(self, process: Process, duration: int, spawned: bool) -> Pr
         )
         appid = application_identifiers.get_python_app_id(process)
         app_metadata = self._metadata.get_metadata(process)
+        if self._container_names_client:
+            container_name = self._container_names_client.get_container_name(process.pid)
+        else:
+            container_name = ""
```
Jongy, line 230: This logic repeats itself - let's encapsulate to a helper. This helper can be a part of the refactor I suggested in https://github.com/Granulate/gprofiler/pull/667/files#r1091286777 - the `ProfilerState`, or `GProfiler`, can expose a method `get_container_name(pid: int)` which gets you the name or `""`.

gprofiler/merge.py
```diff
@@ -311,13 +315,10 @@ def _enrich_pid_stacks(
         application_prefix = f"{idx};"
     else:
         application_prefix = ""
-
```
Jongy, line 314: Bring it back :) Refrain from unrelated formatting changes

Jongy (commented):
Overall looks good. I gave some comments about the design

Jongy:
Please merge from master after https://github.com/Granulate/gprofiler/pull/668, then continue this PR

gprofiler/main.py
```diff
@@ -267,6 +268,7 @@ def stop(self) -> None:
             prof.stop()
 
     def _snapshot(self) -> None:
+        self._container_names_client.refresh_container_names_cache()
```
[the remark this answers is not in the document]
mpozniak95: I think that we need to refresh container cache at the beginning of profiling. If profiled process will be terminated during profiling we won't be able to get his container name because there might be a situation when it will happen before first call of `get_container_name()` and there won't be any container_pid's in cache. I added a test to test_sanity.py to simulate such case.

gprofiler/containers_client.py
```diff
@@ -49,6 +49,7 @@ def get_container_name(self, pid: int) -> str:
             return ""
 
         if pid in self._pid_to_container_name_cache:
+            self._current_container_names.add(self._pid_to_container_name_cache[pid])
             return self._pid_to_container_name_cache[pid]
```
Jongy, line 53:
```suggestion
            name = self._pid_to_container_name_cache[pid]
            self._current_container_names.add(name)
            return name
```

nicer w/o accessing it twice

gprofiler/profiler_state.py
```diff
@@ -40,3 +44,17 @@ def insert_dso_name(self) -> bool:
...
+
+    def refresh_container_cache(self) -> None:
+        if self._container_names_client:
+            self._container_names_client.refresh_container_names_cache()
+
+    def get_container_name(self, pid: int) -> str:
+        if self._container_names_client:
+            return self._container_names_client.get_container_name(pid)
+        else:
+            return ""
```
Jongy, line 60: Any particular reason you moved these wrappers into this class? We only use it to hold the reference - no need to clone methods into it.

if you did it as a way for code sharing (a wrapper, non-None class for the containers client) then it's okay

gprofiler/profiler_state.py
```diff
@@ -40,3 +44,17 @@ def insert_dso_name(self) -> bool:
     @property
     def profiling_mode(self) -> str:
         return self._profiling_mode
+
+    @property
+    def container_names_client(self) -> Optional[ContainerNamesClient]:
+        return self._container_names_client
+
+    def refresh_container_cache(self) -> None:
+        if self._container_names_client:
```
Jongy, line 53:
```suggestion
        if self._container_names_client is not None:
```

Please fix everywhere - no implicit `None`/bool conversions

gprofiler/profilers/perf.py
```diff
@@ -354,7 +354,8 @@ def _generate_profile_data(self, stacks: StackToSampleCount, pid: int) -> Profil
             appid = self._get_appid(pid)
         else:
             appid = None
-        return ProfileData(stacks, appid, metadata)
+        container_name = self._profiler_state.get_container_name(pid)
+        return ProfileData(stacks, appid, metadata, container_name)
```
Jongy, line 358:
```suggestion
        return ProfileData(stacks, appid, metadata, self._profiler_state.get_container_name(pid))
```

No need for the extra variable, if it's only used once

gprofiler/profilers/java.py
```diff
@@ -789,7 +789,7 @@ def _disable_profiling(self, cause: str) -> None:
             self._safemode_disable_reason = cause
 
     def _profiling_skipped_profile(self, reason: str, comm: str) -> ProfileData:
-        return ProfileData(self._profiling_error_stack("skipped", reason, comm), None, None)
+        return ProfileData(self._profiling_error_stack("skipped", reason, comm), None, None, None)
```
Jongy, line 792: This change (generally wherever you pass `None` as container name for a process that we don't profile due to skip / error, but the process is still alive) - because past code would have added the container name in `merge.py`.

Now, you add the name in `perf.py` and in e.g `java.py` for a java process. Then during merging, we select the java `ProfileData`. If the java was `Profiling skipped`, it will have no container name, but the perf one will have!

I suggest: if the process `ProfileData` has no name, but the perf one has, take the perf. This keeps the past behavior for processes we don't profile with a runtime profiler, and gives your enhanced behavior for processes that we do profile with e.g `JavaProfiler`.

tests/test_sanity.py
```diff
@@ -241,3 +243,33 @@ def test_from_container_spawned_process(
         assert_collapsed(collapsed)
 
     assert b"Profiling spawned process" in container.logs()
+
+
+@pytest.mark.parametrize("in_container", [True])
+@pytest.mark.parametrize("profiler_type", ["attach-maps"])
+@pytest.mark.parametrize("runtime", ["nodejs"])
+@pytest.mark.parametrize("application_image_tag", ["without-flags"])
+def test_container_name_when_stopped(
```
Jongy, line 252: Nice test! Good job coming up with it :)

To improve - let's make it non-racey, currently it's racey and might "pass" despite the app not actually going down before the profiler did. Try to write tests that are self-guarding, i.e if you have any assumptions on the flow, then you can guard them.

One way I suggest to do it is to profile a Java app instead of nodejs. For Java, we have logs that indicate a process went down during profiling - `logger.debug(f"Profiled process {ap_proc.process.pid} exited before stopping async-profiler")`.
Your test can run the Java, begin profiling, kill the Java and ensure that this log message appears.
Also, as a best practice let's use `assert_collapsed` here to make sure the profiling of the app really worked.

tests/test_sanity.py
```diff
@@ -241,3 +243,33 @@ def test_from_container_spawned_process(
...
+    profiler_type: str,
+    profiler_flags: List[str],
+    application_docker_container: Container,
+) -> None:
+    profiler_flags.extend(["-d", "20"])
+    container = start_gprofiler_in_container_for_one_session(
+        docker_client, gprofiler_docker_image, output_directory, output_collapsed, runtime_specific_args, profiler_flags
+    )
+    try:
+        wait_event(20, Event(), lambda: re.search(rb"Started perf", container.logs()) is not None)
```
Jongy, line 267: After you cange to Java you can fix this log searching for the log message of profiling a java process.

tests/test_sanity.py
Jongy, line 252: I'd add some docstring explaining the test and linking to the GH ticket. Something short like here:
```
@pytest.mark.parametrize("in_container", [True])
def test_java_jattach_async_profiler_log_output(
    application_pid: int,
    caplog: LogCaptureFixture,
    profiler_state: ProfilerState,
) -> None:
    """
    Tests that AP log is collected and logged in gProfiler's log.
    """
    ...
```

gprofiler/merge.py
```diff
@@ -297,6 +285,11 @@ def _enrich_pid_stacks(
     """
     # generate application name
     appid = profile.appid
+    if profile.container_name:
+        container_prefix = profile.container_name
+    else:
+        container_prefix = ""
+    container_prefix = f"{container_prefix};"
```
Jongy, line 292: Can extract this to a function

gprofiler/main.py
```diff
@@ -142,18 +142,19 @@ def __init__(
         # 2. accessible only by us.
         # the latter can be root only. the former can not. we should do this separation so we don't expose
         # files unnecessarily.
+        if self._enrichment_options.container_names:
+            container_names_client: Optional[ContainerNamesClient] = ContainerNamesClient()
+        else:
+            container_names_client = None
```
Jongy, line 148: Can be a short `ContainerNamesClient() if ... else None` in the call to the ProfilerState c'tor

gprofiler/profiler_state.py
Jongy: Haha I see that it was my own suggestion :sweat_smile: 

>  the ProfilerState, or GProfiler, can expose a method get_container_name(pid: int) which gets you the name or "".

I'm trying to avoid from making ProfilerState a god object. But alright let's go with it here, it's useful.

gprofiler/containers_client.py
```diff
@@ -89,8 +90,11 @@ def _get_container_name(self, container_id: str) -> Optional[str]:
...
         return container_name
 
-    def _refresh_container_names_cache(self) -> None:
+    def refresh_container_names_cache(self) -> None:
         # We re-fetch all of the currently running containers, so in order to keep the cache small we clear it
         self._container_id_to_name_cache.clear()
+        self._pid_to_container_name_cache.clear()
         for container in self._containers_client.list_containers() if self._containers_client is not None else []:
             self._container_id_to_name_cache[container.id] = container.name
+            if container.pid:
```
Jongy, line 99: is not None?

gprofiler/main.py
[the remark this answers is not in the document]
Jongy: I see. Something is still off here for me, I will re-review thoroughly after you fix the other points I commented now

Jongy (changes requested):
Done reviewing, looks very good :)

gprofiler/main.py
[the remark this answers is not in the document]
mpozniak95: This is not needed, I deleted this and check for every process at the beginning of perf snapshot

gprofiler/profilers/dotnet.py
```diff
@@ -93,6 +93,7 @@ def _make_command(self, process: Process, duration: int, output_path: str) -> Li
         ]
 
     def _profile_process(self, process: Process, duration: int, spawned: bool) -> ProfileData:
+        container_name = self._profiler_state.get_container_name(process.pid)
```
Jongy, line 96: Nice next to the app_metadata

Jongy: (in all profilers code, to keep it organized)

gprofiler/profilers/java.py
```diff
@@ -902,6 +902,7 @@ def _check_async_profiler_loaded(self, process: Process) -> bool:
         return False
 
     def _profile_process(self, process: Process, duration: int, spawned: bool) -> ProfileData:
+        container_name = self._profiler_state.get_container_name(process.pid)
```
Jongy, line 905: Can move next to app_metadata and appid as suggested https://github.com/Granulate/gprofiler/pull/667/files#r1116345068

gprofiler/profilers/java.py
```diff
@@ -957,8 +958,7 @@ def _profile_process(self, process: Process, duration: int, spawned: bool) -> Pr
             self._report_meminfo,
         ) as ap_proc:
             stackcollapse = self._profile_ap_process(ap_proc, comm, duration)
-
```
Jongy, line 960: Undo this change (no unrelated formatting changes)

gprofiler/profilers/python.py
```diff
@@ -202,6 +202,7 @@ def _make_command(self, pid: int, output_path: str, duration: int) -> List[str]:
         return command
 
     def _profile_process(self, process: Process, duration: int, spawned: bool) -> ProfileData:
+        container_name = self._profiler_state.get_container_name(process.pid)
```
Jongy, line 205: Next to appid & app_metadata https://github.com/Granulate/gprofiler/pull/667/files#r1116345068

gprofiler/profilers/ruby.py
```diff
@@ -99,6 +99,7 @@ def _make_command(self, pid: int, output_path: str, duration: int) -> List[str]:
         ]
 
     def _profile_process(self, process: Process, duration: int, spawned: bool) -> ProfileData:
+        container_name = self._profiler_state.get_container_name(process.pid)
```
Jongy, line 102: Next to appid and app metadata https://github.com/Granulate/gprofiler/pull/667/files#r1116345068

tests/test_sanity.py
```diff
@@ -241,3 +243,42 @@ def test_from_container_spawned_process(
...
+    """
+    profiler_flags.extend(["-d", "20"])
+    container = start_gprofiler_in_container_for_one_session(
+        docker_client, gprofiler_docker_image, output_directory, output_collapsed, runtime_specific_args, profiler_flags
+    )
+    try:
+        wait_event(
+            20,
+            Event(),
+            lambda: re.search(rb"Profiling .*process \d* with async-profiler", container.logs()) is not None,
```
Jongy, line 274: I know it was my suggestion but I realized it makes no sense to use a Java app here - once the container is dead, async-profiler dies too, as it lives in the Java app... so how can we read its output (it writes the output in the container, but it's dead)? lol. Did you get this test to pass? Because I think it should not.

Another suggestion - let's use a Python app with `profiler_type=py-spy` - it prints an indicative message upon process exit, that is dumped into the profiler log:
```
py-spy> Stopped sampling because process exited
```

You can kill the app, then search for this message.

As a side note, to ensure this test covers exactly the case we want to test, you can build gProfiler from master and run it - it should fail on master and pass on your branch.

Jongy (changes requested):
Done reviewing, one comment about the test and a few other small

tests/test_sanity.py
mpozniak95: Sure I will change it to py-spy. I was able to pass both java and nodeJs test on this PR and I tested both on master and both failed :)

mpozniak95: Changed, Also works on this PR fails on master

gprofiler/profilers/java.py
```diff
@@ -938,8 +941,6 @@ def _profile_process(self, process: Process, duration: int, spawned: bool) -> Pr
             self._profiled_pids.add(process.pid)
 
         logger.info(f"Profiling{' spawned' if spawned else ''} process {process.pid} with async-profiler")
-        app_metadata = self._metadata.get_metadata(process)
```
Jongy, line 941: This was the correct place for app metadata & appid retrieval, I don't want them to be retrieved before the safety checks are done (async-profiler loaded, JVM profiling supported etc). Bring them back and move the container name here as well?

tests/test_sanity.py
Jongy: Awesome

Jongy (changes requested):
One final comment.

I went over existing tests and noticed we don't currently have any tests about container names. You added one here for runtime profilers, but we changed the way container names are added to perf-profiled processes, and that's not tested yet. Can you add testing of container names to a perf test, e.g to `test_perf_fp_dwarf_smart`? You can get the name the same way with `application_docker_container` and ensure it's present in the stack trace (you can also create a helper function `assert_container_names` to be used in both tests)

mpozniak95:
Since we are enriching stack during merging profiles in `main.py` and in `test_perf_fp_dwarf_smart` we use only SystemProfiler object, I added checking if container_name is in process ProfileData instead of checking if it is in stack trace. Is this enough?

tests/test_perf.py
```diff
@@ -67,6 +68,8 @@ def test_perf_fp_dwarf_smart(
         assert is_function_in_collapsed(";_start;__libc_start_main;main;", process_collapsed) ^ bool(
             perf_mode not in ("dwarf", "smart")
         )
+        # Check if container name is added to ProfileData
+        assert application_docker_container.name == snapshot_pid_profile(profiler, application_pid).container_name
```
Jongy, line 72: Avoid calling `snapshot` again - use `snapshot_pid_profiler` in line 54, then compare both the container name & the collapsed of it. Calling `snapshot` again invokes the profiler twice, unnecessarily.

Jongy:
One comment :)

Jongy (approved):

PR #668, opened 2023-01-31
Refactor profiler state parameters to separate object

mpozniak95:
Refactor code to make any future changes in profilers parameters easier. Created another object that keeps "state parameters" for profilers.
As mentioned in https://github.com/Granulate/gprofiler/pull/667,

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -139,7 +139,7 @@ class GProfiler:
             Event(),
             TemporaryDirectoryWithMode(dir=TEMPORARY_STORAGE_PATH, mode=0o755),
             profile_spawned_processes,
-            user_args.get("insert_dso_name"),
+            bool(user_args.get("insert_dso_name")),
             profiling_mode,
         )
         self.system_profiler, self.process_profilers = get_profilers(user_args, profiler_state=self._profiler_state)
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -117,7 +117,6 @@ class GProfiler:
         self._state = state
         self._remote_logs_handler = remote_logs_handler
         self._profile_api_version = profile_api_version
-        self._profile_spawned_processes = profile_spawned_processes
         self._collect_metrics = collect_metrics
         self._collect_metadata = collect_metadata
         self._enrichment_options = enrichment_options
@@ -128,7 +127,6 @@ class GProfiler:
         self._gpid = ""
         self._controller_process = controller_process
         self._duration = duration
-        self._profiling_mode = profiling_mode
         if collect_metadata:
             self._static_metadata = get_static_metadata(spawn_time=self._spawn_time, run_args=user_args)
         self._executor = concurrent.futures.ThreadPoolExecutor(max_workers=10)
@@ -137,9 +135,13 @@ class GProfiler:
         # 2. accessible only by us.
         # the latter can be root only. the former can not. we should do this separation so we don't expose
         # files unnecessarily.
-        self._temp_storage_dir = TemporaryDirectoryWithMode(dir=TEMPORARY_STORAGE_PATH, mode=0o755)
+        temp_dir = TemporaryDirectoryWithMode(dir=TEMPORARY_STORAGE_PATH, mode=0o755)
         self._profiler_state = ProfilerState(
-            self._stop_event, self._temp_storage_dir.name, self._profile_spawned_processes
+            Event(),
+            temp_dir.name,
+            profile_spawned_processes,
+            user_args.get("insert_dso_name"),
+            profiling_mode,
         )
         self.system_profiler, self.process_profilers = get_profilers(user_args, profiler_state=self._profiler_state)
         if self._enrichment_options.container_names:
@@ -300,7 +302,7 @@ class GProfiler:
             if self._collect_metadata
             else {"hostname": get_hostname()}
         )
-        metadata.update({"profiling_mode": self._profiling_mode})
+        metadata.update({"profiling_mode": self._profiler_state.profiling_mode})
         metrics = self._system_metrics_monitor.get_metrics()
         if NoopProfiler.is_noop_profiler(self.system_profiler):
             assert system_result == {}, system_result  # should be empty!
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -120,7 +120,6 @@ class GProfiler:
         self._collect_metrics = collect_metrics
         self._collect_metadata = collect_metadata
         self._enrichment_options = enrichment_options
-        self._stop_event = Event()
         self._static_metadata: Optional[Metadata] = None
         self._spawn_time = time.time()
         self._last_diagnostics = 0.0
@@ -137,7 +136,7 @@ class GProfiler:
         # files unnecessarily.
         self._profiler_state = ProfilerState(
             Event(),
-            TemporaryDirectoryWithMode(dir=TEMPORARY_STORAGE_PATH, mode=0o755),
+            TEMPORARY_STORAGE_PATH,
             profile_spawned_processes,
             bool(user_args.get("insert_dso_name")),
             profiling_mode,
@@ -149,7 +148,7 @@ class GProfiler:
             self._container_names_client = None
         self._usage_logger = usage_logger
         if collect_metrics:
-            self._system_metrics_monitor: SystemMetricsMonitorBase = SystemMetricsMonitor(self._stop_event)
+            self._system_metrics_monitor: SystemMetricsMonitorBase = SystemMetricsMonitor(self._profiler_state.stop_event)
         else:
             self._system_metrics_monitor = NoopSystemMetricsMonitor()
 
@@ -214,7 +213,7 @@ class GProfiler:
                         [resource_path("burn"), "convert", "--type=folded"],
                         suppress_log=True,
                         stdin=stripped_collapsed_data.encode(),
-                        stop_event=self._stop_event,
+                        stop_event=self._profiler_state.stop_event,
                         timeout=10,
                     ).stdout,
                 )
@@ -242,7 +241,7 @@ class GProfiler:
         return "\n".join(lines)
 
     def start(self) -> None:
-        self._stop_event.clear()
+        self._profiler_state.stop_event.clear()
         self._system_metrics_monitor.start()
 
         for prof in list(self.all_profilers):
@@ -260,7 +259,7 @@ class GProfiler:
 
     def stop(self) -> None:
         logger.info("Stopping ...")
-        self._stop_event.set()
+        self._profiler_state.stop_event.set()
         self._system_metrics_monitor.stop()
 
         for prof in self.all_profilers:
@@ -352,7 +351,7 @@ class GProfiler:
         with self:
             self._usage_logger.init_cycles()
 
-            while not self._stop_event.is_set():
+            while not self._profiler_state.stop_event.is_set():
                 self._state.init_new_cycle()
 
                 snapshot_start = time.monotonic()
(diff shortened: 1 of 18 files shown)


gprofiler/profiler_state.py
```diff
@@ -0,0 +1,13 @@
+from threading import Event
+
+
+class ProfilerState:
```
Jongy, line 4: Cool. Let's add documentation explaining (you can follow my comment where I first suggested it).

gprofiler/profiler_state.py
```diff
@@ -0,0 +1,13 @@
...
+    def __init__(
+        self,
+        stop_event: Event,
+        storage_dir: str,
+        profile_spawned_processes: bool,
```
Jongy, line 9: Other fields I think can be added here (because they are generic for all) -
* `insert_dso_name`
* `profiling_mode`

gprofiler/main.py
```diff
@@ -137,12 +138,10 @@ def __init__(
         # the latter can be root only. the former can not. we should do this separation so we don't expose
         # files unnecessarily.
         self._temp_storage_dir = TemporaryDirectoryWithMode(dir=TEMPORARY_STORAGE_PATH, mode=0o755)
-        self.system_profiler, self.process_profilers = get_profilers(
-            user_args,
-            storage_dir=self._temp_storage_dir.name,
-            stop_event=self._stop_event,
-            profile_spawned_processes=self._profile_spawned_processes,
+        self._profiler_state = ProfilerState(
+            self._stop_event, self._temp_storage_dir.name, self._profile_spawned_processes
```
Jongy, line 142: It's nicer if now `GProfiler` uses the fields from `_profiler_state`, i.e - don't store `_stop_event`, `_temp_storage_dir`, `_profile_spawned_processes` on `GProfiler` - but only on the `State` class.

You can pass the `TemporaryDirectoryWithMode(dir=TEMPORARY_STORAGE_PATH, mode=0o755)` as a parameter to the state class, the expose the storage dir with a `@property`?

gprofiler/profilers/java.py
```diff
@@ -321,7 +321,8 @@ def __init__(
         collect_meminfo: bool = True,
     ):
         self.process = process
-        self._stop_event = stop_event
+        self._profiler_state = profiler_state
+        self._stop_event = self._profiler_state.stop_event
```
Jongy, line 325: Don't store it on the class - access directly from the `_profiler_state.stop_event`.

gprofiler/profilers/profiler_base.py
```diff
@@ -207,14 +208,12 @@ def __init__(
...
-        storage_dir: str,
+        profiler_state: ProfilerState,
         insert_dso_name: bool,
-        profile_spawned_processes: bool,
         profiling_mode: str,
     ):
-        super().__init__(frequency, duration, stop_event, storage_dir, insert_dso_name, profiling_mode)
-        self._profile_spawned_processes = profile_spawned_processes
+        super().__init__(frequency, duration, profiler_state, insert_dso_name, profiling_mode)
+        self._profile_spawned_processes = profiler_state.profile_spawned_processes
```
Jongy, line 216: Don't store it - access we store the `_profiler_state` as an instance attribute and access it via that attr.

tests/test_java.py
```diff
@@ -327,11 +326,11 @@ def test_async_profiler_stops_after_given_timeout(
 
     process = psutil.Process(application_pid)
     timeout_s = 5
+    profiler_state = ProfilerState(Event(), str(tmp_path_world_accessible), False)
```
Jongy, line 329: Let's add a fixture that creates this, to be used by various tests?

gprofiler/main.py
mpozniak95: I did it a bit differently. Since we pass string in tests ProfilerState constructor should accept string as storage_dir too.

gprofiler/main.py
```diff
@@ -136,13 +135,14 @@ def __init__(
...
         # the latter can be root only. the former can not. we should do this separation so we don't expose
         # files unnecessarily.
-        self._temp_storage_dir = TemporaryDirectoryWithMode(dir=TEMPORARY_STORAGE_PATH, mode=0o755)
-        self.system_profiler, self.process_profilers = get_profilers(
-            user_args,
-            storage_dir=self._temp_storage_dir.name,
-            stop_event=self._stop_event,
-            profile_spawned_processes=self._profile_spawned_processes,
+        self._profiler_state = ProfilerState(
+            Event(),
```
Jongy, line 139: Note that there is still a `GProfiler._stop_event` initialized and it's initialized to a DIFFERENT `Event()` object. Please use the same object as well - i.e, remove `self._stop_event` and use `self._profiler_state.stop_event`.

gprofiler/profiler_state.py
```diff
@@ -0,0 +1,26 @@
...
+from gprofiler.utils import TemporaryDirectoryWithMode
+
+
+class ProfilerState:
+    # Class for storing generic state parameters. These parameters are the same for each profiler.
+    # Thanks to that class adding new state parameters to profilers won't result changing code in every profiler.
+    def __init__(
+        self,
+        stop_event: Event,
+        storage_dir: Union[TemporaryDirectoryWithMode, str],
```
Jongy, line 13: Why support both cases? Let's support one and be consistent across users of this class (which are either `GProfiler` or the tests code)

gprofiler/profiler_state.py
```diff
@@ -0,0 +1,26 @@
...
+    ) -> None:
+        self.stop_event = stop_event
+        self.profile_spawned_processes = profile_spawned_processes
+        self.insert_dso_name = insert_dso_name
+        self.profiling_mode = profiling_mode
+        if type(storage_dir) == TemporaryDirectoryWithMode:
+            self._temporary_dir = storage_dir
+            self.storage_dir = storage_dir.name
+        else:
+            self.storage_dir = storage_dir
```
Jongy, line 26: These fields can be made private fields, and we expose them with read-only properties. We don't want them to change after the object is constructed by an accident.

gprofiler/profilers/profiler_base.py
```diff
@@ -75,32 +76,28 @@ def __init__(
...
             raise ValueError(
                 f"Minimum duration for {self.__class__.__name__} is {self.MIN_DURATION} (given {duration}), "
                 "raise the duration in order to use this profiler"
             )
         self._duration = duration
-        self._stop_event = stop_event or Event()
-        self._storage_dir = storage_dir
-        self._profiling_mode = profiling_mode
+        self._profiler_state = profiler_state
+        self._stop_event = self._profiler_state.stop_event or Event()
```
Jongy, line 91: Btw - is the `or Event()` actually needed? I think it's legacy code. You can try removing it in one commit and seeing if tests pass. If so, I'm in favor of removing it.

gprofiler/profilers/python.py
```diff
@@ -336,11 +330,7 @@ def __init__(
         self,
         frequency: int,
         duration: int,
-        stop_event: Event,
-        storage_dir: str,
-        insert_dso_name: bool,
-        profiling_mode: str,
-        profile_spawned_processes: bool,
+        profiler_state: ProfilerState,
```
Jongy, line 333: This is so beautiful :nail_care: so much boilerplate code removed

gprofiler/main.py
Jongy: Hmm I commented on that later - https://github.com/Granulate/gprofiler/pull/668#discussion_r1096231221 - it is not a good idea to make the tested code more complex to acommodate tests, instead of makes more sense to make the tests code a bit more complex. In that case, it's just calling the `TemporaryStorageDirectory` on the base path `tmp_path` from pytest, right? I think it's a better solution.

Alternatively - you can move the creation of `TemporaryStorageDirectory` inside `ProfilerState`, then both the tests and `GProfiler` will pass it a `str`. Up to you.

tests/conftest.py
```diff
@@ -76,6 +79,11 @@ def java_args() -> Tuple[str]:
     return cast(Tuple[str], ())
 
 
+@fixture
+def profiler_state(tmp_path_world_accessible: Path) -> ProfilerState:
```
Jongy, line 83: Why `tmp_path_world_accessible`? I see that previously, some tests used just `tmp_path` for this purpose. Or some used `tmp_path` and some used `tmp_path_world_accessible`?

Please explain why you use `tmp_path_world_accessible` here in a comment :+1:

tests/test_perf.py
```diff
@@ -34,14 +35,11 @@ def system_profiler(tmp_path: Path, perf_mode: str, insert_dso_name: bool) -> Sy
 
 
 def make_system_profiler(tmp_path: Path, perf_mode: str, insert_dso_name: bool) -> SystemProfiler:
+    profiler_state = ProfilerState(Event(), str(tmp_path), False, insert_dso_name, CPU_PROFILING_MODE)
```
Jongy, line 38: Code reuse please - you can add a function `make_profiler_state` in the tests, it will have all the default values as your fixture currently has.
The fixture will call the function with no changed arguments.
This test will call it with `insert_dso_name=insert_dso_name`.

tests/test_sanity.py
```diff
@@ -116,8 +113,9 @@ def test_dotnet_trace(
     application_pid: int,
     assert_collapsed: AssertInCollapsed,
     gprofiler_docker_image: Image,
+    profiler_state: ProfilerState,
 ) -> None:
-    with DotnetProfiler(1000, 3, Event(), str(tmp_path), False, False, CPU_PROFILING_MODE, "dotnet-trace") as profiler:
```
Jongy, line 120: Unrelated - but I noticed here that `DotnetProfiler` has the arguments order not in sync with e.g `RbSpyProfiler(1000, 3, Event(), str(tmp_path), False, CPU_PROFILING_MODE, False, "rbspy")`.

Please make sure, after your change now, that the order is synced across all extenders of `ProfilerInterface`.

tests/utils.py
```diff
@@ -206,13 +207,11 @@ def make_java_profiler(
     profiling_mode: str = CPU_PROFILING_MODE,
 ) -> JavaProfiler:
     assert storage_dir is not None
+    profiler_state = ProfilerState(stop_event, storage_dir, profile_spawned_processes, insert_dso_name, profiling_mode)
```
Jongy, line 210: Use the same `make_profiler_state` function that I suggested here - https://github.com/Granulate/gprofiler/pull/668/files#r1096248108

Ideally, the interface of `make_java_profiler` will have `ProfilerState` as well, and not the old arguments, but I assume that'll require a massive change in the tests? Please try it out, if it's a small change then let's fix it here as well to keep a consistent API of `make_..._profiler`. If it's a big change, we can skip, just make sure you use the same `make_profiler_state`.

tests/test_python.py
```diff
@@ -140,10 +139,10 @@ def test_dso_name_in_pyperf_profile(
     profiler_type: str,
     application_image_tag: str,
     insert_dso_name: bool,
+    profiler_state: ProfilerState,
 ) -> None:
-    with PythonProfiler(
-        1000, 2, Event(), str(tmp_path), insert_dso_name, CPU_PROFILING_MODE, False, profiler_type, True, None
-    ) as profiler:
+    profiler_state = ProfilerState(Event(), str(tmp_path), False, insert_dso_name, CPU_PROFILING_MODE)
```
Jongy, line 144: Same as https://github.com/Granulate/gprofiler/pull/668/files#r1096248108 - please reuse `make_profiler_state`.

Jongy (changes requested):
Done reviewing. It looks good. I have some nitpicks :sweat_smile: and we can merge it!

tests/test_sanity.py
mpozniak95: Now every profiler has order: frequency, duration, profiler_state, profiler_specific_parameters

Jongy:
@mpozniak95 can you mark with :rocket: comments that you've addressed + hit "Request review" when it's ready for re-review?

tests/conftest.py
mpozniak95: I used it because some tests in `test_java` used `tmp_path_world_accessible`. I will check if it works with `tmp_path` instead.

mpozniak95: I changed it to `tmp_path` here and just keep requesting `tmp_path_world_accessible` in tests that were using it.

Jongy: Fair enough

Jongy (approved):
LGTM. There are possibly more open things but this is good to go meanwhile.

PR #671, opened 2023-02-04
perf: Remove perf.data & perf.inject intermediates output also when "perf inject/script" fail

Jongy:
Robustness improvement. If those files are leaked due to "perf inject" / "perf script" unexpectedly failing/being killed, then this disk space is not freed.


d3dave (approved):
The logic looks like a classic use-case for a context manager. Python has the `NamedTemporaryFile` class that automatically deletes the file upon `__exit__`. It looks like a good fit. But not important.

Jongy:
> The logic looks like a classic use-case for a context manager. Python has the `NamedTemporaryFile` class that automatically deletes the file upon `__exit__`. It looks like a good fit. But not important.

Good point. I suppose many uses of `remove_path` can be replaced that way. I don't want to deal with it now so I'll open a ticket.

PR #678, opened 2023-02-08
dataproc: collect Spark metrics when spark-metrics attribute is 1

d3dave:
Tested using:

```sh
gsutil cp deploy/dataproc/gprofiler_initialization_action.sh gs://.../gprofiler_initialization_action.sh
export TOKEN='...'
export SERVICE='dataproc'
gcloud dataproc clusters create gprofiler-spark \
   --initialization-actions gs://.../gprofiler_initialization_action.sh \
   --metadata gprofiler-token="$TOKEN",gprofiler-service="$SERVICE",spark-metrics=1,enable-stdout="1" --region us-east1
```

with both `spark-metrics=1` and w/o. Works as expected; with it appends `--collect-spark-metrics` to command, without it does not.

### DIFF
--- a/deploy/dataproc/gprofiler_initialization_action.sh
+++ b/deploy/dataproc/gprofiler_initialization_action.sh
@@ -14,12 +14,20 @@ readonly GPROFILER_SERVICE
 ENABLE_STDOUT=$(/usr/share/google/get_metadata_value attributes/enable-stdout)
 readonly ENABLE_STDOUT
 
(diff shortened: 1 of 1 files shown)


Jongy (approved):
LGTM. @d3dave you tested a run of it w/o passing `spark-metrics`? that's what the `|| true` is for?

d3dave:
> LGTM. @d3dave you tested a run of it w/o passing `spark-metrics`? that's what the `|| true` is for?

yes

PR #681, opened 2023-02-09
Disabled spark_collector applications metrics by default

liorgorb:
Applications metrics is their current form result in too high metrics cardinality. Disabling them for now until we improve it.


Jongy:
Add a flag to control it, or you don't see a use for it anyway now?

Jongy (approved):

PR #689, opened 2023-02-16
spark: Retry finding master address

d3dave:
## Description
If we didn't find the RM/Master address upon launch, retry every cycle until we do.

## How Has This Been Tested?
Ran an EMR cluster with the bootstrap action. Verified data is reported.

### DIFF
--- a/gprofiler/spark/spark_collector.py
+++ b/gprofiler/spark/spark_collector.py
@@ -41,7 +41,7 @@ from gprofiler.utils import get_iso8601_format_time
 from gprofiler.utils.fs import escape_filename
 from gprofiler.utils.process import search_for_process
 
-MAX_FIND_ATTEMPTS = 10
+FIND_CLUSTER_TIMEOUT_SECS = 10 * 60
 
 # Application type and states to collect
 YARN_SPARK_APPLICATION_SPECIFIER = "SPARK"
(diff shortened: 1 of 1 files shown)


gprofiler/spark/spark_collector.py
```diff
@@ -41,6 +41,8 @@
 from gprofiler.utils.fs import escape_filename
 from gprofiler.utils.process import search_for_process
 
+MAX_FIND_ATTEMPTS = 10
```
Jongy, line 44: We want 10 minutes but the interval might be e.g 30s in which case max attempts is 20.

Instead - get the monotonic time when starting and try up to 10 minutes?

gprofiler/spark/spark_collector.py
```diff
@@ -781,42 +784,46 @@ def _find_spark_cluster(self) -> Optional[Tuple[str, str]]:
...
-                self._client.submit_spark_metrics(timestamp, metrics)
+            if self._spark_sampler is None:
+                spark_cluster_conf = self._find_spark_cluster()
+                if spark_cluster_conf is not None:
+                    master_address, cluster_mode = spark_cluster_conf
+                    self._spark_sampler = SparkCollector(cluster_mode, master_address)
+                else:
+                    self._find_attempts += 1
+                    if self._find_attempts == MAX_FIND_ATTEMPTS:
+                        logger.info("Did not identify Spark cluster. Stopping Spark collector.")
```
Jongy, line 805:
```suggestion
                        logger.info("Did not identify Spark cluster after max retry attempts. Stopping Spark collector.")
```

Jongy (changes requested):
Done reviewing

Jongy (approved):

Issue #697, opened 2023-02-20
dotnet: test_executable_not_privileged failure


Jongy:
Example run: https://github.com/Granulate/gprofiler/actions/runs/4217195658/jobs/7321069929

Failure:
```
E       Provider Name                           Keywords            Level               Enabled By
E       Microsoft-DotNETCore-SampleProfiler     0x0000F00000000000  Informational(4)    --profile 
E       Microsoft-Windows-DotNETRuntime         0x00000014C14FCCBD  Informational(4)    --profile 
E       
E       Process        : Process1
E       Output File    : /tmp/gprofiler_tmp/tmp3o1y169p/dotnet-trace-hkYnZMwctShckhRe-1
E       Trace Duration : 00:00:01:00
E       
E       
E       
E       Stopping the trace. This may take several minutes depending on the application being traced.
E       
E       Trace completed.
E       Writing:	/tmp/gprofiler_tmp/tmp3o1y169p/dotnet-trace-hkYnZMwctShckhRe-1.speedscope.json
E       System.IO.FileNotFoundException: Could not load file or assembly 'System.Net.Http, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
E       
E       File name: 'System.Net.Http, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
E          at Microsoft.Diagnostics.Tools.Trace.TraceFileFormatConverter.Convert(TraceFileFormat format, String fileToConvert, String outputFilename, Boolean continueOnError)
E          at Microsoft.Diagnostics.Tools.Trace.TraceFileFormatConverter.ConvertToFormat(TraceFileFormat format, String fileToConvert, String outputFilename)
E       Conversion complete
E       )
```

Underlying error: `System.IO.FileNotFoundException: Could not load file or assembly 'System.Net.Http, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.`.

Jongy:
@pfilipko1 can you please take a look?

PR #698, opened 2023-02-20
tests: dotnet: xfail test_executable_not_privileged

Jongy:
See https://github.com/Granulate/gprofiler/issues/697

### DIFF
--- a/tests/test_executable.py
+++ b/tests/test_executable.py
@@ -83,7 +83,11 @@ def test_executable(
         ("python", "py-spy"),
         ("ruby", "rbspy"),
         ("php", "phpspy"),
-        ("dotnet", "dotnet-trace"),
+        pytest.param(
+            "dotnet",
+            "dotnet-trace",
+            marks=pytest.mark.xfail(reason="https://github.com/Granulate/gprofiler/issues/697"),
+        ),
     ],
 )
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


Jongy:
Hmm need to xfail also tests in container

d3dave (approved):

tests/test_executable.py
```diff
@@ -83,7 +83,11 @@ def test_executable(
         ("python", "py-spy"),
         ("ruby", "rbspy"),
         ("php", "phpspy"),
-        ("dotnet", "dotnet-trace"),
+        pytest.param(
+            "dotnet",
+            "dotnet-trace",
+            marks=pytest.mark.xfail(reason="https://github.com/Granulate/gprofiler/issues/697"),
```
uriariel, line 89: `, strict=False`
it fails for me locally because its strict

Jongy: Skip it locally then. We will fix the issue in the CI but we will not debug why it does or doesn't reproduce on everyone's machine lol

uriariel: But if it fails sometimes it should be with `strict=False` no?

Jongy: Nope it's determinstic, it's a bug in the build of dotnet-trace, see the ticket https://github.com/Granulate/gprofiler/issues/697

IDK why it doesn't fail on local builds - perhaps because of cached layers? i.e in CI it rebuilds and something changed in our base dotnet image.

uriariel: I ran `docker system prune` before building the image from this branch

Jongy: It doesn't prune the cache layers of buildx

Issue #570, opened 2022-11-02
java: async-profiler: include method modifiers in name


Jongy:
Method modifiers (`ACC_*` values such as `static`, `public` etc) are interesting and can provide useful insights for profile readers. It's easy to get them with `jvmti->GetMethodModifiers`.

For MVP, I'd like it this way:
* We add an option `--java-include-method-modifiers`, disabled by default.
* If above option is given, we pass a matching option to async-profiler (that we add in async-profiler) - `includemm` (mm for method modifiers).
* When `includemm` is given, we will call `GetMethodModifiers` and append the data to the frame name as a prefix, in a format that would match the way it was defined in the Java file. So if we got the modifiers `ACC_PUBLIC`, `ACC_FINAL`, `ACC_STATIC` - we will add `public static final` (maintain the same order as common in Java files), e.g `public final String.replace(....);` (just an example).
* When `includemm` is not in use, async-profiler's behavior should not be changed at all (i.e avoid calling any new methods unless necessary).

PR #712, opened 2023-03-07
java: Add option to include method modifiers

mpozniak95:
## Description
Added an option for java profiling: --java-include-method-modifiers, disabled by default.
When flag provided we use: jvmti->GetMethodModifiers in async-profiler to get method modifiers.

## Related Issue
https://github.com/Granulate/gprofiler/issues/570

## How Has This Been Tested?
Added test that tests this case


mpozniak95:
Needs this: https://github.com/Granulate/async-profiler/pull/5 to be merged first, and then need to make a change in async_profiler_shared_build.sh

gprofiler/profilers/java.py
```diff
@@ -482,6 +483,10 @@ def __init__(
         self._jattach_timeout = jattach_timeout
         self._mcache = mcache
         self._collect_meminfo = collect_meminfo
+        if include_method_modifiers:
+            self._include_method_modifiers = ",includemm"
+        else:
+            self._include_method_modifiers = ""
```
Jongy, line 489: Make it a oneliner?

Jongy (changes requested):
One small comment here, fix the AP-side comments and then update the branch so the test matches what I requested

scripts/async_profiler_build_shared.sh
```diff
@@ -5,11 +5,12 @@
...
-VERSION=v2.9g5
-GIT_REV="60cbc128a752c657e8a97300784686e08c8dbe9d"
+VERSION=add_method_modifiers
+GIT_REV="664d04f18da2205b4f34ce422ec144dd2974906e"
 
-git clone --depth 1 -b "$VERSION" https://github.com/Granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
-make all
+git clone --depth 1 -b "$VERSION" https://github.com/mpozniak95/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
+
+make all CXXFLAGS="-O3 -fno-omit-frame-pointer -fvisibility=hidden -std=c++11"
```
mpozniak95, line 13: I had to add `-std=c++11` because build was failing on centos glibc builder. It seems that by default centos gcc compile with c++98 standard. I am not sure if that change is ok?

tests/test_java.py
```diff
@@ -1133,3 +1133,18 @@ def test_collect_none_jvm_flags(
...
+    application_pid: int,
+    profiler_state: ProfilerState,
+    include_mmm: bool,
+) -> None:
+    with make_java_profiler(profiler_state, java_include_method_modifiers=include_mmm) as profiler:
+        collapsed = snapshot_pid_collapsed(profiler, application_pid)
+        if include_mmm:
+            assert is_function_in_collapsed("private static Fibonacci.fibonacci(I)J_[j]", collapsed)
+        else:
+            assert not is_function_in_collapsed("private static Fibonacci.fibonacci(I)J_[j]", collapsed)
```
Jongy, line 1150: Good negative test.

scripts/async_profiler_build_shared.sh
Jongy: Ideally we should not make wide changes in the build such as modifying default flags. I'm not sure if it has other side effects on AP's built code. Can you do it without? i.e if the problem is `const std::vector<std::pair<int, std::string>> access_flags` then use a different construct and not a vector of pairs?

Jongy: I prefer the changes we do on upstream to remain local, i.e when the includemm option is not given, the behavior should be exact. If we change the global build flags, it's maybe not exact anymore :/

scripts/async_profiler_build_shared.sh
```diff
@@ -5,10 +5,11 @@
...
 set -euo pipefail
 
-VERSION=v2.9g5
-GIT_REV="60cbc128a752c657e8a97300784686e08c8dbe9d"
+VERSION=add_method_modifiers
+GIT_REV="3b60db24f5825a55f7ad9dd0aa3487049ad3bb77"
+
+git clone --depth 1 -b "$VERSION" https://github.com/mpozniak95/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
 
-git clone --depth 1 -b "$VERSION" https://github.com/Granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
```
Jongy, line 11: Can revert the changes here

Jongy: (I mean, reminder to do so after we merge the other)

Jongy:
Can update the commit, and we're good :)

Jongy:
Pushed tag `v2.9g6`

scripts/async_profiler_build_shared.sh
```diff
@@ -5,10 +5,11 @@
 #
 set -euo pipefail
 
-VERSION=v2.9g5
-GIT_REV="60cbc128a752c657e8a97300784686e08c8dbe9d"
+VERSION=v2.9g6
+GIT_REV="4938ce815be597fafc6a7a185a16ff394b9d7f41"
 
 git clone --depth 1 -b "$VERSION" https://github.com/Granulate/async-profiler.git && cd async-profiler && git reset --hard "$GIT_REV"
+
```
Jongy, line 12:
```suggestion
```

mpozniak95: Sorry for that :(

Jongy:
Looks good, I will merge once tests pas..

Jongy (approved):

PR #716, opened 2023-03-08
aarch64 java tests

pfilipko1:
## Checklist:


tests/conftest.py
```diff
@@ -349,6 +350,13 @@ def application_docker_image(
     runtime: str,
     application_image_tag: str,
 ) -> Iterable[Image]:
+    if platform.machine() == "aarch64":
```
Jongy, line 353: Can you move all `platform.machine() == "aarch64"` and `platform.machine() != "aarch64"` checks to a `is_aarch64` function it tests' utils, and use it?

tests/test.sh
```diff
@@ -8,8 +8,20 @@ set -e
...
 if [ -z ${NO_APT_INSTALL+x} ]; then
- sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update
- sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends openjdk-8-jdk python3 python3-pip docker.io php ruby2.7 dotnet-sdk-6.0
+   if [ "$(uname -m)" = "aarch64" ]; then
+    curl -SL -o dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-sdk-latest-linux-arm64.tar.gz
+    sudo mkdir -p /usr/share/dotnet
+    sudo tar -zxf dotnet.tar.gz -C /usr/share/dotnet
+    sudo ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
+    sudo DEBIAN_FRONTEND=noninteractive apt-get update
+    sudo DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends openjdk-8-jdk python3 python3-pip docker.io php ruby3.0
```
Jongy, line 17: Can you keep shared what can be shared between this one and the x86_64 one? e.g the openjdk, the python ....

tests/test.sh
```diff
@@ -8,8 +8,20 @@ set -e
...
+    mkdir -vp ~/.docker/cli-plugins/
```
Jongy, line 18: Do we need buildx here?

tests/test.sh
```diff
@@ -8,8 +8,20 @@ set -e
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"  # https://stackoverflow.com/a/246128
 
 if [ -z ${NO_APT_INSTALL+x} ]; then
- sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update
- sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends openjdk-8-jdk python3 python3-pip docker.io php ruby2.7 dotnet-sdk-6.0
+   if [ "$(uname -m)" = "aarch64" ]; then
+    curl -SL -o dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-sdk-latest-linux-arm64.tar.gz
```
Jongy, line 12: No dotnet in ubuntu apt of aarch64?

Jongy (changes requested):
Done reviewing

tests/test.sh
Jongy, line 12: For all non-apt installers, please make sure they handle re-runs of the script. For example this installation of `dotnet` fails because `ln -s` fails as `dotnet` already exists. The installer logic can check if the installed is already here, and skip, like apt does when you run `apt install -y`.

tests/test.sh
```diff
@@ -8,8 +8,20 @@ set -e
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"  # https://stackoverflow.com/a/246128
 
 if [ -z ${NO_APT_INSTALL+x} ]; then
- sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update
- sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends openjdk-8-jdk python3 python3-pip docker.io php ruby2.7 dotnet-sdk-6.0
+   if [ "$(uname -m)" = "aarch64" ]; then
```
Jongy, line 11: Also missing on aarch64 that I ran were - insallation of `build-essential` and `python3-devel` to build python packages we install and do not have `.whl` files. Please add here under the aarch64 branch.

tests/conftest.py
```diff
@@ -349,6 +350,13 @@ def application_docker_image(
...
+        if application_image_tag == "j9" or application_image_tag == "zing":
+            pytest.xfail(
+                "Different JVMs are not supported on aarch64, see https://github.com/Granulate/gprofiler/issues/717"
+            )
+        if application_image_tag == "musl":
+            pytest.xfail("musl is not supported on aarch64, see https://github.com/Granulate/gprofiler/issues/717")
```
Jongy, line 359: Is this the right link? https://github.com/Granulate/gprofiler/issues/717 doesn't mention anything about musl.

tests/test.sh
pfilipko1: Unfortunately I don't think so. I referred to this stack overflow topic:
https://stackoverflow.com/questions/73022340/net-sdk-installation-on-arm64-ubuntu-instance

tests/test.sh
```diff
@@ -8,8 +8,19 @@ set -e
...
 if [ -z ${NO_APT_INSTALL+x} ]; then
- sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update
- sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends openjdk-8-jdk python3 python3-pip docker.io php ruby2.7 dotnet-sdk-6.0
+  sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update
+  sudo DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends openjdk-8-jdk python3 python3-pip docker.io php
+  if [ "$(uname -m)" = "aarch64" ]; then
+    sudo DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends  python3-dev  ruby3.0 build-essential
+    curl -SL -o dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-sdk-latest-linux-arm64.tar.gz
+    sudo mkdir -p /usr/share/dotnet
+    sudo tar -zxf dotnet.tar.gz -C /usr/share/dotnet
```
Jongy, line 17: Move all these into the `if` of "dotnet doesn't exist"?

tests/test_app_metadata.py
```diff
@@ -201,6 +201,9 @@ def test_app_metadata(
         # don't check JVM flags in direct comparison, as they might change a bit across machines due to ergonomics
         actual_jvm_flags = metadata["application_metadata"][idx].pop("jvm_flags")
         expected_jvm_flags = expected_metadata.pop("jvm_flags")
+        if is_aarch64():
+            metadata["application_metadata"][idx].pop("libjvm_elfid")
+            expected_metadata.pop("libjvm_elfid")
```
Jongy, line 206: Explain in comment why you pop, pls (I suppose because the hash we compare is for x86_64 not for aarch64, so you don't want to compare it?)

Jongy (changes requested):
Done re-reviewing

tests/test.sh
```diff
@@ -8,8 +8,20 @@ set -e
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"  # https://stackoverflow.com/a/246128
 
 if [ -z ${NO_APT_INSTALL+x} ]; then
- sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update
- sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends openjdk-8-jdk python3 python3-pip docker.io php ruby2.7 dotnet-sdk-6.0
+  sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update
+  sudo DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends openjdk-8-jdk python3 python3-pip docker.io php
```
Jongy, line 12:
```suggestion
  sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends openjdk-8-jdk python3 python3-pip docker.io php
```
Bring it back?

Issue #670, opened 2023-02-01
java: Add an option to collect full hs_err file


Jongy:
In `_check_hotspot_error` we check, given a Java process, if it has an hs_err file. If we find one, we read only relevant data from it and log it. In some cases the full file is needed - I'd like it configurable with a flag `--java-full-hserr` which will replace these lines
https://github.com/Granulate/gprofiler/blob/fded6bc0f333b386813928139a394c5c94178924/gprofiler/profilers/java.py#L1014-L1029
with just logging the full file: `Found Hotspot error log for pid {pid} at {error_file}:\n{contents}`.

PR #719, opened 2023-03-09
java: Add option to collect full hserr

mpozniak95:
## Related Issue
https://github.com/Granulate/gprofiler/issues/670


gprofiler/profilers/java.py
```diff
@@ -795,6 +795,13 @@ def read_output(self) -> Optional[str]:
             "'default' for default flag filtering settings; see default_collection_filter_jvm_flag(), or 'none' to "
             "disable collection of JVM flags. Defaults to '%(default)s'",
         ),
+        ProfilerArgument(
+            "--java-full-hserr",
+            dest="java_full_hserr",
+            action="store_true",
+            default=False,
+            help="Show full hs_err log if it occurs",
```
Jongy, line 803:
```suggestion
            help="Log the full hs_err instead of excerpts only, if one is found for a profiled Java application",
```

gprofiler/profilers/java.py
```diff
@@ -1163,24 +1172,27 @@ def _check_hotspot_error(self, ap_proc: AsyncProfiledProcess) -> None:
...
-        logger.warning(
-            f"Found Hotspot error log for pid {pid} at {error_file}:\n"
-            f"VM info: {vm_info}\n"
-            f"siginfo: {siginfo}\n"
-            f"Problematic frame: {problematic_frame}\n"
-            f"native frames:\n{native_frames}\n"
-            f"container info:\n{container_info}"
-        )
+        if self._java_full_hserr:
+            logger.warning(f"Found Hotspot error log for pid {pid} at {error_file}:\n{contents}")
```
Jongy, line 1176: Please ensure the core of the message (i.e `f"Found Hotspot error log for pid {pid} at {error_file}:\n`) is exactly the same for both branches, this can be done by using a single variable for the log, e.g
```
msg = "Found hotspot..."
```

then use msg in both sites

Jongy, line 1176: Additionally - let's improve the message to use logging extras to be consistent with other code, so for example
```
logger.warning("Found Hotspot error log" pid=pid, hs_err_file=error_file, hs_err=contents)
```
same for the other branch (there you can concatenate the excerpts into `contents` variable and then log it)

Jongy (changes requested):
Done reviewing

Jongy (approved):

PR #724, opened 2023-03-12
metrics: Send samples per series instead of per metric name

d3dave:
Currently samples are only generated once per metric name, ignoring additional labels which might result in additional series. This PR standardizes the data classes to be used for samples and the entire snapshot, and fixes the code to report samples for all series even if they have the same metric name but different labels.

### DIFF
--- a/gprofiler/spark/spark_collector.py
+++ b/gprofiler/spark/spark_collector.py
@@ -12,7 +12,7 @@ import time
 import xml.etree.ElementTree as ET
 from datetime import datetime, timezone
 from threading import Event, Thread
-from typing import Any, Dict, Generator, List, Optional, Tuple, Union
+from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 from urllib.parse import urljoin, urlparse
 
 import psutil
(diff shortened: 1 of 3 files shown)


gprofiler/spark/spark_collector.py
```diff
@@ -806,15 +754,18 @@ def _collect_loop(self) -> None:
                         break
 
             if self._spark_sampler is not None:
-                metrics = list(self._spark_sampler.collect())
-                timestamp = datetime.now(tz=timezone.utc)
+                collected = self._spark_sampler.collect()
+                # No need to submit samples that don't actually have a value:
```
Jongy, line 758: Where does it actually happen? Or when?

gprofiler/spark/spark_collector.py
```diff
@@ -806,15 +754,18 @@ def _collect_loop(self) -> None:
...
+                samples = tuple(filter(lambda s: s.value is not None, collected))
+                snapshot = MetricsSnapshot(datetime.now(tz=timezone.utc), samples)
+                # _client and _storage_dir are mutually exclusive, so we will never bake the payload twice
```
Jongy, line 761: Are they? You can pass both.

Jongy (changes requested):
Done reviewing

gprofiler/spark/spark_collector.py
d3dave: `_client` => `--upload-results`, and `_storage_dir` => `--output-dir`, and the two flags are mutually exclusive, so de facto so are the fields. No?

gprofiler/spark/spark_collector.py
d3dave: If a json field specified in one of the `METRICS` dicts is not present in a json response returned from the yarn/spark server, then a sample can be reported for that metric with value `None`.

Jongy: I see. We have logs for that now though, so we'll need to add if we ever encounter..

gprofiler/spark/spark_collector.py
Jongy: They are not mutual exclusive o.O

Jongy (approved):

PR #728, opened 2023-03-18
Upgrade py-spy to v0.3.14 (which supports Python 3.11)

Jongy:
This is the latest version of py-spy & supports 3.11, which brings 3.11 support to gProfiler.

I had to upgrade Rust because one of the py-spy components required it. I didn't upgrade to latest for the same reasons described [here](https://github.com/Granulate/gprofiler/pull/396) - it breaks the rbspy build.

There were a bit conflicts during rebasing:
```
$ git range-diff v0.3.12..v0.3.12g1 v0.3.14..v0.3.14g1
1:  10feb0f = 1:  761491d Don't gather thread activity for all threads when --nonblocking is provided
2:  8b041f9 ! 2:  513e9a9 Don't grab stack trace of non-GIL threads when --gil is provided
    @@ src/python_spy.rs: impl PythonSpy {
     +                continue;
     +            }
     +
    -             let mut os_thread_id = self._get_os_thread_id(python_thread_id, &interp)?;
    ++            // Get the stack trace of the python thread
    ++            let mut trace = get_stack_trace(&thread, &self.process, self.config.dump_locals > 0, self.config.lineno)?;
      
    -             // linux can see issues where pthread_ids get recycled for new OS threads,
    +             // python 3.11+ has the native thread id directly on the PyThreadState object,
    +             // so use that if available
     @@ src/python_spy.rs: impl PythonSpy {
    -                 }
                  }
      
    -+            // Get the stack trace of the python thread
    -+            let mut trace = get_stack_trace(&thread, &self.process, self.config.dump_locals > 0, self.config.lineno)?;
    -             trace.os_thread_id = os_thread_id.map(|id| id as u64);
                  trace.thread_name = self._get_python_thread_name(python_thread_id);
     -            trace.owns_gil = trace.thread_id == gil_thread_id;
     +            trace.owns_gil = owns_gil;
3:  b45bd56 = 3:  6d4cee6 Add a suffix to the stacks to easily identify Python stacks (#1)
```

Nothing too major, things moved around a bit.


d3dave (dismissed):

d3dave (approved):

Jongy:
Tested build on Aarch64 - works fine.

PR #729, opened 2023-03-19
metrics: send time as ISO8601

d3dave:
~The existing `get_iso8601_format_time` cannot be changed because the server might not support Z-suffixed timestamps, which we've agreed to use for the spark metrics :/~

Agreed to send in long format.

### DIFF
--- a/gprofiler/client.py
+++ b/gprofiler/client.py
@@ -241,7 +241,7 @@ class APIClient(BaseAPIClient):
 
 def bake_metrics_payload(snapshot: MetricsSnapshot) -> Dict[str, Any]:
     return {
-        "format_version": 0,
+        "format_version": 2,
         "timestamp": get_iso8601_format_time(snapshot.timestamp),
         "metrics": [sample.__dict__ for sample in snapshot.samples],
     }
--- a/gprofiler/client.py
+++ b/gprofiler/client.py
@@ -16,7 +16,7 @@ from gprofiler.exceptions import APIError
 from gprofiler.log import get_logger_adapter
 from gprofiler.metadata.system_metadata import get_hostname
 from gprofiler.metrics import MetricsSnapshot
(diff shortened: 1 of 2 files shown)


gprofiler/client.py
```diff
@@ -242,6 +242,6 @@ def submit_spark_metrics(self, snapshot: MetricsSnapshot) -> Dict:
 def bake_metrics_payload(snapshot: MetricsSnapshot) -> Dict[str, Any]:
     return {
         "format_version": 0,
-        "timestamp": int(snapshot.timestamp.timestamp()),
+        "timestamp": format_as_short_iso8601(snapshot.timestamp),
```
Jongy, line 245: As discussed let's change it to the format of `+00:00` instead of `Z`.

Jongy: i.e `get_iso8601_format_time`

Jongy (changes requested):
Also - need to change URL to `v2` and remove `format_version`?

gprofiler/client.py
d3dave: fixed.

d3dave:
> Also - need to change URL to `v2` and remove `format_version`?

We agreed to keep and increment format_version to 2, right?

gprofiler/client.py
```diff
@@ -241,7 +241,7 @@ def submit_spark_metrics(self, snapshot: MetricsSnapshot) -> Dict:
 
 def bake_metrics_payload(snapshot: MetricsSnapshot) -> Dict[str, Any]:
     return {
-        "format_version": 0,
-        "timestamp": int(snapshot.timestamp.timestamp()),
+        "format_version": 2,
```
Jongy, line 244:
```suggestion
        "format_version": 1,
```

Jongy (approved):
Let's merge it @d3dave

Issue #732, opened 2023-03-21
Disable coredump creation for binaries gProfiler runs


Jongy:
Binaries we run may crash and create unwanted corefiles, which waste disk space. I saw this on one of our test deployments:
```
...
-rw-------   1 root root 11145216 Mar  5 09:52 core.rbspy.3654398.1678009124
-rw-------   1 root root 11145216 Mar  6 02:26 core.rbspy.3965157.1678069334
-rw-------   1 root root 11145216 Mar 12 22:35 core.rbspy.572769.1678660576
....
```
We should set RLIMIT such that binaries we execute do not emit a core file (and possibly make it an option enabled by default, so if you actually want to debug and understand why things crash, you can disable it and get the core files back).

PR #735, opened 2023-03-22
Set RLIMIT_CORE=0 to disable generation of core files

Jongy:
Closes: https://github.com/Granulate/gprofiler/issues/732

I tested it by:
* Setting `/proc/sys/kernel/core_pattern` on my box to `/core.%p`.
* Running gProfiler from this branch with `--dont-disable-core-files`
* `pkill -SIGSEGV PyPerf`
* Core file appears
* Running gProfiler from this branch **without** `--dont-disable-core-files`
* `pkill -SIGSEGV PyPerf`
* Core file doesn't appear
* Also ran `cat /proc/gprofiler_pid/limits` and ensured that core file has limits `0 0`.

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -57,6 +57,7 @@ from gprofiler.usage_loggers import CgroupsUsageLogger, NoopUsageLogger, UsageLo
 from gprofiler.utils import (
     TEMPORARY_STORAGE_PATH,
     atomically_symlink,
+    disable_core_files,
     get_iso8601_format_time,
     grab_gprofiler_mutex,
     is_root,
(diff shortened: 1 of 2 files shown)


gprofiler/main.py
```diff
@@ -572,6 +576,14 @@ def parse_cmd_args() -> configargparse.Namespace:
         help="Whether to upload the profiling results to the server",
     )
 
+    parser.add_argument(
+        "--dont-disable-core-files",
+        action="store_false",
+        dest="disable_core_files",
+        help="Do not disable creation of coredumps for processes started by the profiler. By default, we disable,"
+        " to refrain from generating core files which consume disk space, in case any of the executed profilers crash.",
+    )
```
d3dave, line 585: where is the default?

Jongy: If you use `store_false` then it defaults to True

d3dave (approved):

PR #736, opened 2023-03-22
java: Guard from enabling async-profiler over Zing ZVM-17168

Jongy:
Needs https://github.com/Granulate/granulate-utils/pull/133

I tested manually the different versions & expected result of `_is_zing_vm_supported` - works fine.

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -946,7 +946,7 @@ class JavaProfiler(SpawningProcessProfilerBase):
     def _is_zing_vm_supported(self, jvm_version: JvmVersion) -> bool:
         # Zing >= 18 is assumed to support AsyncGetCallTrace per
         # https://github.com/jvm-profiling-tools/async-profiler/issues/153#issuecomment-452038960
-        assert jvm_version.zing_major is not None  # it's Zing so should be non-None.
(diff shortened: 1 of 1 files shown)


gprofiler/profilers/java.py
```diff
@@ -946,8 +946,25 @@ def _is_jvm_type_supported(self, java_version_cmd_output: str) -> bool:
...
         # Zing >= 18 is assumed to support AsyncGetCallTrace per
         # https://github.com/jvm-profiling-tools/async-profiler/issues/153#issuecomment-452038960
-        assert jvm_version.zing_major is not None  # it's Zing so should be non-None.
-        return jvm_version.zing_major >= 18
+        assert jvm_version.zing_version is not None  # it's Zing so should be non-None.
+        if jvm_version.zing_version.major < 18:
+            return False
+
+        # try to filter versions exhibiting ZVM-17168, from Zing release notes https://docs.azul.com/prime/release-notes
+        # it seems that the Zing 20 product line has it, so we filter it out here.
```
YoniKF, line 954: Do you think that the issues does not exist on Zing 19?

Jongy: Hmm, I didn't see a bugfix for the 19 line. But not sure.

PR #749, opened 2023-03-27
Big data metrics utils integration

roi-granulate:
Works with https://github.com/Granulate/granulate-utils/pull/132/commits/1341cea4a5641d22e1c56828e8e9a1ded804e446

Please note that https://github.com/Granulate/granulate-utils/pull/132 is still WIP so this PR will be CRable only after it will merge

## Checklist:


requirements.txt
```diff
@@ -14,3 +14,4 @@ netifaces==0.11.0; sys.platform == "win32"
 WMI==1.5.1; sys.platform == "win32"
 ./granulate-utils/
 humanfriendly==10.0
+beautifulsoup4==4.11.1
```
Jongy, line 17: If moved to granulate-utils, do we still need it here?

Jongy:
@roi-granulate - build error is a real thing, please fix
```
#144 73.43 missing module named 'granulate_utils.metrics.sampler' - imported by gprofiler.spark.sampler (top-level)
```

Jongy:
I think you need to update the submodule? This PR lacks it.

gprofiler/spark/sampler.py
```diff
@@ -47,204 +39,19 @@ def __init__(
...
-
-        if spark_master_process is None or webapp_url is None or spark_cluster_mode == "unknown":
-            logger.warning("Could not get proper Spark cluster configuration")
-            return None
-
-        logger.info("Guessed settings are", cluster_mode=spark_cluster_mode, webapp_url=webapp_url)
-
-        return webapp_url, spark_cluster_mode
+        self._spark_sampler = BigDataSampler(
+            logger, get_hostname(), self._master_address, self._spark_mode, self._applications_metrics
```
Jongy, line 53: As `master_address` is not configurable - just pass None?

gprofiler/spark/sampler.py
```diff
@@ -47,204 +39,19 @@ def __init__(
         self._sample_period = sample_period
         # not the same instance as GProfiler._stop_event
         self._stop_event = Event()
-        self._spark_sampler: Optional[SparkCollector] = None
+        self._spark_sampler: Optional[BigDataSampler] = None
```
Jongy, line 42:
```suggestion
        self._big_data_sampler: Optional[BigDataSampler] = None
```

Jongy: Match the name?

Jongy, line 42: Any reason you initialize it to None here then override it 5 lines later? Remove this line?

gprofiler/spark/sampler.py
Jongy: Same for `self._spark_mdoe` and `self._applications_metrics` - Just remove them and pass None/False, explaining that the autofill logic in BIgDataSampler will handle?

gprofiler/spark/sampler.py
```diff
@@ -47,204 +39,19 @@ def __init__(
...
-            webapp_url = self._guess_mesos_master_webapp_address(spark_master_process)
-
-        if spark_master_process is None or webapp_url is None or spark_cluster_mode == "unknown":
-            logger.warning("Could not get proper Spark cluster configuration")
-            return None
-
-        logger.info("Guessed settings are", cluster_mode=spark_cluster_mode, webapp_url=webapp_url)
-
-        return webapp_url, spark_cluster_mode
+        self._spark_sampler = BigDataSampler(
```
Jongy, line 52: See https://github.com/Granulate/granulate-utils/pull/132/files#r1157792829. This c'tor is called in the main thread of gProfiler. I don't want the guess logic to happen at this time (for it can block + for it can crash the thread here).

All logic should happen in the separate thread. Please avoid calling `guess` in the c'tor.

gprofiler/spark/sampler.py
```diff
@@ -259,30 +66,26 @@ def _collect_loop(self) -> None:
...
-                    base_filename = os.path.join(self._storage_dir, f"spark_metric_{escape_filename(now)}")
-                    with open(base_filename, "w") as f:
-                        json.dump(bake_metrics_payload(snapshot), f)
-                if self._client is not None:
-                    self._client.submit_spark_metrics(snapshot)
-
-            self._stop_event.wait(self._sample_period)
+                elif discovered:
+                    snapshot = self._spark_sampler.snapshot()
+                    logger.debug("Collected Spark metrics", snapshot=snapshot)
```
Jongy, line 78: Remove

gprofiler/spark/sampler.py
```diff
@@ -259,30 +66,26 @@ def _collect_loop(self) -> None:
...
+                    # No need to submit samples that don't actually have a value:
```
Jongy, line 79: Can `snapshot` be None here? It can't happen if discovery passed, can it? Then don't check it and just `assert snapshot is not None`, or alternatively, in the `snapshot` value, you can `assert` that it's called only for discovered clusters, and not if `discover` failed.

Also this code doesn't make much sense - you refrain from writing locally if it's None, but not refraining from submitting the metrics?

gprofiler/spark/sampler.py
```diff
@@ -259,30 +66,26 @@ def _collect_loop(self) -> None:
...
         start_time = timefn()
         while not self._stop_event.is_set():
-            if self._spark_sampler is None:
-                spark_cluster_conf = self._find_spark_cluster()
-                if spark_cluster_conf is not None:
-                    master_address, cluster_mode = spark_cluster_conf
-                    self._spark_sampler = SparkCollector(cluster_mode, master_address)
-                else:
+            if self._spark_sampler is not None:
+                discovered = self._spark_sampler.discover()
```
Jongy, line 70: Why do you call `discover` every time and not cache it? `discover` logs every call:
```
            self._logger.info(
                "We already know cluster mode and master address",
                cluster_mode=self._cluster_mode,
                master_address=self._master_address,
            )
```

This is bad API. Do not allow it - instead of logging, you should `assert` discover is not called after discovery succeeded, and have your caller handle the state machine; or alternatively, handle it silently yourself without any side effect such as logs etc.

Jongy (changes requested):
Done reviewing.

gprofiler/spark/sampler.py
roi-granulate: IDC to pass `None`
btw TODO is to have the option to configure them in the future

Jongy: Mhm, you can leave such a TODO in code.

gprofiler/spark/sampler.py
roi-granulate: In the meantime I'll cache `discover`, it'll won't be called it returned `True` once
I'll remove the logs in a separate PR.

gprofiler/spark/sampler.py
roi-granulate: https://github.com/Granulate/granulate-utils/pull/138/files#diff-e8dd154b7df986eb768fe56114e039f44a244a6a5d75f464611aa95fcb21866fL73

gprofiler/spark/sampler.py
roi-granulate: https://github.com/Granulate/granulate-utils/pull/138/files#diff-e8dd154b7df986eb768fe56114e039f44a244a6a5d75f464611aa95fcb21866fL298

gprofiler/spark/sampler.py
roi-granulate: https://github.com/Granulate/gprofiler/pull/749/files#diff-18dce65a9628cdb97bacc94da9b0401f8686d3326c5a71c91a23eb0bf4cbb919R49-R52

gprofiler/spark/sampler.py
roi-granulate: https://github.com/Granulate/gprofiler/pull/749/files#diff-18dce65a9628cdb97bacc94da9b0401f8686d3326c5a71c91a23eb0bf4cbb919R50

gprofiler/spark/sampler.py
roi-granulate: https://github.com/Granulate/gprofiler/pull/749/files#diff-18dce65a9628cdb97bacc94da9b0401f8686d3326c5a71c91a23eb0bf4cbb919R78

gprofiler/spark/sampler.py
roi-granulate: caching discover and it called until returns `True`
https://github.com/Granulate/gprofiler/pull/749/files#diff-18dce65a9628cdb97bacc94da9b0401f8686d3326c5a71c91a23eb0bf4cbb919R71-R72

roi-granulate:
currently there is an issue that [this](https://github.com/Granulate/gprofiler/pull/749/files#diff-18dce65a9628cdb97bacc94da9b0401f8686d3326c5a71c91a23eb0bf4cbb919R78) assert happens, work to resolve why

gprofiler/spark/sampler.py
```diff
@@ -57,194 +46,8 @@ def __init__(
...
-
-        if spark_master_process is None or webapp_url is None or spark_cluster_mode == "unknown":
-            logger.warning("Could not get proper Spark cluster configuration")
-            return None
-
-        logger.info("Guessed settings are", cluster_mode=spark_cluster_mode, webapp_url=webapp_url)
-
-        return webapp_url, spark_cluster_mode
+        # TODO: master address, spark mode and applications metrics should be configurable in the future.
+        self._big_data_sampler: BigDataSampler = BigDataSampler(logger, get_hostname(), None, None, False)
```
Jongy, line 50:
```suggestion
        self._big_data_sampler = BigDataSampler(logger, get_hostname(), None, None, False)
```

No need to specify the type in such case, that's what mypy would derive, no?

gprofiler/spark/sampler.py
```diff
@@ -258,31 +61,31 @@ def _collect_loop(self) -> None:
...
+
         while not self._stop_event.is_set():
-            if self._spark_sampler is None:
-                spark_cluster_conf = self._find_spark_cluster()
-                if spark_cluster_conf is not None:
-                    master_address, cluster_mode = spark_cluster_conf
-                    self._spark_sampler = SparkCollector(cluster_mode, master_address)
-                else:
-                    if timefn() - start_time >= FIND_CLUSTER_TIMEOUT_SECS:
+            if self._big_data_sampler is not None:
```
Jongy, line 67: What is this check for? It is never None. You initialize it in the c'tor.

gprofiler/spark/sampler.py
```diff
@@ -258,31 +61,31 @@ def _collect_loop(self) -> None:
...
+                if not discovered:
+                    if (
+                        not (discovered := self._big_data_sampler.discover())
+                        and timefn() - start_time >= FIND_CLUSTER_TIMEOUT_SECS
+                    ):
                         logger.info("Timed out identifying Spark cluster. Stopping Spark collector.")
                         break
```
Jongy, line 74: Nicer as 2 separate `if`s, move readable:
```suggestion
                    discovered = self._big_data_sampler.discover()

                if not discovered and timefn() - start_time >= FIND_CLUSTER_TIMEOUT_SECS:
                    logger.info("Timed out identifying Spark cluster. Stopping Spark collector.")
                    break
```

Jongy (changes requested):

gprofiler/spark/sampler.py
roi-granulate: https://github.com/Granulate/gprofiler/pull/749/files#diff-18dce65a9628cdb97bacc94da9b0401f8686d3326c5a71c91a23eb0bf4cbb919R67-R72

gprofiler/spark/sampler.py
roi-granulate: f1410bab8ad0a9af9831f554578de4e02db80303

gprofiler/spark/sampler.py
roi-granulate: 3258c059d2a6eb1181962c68eff1bae54fd881e3

gprofiler/spark/sampler.py
```diff
@@ -41,13 +33,10 @@ def __init__(
...
         api_client: Optional[APIClient] = None,
     ):
-        self._master_address: Optional[str] = None
-        self._spark_mode: Optional[str] = None
         self._collection_thread: Optional[Thread] = None
         self._sample_period = sample_period
         # not the same instance as GProfiler._stop_event
         self._stop_event = Event()
-        self._spark_sampler: Optional[SparkCollector] = None
         self._stop_collection = False
```
Jongy, line 40: Not used - remove?

roi-granulate: a1bf1d0e7580c5afd37133462d23778deeeff816

gprofiler/spark/sampler.py
Jongy: Logs are not removed yet. I think a saner API is allowing `discover` to be called only once, or at the very least, ignoring futher `discover` calls after discovery succeeded once. Not required for this PR, anyway.

Jongy:
Hmm reopen

Jongy (approved):
LGTM

Issue #673, opened 2023-02-04
perf: _parse_perf_script parses partial function names


Jongy:
We have `_parse_perf_script` which parses the outputs from `perf script`. It uses `SAMPLE_REGEX` to match a stack from `perf script` and convert it to a collapsed format.

I found a case where it fails to capture the full function names - stacks collected from a C++ program (specifically `envoy`).
Example frame in raw output of `perf script`:
```
std::__1::__function::__func<Envoy::Event::DispatcherImpl::createFileEvent(int, std::__1::function<void (unsigned int)>, Envoy::Event::FileTriggerType, unsigned int)::$_5, std::__1::allocator<Envoy::Event::DispatcherImpl::createFileEvent(int, std::__1::function<void (unsigned int)>, Envoy::Event::FileTriggerType, unsigned int)::$_5>, void (unsigned int)>::operator()
```
How this frame is extracted by gProfiler:
```
std::__1::__function::__func<Envoy::Event::DispatcherImpl::createFileEvent(int, std::__1::function<void
```

This seems to happen for other functions which follow the `function<void[SPACE]....` format, so I suppose we have a bug in the regex pattern.

We need to fix the pattern & introduce tests (can add a testcase in `test_collapse_stack_consider_dso`) for it.

PR #753, opened 2023-03-29
perf: Correctly parse frames with " (" in them

Jongy:
Closes: https://github.com/Granulate/gprofiler/issues/673


Jongy:
This now fails another test:
```
    def test_collapse_stack_consider_dso(stack: str, insert_dso_name: bool, outcome_dict: Dict[str, str]) -> None:
        expected = f"program;{outcome_dict['dso_true' if insert_dso_name else 'dso_false']}"
>       assert expected == _collapse_stack("program", stack, insert_dso_name)
E       AssertionError: assert 'program;(/li...so (deleted))' == 'program;[unk...so (deleted))'
E         - program;[unknown] (/lib/libudev.so (deleted))
E         ?         ----------
E         + program;(/lib/libudev.so (deleted))

```

gprofiler/profilers/perf.py
```diff
@@ -44,7 +44,7 @@
 # ffffffff81082227 mmput+0x57 ([kernel.kallsyms])
 # 0 [unknown] ([unknown])
 # 7fe48f00faff __poll+0x4f (/lib/x86_64-linux-gnu/libc-2.31.so)
-FRAME_REGEX = re.compile(r"^\s*[0-9a-f]+ (.*?) \(\[?(.*?)\]?\)$")
+FRAME_REGEX = re.compile(r"^\s*[0-9a-f]+ (.*) \(\[?(.*?)\]?\)$")
```
marcin-ol, line 47: The regex needs to get more complex.
We need hints to avoid matching parenthesis in dso name while allowing `(deleted)` output in library path resolution.
I propose this, as a starting point:
```suggestion
FRAME_REGEX = re.compile(r"^\s*[0-9a-f]+ (?P<symbol>.*?) \(\[?(?P<library_name>[^(]*?(?:\(deleted\))?)\]?\)$")
```

marcin-ol (changes requested):

gprofiler/profilers/perf.py
Jongy: Hmm, might work. Can you check how this affects performance? e.g take a few example stacks from `perf scripts` and use `timeit` to see how much time `_collapse_stack` takes with the changed regex. It is a very critical part in the hot path so I want to make sure we are not worsening performance.

marcin-ol: That's 2us vs 1.4us for improved vs the old regex.
Sample file: 520k lines from perf script.

gprofiler/profilers/perf.py
```diff
@@ -121,6 +132,7 @@ def _collapse_stack(comm: str, stack: str, insert_dso_name: bool = False) -> str
         m = FRAME_REGEX.match(line)
         assert m is not None, f"bad line: {line}"
         sym, dso = m.groups()
+        dso = dso.strip("[]")
```
Jongy, line 135: I haven't fully read the new regex pattern yet, but trying to understand here - are you expecting `[]` to be optionally around the dso? To be certainly around the dso?

I prefer the code to be explicit. If you always expect - then assert they are there, and remove them. If they are optional (but one set of `[]`) then remove at most one (`strip` might remove multiple, ...)

marcin-ol: My observation is that style of dso name in frames is much more well-defined.
It's either `([vdso])` or `(library_path)` or `(library_path (deleted))`. 3 well defined patterns we can build upon.

But that's good idea to make this pattern more explicit. I will also check if it still doesn't affect performance negatively.

marcin-ol: Checked performance. New regex is 15% faster on a large perf script (1.29us vs 1.51us per frame on average).

Jongy:
Looks good.

Issue #605, opened 2022-11-17
Filter PIDs to profile


Jongy:
Filtering processes to profile is a feature that was requested in multiple cases.https://github.com/Granulate/gprofiler/issues/482 https://github.com/Granulate/gprofiler/issues/298 https://github.com/Granulate/gprofiler/issues/274 https://github.com/Granulate/gprofiler/issues/573. This ticket is for the first step - filter by PIDs (relevant only in cases where processes remain running).

I suggest:
* `--pids` parameter that is a number or comma-separated list of numbers, and can also be given multiple times e.g `--pids 1,2 --pids 5 --pids 76,76576`. All are accumulated.
* At this point https://github.com/Granulate/gprofiler/blob/914e05081bf90f96059f1a3b5995dec71b334b7c/gprofiler/profilers/profiler_base.py#L177  
PIDs are selected for runtime profilers. We can intersect the "selected PIDs" here with the PIDs passed via `--pids`. This class is extended by all runtime profilers classes, so the "easy" way to pass more parameters is to add it in all constructors of `JavaProfiler` etc. I prefer if we find a clean way to do so, so we don't need to modify all runtime profilers, this needs to be decoupled.
* There are 3 profilers which select their PIDs "automatically" and not via gProfiler decision: `perf`, `PyPerf` and `phpspy`. They need to be modified. At the very least, `perf` and `PyPerf` support receiving target PID(s) as arguments. We need to check about `phpspy`. Their respective classes should receive the `pids` argument and make use of it if passed.
* UX wise, I think gProfiler should be fine if it gets `--pids a,b,c` and only e.g `b`, `c` are running and `a` is not running.
* By default, gProfiler will continue in its standard behavior of profiling everything. So if no PIDs are given, an empty set of filter PIDs will be considered ALL processes.
* Please add adequate tests - 2 running Java apps, pass the PID of one of them and ensure that async-profiler was invoked only on one of them.
* Deny passing `--profile-spawned-processes` with `--pids`, because it is meaningless, we target live processes.

Jongy:
On top of this task we could have `--containers=` which accepts multiple container names, and will accept all PIDs originating from those containers. It can also accept the container name as a regex pattern. Not required for MVP but nice to have :)

mpozniak95:
> On top of this task we could have `--containers=` which accepts multiple container names, and will accept all PIDs originating from those containers. It can also accept the container name as a regex pattern. Not required for MVP but nice to have :)

[...]

PR #761, opened 2023-04-05
Filter PIDs to profile

mpozniak95:
## Description
This PR:
- Add flag --pid that enables providing chosen pids to profile
- Checks if at least one of processes provided is alive
- pids are passed to profilers via ProfilerState
- Adds test for checking if only provided pid is profiled with async-profiler

## Related Issue
https://github.com/Granulate/gprofiler/issues/605

### DIFF
--- a/gprofiler/gprofiler_types.py
+++ b/gprofiler/gprofiler_types.py
@@ -6,7 +6,7 @@
 import re
 from collections import Counter
 from dataclasses import dataclass
-from typing import Any, Callable, Dict, MutableMapping, Optional, Union
+from typing import Any, Callable, Dict, List, MutableMapping, Optional, Union
 
 import configargparse
 
@@ -74,6 +74,16 @@ def nonnegative_integer(value_str: str) -> int:
     if value < 0:
         raise configargparse.ArgumentTypeError("invalid non-negative integer value: {!r}".format(value))
     return value
+
+
+def integers_list(value_str: str) -> List[int]:
+    try:
+        pids = [int(pid) for pid in value_str.split(",")]
+    except ValueError:
+        raise configargparse.ArgumentTypeError(
+            "IntegerList should be a single integer, or comma separated list of integers f.e. 13,452,2388"
+        )
+    return pids
 
 
 def integer_range(min_range: int, max_range: int) -> Callable[[str], int]:
--- a/gprofiler/gprofiler_types.py
+++ b/gprofiler/gprofiler_types.py
@@ -81,7 +81,7 @@ def integers_list(value_str: str) -> List[int]:
         pids = [int(pid) for pid in value_str.split(",")]
     except ValueError:
         raise configargparse.ArgumentTypeError(
-            "IntegerList should be a single integer, or comma separated list of integers f.e. 13,452,2388"
+            "Integer list should be a single integer, or comma separated list of integers f.e. 13,452,2388"
         )
     return pids
 
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -786,7 +786,7 @@ def parse_cmd_args() -> configargparse.Namespace:
     if args.nodejs_mode in ("perf", "attach-maps") and args.perf_mode not in ("fp", "smart"):
         parser.error("--nodejs-mode perf or attach-maps requires --perf-mode 'fp' or 'smart'")
 
-    if args.profile_spawned_processes and len(args.pids_to_profile) > 0:
+    if args.profile_spawned_processes and args.pids_to_profile is not None:
         parser.error("--pids is not allowed when profiling spawned processes")
 
     return args
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -15,7 +15,7 @@ import traceback
 from pathlib import Path
 from threading import Event
 from types import FrameType, TracebackType
-from typing import Any, Iterable, Optional, Type, cast
+from typing import Iterable, List, Optional, Type, cast
 
 import configargparse
 import humanfriendly
@@ -102,26 +102,46 @@ def sigint_handler(sig: int, frame: Optional[FrameType]) -> None:
 
 
 class GProfiler:
-    def __init__(self, **kwargs: Any):
-        self._output_dir: str = kwargs.pop("output_dir")
-        self._flamegraph: bool = kwargs.pop("flamegraph")
-        self._rotating_output: bool = kwargs.pop("rotating_output")
-        self._profiler_api_client: Optional[ProfilerAPIClient] = kwargs.pop("profiler_api_client")
-        self._state: State = kwargs.pop("state")
-        self._remote_logs_handler: Optional[RemoteLogsHandler] = kwargs.get("remote_logs_handler", None)
-        self._profile_api_version: str = kwargs.pop("profile_api_version")
-        self._collect_metrics: bool = kwargs.pop("collect_metrics")
-        self._collect_metadata: bool = kwargs.pop("collect_metadata")
-        self._enrichment_options: EnrichmentOptions = kwargs.pop("enrichment_options")
-        self._user_args: UserArgs = kwargs.pop("user_args")
+    def __init__(
+        self,
+        *,
+        output_dir: str,
+        flamegraph: bool,
+        rotating_output: bool,
+        profiler_api_client: Optional[ProfilerAPIClient],
+        collect_metrics: bool,
+        collect_metadata: bool,
+        enrichment_options: EnrichmentOptions,
+        state: State,
+        usage_logger: UsageLoggerInterface,
+        user_args: UserArgs,
+        duration: int,
+        profile_api_version: str,
+        profiling_mode: str,
+        processes_to_profile: Optional[List[Process]],
+        profile_spawned_processes: bool = True,
+        remote_logs_handler: Optional[RemoteLogsHandler] = None,
+        controller_process: Optional[Process] = None,
+        spark_sampler: Optional[SparkSampler] = None,
+    ):
+        self._output_dir = output_dir
+        self._flamegraph = flamegraph
+        self._rotating_output = rotating_output
+        self._profiler_api_client = profiler_api_client
+        self._state = state
+        self._remote_logs_handler = remote_logs_handler
+        self._profile_api_version = profile_api_version
+        self._collect_metrics = collect_metrics
+        self._collect_metadata = collect_metadata
+        self._enrichment_options = enrichment_options
         self._static_metadata: Optional[Metadata] = None
         self._spawn_time = time.time()
         self._last_diagnostics = 0.0
         self._gpid = ""
-        self._controller_process: Optional[Process] = kwargs.get("controller_process", None)
-        self._duration: int = kwargs.pop("duration")
+        self._controller_process = controller_process
+        self._duration = duration
         if self._collect_metadata:
-            self._static_metadata = get_static_metadata(spawn_time=self._spawn_time, run_args=self._user_args)
+            self._static_metadata = get_static_metadata(spawn_time=self._spawn_time, run_args=user_args)
         self._executor = concurrent.futures.ThreadPoolExecutor(max_workers=10)
         # TODO: we actually need 2 types of temporary directories.
         # 1. accessible by everyone - for profilers that run code in target processes, like async-profiler
@@ -129,20 +149,17 @@ class GProfiler:
         # the latter can be root only. the former can not. we should do this separation so we don't expose
         # files unnecessarily.
         container_names_client = ContainerNamesClient() if self._enrichment_options.container_names else None
-        profiler_state_kwargs = {
-            "stop_event": Event(),
-            "profile_spawned_processes": kwargs.get("profile_spawned_processes", True),
-            "insert_dso_name": bool(self._user_args.get("insert_dso_name")),
-            "profiling_mode": kwargs.pop("profiling_mode"),
-            "container_names_client": container_names_client,
-            "pids_to_profile": kwargs.pop("pids_to_profile"),
-            "storage_dir": TEMPORARY_STORAGE_PATH,
-        }
-        self._profiler_state = ProfilerState(**profiler_state_kwargs)
-        self.system_profiler, self.process_profilers = get_profilers(
-            self._user_args, profiler_state=self._profiler_state
+        self._profiler_state = ProfilerState(
+            stop_event=Event(),
+            storage_dir=TEMPORARY_STORAGE_PATH,
+            profile_spawned_processes=profile_spawned_processes,
+            insert_dso_name=bool(user_args.get("insert_dso_name")),
+            profiling_mode=profiling_mode,
+            container_names_client=container_names_client,
+            processes_to_profile=processes_to_profile,
         )
-        self._usage_logger: UsageLoggerInterface = kwargs.pop("usage_logger")
+        self.system_profiler, self.process_profilers = get_profilers(user_args, profiler_state=self._profiler_state)
+        self._usage_logger: UsageLoggerInterface = usage_logger
         if self._collect_metrics:
             self._system_metrics_monitor: SystemMetricsMonitorBase = SystemMetricsMonitor(
                 self._profiler_state.stop_event
@@ -150,7 +167,7 @@ class GProfiler:
         else:
             self._system_metrics_monitor = NoopSystemMetricsMonitor()
 
-        self._spark_sampler: Optional[SparkSampler] = kwargs.get("spark_sampler", None)
+        self._spark_sampler: Optional[SparkSampler] = spark_sampler
 
         if isinstance(self.system_profiler, NoopProfiler) and not self.process_profilers and not self._spark_sampler:
             raise NoProfilersEnabledError()
@@ -818,7 +835,7 @@ def _add_profilers_arguments(parser: configargparse.ArgumentParser) -> None:
             arg_group.add_argument(name, **profiler_arg_kwargs)
 
 
-def verify_preconditions(args: configargparse.Namespace) -> None:
+def verify_preconditions(args: configargparse.Namespace, processes_to_profile: Optional[List[Process]]) -> None:
     if not is_root():
         print("Must run gprofiler as root, please re-run.", file=sys.stderr)
         sys.exit(1)
@@ -849,16 +866,8 @@ def verify_preconditions(args: configargparse.Namespace) -> None:
         print("--log-usage is available only when run as a container!", file=sys.stderr)
         sys.exit(1)
 
-    if args.pids_to_profile is not None:
-        pids_to_profile = []
-        for pid in args.pids_to_profile:
-            try:
-                Process(pid)
-                pids_to_profile.append(pid)
-            except NoSuchProcess:
-                continue
-        setattr(args, "pids_to_profile", pids_to_profile)
-        if len(pids_to_profile) == 0:
+    if processes_to_profile is not None:
+        if len(processes_to_profile) == 0:
             print("There aren't any alive processes from provided via --pid PID list")
             sys.exit(1)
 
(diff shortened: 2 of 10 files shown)


mpozniak95:
I'll create two more issues. It seems that flags --pid/-p doesn't work as intendent in pyperf and phpspy. Pyperf seems to ignore this parameter and phpspy is getting stuck with this flag

gprofiler/main.py
```diff
@@ -153,6 +154,7 @@ def __init__(
             profile_spawned_processes,
             bool(user_args.get("insert_dso_name")),
             profiling_mode,
+            pids_to_profile,
```
Jongy, line 157: When methods have lots of arguments, using kwargs is nicer.

Can you change the c'tor of `ProfilerState` to be kwargs-only? (i.e `__init__(self, *, ...)`) and use kwargs here? This is more readable.

gprofiler/main.py
```diff
@@ -521,6 +523,14 @@ def parse_cmd_args() -> configargparse.Namespace:
     parser.add_argument(
         "--rotating-output", action="store_true", default=False, help="Keep only the last profile result"
     )
+    parser.add_argument(
+        "--pids",
+        dest="pids_to_profile",
+        action="append",
+        default=list(),
+        type=str,
+        help="Coma separated list of processes that will be filtered to profile",
```
Jongy, line 532: Mention that it can be given multiple times as well

gprofiler/main.py
```diff
@@ -791,6 +801,26 @@ def parse_cmd_args() -> configargparse.Namespace:
...
+                    try:
+                        Process(pid)
+                        pids_to_profile.append(pid)
+                    except NoSuchProcess:
+                        continue
+            if len(pids_to_profile) == 0:
+                parser.error("There aren't any alive processes from provided PID list")
+    except ValueError:
+        parser.error("--pids should be integer or coma separated list of integers f.e. --pids 13,452,2388")
+    setattr(args, "pids_to_profile", pids_to_profile)
```
Jongy, line 822: Nicer if you create an argparse type for it, e.g like `positive_integer`. You can create `integers_list` and use it as the `type=`. It will do the comma separation and parsing. Then here you can filter out only alive PIDs.

gprofiler/main.py
```diff
@@ -791,6 +801,26 @@ def parse_cmd_args() -> configargparse.Namespace:
...
+            for input in args.pids_to_profile:
+                pids = [int(pid) for pid in input.split(",")]
+                for pid in pids:
+                    try:
+                        Process(pid)
+                        pids_to_profile.append(pid)
+                    except NoSuchProcess:
+                        continue
+            if len(pids_to_profile) == 0:
+                parser.error("There aren't any alive processes from provided PID list")
```
Jongy, line 819: This needn't be a `parser.error` - the arguments are correct but the runtime checks fail.

Logically I want this check to be in `verify_preconditions` and result in a print to stderr, then `sys.exit(1)`.

gprofiler/main.py
```diff
@@ -791,6 +801,26 @@ def parse_cmd_args() -> configargparse.Namespace:
...
+    except ValueError:
+        parser.error("--pids should be integer or coma separated list of integers f.e. --pids 13,452,2388")
```
Jongy, line 821:
```suggestion
        parser.error("--pids should be a single integer, or a comma separated list of integers f.e. --pids 13,452,2388")
```

gprofiler/main.py
```diff
@@ -1048,6 +1078,7 @@ def main() -> None:
             args.duration,
             args.profile_api_version,
             args.profiling_mode,
+            args.pids_to_profile,
```
Jongy, line 1081: This can also be made kwargs-only if you'd like in this PR :) Can also do in another.

gprofiler/profilers/profiler_base.py
```diff
@@ -171,6 +171,11 @@ def _profiling_error_stack(
     def snapshot(self) -> ProcessToProfileData:
         processes_to_profile = self._select_processes_to_profile()
         logger.debug(f"{self.__class__.__name__}: selected {len(processes_to_profile)} processes to profile")
+        if self._profiler_state.pids_to_profile is not None:
+            processes_to_profile = [
+                process for process in processes_to_profile if process.pid in self._profiler_state.pids_to_profile
+            ]
+            logger.debug(f"Processes left after filtering: {len(processes_to_profile)}")
```
Jongy, line 178:
```suggestion
            logger.debug(f""{self.__class__.__name__}:pProcesses left after filtering: {len(processes_to_profile)}")
```

tests/conftest.py
```diff
@@ -90,7 +90,9 @@ def java_args() -> Tuple[str]:
 
 @fixture()
 def profiler_state(tmp_path: Path, insert_dso_name: bool) -> ProfilerState:
-    return ProfilerState(Event(), str(tmp_path), False, insert_dso_name, CPU_PROFILING_MODE, ContainerNamesClient())
+    return ProfilerState(
```
Jongy, line 93: kwargs here :)

tests/test_sanity.py
```diff
@@ -281,3 +281,35 @@ def test_container_name_when_stopped(
...
+                docker_client,
+                gprofiler_docker_image,
+                output_directory,
+                output_collapsed,
+                runtime_specific_args,
+                profiler_flags,
+            )
+            wait_for_gprofiler_container(container, output_collapsed)
+            assert f"Profiling process {pid1} with async-profiler" in container.logs().decode()
+            assert f"Profiling process {pid2} with async-profiler" not in container.logs().decode()
```
Jongy, line 315: You can add here logs checking that `Processes left after filtering` shows 1 and `selected {len(processes_to_profile)} processes to profile` shows 2.

gprofiler/profiler_state.py
```diff
@@ -49,6 +51,13 @@ def profiling_mode(self) -> str:
     def container_names_client(self) -> Optional[ContainerNamesClient]:
         return self._container_names_client
 
+    @property
+    def pids_to_profile(self) -> Optional[List[int]]:
+        if len(self._pids_to_profile) > 0:
```
Jongy, line 56: API wise, I think it's nicer if the list is either None (if we're not in `--pids` mode) or non-None. i.e, so this class needn't check if length is > 0. If it got a list - use the list. If not - don't use.

gprofiler/profilers/perf.py
```diff
@@ -397,20 +404,21 @@ def __init__(
...
         if perf_mode in ("dwarf", "smart"):
+            extra_args.extend(["--call-graph", f"dwarf,{perf_dwarf_stack_size}"])
             self._perf_dwarf: Optional[PerfProcess] = PerfProcess(
                 self._frequency,
                 self._profiler_state.stop_event,
                 os.path.join(self._profiler_state.storage_dir, "perf.dwarf"),
                 True,
                 False,  # no inject in dwarf mode, yet
-                ["--call-graph", f"dwarf,{perf_dwarf_stack_size}"],
+                extra_args,
```
Jongy, line 421:
```suggestion
                extra_args + ["--call-graph", f"dwarf,{perf_dwarf_stack_size}"],
```
Since extra_args is used for other purposes in this function, editing it with `extend` will affect code that we might add later in the function... Making this piece more fragile. In such cases it's preferable not to edit

gprofiler/profilers/perf.py
```diff
@@ -389,6 +392,10 @@ def __init__(
         self._node_processes: List[Process] = []
         self._node_processes_attached: List[Process] = []
         self._perf_memory_restart = perf_memory_restart
+        extra_args: List[str] = []
```
Jongy, line 395: This can be `pid_args` if we do the other change I suggested on this file.

gprofiler/profilers/perf.py
```diff
@@ -160,7 +160,10 @@ def merge_global_perfs(
 
     total_fp_samples = sum([sum(stacks.values()) for stacks in fp_perf.values()])
     total_dwarf_samples = sum([sum(stacks.values()) for stacks in dwarf_perf.values()])
-    fp_to_dwarf_sample_ratio = total_fp_samples / total_dwarf_samples
+    if total_dwarf_samples == 0:
```
Jongy, line 163: How did you encounter it now? You had a case when the fp perf produced samples but dwarf one did not?

Jongy (changes requested):
Looks good :) I gave some comments. Did not try it out locally yet, will do that after the next CR round.

gprofiler/profilers/perf.py
mpozniak95: Yes I had `ZeroDivisionError` when I run gProfiler to profile single `docker run --rm python-listener` process (container built from tests/containers/python)

tests/test_sanity.py
mpozniak95: I changed it but I am not sure about asserting 

> selected {len(processes_to_profile)} processes to profile shows 2

I feel that this might lead to some misleading failures while testing. F.e If there is some background java process running in the environment it will result in failing the test

mpozniak95:
Two tests failed in CI/CD but I think that it isn't connected with these changes

gprofiler/profiler_state.py
```diff
@@ -8,22 +8,15 @@
...
-        insert_dso_name: bool,
-        profiling_mode: str,
-        container_names_client: Optional[ContainerNamesClient],
-    ) -> None:
-        self._stop_event = stop_event
-        self._profile_spawned_processes = profile_spawned_processes
-        self._insert_dso_name = insert_dso_name
-        self._profiling_mode = profiling_mode
-        self._temporary_dir = TemporaryDirectoryWithMode(dir=storage_dir, mode=0o755)
+    def __init__(self, **kwargs: Any) -> None:
```
Jongy, line 11: Instead of using kwargs (which prevents using types for arguments and is more error prone because unwanted names can be given), what I meant here:

> Can you change the c'tor of ProfilerState to be kwargs-only? (i.e __init__(self, *, ...)) and use kwargs here? This is more readable.

is to use [keyword-only arguments](https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html) , so:

```
def __init__(self, *, stop_event: Event, storage_dir: str, ....)
```

gprofiler/main.py
```diff
@@ -102,71 +102,57 @@ def sigint_handler(sig: int, frame: Optional[FrameType]) -> None:
...
-        self._flamegraph = flamegraph
-        self._rotating_output = rotating_output
-        self._profiler_api_client = profiler_api_client
-        self._state = state
-        self._remote_logs_handler = remote_logs_handler
-        self._profile_api_version = profile_api_version
-        self._collect_metrics = collect_metrics
-        self._collect_metadata = collect_metadata
-        self._enrichment_options = enrichment_options
+    def __init__(self, **kwargs: Any):
```
Jongy, line 105: Same here - https://github.com/Granulate/gprofiler/pull/761/files#r1162452661

gprofiler/profiler_state.py
Jongy: I'm sorry for the confusion, should've written kwargs-*only* explicitly haha

tests/test_sanity.py
Jongy: You're right. It can make the test flaky.

Let's remove `assert "selected 2 processes to profile" in container.logs().decode()` and keep only `assert "processes left after filtering: 1" in container.logs().decode()` which is always correct.

gprofiler/gprofiler_types.py
```diff
@@ -76,6 +76,16 @@ def nonnegative_integer(value_str: str) -> int:
     return value
 
 
+def integers_list(value_str: str) -> List[int]:
+    try:
+        pids = [int(pid) for pid in value_str.split(",")]
+    except ValueError:
+        raise configargparse.ArgumentTypeError(
+            "IntegerList should be a single integer, or comma separated list of integers f.e. 13,452,2388"
```
Jongy, line 84:
```suggestion
            "integers list should be a single integer, or comma separated list of integers f.e. 13,452,2388"
```

gprofiler/main.py
```diff
@@ -851,6 +849,19 @@ def verify_preconditions(args: configargparse.Namespace) -> None:
...
+        for pid in args.pids_to_profile:
+            try:
+                Process(pid)
+                pids_to_profile.append(pid)
+            except NoSuchProcess:
+                continue
+        setattr(args, "pids_to_profile", pids_to_profile)
+        if len(pids_to_profile) == 0:
+            print("There aren't any alive processes from provided via --pid PID list")
+            sys.exit(1)
```
Jongy, line 863: This check itself needs to be in `verify_preconditions`. However, the translation part (PIDs -> Process objects) should not be here - this function is only for verification.

I think the correct place to put it is in `main()`, after the `parse_cmd_args()` call, add a function `pids_to_processes` which modifies `args` to keep only the live processes.

gprofiler/main.py
```diff
@@ -851,6 +849,19 @@ def verify_preconditions(args: configargparse.Namespace) -> None:
...
 
+    if args.pids_to_profile is not None:
+        pids_to_profile = []
+        for pid in args.pids_to_profile:
+            try:
+                Process(pid)
+                pids_to_profile.append(pid)
+            except NoSuchProcess:
+                continue
+        setattr(args, "pids_to_profile", pids_to_profile)
```
Jongy, line 860: Why using `setattr` and not just `args.pids_to_profile = ...`?

gprofiler/main.py
```diff
@@ -851,6 +849,19 @@ def verify_preconditions(args: configargparse.Namespace) -> None:
         print("--log-usage is available only when run as a container!", file=sys.stderr)
         sys.exit(1)
 
+    if args.pids_to_profile is not None:
+        pids_to_profile = []
+        for pid in args.pids_to_profile:
+            try:
+                Process(pid)
```
Jongy, line 856: One improvement idea that I have - instead of storing the "pids to profile" in run time, you can store the `Process` objects that you create here, and then wherever you can compare `Process` object and not just pid, you can compare the object. It works better because it will handle PID reuse (and correctly refrain from profiling if PID was reused, because two Process objects are == if the PID was not reused, and != if the PID was reused, even if it's the same PID). In sites where we compare only by PID (for example, in `perf.py` where we pass the PID to profile) then you can use `Process.pid`.

gprofiler/profilers/perf.py
Jongy: kk

Jongy (changes requested):
Done reviewing

gprofiler/main.py
Jongy: kk

Jongy (changes requested):
Awesome. Gave a few minor comments. I'll play with it a bit locally now and write my impression, currently no other comment besides those few and then it's good to merge.

gprofiler/profilers/profiler_base.py
```diff
@@ -171,6 +171,11 @@ def _profiling_error_stack(
     def snapshot(self) -> ProcessToProfileData:
         processes_to_profile = self._select_processes_to_profile()
         logger.debug(f"{self.__class__.__name__}: selected {len(processes_to_profile)} processes to profile")
+        if self._profiler_state.processes_to_profile is not None:
+            processes_to_profile = [
+                process for process in processes_to_profile if process in self._profiler_state.processes_to_profile
+            ]
+            logger.debug(f"{self.__class__.__name__}: processes left after filtering: {len(processes_to_profile)}")
```
Jongy, line 178: UX wise - no need to log this line if `len(processes_to_profile) == 0` to begin with, let's skip this `if` is that's the case?

gprofiler/main.py
```diff
@@ -906,13 +926,28 @@ def setup_env(should_disable_core_files: bool, pid_file: str) -> None:
...
+def pids_to_processes(args: configargparse.Namespace) -> Optional[List[Process]]:
+    if args.pids_to_profile is not None:
+        processes_to_profile = []
+        for pid in args.pids_to_profile:
+            try:
+                process = Process(pid)
+                processes_to_profile.append(process)
+            except NoSuchProcess:
+                continue
+        return processes_to_profile
```
Jongy, line 938: Let's add a log here, in the `is not None` branch - INFO level - of all PIDs that gProfiler will profile this run. Something like "Target PIDs given by --pids: 1,2,3,4".

mpozniak95: Added log here but I had to move `pids_to_processes()` and `verify_preconditions()` in `main()` to be called after initializing` logger`. I am not sure if that is ok? The other option is instead of logging it here we can add one more check after `logger` is initialized in `main()`

mpozniak95: Ok I see that one of the test from test_preconditions.py fails I guess its not ok then :D

mpozniak95: Changed to second option

Jongy: > The other option is instead of logging it here we can add one more check after logger is initialized in main() and log it there

That's indeed the correct choice. Looks good now.

gprofiler/profilers/perf.py
```diff
@@ -397,7 +404,7 @@ def __init__(
                 os.path.join(self._profiler_state.storage_dir, "perf.fp"),
                 False,
                 perf_inject,
-                [],
+                pid_args,
```
Jongy, line 407: In `PerfProcess`, we always pass `-a` to perf (which means "profile all"). This flag doesn't go with `-p`, I get this warning when I run locally:
```
Warning:
PID/TID switch overriding SYSTEM
```

I suppose we can move the decision of which flag to use into `PerfProcess` - it will get the PID list, then create the `--pid ....` if PIDs are given, or use `-a` if no PIDs are given (default, old behavior)

gprofiler/profilers/profiler_base.py
```diff
@@ -171,6 +171,12 @@ def _profiling_error_stack(
     def snapshot(self) -> ProcessToProfileData:
         processes_to_profile = self._select_processes_to_profile()
         logger.debug(f"{self.__class__.__name__}: selected {len(processes_to_profile)} processes to profile")
+        if self._profiler_state.processes_to_profile is not None:
+            processes_to_profile = [
+                process for process in processes_to_profile if process in self._profiler_state.processes_to_profile
+            ]
+            if len(processes_to_profile) > 0:
```
Jongy, line 178: This `if` should be before filtering. What I suggested is, if we have 0 selected processes to begin with, then obviously we're left with 0 after filtering. But if we have 2 selected processes and both are filtered out, then the current code won't print that 0 are left.

The condition can be `if self._profiler_state.processes_to_profile is not None and len(processes_to_profile) > 0:`.

gprofiler/main.py
```diff
@@ -926,6 +963,8 @@ def main() -> None:
         args.log_rotate_backup_count,
         remote_logs_handler,
     )
+    if processes_to_profile is not None:
```
Jongy, line 966: Let's move this log after the `Running gProfiler` log message.

Jongy (changes requested):
Done reviewing and trying it out. 2 UX comments and we can merge

gprofiler/main.py
```diff
@@ -943,7 +980,8 @@ def main() -> None:
         logger.info(
             "Running gProfiler", version=__version__, commandline=" ".join(sys.argv[1:]), arguments=args.__dict__
         )
-
+        if processes_to_profile is not None:
+            logger.info("Target PIDs given by --pids: ", pids=[process.pid for process in processes_to_profile])
```
Jongy, line 984: This looks like:
```
[2023-04-12 15:56:36,138] INFO: gprofiler: Target PIDs given by --pids:  (pids=[1])
```

No need for the `:` and the space, since logging adds it anyway when formatting the extras.

Jongy, line 984:
```suggestion
            logger.info("Target PIDs given by --pids", pids=[process.pid for process in processes_to_profile])
```

Jongy (approved):

PR #772, opened 2023-04-13
aarch64 executable tests

pfilipko1:
## Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -35,7 +35,7 @@ from gprofiler.metadata.enrichment import EnrichmentOptions
 from gprofiler.profiler_state import ProfilerState
 from gprofiler.profilers.java import AsyncProfiledProcess, JattachJcmdRunner
 from gprofiler.state import init_state
-from tests import CONTAINERS_DIRECTORY, PARENT, PHPSPY_DURATION
+from tests import CONTAINERS_DIRECTORY, PHPSPY_DURATION
 from tests.utils import (
     _application_docker_container,
     _application_process,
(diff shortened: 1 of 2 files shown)


pfilipko1:
I had to install staticx manually as root before, but I think it is a matter of my testing environment, so I decided not to add this change to this PR. If the problem turns out to be aarch64 specific, I'll add this installation to the tests/test.sh file.

Jongy:
It doesn't make sense to me that specifically the "not_privileged" tests will fail on Aarch64, this is most likely something with the environment

tests/conftest.py
```diff
@@ -186,18 +186,8 @@ def application_executable(runtime: str) -> str:
 @fixture
 def gprofiler_exe(request: FixtureRequest, tmp_path: Path) -> Path:
     precompiled = request.config.getoption("--executable")
-    if precompiled is not None:
-        return Path(precompiled)
-
-    with chdir(PARENT):
```
Jongy, line 192: For the record, we are removing it here bcause this diverges from the pyi.Dockerfile. We follow the dockerfile build, not the logic here.

Jongy:
I ran this on my Aarch64 server:
```
sudo ./tests/test.sh --executable build/aarch64/gprofiler
```
and got:
```
=========================================================== 184 passed, 26 skipped, 27 xfailed, 676 warnings in 2397.70s (0:39:57) ===========================================================
```

Will merge this soon

Jongy:
Well, the Aarch64 executable pyperf test is marked as passing
```
tests/test_executable.py::test_executable[False-python-pyperf] PASSED                                                                                                                  [ 18%]
```

because `PythonProfiler` implicitly falls back to py-spy despite pyperf being requested, this needs to be fixed + the test should check for kernel + native stacks like the container test. I'll fix in a separate PR.

PR #771, opened 2023-04-13
Upgrade gProfiler exe to use Python 3.11

mpozniak95:
## Related Issue
https://github.com/Granulate/gprofiler/issues/766

PR #775, opened 2023-04-17
Upgrade granulate-utils to upgrade grpcio

Jongy:
https://github.com/Granulate/granulate-utils/pull/139

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,13 +1,13 @@
 pytest==7.0.1
-flake8==5.0.4
-black==22.8.0
+flake8==6.0.0
+black==23.3.0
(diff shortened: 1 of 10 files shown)


Jongy:
The upgraded grpcio breaks Python and causes it to hang. I tested locally - commented out the `import grpc` and the use of `ContainersNameClient`, and `test_sanity.py::test_from_container[True-python-py-spy]` passes. Without commenting them out, the test hangs.

I'm not sure if the problem is in gProfiler's usage or a bug in Python. We are now working on [3.11 upgrade](https://github.com/Granulate/gprofiler/pull/771) - which requires the updated grpcio - I suggest we try to address them together.

PR #778, opened 2023-04-18
Update remote logs service name after --databricks-job-name-as-service-name

Jongy:
Without it, we will forever use `null` as service name when `--databricks-job-name-as-service-name` is used.


roi-granulate (approved):
LGTM

Issue #674, opened 2023-02-04
tests: ensure all processes go down if gProfiler is brutally killed


Jongy:
When stopped gracefully, gProfiler will ask all child processes to go down, and wait until they do; so once gProfiler exited, we no long have perfs/pyperf/phpspy etc.

When gProfiler is killed brutally (e.g `kill -9`), it cannot ask the childs to go down. We utilize a Linux feature called `PDEATHSIG`. It was added in https://github.com/Granulate/gprofiler/pull/247. I'd like to add automatic tests for this feature -
* The tests need to run gProfiler exe & not container, because once gProfiler in a container is killed, the container exits and all other processes are killed automatically. The case we're testing is exe therefore.
* The test can run gProfiler exe in a clean container.
* `kill -9` it and ensure that no processes remain in the container besides e.g a `bash` which has started the profiler exe.
* The test should ensure that gProfiler was indeed brutally killed - i.e, no cleanups were performed, the process exited with `SIGKILL` exit code, the temporary directory at `/tmp/gprofiler_tmp/tmpxxxxx` remains on the filesystem.

PR #780, opened 2023-04-21
Add test to ensure all profiler-spawned processes go down if gProfiler is brutally killed

mpozniak95:
## Description
Test checks if py-spy will go down if gprofiler processes are brutally killed.

## Related Issue
https://github.com/Granulate/gprofiler/issues/674

### DIFF
--- a/tests/test_executable.py
+++ b/tests/test_executable.py
@@ -4,8 +4,10 @@
 #
 import os
 import shutil
+import signal
 from pathlib import Path
 from subprocess import Popen
+from threading import Event
 from typing import Callable, List, Mapping, Optional
 
 import pytest
@@ -13,6 +15,7 @@ from docker import DockerClient
 from docker.models.containers import Container
 from docker.models.images import Image
 
+from gprofiler.utils import wait_event
 from gprofiler.utils.collapsed_format import parse_one_collapsed
 from tests.utils import RUNTIME_PROFILERS, _no_errors, run_gprofiler_in_container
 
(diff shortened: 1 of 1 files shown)


tests/test_executable.py
```diff
@@ -137,3 +140,55 @@ def test_executable_not_privileged(
...
+@pytest.mark.parametrize("application_docker_capabilities", ["SYS_PTRACE"])
+def test_killing_spawned_processes(
+    gprofiler_exe: Path,
+    application_docker_container: Container,
+    runtime_specific_args: List[str],
+    output_directory: Path,
+    profiler_flags: List[str],
+    application_docker_mount: bool,
+) -> None:
+    """Tests if killing gprofiler with -9 results in killing py-spy"""
```
Jongy, line 162: I'd give a link to the issue https://github.com/Granulate/gprofiler/issues/674 here which explains a bit more about the reasoning.

tests/test_executable.py
```diff
@@ -137,3 +140,55 @@ def test_executable_not_privileged(
...
+    for pid in gprofiler_pids:
+        os.kill(int(pid), signal.SIGKILL)
+    processes_in_container = application_docker_container.top().get("Processes")
+    processes_in_container = [process for process in processes_in_container if "<defunct>" not in process[-1]]
+    print(f"Processes left in container: {processes_in_container}")
+    assert len(processes_in_container) == 1
+    assert "py-spy record" not in str(processes_in_container)
+    command = ["ls", "/tmp/gprofiler_tmp"]
+    e, ls_output = application_docker_container.exec_run(cmd=command, privileged=True, user="root:root", detach=False)
+    assert "tmp" in ls_output.decode()
```
Jongy, line 194: This means to ensure that part?

> The test should ensure that gProfiler was indeed brutally killed - i.e, no cleanups were performed, the process exited with SIGKILL exit code, the temporary directory at /tmp/gprofiler_tmp/tmpxxxxx remains on the filesystem.

Please add a comment, the purpose isn't clear. Also the test can be made more robust, this is not enough - even if the directory is missing, output will be `ls: cannot access '/tmp/....': no such file...` and your assert still passes, won't it?

Should check either for the exact directory `/tmp/gprofiler_tmp/tmpxxxx` OR for the unpacked staticx executable (which is removed after gprofiler exits gracefully, but not when terminated). I think checking the directory is nicer, but the exact one. You can find it by running `ls` earlier and finding it.

Jongy:
With https://github.com/Granulate/gprofiler/pull/790 merged, this can be revived @mpozniak95 if you have the time for it.

PR #783, opened 2023-04-24
WIP: NamedTemporaryFile usage

pfilipko1:
## Checklist:

### DIFF
--- a/gprofiler/profilers/perf.py
+++ b/gprofiler/profilers/perf.py
@@ -7,11 +7,13 @@ import os
 import re
 import signal
 import time
+import glob
 from collections import Counter, defaultdict
 from pathlib import Path
 from subprocess import Popen
 from threading import Event
 from typing import Any, Dict, Iterable, List, Optional
+from tempfile import NamedTemporaryFile
 
 from granulate_utils.golang import get_process_golang_version, is_golang_process
 from granulate_utils.linux.elf import is_statically_linked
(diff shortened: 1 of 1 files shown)


gprofiler/utils/fs.py
```diff
@@ -27,29 +26,27 @@ def is_rw_exec_dir(path: str) -> bool:
...
     """
-    # randomize the name - this function runs concurrently on paths of in same mnt namespace.
-    test_script = Path(path) / f"t-{token_hex(10)}.sh"
 
     # try creating & writing
     try:
         os.makedirs(path, 0o755, exist_ok=True)
-        test_script.write_text("#!/bin/sh\nexit 0")
-        test_script.chmod(0o755)
+        test_script = NamedTemporaryFile(dir=path, suffix=".sh")
```
Jongy, line 33: I think that to have `NamedTemporaryFile` actually remove the file, you need `__exit__` to be called. So the usage should be:
```
with NamedTemporaryFile(...) as test_script:
    do function logic
```

gprofiler/profilers/perf.py
```diff
@@ -322,27 +323,26 @@ def wait_and_script(self) -> str:
             # (unlike Popen.communicate())
             assert self._process is not None and self._process.stderr is not None
             logger.debug(f"{self._log_name} run output", perf_stderr=self._process.stderr.read1())  # type: ignore
-
```
Jongy, line 325: Bring it back - these are 2 separate code blocks, can keep them separate.

gprofiler/profilers/perf.py
```diff
@@ -322,27 +323,26 @@ def wait_and_script(self) -> str:
...
             perf_data.unlink()
-            if self._inject_jit:
-                # might be missing if it's already removed.
-                # might be existing if "perf inject" itself fails
-                remove_path(inject_data, missing_ok=True)
+            # always read its stderr
+            # using read1() which performs just a single read() call and doesn't read until EOF
+            # (unlike Popen.communicate())
+            assert self._process is not None and self._process.stderr is not None
+            logger.debug(f"{self._log_name} run output", perf_stderr=self._process.stderr.read1())  # type: ignore
```
Jongy, line 345: Are these related?

gprofiler/profilers/perf.py
```diff
@@ -322,27 +323,26 @@ def wait_and_script(self) -> str:
...
-            inject_data = Path(f"{str(perf_data)}.inject")
-            if self._inject_jit:
-                run_process(
-                    [perf_path(), "inject", "--jit", "-o", str(inject_data), "-i", str(perf_data)],
+            with NamedTemporaryFile(dir=os.path.dirname(self._output_path), suffix=".inject") as inject_data:
+                if self._inject_jit:
+                    run_process(
+                        [perf_path(), "inject", "--jit", "-o", str(inject_data), "-i", str(perf_data)],
+                    )
+                    perf_data = Path(inject_data.name)
```
Jongy, line 332: This creates a problematic flow:
1. `perf_data = temporary file`
2. Exception is raised during `perf script`
3. `finally` runs and tries to unlink `perf_data` which is already unlinked.

I suggest you use a 3rd variable (like `perf_script_input`), then set it to either `perf_data` OR `inject_data`, and only unlink `perf_data` as `inject_data` is automatically removed.

Jongy (changes requested):
Done reviewing.

Please link the PR to the relevant issue.

gprofiler/utils/fs.py
pfilipko1: I wrote a simple test to check it and it gets closed everytime, even with "sys.exit(1) between these lines:
``` 
tf = tempfile.NamedTemporaryFile(dir="/home/kitter/Documents/june-gprof/my-gprof-fork/gprofiler", suffix=".tester")

with open(tf.name, "w") as f:
    f.write("my-tester")
print(Path(tf.name).read_text())
time.sleep(5)
``` 
but I changed it accordingly anyway :)

PR #784, opened 2023-04-27
Spark sa tests

roi-granulate:
## Checklist:


tests/test_spark.py
```diff
@@ -0,0 +1,123 @@
...
+    SPARK_APPLICATION_DIFF_METRICS,
+    SPARK_APPLICATION_GAUGE_METRICS,
+    SPARK_EXECUTORS_METRICS,
+)
+from granulate_utils.metrics.sampler import BigDataSampler
+from pytest import LogCaptureFixture
+
+from gprofiler.log import get_logger_adapter
+from tests.conftest import _build_image
+from tests.utils import _wait_container_to_start
```
d3dave, line 26: remove the underscore prefix from these function names

tests/test_spark.py
```diff
@@ -0,0 +1,123 @@
...
+    for metric in metrics_dict.values()
+]
+
+SPARK_MASTER_HOST = "127.0.0.1"
+
+DISCOVER_INTERVAL_SECS = 5
+DISCOVER_TIMEOUT_SECS = 60
+
+
+@pytest.fixture(scope="function")
```
d3dave, line 49: function scope is default, no need to specify it explicitly

tests/test_spark.py
```diff
@@ -0,0 +1,123 @@
...
+def sparkpi_container(docker_client: DockerClient, application_docker_mounts: List[Mount]) -> Container:
+    """
+    This fixture is responsible for running SparkPi application in a container.
+    See `containers/spark/Dockerfile`
+    """
+    spark_image = _build_image(docker_client=docker_client, runtime="spark")
```
d3dave, line 55: I see that _build_image passes rm=True to docker build, meaning cache is not retained. So this rebuilds the image for every test. The image should be a separate fixture with wider scope.

tests/test_spark.py
```diff
@@ -0,0 +1,123 @@
...
+    In discover mode we do not know what's the cluster mode and master address.
+    """
+    discover_timeout = DISCOVER_TIMEOUT_SECS
+    caplog.set_level(logging.DEBUG)
+    sampler = BigDataSampler(logger, "", None, None, False)
+    # We want to make the discovery process as close as possible to the real world scenario.
+    while not (discovered := sampler.discover()) and discover_timeout > 0:
+        sleep(DISCOVER_INTERVAL_SECS)
+        discover_timeout -= DISCOVER_INTERVAL_SECS
+    assert discovered, "Failed to discover cluster mode and master address"
```
d3dave, line 95: Add assert on the actual expected discovered cluster settings

d3dave (dismissed):

tests/test_spark.py
roi-granulate: b29f95b

tests/test_spark.py
roi-granulate: 57adb59

tests/test_spark.py
```diff
@@ -0,0 +1,123 @@
...
+]
+
+SPARK_MASTER_HOST = "127.0.0.1"
+
+DISCOVER_INTERVAL_SECS = 5
+DISCOVER_TIMEOUT_SECS = 60
+
+
+@pytest.fixture
+def sparkpi_container(docker_client: DockerClient, application_docker_mounts: List[Mount]) -> Container:
```
Jongy, line 50: I don't think you need `application_docker_mounts`? Don't accept the fixture & pass empty mounts

tests/test_spark.py
```diff
@@ -0,0 +1,123 @@
...
+    Validates that the snapshot contains all the expected metrics.
+    """
+    samples = snapshot.samples
+    assert len(samples) != 0, "No samples found in snapshot"
+    metric_keys = [sample.name for sample in samples]
+    for key in EXPECTED_SA_METRICS_KEYS:
+        assert key in metric_keys, f"Metric {key} not found in snapshot"
+
+
+def test_sa_spark_discovered_mode(caplog: LogCaptureFixture, sparkpi_container: Container) -> None:
```
Jongy, line 83: Test names - use `test_spark_XXX` (as "spark" is the category here, not "sa"). Other tests follow this pattern, e.g `test_java`, `test_python` etc.

tests/test_spark.py
roi-granulate: 7b978aa

tests/test_spark.py
roi-granulate: 68b77d3

tests/test_spark.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    discover_timeout = DISCOVER_TIMEOUT_SECS
+    caplog.set_level(logging.DEBUG)
+    sampler = BigDataSampler(logger, "", None, None, False)
+    # We want to make the discovery process as close as possible to the real world scenario.
+    while not (discovered := sampler.discover()) and discover_timeout > 0:
+        sleep(DISCOVER_INTERVAL_SECS)
+        discover_timeout -= DISCOVER_INTERVAL_SECS
+    assert discovered, "Failed to discover cluster mode and master address"
+    # Sleeping before calling `snapshot()` to make sure SparkPi application is submitted and recognized by Master.
```
Jongy, line 96: This is racy. No way to check for some actual event and then sample? For example, if we can run the app to completion and THEN sample the metrics (and we'd still see them).

tests/test_spark.py
roi-granulate: 03972fb

tests/test_spark.py
roi-granulate: seems like a broader issue - `build_image` is used by `sparkpi_container` and `application_docker_image`, both are function scoped..
WDYT @Jongy

tests/test_spark.py
roi-granulate: https://github.com/Granulate/gprofiler/pull/784/files#diff-09cfb6f133292348c331f171e17cced53e5dd31dc6c777a92e2241837b253a25R73-R85

tests/test_spark.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    Waits for SparkPi to be recognized by Master.
+    Doing so using `SparkRunningApps` class, `get_running_apps()` method.
+    """
+    running_apps = SparkRunningApps(SPARK_STANDALONE_MODE, f"http://{SPARK_MASTER_HOST}:8080", logger)
+    apps: Dict[str, Tuple[str, str]] = {}
+    while not apps:
+        try:
+            apps = running_apps.get_running_apps()
+        except Exception:
```
Jongy, line 83: What exception does it raise until apps are found?

tests/test_spark.py
```diff
@@ -0,0 +1,146 @@
...
+from granulate_utils.metrics.sampler import BigDataSampler
+from granulate_utils.metrics.spark import SparkRunningApps
+from pytest import LogCaptureFixture
+
+from gprofiler.log import get_logger_adapter
+from tests.conftest import build_image
+from tests.utils import wait_container_to_start
+
+# `BigDataSampler` receives a logger as an argument.
+logger = get_logger_adapter("gprofiler_spark_test")
```
Jongy, line 31:
```suggestion
logger = get_logger_adapter(__name__)
```
like others?

tests/test_spark.py
roi-granulate: it's outdated

Jongy: Yeah more clear now.

tests/test_spark.py
roi-granulate: tried, throws:
```
tests/test_spark.py:30: in <module>
gprofiler/log.py:30: in get_logger_adapter
    assert LOGGER_NAME_RE.match(logger_name) is not None, "logger name must start with 'gprofiler'"
E   AssertionError: logger name must start with 'gprofiler'
========================================================================================================= short test summary info ==========================================================================================================
ERROR tests/test_spark.py - AssertionError: logger name must start with 'gprofiler'```

File: tests/test_spark.py [hunk 10]
```diff
@@ -0,0 +1,146 @@
...
+        try:
+            apps = running_apps.get_running_apps()
+        except requests.exceptions.ConnectionError as e:
+            if "Max retries exceeded" in str(e):
+                # Spark Master is not ready yet.
+                pass
+            sleep(1)
+
+
+def _validate_spark_sa_metricssnapshot(snapshot: MetricsSnapshot) -> None:
```
Jongy, line 93: Can you test the values as well? For example, number of apps, number of executors, ...

tests/test_spark.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def test_spark_sa_discovered_mode(caplog: LogCaptureFixture, sparkpi_container: Container) -> None:
+    """
+    Validates `BigDataSampler`s' `discover()` and `snapshot()` API's in discover mode.
+    In discover mode we do not know what's the cluster mode and master address.
+    """
+    discover_timeout = DISCOVER_TIMEOUT_SECS
+    caplog.set_level(logging.DEBUG)
+    sampler = BigDataSampler(logger, "", None, None, False)
```
Jongy, line 111: Use kwargs pls :sweat_smile: too many positional arguments.

tests/test_spark.py
Jongy: It's not what `rm` means:
> rm (bool) – Remove intermediate containers. The docker build command now defaults to --rm=true, but we have kept the old default of False to preserve backward compatibility

Not intermediate images but intermediate containers, we do want to remove them

tests/test_spark.py
Jongy: Aha, because it's a tests module. Okie

tests/test_spark.py
roi-granulate: named arguments
fcbc91f

tests/test_spark.py
roi-granulate: 6ae02d5

tests/test_spark.py
roi-granulate: querying immediately after we know `SparkPi` is running is also problematic, `Master` might take a moment or two to provision.
exceptions were thrown from time to time (e.g. `_spark_application_metrics()`).
sleeping for 5 seconds after we know `SparkPi` is recognize is the best option IMO

tests/test_spark.py
```diff
@@ -0,0 +1,164 @@
...
+    samples = snapshot.samples
+    assert len(samples) != 0, "No samples found in snapshot"
+    actual_metrics = {sample.name: sample.value for sample in samples}
+    # Creating a dict of the expected samples.
+    expected_metrics = dict.fromkeys(EXPECTED_SA_METRICS_KEYS, 0)
+    # Manually setting the expected values of some metrics.
+    expected_metrics["spark_aggregated_stage_active_stages"] = 1
+    expected_metrics["spark_num_applications_running"] = 1
+    expected_metrics["spark_job_diff_numTasks"] = 1000
+    expected_metrics["spark_job_numActiveStages"] = 1
```
Jongy, line 108: Instead of specifying them like that, add code like
```
if key not in ("spark_aggregated_stage_active_stages", ...):
    assert value == actual_metrics[key], ...
```
and explain in comment why you DON'T compare them.

tests/test_spark.py
```diff
@@ -0,0 +1,164 @@
...
+    for key, value in expected_metrics.items():
+        assert key in actual_metrics.keys(), f"Metric {key} not found in snapshot"
+        assert value == actual_metrics[key], f"Metric {key} has unexpected value"
```
Jongy, line 111: No need to add assert messages here - pytest will do it better for you (with both compared values, etc).

tests/test_spark.py
roi-granulate: we spoke about that f2f

Jongy: Yeah it's my misunderstanding.

tests/test_spark.py
roi-granulate: is that what you meant?
```suggestion
        assert key in actual_metrics.keys() and value == actual_metrics[key], f"Metric key or value was not as expected"
```

Jongy:
```suggestion
        assert key in actual_metrics.keys()
        assert value == actual_metrics[key]
```

Jongy (dismissed):
One small, then we can merge

Issue #676, opened 2023-02-08
Nicer error message for invalid token


Jongy:
From the profiler server we get -
```
[2023-02-08 07:19:07,476] ERROR: gprofiler: Server error: Invalid API key
```

and the agent exits.

For the API server of glogger, we get this:
```
2023-02-08 07:19:08,434 - ERROR - Error posting to server
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/glogger/sender.py", line 122, in send
    batch = self._send_once()
  File "/usr/local/lib/python3.8/dist-packages/glogger/sender.py", line 153, in _send_once
    self._send_once_to_server(data)
  File "/usr/local/lib/python3.8/dist-packages/glogger/sender.py", line 174, in _send_once_to_server
    response.raise_for_status()
  File "/usr/local/lib/python3.8/dist-packages/requests/models.py", line 960, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://api.granulate.io/api/v1/logs
2023-02-08 07:19:08,434 - ERROR - Error posting to server
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/glogger/sender.py", line 122, in send
    batch = self._send_once()
  File "/usr/local/lib/python3.8/dist-packages/glogger/sender.py", line 153, in _send_once
    self._send_once_to_server(data)
  File "/usr/local/lib/python3.8/dist-packages/glogger/sender.py", line 174, in _send_once_to_server
    response.raise_for_status()
  File "/usr/local/lib/python3.8/dist-packages/requests/models.py", line 960, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://api.granulate.io/api/v1/logs
```

This clutters the log and hides the simpler error message. We should catch the error and replace with a nicer print.

At the same time we can change the "server error: invalid API key" log to be CRITICAL.

PR #788, opened 2023-05-04
logger message severity changed 

pfilipko1:
Related to https://github.com/Granulate/granulate-utils/pull/146
and https://github.com/Granulate/gprofiler/issues/676

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -1032,7 +1032,7 @@ def main() -> None:
                 else None
             )
         except APIError as e:
-            logger.error(f"Server error: {e}")
+            logger.critical(f"Server error: {e}")
             sys.exit(1)
         except RequestException as e:
             proxy = get_https_proxy()


gprofiler/main.py
```diff
@@ -1032,7 +1032,7 @@ def main() -> None:
                 else None
             )
         except APIError as e:
-            logger.error(f"Server error: {e}")
+            logger.critical(f"Server error: {e}")
```
marcin-ol, line 1035: Wasn't the critical level applicable only for invalid API token? 
> At the same time we can change the "server error: invalid API key" log to be CRITICAL.
(see [this comment](https://github.com/Granulate/gprofiler/issues/676#issue-1575618054))

A known-message check might be needed here to trigger the critical level message in that specific case.
@Jongy , take a look?

marcin-ol (approved):
LGTM

PR #789, opened 2023-05-09
using granulate-utils DatabricksClient

roi-granulate:
## Checklist:


gprofiler/main.py
```diff
@@ -35,7 +36,6 @@
 )
 from gprofiler.consts import CPU_PROFILING_MODE
 from gprofiler.containers_client import ContainerNamesClient
-from gprofiler.databricks_client import DatabricksClient
```
Jongy, line 38: Remove the old file please

roi-granulate: 64021b3

granulate-utils
Jongy, line 1: This commit is the unmerged 2nd PR - we won't merge it as is. We can first merge the 2nd PR, the update the ref to granulate-utils updated master, OR update it now to the master after the 1st PR. I suggest we wait, so let's first merge the 2nd PR, https://github.com/Granulate/granulate-utils/pull/149

Jongy:
This updates https://github.com/Granulate/granulate-utils/pull/149

PR #790, opened 2023-05-09
Consolidate image & executable build into exe-in-image

mpozniak95:
## Related Issue
https://github.com/Granulate/gprofiler/issues/582


gprofiler/main.py
```diff
@@ -769,7 +785,6 @@ def parse_cmd_args() -> configargparse.Namespace:
     )
 
     args = parser.parse_args()
-
```
Jongy, line 772: Bring back :)

gprofiler/main.py
```diff
@@ -769,7 +785,6 @@ def parse_cmd_args() -> configargparse.Namespace:
     )
 
     args = parser.parse_args()
```
Jongy, line 787: I think you can just check `if args.subcommand == extract-resources` then call `copy_resources` then `return`. This way, we don't need to make any other changes in the `main` function

tests/test.sh
```diff
@@ -30,7 +30,7 @@ fi
 PERF_RESOURCE="$SCRIPT_DIR/../gprofiler/resources/perf"
 if [ ! -f "$PERF_RESOURCE" ] && [[ "$*" != *"--executable"* ]]; then
     echo "perf resource not found: $(readlink -f "$PERF_RESOURCE")"
-    echo "Please run $(readlink -f "$SCRIPT_DIR/../scripts/copy_resources_from_image.sh") to get all resources"
+    echo "Please run gprofiler_exe extract_resources --resources-dest=$SCRIPT_DIR/../gprofiler/resources/ to get all resources"
```
Jongy, line 33:
```suggestion
    echo "Please run \"gprofiler extract-resources --resources-dest=$SCRIPT_DIR/../gprofiler/resources/\" to get all resources"
```

tests/test_sanity.py
```diff
@@ -196,7 +196,14 @@ def test_from_container(
     in_container: bool,
 ) -> None:
     if runtime == "php" and not in_container:
-        pytest.skip("Flaky https://github.com/Granulate/gprofiler/issues/630")
+        pytest.skip(
+            "Flaky https://github.com/Granulate/gprofiler/issues/630 and issue"
+            "with phpspy grep: https://github.com/Granulate/gprofiler/issues/794"
```
Jongy, line 201:
```suggestion
            " with phpspy grep: https://github.com/Granulate/gprofiler/issues/794"
```

pyi.Dockerfile
```diff
@@ -115,14 +115,15 @@ RUN ./build_node_package.sh
 # built on newer Ubuntu because they require new clang (newer than available in GPROFILER_BUILDER's CentOS 7)
 # these are only relevant for modern kernels, so there's no real reason to build them on CentOS 7 anyway.
 FROM ubuntu${PYPERF_BUILDER_UBUNTU} AS bcc-build
+ARG STATICX=""
```
Jongy, line 118: No need to modify the staticx config for pyperf  - the idea is to make gprofiler build faster when modifying only python code. I don't want to change the build logic of pyperf itself, only of the pyinstaller of gprofiler.

Dockerfile
```diff
@@ -1,222 +1,7 @@
...
-COPY granulate-utils/setup.py granulate-utils/requirements.txt granulate-utils/README.md granulate-utils/
-COPY granulate-utils/granulate_utils granulate-utils/granulate_utils
-COPY granulate-utils/glogger granulate-utils/glogger
-RUN pip3 install --no-cache-dir -r requirements.txt
-
-COPY LICENSE.md MANIFEST.in README.md setup.py  ./
-COPY gprofiler gprofiler
-RUN pip3 install --no-cache-dir -e .
-
-# lets gProfiler know it is running in a container
```
Jongy, line 219: Please keep tis comment over `GPROFILER_IN_CONTAINER`.

Jongy (changes requested):
I reviewed partially. i still need to go over the CI changes and ensure that we didn't miss anything in the removed `Dockerfile`.

.dockerignore
```diff
@@ -1,5 +1,4 @@
 /bin/
-/build/
```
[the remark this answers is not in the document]
mpozniak95: Executable that is copied into docker image is located in build dir, thats why it is not ignored anymore

gprofiler/main.py
mpozniak95: It was never deleted, not sure why git shows this deletion here :o

marcin-ol (commented):
Reviewed the changes, compared  `pyi.Dockerfile` with `Dockerfile` from master, according to suggestion from @Jongy:
- Found no key parts missing from Dockerfile (former container-build), step-wise.
- Verified resources referenced in `build-stage` with final layer of container build.

No remarks wrt build on x86.

marcin-ol:
Verified executable build and test on aarch64.
Executable tests complete without issues.

marcin-ol:
Building and running container tests on aarch64 results in error:
```
E           docker.errors.ContainerError: Command '['-v', '-d', '3', '-o', '/tmp/gprofiler', '-d', '3', '--no-java', '--no-php', '--no-ruby', '--no-nodejs', '--no-dotnet', '--python-mode=pyperf']' in image '<Image: 'gprofiler:latest'>' returned non-zero exit status 1: b'

exec /gprofiler: no such file or directory\n'
```

Seems that gprofiler bootstrap in container needs attention:
```
/ # ldd /gprofiler
        /lib/ld-linux-aarch64.so.1 (0xffffa7d5d000)
        libdl.so.2 => /lib/ld-linux-aarch64.so.1 (0xffffa7d5d000)
        libz.so.1 => /lib/libz.so.1 (0xffffa7d2c000)
        libpthread.so.0 => /lib/ld-linux-aarch64.so.1 (0xffffa7d5d000)
        libc.so.6 => /lib/ld-linux-aarch64.so.1 (0xffffa7d5d000)
Error relocating /gprofiler: __realpath_chk: symbol not found
Error relocating /gprofiler: __strdup: symbol not found
```

pyi.Dockerfile
```diff
@@ -290,7 +293,7 @@ COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # using scl here to get the proper LD_LIBRARY_PATH set
 # hadolint ignore=SC2046,SC2086
 RUN set -e; \
-    if [ $(uname -m) != "aarch64" ]; then \
+    if [ $(uname -m) != "aarch64" ] && [ "$STATICX" = "true" ]; then \
```
marcin-ol, line 296:
```suggestion
    if [ "$STATICX" = "true" ]; then \
```
`staticx` should be run on aarch64 for gprofiler to ensure correct loading.

Jongy: Yup, it should :( we broke it. It is fixed in https://github.com/Granulate/gprofiler/pull/806

gprofiler/main.py
Jongy: In master it looks like:
```
    args = parser.parse_args()

    args.perf_inject = args.nodejs_mode == "perf"
    args.perf_node_attach = args.nodejs_mode == "attach-maps"
```

how does it look in your branch?

gprofiler/main.py
```diff
@@ -1063,9 +1085,13 @@ def main() -> None:
...
-            assert args.subcommand == "upload-file"
-            args.func(args, profiler_api_client)
-            return
+            if args.subcommand == "extract-resources":
+                args.func(args.resources_dest)
+                return
+            else:
+                assert args.subcommand == "upload-file"
+                args.func(args, profiler_api_client)
+                return
```
Jongy, line 1094:
```suggestion
            if args.subcommand == "extract-resources":
                args.func(args.resources_dest)
            else:
                assert args.subcommand == "upload-file"
                args.func(args, profiler_api_client)
            return
```

pyi.Dockerfile
```diff
@@ -290,7 +293,7 @@ COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # using scl here to get the proper LD_LIBRARY_PATH set
 # hadolint ignore=SC2046,SC2086
 RUN set -e; \
-    if [ $(uname -m) != "aarch64" ]; then \
+    if [ "$STATICX" = "true" ]; then \
```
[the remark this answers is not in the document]
Jongy: Or actually, yeah, only if not `--fast`.

gprofiler/main.py
mpozniak95: Oh, ok. I changed it back.

Jongy (commented):
Getting this build error:
```
 > [bcc-build 14/14] RUN ./pyperf_build.sh --with-staticx:
#59 0.203 Cloning into 'bcc'...
#59 7.473 fatal: Could not parse object '598855b0e297e659f4f880a6fcf49dc173eb5631'.
```

Which is fixed once you merge master, meanwhile I did it locally as I don't have push permissions to your repo

gprofiler/main.py
```diff
@@ -639,6 +645,15 @@ def parse_cmd_args() -> configargparse.Namespace:
 
     upload_file.set_defaults(func=send_collapsed_file_only)
 
+    extract_resources = subparsers.add_parser("extract-resources")
+    extract_resources.set_defaults(func=copy_resources)
+    extract_resources.add_argument(
+        "--resources-dest",
+        dest="resources_dest",
+        default="",
```
Jongy, line 653: `None` is a saner sentinel - so default to `None` and check `is not None`.

gprofiler/main.py
```diff
@@ -1072,8 +1095,11 @@ def main() -> None:
             spark_sampler = None
 
         if hasattr(args, "func"):
-            assert args.subcommand == "upload-file"
-            args.func(args, profiler_api_client)
+            if args.subcommand == "extract-resources":
```
Jongy, line 1098: `extract-resources`, unlike `upload-file` is a subcommand which requires much less preparation/verification of preconditions. By putting its handling this late in the code, you have to fix various sites like
```
    if not args.upload_results and not args.output_dir and not args.extract_resources:
        parser.error("Must pass at least one output method (--upload-results / --output-dir)")
```
and there are also some unfixed sites - for example, `verify_preconditions` still runs and unnecessarily checks for root, etc.

Please move `extract-resources` handling to be earlier, right after `parse_args()`. Also don't forget to move this check:
```
    if args.extract_resources and args.resources_dest == "":
        parser.error("Must provide --resources-dest when extract-resources")
```

Dockerfile
```diff
@@ -1,222 +1,8 @@
...
+FROM alpine as gprofiler
+ARG ARCH
+ARG EXE_PATH=build/${ARCH}/gprofiler
 # lets gProfiler know it is running in a container
 ENV GPROFILER_IN_CONTAINER=1
-
-ENTRYPOINT [ "python3", "-m", "gprofiler" ]
+COPY ${EXE_PATH} /gprofiler
+RUN chmod +x /gprofiler
+ENTRYPOINT ["/gprofiler"]
```
Jongy, line 8: Add newline before EOF (so after line 8)

Dockerfile
```diff
@@ -1,222 +1,8 @@
...
-COPY granulate-utils/setup.py granulate-utils/requirements.txt granulate-utils/README.md granulate-utils/
-COPY granulate-utils/granulate_utils granulate-utils/granulate_utils
-COPY granulate-utils/glogger granulate-utils/glogger
-RUN pip3 install --no-cache-dir -r requirements.txt
-
-COPY LICENSE.md MANIFEST.in README.md setup.py  ./
-COPY gprofiler gprofiler
-RUN pip3 install --no-cache-dir -e .
-
+FROM alpine as gprofiler
```
Jongy, line 1: Some spacing would make in this file to make it readable. between the `FROM` to next lines, before `ENTRYPOINT`.

Dockerfile
```diff
@@ -1,222 +1,8 @@
...
-ARG PYPERF_BUILDER_UBUNTU=@sha256:cf31af331f38d1d7158470e095b132acd126a7180a54f263d386da88eb681d93
-# perf - ubuntu:16.04
-ARG PERF_BUILDER_UBUNTU=@sha256:d7bb0589725587f2f67d0340edb81fd1fcba6c5f38166639cf2a252c939aa30c
-# phpspy - ubuntu:20.04
-ARG PHPSPY_BUILDER_UBUNTU=@sha256:cf31af331f38d1d7158470e095b132acd126a7180a54f263d386da88eb681d93
-# dotnet builder - mcr.microsoft.com/dotnet/sdk:6.0-focal
-ARG DOTNET_BUILDER=@sha256:749439ff7a431ab4bc38d43cea453dff9ae1ed89a707c318b5082f9b2b25fa22
-# async-profiler glibc build
-# requires CentOS 7 so the built DSO can be loaded into machines running with old glibc (tested up to centos:6),
-# we do make some modifications to the selected versioned symbols so that we don't use anything from >2.12 (what centos:6
```
Jongy, line 15: Please move this comment to its appropriate location in `build_x86_64_executable.sh`, I don't want this information forgotten.

Dockerfile
Jongy, line 1: On this same occasion, let's rename this file to `container.Dockerfile` and fix related build scripts.

pyi.Dockerfile
Jongy, line 1: Continuing my suggestion about `container.Dockerfile`, this should be `executable.Dockerfile`.

tests/test_app_metadata.py
```diff
@@ -191,6 +191,8 @@ def test_app_metadata(
     expected_metadata: Dict,
     application_executable: str,
 ) -> None:
+    if runtime == "dotnet":
+        pytest.xfail("Dotnet-trace doesn't work with alpine: https://github.com/Granulate/gprofiler/issues/795")
```
Jongy, line 195: So by switching the base layer of the container to alpine, we break .NET completely for the container build? (not saying it's a problem, as .NET hasn't been ours/users' focus lately, just trying to understand)

scripts/build_aarch64_container.sh
Jongy, line 1: To make sure this one (and `build_x86_64_container.sh`) always use a fresh build of the executable, and not a stale result - please invoke the matching executable build script here before rebuilding the image itself.

Jongy (commented):
Got this failure trying to build in `--fast` mode:
```
 => [build-stage 41/41] RUN set -e;     if [ "false" = "true" ]; then         LIBS=$(./scripts/list_needed_libs.sh) &&         staticx $LIBS dist/gprofiler dist/gprofiler.static ;      0.3s
 => ERROR [export-stage 1/1] COPY --from=build-stage /app/dist/gprofiler.static /gprofiler                                                                                               0.0s
------
 > [export-stage 1/1] COPY --from=build-stage /app/dist/gprofiler.static /gprofiler:
------
failed to compute cache key: "/app/dist/gprofiler.static" not found: not found
```

If `--fast` you need to copy `/app/dist/gprofiler`, not the `.static` one

.github/workflows/build-executable-test-deploy.yml
Jongy, line 1: This file has gotten very long and condensed. I'm not sure there's a way to split it to separate workflows, because all jobs here are dependent in some way. Please spend some time thinking if this can be split somehow.

If we don't split - let's at least give this file a shorter filename (and `name:` in the yaml). `build-test-deploy.yml` and `name: Build, test and deploy`

.github/workflows/build-executable-test-deploy.yml
```diff
@@ -155,3 +155,232 @@ jobs:
...
+
+    # used in the tests
+    - name: Import gProfiler image
+      run: docker image load < output/gprofiler.img
+
+    # TODO: Add docker layer caching when GitHub Actions cache is stabilized and works good with "satackey/action-docker-layer-caching@v0.0.11"
+    - name: Run gProfiler tests
+      run: NO_APT_INSTALL=1 ./tests/test.sh --ignore=tests/test_executable.py
+
+  build-docker-x64:
```
Jongy, line 255: Isn't this job & `build-container-x64` the same? This seems like some duplication.

.github/workflows/build-executable-test-deploy.yml
Jongy: Also, if we can't split, please document in a comment in the top of this file WHY we can't split.

Jongy, line 1: As the file has many jobs now, I suggest we maintain a stricter naming convention to make things clear.

Let's have these job names (I'm using shell patterns to convey permutations)
* build-{executable,container}-{x64,aarch64} - no "build-docker".
* test-{executable,container-and-unit}-x64 - add x64 as a suffix, since all tests right now are "x64". @pfilipko1 is working on aarch64 tests in https://github.com/Granulate/gprofiler/pull/822, I believe we will first merge this PR and then @pfilipko1 will add the -aarch64 jobs.
* push-manifest should be renamed to deploy-container, to match deploy-executable. Both don't require arch suffix, because they deploy both arches.

Jongy (changes requested):
I finished reviewing everything :)

Besides CI that runs on pull request, I'd want to test the new deployment jobs after all fixes. When you've fixed everything please coordinate with me a test. We'll need to edit some stuff to avoid pushing the `latest` image.

tests/test_app_metadata.py
mpozniak95: Unfortunately yes :(

.github/workflows/build-executable-test-deploy.yml
mpozniak95: Some steps of job `build-docker-x64` were the same as steps in `build-container-x64`, and I moved other steps from that job to `deploy-container` job

executable.Dockerfile
```diff
@@ -289,9 +292,14 @@ COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # and make staticx pack them as well.
 # hadolint ignore=SC2086
 RUN set -e; \
-    LIBS=$(./scripts/list_needed_libs.sh) && \
-    staticx $LIBS dist/gprofiler dist/gprofiler.static
+    if [ "$STATICX" = "true" ]; then \
+        LIBS=$(./scripts/list_needed_libs.sh) && \
+        staticx $LIBS dist/gprofiler dist/gprofiler.static &&  \
+        mv dist/gprofiler.static dist/gprofiler.output ; \
```
Jongy, line 298:
```suggestion
        staticx $LIBS dist/gprofiler dist/gprofiler.output ; \  
```

Jongy:
I finished re-reviewing all past comments, I want to have one last go at the CI file and we'll be able to merge it.

Jongy (approved):
Good to merge. I have some final nitpicks that I'll handle in a separate PR after this.

Good job @mpozniak95 , this is a core change with many different aspects and subtleties and you've handled all of them thoroughly and properly :)

Issue #500, opened 2022-09-26
Refactor py-spy & pyperf to separate ProfilerInterface classes


Jongy:
We have `PythonProfiler` which is a class that hides the fact that we have 2 profiler types behind - py-spy or pyperf, and decides itself which one is used.

The decision needs to move out, and each of py-spy & pyperf should implement the interface and be registered separately. This way they can have different settings in their `register_profiler` decorator, as suggested here: https://github.com/Granulate/gprofiler/blob/4bcb3aedae1c286d3058a0dc0f559cf88fe5e940/gprofiler/profilers/python.py#L279

This also removes the need for the `PythonProfiler` itself (which directs calls at the different underlying classes).

Instead, for each "runtime" such as "python", we'll have a function `select_profiler` that gets the mode requested by the user and the list of available profilers (given the architecture & platform) and will run the necessary tests to decide (what is currently implemented in `PythonEbpfProfiler.test`)

Jongy:
https://github.com/Granulate/gprofiler/issues/448 is caused by this merge into `PythonProfiler` - it introduces room for errors.

marcin-ol:
> Instead, for each "runtime" such as "python", we'll have a function select_profiler
So a central authority to select proper profiler?

How about letting candidate profilers report their readiness, via e.g, `readiness_check()`?

gProfiler would group profilers by their runtime, let them report their status and select the one that is ready.

marcin-ol:
I believe we would also need to mark one profiler as preferred for runtimes that have more than one.

Jongy:
> I believe we would also need to mark one profiler as preferred for runtimes that have more than one.

Agree.

Issue #448, opened 2022-08-28
get_stack_offset failure results in pyspy profiler not being initialized


Jongy:
See https://github.com/Granulate/gprofiler/runs/8046863788?check_suite_focus=true - initialization of `PythonEbpfProfiler` fails but pyspy profiler isn't initialized. See `test_app_metadata[True-python-pyperf-expected_metadata0] ` specifically.

Jongy:
This needs to be investigated - perhaps it is a no issue anymore. I'd begin by trying to reproduce it, and report the results here.

Jongy:
This still happens - let's try reproducing it to make sure we're able to test the fix, then fix

PR #805, opened 2023-06-20
Support grouping profilers under common runtime

marcin-ol:
Refactor py-spy & pyperf to separate ProfilerInterface classes.

## Description
Group profilers by common runtime (e.g.: Python, Java, etc.).
Support automatic selection of best available profiler for each runtime.

## Related Issue
#500 

## Motivation and Context
We have PythonProfiler which is a class that hides the fact that we have 2 profiler types behind - py-spy or pyperf, and decides itself which one is used.
There was no mechanism to support multiple profiler options and keep them separate. More profilers will arrive in future that will benefit from this.

Also: #448 

## How Has This Been Tested?
Additional tests developed for new mechanism, added tests to ensure correct behavior of Python profiling.

## Checklist:


marcin-ol:
Remaining actions for this PR:
- [ ] Register PySpyProfiler and EbpfProfiler separately for python, using developed mechanism.
- [ ] Organize profilers options: supported modes, runtime arguments should be either well separated or declared with runtime (Python) itself.

tests/conftest.py
```diff
@@ -580,6 +583,33 @@ def profiler_flags(runtime: str, profiler_type: str) -> List[str]:
     return flags
 
 
+@fixture
+def make_profiler_instance(
+    profiler_flags: List[str], monkeypatch: MonkeyPatch, output_directory: Path, profiler_state: ProfilerState
+) -> Callable[[], _GeneratorContextManager]:
+    """
+    Deliver an instance of profiler selected by runtime and profiler_type.
```
pfilipko1, line 591: Please leave more in-depth comment, as this method is very complex.

marcin-ol: 👍

tests/conftest.py
```diff
@@ -580,6 +583,37 @@ def profiler_flags(runtime: str, profiler_type: str) -> List[str]:
...
+    """
+
+    # Reuse gProfiler flow to initialize specific profiler instance - one selected by runtime fixture.
+    @contextmanager
+    def _make_profiler_instance() -> Iterator[ProfilerBase]:
```
Jongy, line 596: What's the reason to create profiler instances in such implicit way and not directly as we did before?

Personally I prefer the explicit way. This new flow is harder to understand and is more error prone :/

tests/test_python.py
```diff
@@ -165,3 +182,39 @@ def test_dso_name_in_pyperf_profile(
...
+    elif profiler_type == "enabled":
+        # make sure the default behavior, with implicit enabled mode leads to auto selection
+        profiler_flags.remove(f"--python-mode={profiler_type}")
+    profiler_flags.extend(["--no-perf"])
+    gprofiler = start_gprofiler_in_container_for_one_session(
+        docker_client, gprofiler_docker_image, output_directory, output_collapsed, runtime_specific_args, profiler_flags
+    )
+    wait_for_log(gprofiler, "gProfiler initialized and ready to start profiling", 0, timeout=7)
+    assert f"Initialized {profiler_class_name}".encode() in gprofiler.logs()
+    gprofiler.remove(force=True)
```
Jongy, line 220: `force=True` uses SIGKILL, preventing gprofiler from doing proper cleanup. Why?

gprofiler/profilers/python_ebpf.py
```diff
@@ -41,6 +42,47 @@ class PythonEbpfError(CalledProcessError):
     """
 
 
+@register_profiler(
+    "Python",
+    profiler_name="PyPerf",
+    is_preferred=True,
+    # py-spy is like pyspy, it's confusing and I mix between them
```
Jongy, line 49: Comment not relevant here anymore

gprofiler/profilers/registry.py
```diff
@@ -80,26 +95,88 @@ def register_profiler(
...
+
+
+def get_profilers_by_name() -> Dict[str, ProfilerConfig]:
+    return {config.profiler_name: config for configs in profilers_config.values() for config in configs}
+
+
+def get_runtime_possible_modes(runtime: str) -> List[str]:
+    possible_modes: List[str] = [ProfilerConfig.ENABLED_MODE] if len(profilers_config[runtime]) > 1 else []
+    for config in profilers_config[runtime]:
+        possible_modes += [m for m in config.get_active_modes() if m not in possible_modes]
```
Jongy, line 138: Instead of checking for existence you can build it as a `set` and then convert to a list.

gprofiler/profilers/registry.py
```diff
@@ -80,26 +95,88 @@ def register_profiler(
...
+
+
+def get_sorted_profilers(runtime: str) -> List[ProfilerConfig]:
+    """
+    Get all profiler configs registered for given runtime sorted by preference.
+    """
+    arch = get_arch()
+    profiler_configs = sorted(
+        profilers_config[runtime],
+        key=lambda c: (arch in c.get_supported_archs(), c.is_preferred, c.profiler_name),
```
Jongy, line 150: unsupported arches should be filtered out, not sorted away, no?

gprofiler/profilers/factory.py
```diff
@@ -24,44 +28,67 @@ def get_profilers(
...
-            supported_archs = (
-                profiler_config.supported_windows_archs if is_windows() else profiler_config.supported_archs
-            )
-            if arch not in supported_archs:
+    if profiling_mode == "none":
+        return system_profiler, process_profilers_instances
+    arch = get_arch()
+    for runtime in get_profilers_registry():
+        runtime_args_prefix = runtime.lower().replace("-", "_")
+        runtime_mode = user_args.get(f"{runtime_args_prefix}_mode")
```
Jongy, line 36: 1. Why do you need `replace(-, _)` on the runtime name?
2. This logic (these 2 lines) repeat more than once, extract to a function?

gprofiler/profilers/factory.py
```diff
@@ -24,44 +28,67 @@ def get_profilers(
...
+        if runtime_mode != ProfilerConfig.ENABLED_MODE:
+            requested_configs = [c for c in requested_configs if runtime_mode in c.get_active_modes()]
+        # select profilers that support this architecture and profiling mode
+        selected_configs: List[ProfilerConfig] = []
+        for config in requested_configs:
+            profiler_name = config.profiler_name
+            if arch not in config.get_supported_archs() and len(requested_configs) == 1:
                 logger.warning(f"Disabling {profiler_name} because it doesn't support this architecture ({arch})")
                 continue
-
```
Jongy, line 41: I'd keep this blank line for clarity

gprofiler/profilers/factory.py
```diff
@@ -24,44 +28,67 @@ def get_profilers(
...
-            if profiler_mode in ("none", "disabled"):
-                continue
-
-            supported_archs = (
-                profiler_config.supported_windows_archs if is_windows() else profiler_config.supported_archs
-            )
-            if arch not in supported_archs:
+    if profiling_mode == "none":
+        return system_profiler, process_profilers_instances
+    arch = get_arch()
```
Jongy, line 33: I'd add a blank line between these 2 lines for clarity

gprofiler/profilers/factory.py
```diff
@@ -24,44 +28,67 @@ def get_profilers(
...
-                if isinstance(profiler_instance, SystemProfiler):
-                    system_profiler = profiler_instance
-                else:
-                    process_profilers_instances.append(profiler_instance)
-
+                if len(requested_configs) == 1:
+                    logger.critical(
+                        f"Couldn't create the {profiler_name} profiler for runtime {runtime}, not continuing."
+                        f" Request different profiler for runtime with --{runtime_args_prefix}-mode, or disable"
+                        f" {runtime} profiling with --no-{runtime_args_prefix} to disable this profiler",
```
Jongy, line 84:
```suggestion
                        f" {runtime} profiling with --{runtime_args_prefix}-mode=none to disable this profiler",
```

I prefer this way of expressing it (`--no-...` is legacy to me)

gprofiler/profilers/factory.py
```diff
@@ -24,44 +28,67 @@ def get_profilers(
...
-                )
-                sys.exit(1)
-            else:
-                if isinstance(profiler_instance, SystemProfiler):
-                    system_profiler = profiler_instance
-                else:
-                    process_profilers_instances.append(profiler_instance)
-
+                if len(requested_configs) == 1:
+                    logger.critical(
```
Jongy, line 81: This is a new fatal error. In which situations can it happen? When I force `--python-mode=pyperf` but PyPerf can't be used?

gprofiler/profilers/python.py
```diff
@@ -163,6 +152,19 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
         return metadata
 
 
+@register_profiler(
+    "Python",
+    profiler_name="PySpy",
+    # py-spy is like pyspy, it's confusing and I mix between them
+    possible_modes=["auto", "pyspy", "py-spy"],
```
Jongy, line 159: What does "auto" mean? This profiler no longer supports any "modes", it's just py-spy.

Jongy (changes requested):
I reviewed most of it.

Issues/problems I spot:
1. Profilers still need to specify `default_mode` and `possible_modes`, which no longer makes sense as each profiler class now represents a single profiler and a single mode. IMO this is one of the problems we were trying to solve in this PR.
2. As you mentioned, we lack the concept of "runtime arguments". This is visible for Python which has 2 profilers, and will be visible in others as well if we ever add more profielrs (e.g in Java, some of the arguments are truly async-profiler specific and some are not and will be relevant to any java profiler, e.g `--java-collect-jvm-flags`). "Runtime arguments" is where we could have placed the `default_mode` and `possible_mode` arguments.

My proposal:
1. let's add the concept of a runtime object, and profilers will be attached to their runtime object. When generating commandline arguments,  they will be taken into account separately (i.e there will be a section "Java arguments" and then "Java async-profiler arguments").
2. `possible_modes` will be the default ones + the different registered runtime profilers for this runtime.
3. At the same time we can rename `JavaProfiler` to `JavaAsyncProfiler`
4. This simplifies `get_profiler_arguments`

tests/test_python.py
marcin-ol: No specific reason, I think I made a shortcut here. 
Will replace it.

marcin-ol:
> I reviewed most of it.
> 
> Issues/problems I spot:
> 
> 1. Profilers still need to specify `default_mode` and `possible_modes`, which no longer makes sense as each profiler class now represents a single profiler and a single mode. IMO this is one of the problems we were trying to solve in this PR.

That would be okay but I see a few opportunities:
- `SystemProfiler` supports 3 possible on-modes: smart, dwarf, and fp. Either we refactor it with this PR (but that wouldn't align with the intention of this PR) or keep possible_modes **for now**, but add a new issue to solve it.
- python profiling won't be possible with "auto" mode, but I suppose it wasn't used frequently?

And issue with `default_mode`:
- `DotnetProfiler` is disabled by default, with use of `default_mode` parameter; I think we might still need it for newly introduced profilers. I propose to keep this argument but at the Runtime level, renamed to: `enabled_by_default`

> 2. As you mentioned, we lack the concept of "runtime arguments". This is visible for Python which has 2 profilers, and will be visible in others as well if we ever add more profielrs (e.g in Java, some of the arguments are truly async-profiler specific and some are not and will be relevant to any java profiler, e.g `--java-collect-jvm-flags`). "Runtime arguments" is where we could have placed the `default_mode` and `possible_mode` arguments.
> 
> My proposal:
> 
> 1. let's add the concept of a runtime object, and profilers will be attached to their runtime object. When generating commandline arguments,  they will be taken into account separately (i.e there will be a section "Java arguments" and then "Java async-profiler arguments").

I'm all for it. I wanted to make it clear (and convince myself too) that it's really needed, that's why I paused implementation at this stage.

> 2. `possible_modes` will be the default ones + the different registered runtime profilers for this runtime.

I agree we still need default mode - for the likes of DotnetProfiler and experimental ones.

> 3. At the same time we can rename `JavaProfiler` to `JavaAsyncProfiler`

And register with `profiler_name=ap` to support auto-generated `possible_mode=ap`?

> 4. This simplifies `get_profiler_arguments`

Sure.

gprofiler/profilers/registry.py
marcin-ol: On a second thought, this has one implication - we won't be able to explain to the user why profiler wasn't selected:
- `factory.get_profilers()` is walking through this list of sorted profilers and will warn if some profilers are unavailable (on given architecture),
- if we get an already-filtered list, user will only learn that `no profilers were selected`.

To be on the safe side we should filter available profilers when building list of command-line arguments. Therefore another function needs changing - `_add_profilers_arguments()`, to do this filtering.

marcin-ol:
As discussed earlier, I have introduced runtime object in form of `ProfilingRuntime`.
Runtime defines shared attributes of profilers: default_mode, mode_help, disablement_help and common_arguments.
- Few things were simplified - get_profiler_arguments, get_preferred_or_first_profiler were discarded, as now the common and profiler-specific arguments are separate.
- New concept was added - `InternalArgument` to handle arguments that are set in the code, not by arg parser, i.e.: `perf_node_attach`, `perf_inject`.

gprofiler/profilers/python.py
```diff
@@ -163,6 +152,40 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
...
+)
+class PythonRuntime(ProfilingRuntime):
+    pass
+
+
+@register_profiler(
+    "PySpy",
+    runtime_class=PythonRuntime,
+    possible_modes=["auto", "pyspy", "py-spy"],
+    # we build pyspy for both,.
```
Jongy, line 182:
```suggestion
    # we build pyspy for both
```

tests/conftest.py
```diff
@@ -626,7 +626,6 @@ def python_version(in_container: bool, application_docker_container: Container)
     else:
         # If not running in a container the test application runs on host
         output = subprocess.check_output("python3 --version", stderr=subprocess.STDOUT, shell=True)
-
```
Jongy, line 629: Undo diff?

Jongy:
I only briefly reviewed now.

Issue #768, opened 2023-04-11
tests: aarch64: Run in CI


Jongy:
After https://github.com/Granulate/gprofiler/issues/506 we have Aarch64 container tests, and after https://github.com/Granulate/gprofiler/issues/767 we'll have executable tests as well. Currently they aren't run in our CI - AFAIK GH don't have builtin Aarch64 runners, so we'll need a self-hosted one. Once we get to implement this ticket, let's search the web again for current solutions.

Jongy:
We added some self-hosted x86_64 and Aarch64 runners. See my commit [here](https://github.com/Granulate/gprofiler/pull/810/commits/00338eb86d070cbd1f8efaf94b8d0dd81ab8b2e1) which utilized them using the necessary labels.

We can try using those runners now :) I suggest we start with `build-container-and-test.yml`. The 2 steps (`build-container` and `test`) can be cloned to `build-container-x86` and `test-x86`, then create a copy for Aarch64 in the same file, with the necessary tags so it runs on Aarch64 machines.

PR #806, opened 2023-06-20
Reenable staticx for arm

d3dave:
## Description
PR #622 accidentally removed invocation of staticx on ARM builds.

### DIFF
--- a/pyi.Dockerfile
+++ b/pyi.Dockerfile
@@ -287,6 +287,7 @@ COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # in some cases, when the application is lazily loading some DSOs, staticx doesn't handle it.
 # we use list_needed_libs.sh to list the dynamic dependencies of *all* of our resources,
 # and make staticx pack them as well.
+# hadolint ignore=SC2086
 RUN set -e; \
     LIBS=$(./scripts/list_needed_libs.sh) && \
     staticx $LIBS dist/gprofiler dist/gprofiler.static


gprofiler/__init__.py
```diff
@@ -1,5 +1,6 @@
+
 #
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
-__version__ = "1.29.0"
+__version__ = "1.30.0-test-static-arm"
```
Jongy, line 5: 1.30.0

pyi.Dockerfile
```diff
@@ -290,10 +290,8 @@ COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # using scl here to get the proper LD_LIBRARY_PATH set
 # hadolint ignore=SC2046,SC2086
 RUN set -e; \
-    if [ $(uname -m) != "aarch64" ]; then \
-        LIBS=$(./scripts/list_needed_libs.sh) && \
-        staticx $LIBS dist/gprofiler dist/gprofiler ; \
-    fi
+    LIBS=$(./scripts/list_needed_libs.sh) && \
+    staticx $LIBS dist/gprofiler dist/gprofiler
```
Jongy: As discussed, use a different filename for the post-staticx file

Jongy (changes requested):

pyi.Dockerfile
```diff
@@ -287,14 +287,11 @@ COPY ./scripts/list_needed_libs.sh ./scripts/list_needed_libs.sh
 # in some cases, when the application is lazily loading some DSOs, staticx doesn't handle it.
 # we use list_needed_libs.sh to list the dynamic dependencies of *all* of our resources,
 # and make staticx pack them as well.
-# using scl here to get the proper LD_LIBRARY_PATH set
 # hadolint ignore=SC2046,SC2086
```
Jongy, line 290: Can remove the hadolint ignore I think

Jongy (approved):
Sanity-tested this way - build 1.29.0 and ran:
```
$ docker run --rm -v $(pwd):/gprofiler centos:7 /gprofiler/build/aarch64/gprofiler --version
[7] Error loading Python lib '/tmp/_MEI3yzWHp/libpython3.10.so.1.0': dlopen: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/_MEI3yzWHp/libpython3.10.so.1.0)
```

Built this and ran:
```
$ docker run --rm -v $(pwd):/gprofiler centos:7 /gprofiler/build/aarch64/gprofiler --version
1.30.0
```

I still want to check if current master fails with `test_executable` when ran on `centos:7`, because that's what should hav ecaught this problem after we have https://github.com/Granulate/gprofiler/issues/768...

Issue #798, opened 2023-05-24
perf: improve "smart" perf heuristic to ignore [unknown] frames


Jongy:
`--perf-mode=smart` is a mode which runs two `perf`s, in DWARF and FP mode, and per-PID selects either the DWARF or FP stacks, depending on which are "deeper", under the assumption that if either DWARF or FP fail, the stack will be shallow. (Just as a note, many compiled programs are not built with FP so FP perf fails, while DWARF succeeds if debugging information is present; and Golang programs, from our experience, succeed with FP but fail with DWARF. That's the reason we have both).

The heuristic is implemented in `get_average_frame_count` and it's a simple check of depth. However, from what I've recently seen, DWARF stacks can be rather deep but all are `[unknown]`. In such cases, we don't want to account them.

1. Fix `get_average_frame_count` to avoid accounting `[unknown]` frames in the depth check
2. Add a test that shows that if DWARF is deeper than FP (or vice versa) but the depth is mostly `[unknown]`. then FP is selected (or DWARF in the opposite case).

PR #807, opened 2023-06-21
Ignore unknown frames in perf get_average_frame_count()

mpozniak95:
## Related Issue
https://github.com/Granulate/gprofiler/issues/798

### DIFF
--- a/gprofiler/profilers/perf.py
+++ b/gprofiler/profilers/perf.py
@@ -73,14 +73,19 @@ def get_average_frame_count(samples: Iterable[str]) -> float:
         if len(kernel_split) == 1:
             kernel_split = sample.split("_[k] ", 1)
 
+        # Check if there are any "[unknown]" frames in this sample, if they are ignore it. This is helpful
+        # in perf_mode="smart", because DWARF stack can be deep but some of the frames might be unknown. In
+        # that situation we want to avoid counting unknown frames and then choose stacks that has better
+        # sum(frame_count_per_samples)/len(frame_count_per_samples) ratio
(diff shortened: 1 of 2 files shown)


gprofiler/profilers/perf.py
```diff
@@ -77,10 +77,10 @@ def get_average_frame_count(samples: Iterable[str]) -> float:
         if len(kernel_split) > 1:
             # example: "a;b;c;d_[k];e_[k] 1" should return the same value as "a;b;c 1", so we don't
             # add 1 to the frames count like we do in the other branch.
-            frame_count_per_samples.append(kernel_split[0].count(";"))
+            frame_count_per_samples.append(kernel_split[0].count(";") - kernel_split[0].count("[unknown]"))
```
Jongy, line 80: 1. Document what is the purpose for this `count`.
2. Now we added more logic to both branches, perhaps share it? i.e you can run both `count` calls of `;` and `unknown` before the branch, then use the same variable.

tests/test_perf.py
```diff
@@ -260,6 +260,7 @@ def test_perf_restarted_if_killed(
         (["d_[k] 1"], 0),
         (["d_[k];e_[k] 1"], 0),
         (["a;b;c;d_[k] 1"], 3),
+        (["[unknown];[unknown];[unknown];a;b;c;d_[k] 1"], 3),
```
Jongy, line 263: Let's document this one and explain, because it's not obvious.

Jongy (changes requested):
Done reviewing

tests/test_perf.py
Jongy: I think it's clear enough.

Jongy (approved):

PR #809, opened 2023-06-21
test installation tweak on aarch64

pfilipko1:
There were more tweaks than ruby name change, but it turned out to be caused by my setup :)

### DIFF
--- a/tests/test.sh
+++ b/tests/test.sh
@@ -10,14 +10,14 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"  # h
 if [ -z ${NO_APT_INSTALL+x} ]; then
   sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update
   sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends openjdk-8-jdk python3 python3-pip php
+  sudo apt-get install ca-certificates curl gnupg
+  curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+  sudo chmod a+r /etc/apt/keyrings/docker.gpg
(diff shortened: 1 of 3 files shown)


tests/test.sh
```diff
@@ -11,7 +11,7 @@ if [ -z ${NO_APT_INSTALL+x} ]; then
   sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update
   sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends openjdk-8-jdk python3 python3-pip docker.io php
   if [ "$(uname -m)" = "aarch64" ]; then
-    sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends python3-dev ruby3.0 build-essential nodejs
+    sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends python3-dev ruby build-essential nodejs
```
Jongy, line 14: I think this depends on the Ubuntu version? What version are you using?

On my ubuntu 22.04 I see that ruby3.0 needs ruby and ruby needs ruby3.0, how did you encounter this problem?

Jongy: Okay, in any case I think it's fine to change to `ruby`.

tests/test_app_metadata.py
```diff
@@ -187,9 +187,12 @@ def test_app_metadata(
     assert_collapsed: AssertInCollapsed,
     profiler_flags: List[str],
     runtime: str,
+    profiler_type: str,
     expected_metadata: Dict,
     application_executable: str,
 ) -> None:
+    if profiler_type == "pyperf":
```
Jongy, line 194: Need to check architecture here, don't xfail for x86_64

tests/test_sanity.py
```diff
@@ -193,8 +193,11 @@ def test_from_container(
     assert_app_id: Callable,
     profiler_flags: List[str],
     runtime: str,
+    profiler_type: str,
     in_container: bool,
 ) -> None:
+    if profiler_type == "pyperf":
```
Jongy, line 199: Need to check architecture here, don't xfail for x86_64

Jongy (approved):

PR #813, opened 2023-06-25
ECS: Impose hard limits for memory and cpu

mirikl:
## Description
Impose hard limits for memory and cpu in gprofiler ECS task definition

### DIFF
--- a/deploy/ecs/gprofiler_task_definition.json
+++ b/deploy/ecs/gprofiler_task_definition.json
@@ -6,9 +6,7 @@
         {
             "name": "granulate-gprofiler",
             "image": "index.docker.io/granulate/gprofiler:latest",
-            "memory": "1024",
             "memoryReservation": "256",
-            "cpu": "100",
             "essential": true,
             "command": [
                 "-cu",
@@ -19,6 +17,8 @@
             "privileged": true
         }
     ],
+    "memory": "1024",
+    "cpu": "1024",
     "networkMode": null,
     "pidMode": "host",
     "ipcMode": "host",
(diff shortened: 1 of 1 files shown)


deploy/ecs/gprofiler_task_definition.json
```diff
@@ -6,9 +6,7 @@
         {
             "name": "granulate-gprofiler",
             "image": "index.docker.io/granulate/gprofiler:latest",
-            "memory": "1024",
```
Jongy, line 9: Is that the case for memory as well? I was under the impression that `memory` is respected under the container, but `cpu` is not.

deploy/ecs/gprofiler_task_definition.json
```diff
@@ -6,9 +6,7 @@
...
             "memoryReservation": "256",
-            "cpu": "100",
```
Jongy, line 11: Value is still required here, to serve as the "requests" for CPU. Keep it

deploy/ecs/gprofiler_task_definition.json
mirikl: You're right I'll keep it in container level.

Jongy (approved):

deploy/ecs/gprofiler_task_definition.json
```diff
@@ -19,6 +19,7 @@
             "privileged": true
         }
     ],
+    "cpu": "1024",
```
Jongy, line 22: Actually, let's use 512 to match k8s 500

Jongy (changes requested):
.

Jongy (approved):

Issue #704, opened 2023-02-27
python: gunicorn appids improvements


Jongy:
This assumption is wrong - 
https://github.com/Granulate/gprofiler/blob/7d9ffb2c80d0a4ffa9030ed9f44ab56b5317ba5a/gprofiler/metadata/application_identifiers.py#L119-L120

and I'm seeing many sites where we select a random argument to `gunicorn` due to it.

Let's try improving it with heuristics:
1. all app specs will have `:` in them, so don't try arguments that don't have `:`.
2. if multiple args with `:` exist, prefer the one not preceeded by a `--` or `-` argument, i.e in a commandline like `gunicorn -k 4 myapp:app -b 0.0.0.0:80` it will prefer the first `:` occurrence because it's preceeded by arg `4` not by arg `-b`.

The same probably applies for #693 .

PR #816, opened 2023-06-27
gunicorn appid improvement

pfilipko1:
Works properly, not sure about the naming convention though.
https://github.com/Granulate/gprofiler/issues/704

### DIFF
--- a/gprofiler/metadata/application_identifiers.py
+++ b/gprofiler/metadata/application_identifiers.py
@@ -111,6 +111,7 @@ class _GunicornApplicationIdentifierBase(_ApplicationIdentifier):
                 if arg == colon_arg and "-" not in cmdline_list[index - 1]:
                     return arg
 
+
 class _GunicornApplicationIdentifier(_GunicornApplicationIdentifierBase):
     def get_app_id(self, process: Process) -> Optional[str]:
         # As of gunicorn documentation the WSGI module name most probably will come from the cmdline and not from the
@@ -123,7 +124,6 @@ class _GunicornApplicationIdentifier(_GunicornApplicationIdentifierBase):
             return None
 
         # wsgi app specification will come always as the last argument (if hasn't been specified config file)
-        print(process.cmdline())
         return self.gunicorn_to_app_id(self.gunicorn_get_app_name(process.cmdline()), process)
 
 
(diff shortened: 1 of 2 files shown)


Jongy (changes requested):
Before I begin reviewing, can you please add tests next to `test_gunicorn` in `test_appids.py`? (tests that show this change now supports the cases described in the ticket)

pfilipko1:
Added two test cases (hopefully correct) containing commands with multiple arguments containing ":"

gprofiler/metadata/application_identifiers.py
```diff
@@ -116,8 +126,8 @@ def get_app_id(self, process: Process) -> Optional[str]:
         ) and "gunicorn" != os.path.basename(_get_cli_arg_by_index(process.cmdline(), 1)):
             return None
 
-        # wsgi app specification will come always as the last argument (if hasn't been specified config file)
-        return self.gunicorn_to_app_id(process.cmdline()[-1], process)
+        # wsgi app specification must not be the last argument, thus the method to extract it
```
Jongy, line 129:
```suggestion
        # wsgi app specification might not be the last argument, thus the method to extract it
```

gprofiler/metadata/application_identifiers.py
```diff
@@ -104,6 +104,16 @@ def gunicorn_to_app_id(self, wsgi_app_spec: str, process: Process) -> str:
...
         return f"gunicorn: {wsgi_app_spec} ({_append_python_module_to_proc_wd(process, wsgi_app_file)})"
 
+    def gunicorn_get_app_name(self, cmdline_list: List[str]) -> str:
+        # method improving appid selection:
+        # https://github.com/Granulate/gprofiler/issues/704
+        cmdline_colon_list = list(filter(lambda x: ":" in x, cmdline_list))
+        for index, arg in enumerate(cmdline_list):
+            for colon_arg in cmdline_colon_list:
+                if arg == colon_arg and not cmdline_list[index - 1].startswith("-"):
+                    return arg
```
Jongy, line 114: Why do you require `cmdline_colon_list`? Can't you replace `arg == colon_arg` with `":" in arg`?

tests/test_appids.py
Jongy, line 1: Let's add one test case for the negative case, i.e for `return "unknown app name"`.

Jongy (changes requested):
.

Jongy (approved):

PR #822, opened 2023-07-06
tests: aarch64: Run in CI

pfilipko1:
https://github.com/Granulate/gprofiler/issues/768


Jongy:
@pfilipko1 based on this
```
Run ./scripts/build_aarch64_container.sh
unknown flag: --platform
See 'docker --help'.

Usage:  docker [OPTIONS] COMMAND

A self-sufficient runtime for containers
```
I think we lack docker buildx on the runner, I suppose there's a publicly-available action to install it.

Jongy:
@pfilipko1 2 more notes:
1. I see space ran out on the runner - I'll work to get that fixed:
```
ERROR: failed to solve: failed to compute cache key: write /var/lib/buildkit/runc-overlayfs/snapshots/snapshots/40/fs/usr/local/go/src/strconv/quote_test.go: no space left on device
Error: Process completed with exit code 1.
```
2. I noticed a script I wrote in the past `setup_runner_requirements.sh` that we currently (in master) execute for the aarch64 build - the only one running on blank runners. You might need to add that as a step now in all jobs, keep that in mind.

Jongy:
@pfilipko1 to solve the dockerhub ratelimiting when using self-hosted runners, we can log in with the users configured in secrets. So basically, in all jobs - build exe, build container, test exe, test container - add these lines before the step to have it log in:
```
      - name: Login to DockerHub
        uses: docker/login-action@v1
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_TOKEN }}
```

PR #825, opened 2023-07-14
Fix getaddrinfo() EBUSY on ARM

d3dave:
CPython uses glibc's `getaddrinfo()` to resolve hostnames. glibc tries to dynamically load the backend libs that implement the actual resolution, `libnss_files.so` and `libnss_dns.so` (as configured by staticx), which are unfortunately missing. When running staticx, it determines the libs to include by examining dependencies using `ldd`. For this reason, staticx adds a "fake" dependency on these two libs by linking `libnssfix` -- staticx's injected lib to configure nss to use only `files` and `dns` backends and ignore system configuration -- with `nss_files` and `nss_dns`. When building for ARM, we recompile staticx v0.13.6 with a patch that fixes some other static NSS issues (resolved in v0.13.7+) instead of using the prebuilt version from pypi.

The nss libs were never intended to be loaded by anything other than glibc itself, thus the `lib.so -> lib.so.X` symlinks for the linker are not installed. In turn, the linker complains that it cannot find `-lnss_files` and `-lnss_dns`. It appears that this was overcome by linking against fake _static_ libraries, thus preventing the creation of dependencies for `ldd` to discover.

The solution is simply to replace these with the `*.so` symlinks `ld` expects.

## How Has This Been Tested?
Built and tested on graviton instance.

## Checklist:

### DIFF
--- a/pyi.Dockerfile
+++ b/pyi.Dockerfile
@@ -212,13 +212,6 @@ COPY granulate-utils/glogger granulate-utils/glogger
 RUN python3 -m pip install --no-cache-dir -r requirements.txt
 
 COPY exe-requirements.txt exe-requirements.txt
-# build on centos:8 of Aarch64 requires -lnss_files and -lnss_dns. the files are missing but the symbols
-# seem to be provided from another archive (e.g libc.a), so this "fix" bypasses the ld error of "missing -lnss..."
-# see https://github.com/JonathonReinhart/staticx/issues/219
-RUN if grep -q "CentOS Linux 8" /etc/os-release ; then \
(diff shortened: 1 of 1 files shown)


Jongy (approved):
Fix is understood, and is what was suggested in https://github.com/JonathonReinhart/staticx/issues/219 after all.

I'm still curious about - why did it trigger now? A change in libc used during static?

Jongy:
I assume you've verified that this fixes names resolution on Aarch64, right?

d3dave:
> I assume you've verified that this fixes names resolution on Aarch64, right?

yes

PR #828, opened 2023-07-18
main: disable pid ns check and disable perf when running on Fargate

d3dave:
Automatically disable pidns check and perf when running on Fargate.

## Checklist:

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -22,6 +22,7 @@ import humanfriendly
 from granulate_utils.linux.ns import is_running_in_init_pid
 from granulate_utils.linux.process import is_process_running
 from granulate_utils.metadata import Metadata
+from granulate_utils.metadata.cloud import get_aws_execution_env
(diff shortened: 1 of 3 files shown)


gprofiler/main.py
```diff
@@ -950,7 +951,7 @@ def main() -> None:
     args = parse_cmd_args()
     processes_to_profile = pids_to_processes(args)
 
-    if is_windows():
+    if is_windows() or get_aws_execution_env() == 'AWS_ECS_FARGATE':
```
Jongy, line 954: verify_preconditions still needs to run though

Jongy (changes requested):
Please also update the README - remove those flags from the suggested instructions, and in the appropriate location explain that gProfiler understand that it's running on Fargate and applies these 2 flags.

Jongy (approved):

gprofiler/main.py
Jongy: irrevalant - it's `if` not `if else`

Jongy:
merging master will resolve ci

PR #829, opened 2023-07-19
aarch64 CI test

pfilipko1:
https://github.com/Granulate/gprofiler/issues/768


.github/workflows/build-container-and-test-aarch64.yml
```diff
@@ -0,0 +1,83 @@
...
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+    # the tests need the gprofiler image built (from Dockerfile). I run it separately here, because "docker build" prints the build logs
+    # more nicely. the tests will then be able to use the built image.
+    - name: Build gProfiler image
+      run: ./scripts/build_aarch64_container.sh --output type=docker -t gprofiler 
+
+    - name: Export gProfiler image
+      run: mkdir -p output && docker image save gprofiler > output/gprofiler.img
```
Jongy, line 36:
```suggestion
      run: mkdir -p output && docker image save gprofiler > output/gprofiler_aarch64.img
```

.github/workflows/build-container-and-test-aarch64.yml
```diff
@@ -0,0 +1,83 @@
...
+
+    - name: Upload the image artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: gprofiler.img
```
Jongy, line 41:
```suggestion
        name: gprofiler_aarch64.img
```

tests/test.sh
```diff
@@ -35,6 +35,7 @@ if [ ! -f "$PERF_RESOURCE" ] && [[ "$*" != *"--executable"* ]]; then
 fi
 
 python3 -m pip install -q --upgrade setuptools pip
-python3 -m pip install -r ./requirements.txt -r ./exe-requirements.txt -r ./dev-requirements.txt
+python3 -m pip install --ignore-installed -r ./requirements.txt -r ./exe-requirements.txt -r ./dev-requirements.txt
```
Jongy, line 38: Why is `--ignore-installed` required?

.github/workflows/build-container-and-test-aarch64.yml
```diff
@@ -0,0 +1,83 @@
...
+
+jobs:
+  build-container-aarch64:
+    runs-on:
+      - self-hosted
+      - public
+      - ARM64
+
+    steps:
+    - name: change pycache permissions
```
Jongy, line 13: Why is this step required?

Jongy (changes requested):
I reviewed the changes. Note that I've merged #790 now, and added some small fixes in https://github.com/Granulate/gprofiler/pull/842.

Please merge https://github.com/Granulate/gprofiler/pull/842 into your branch (or merge master after https://github.com/Granulate/gprofiler/pull/842 is merged) and apply your relevant fixes on the new CI file `build-test-deploy.yml`.

The changes I see we'll need:
1. The current `build-container-aarch64` runs only on tags - remove that condition.
2. The current `build-container-aarch64` pushes the images immediately - instead, it should store it as an artifact `gprofiler_aarch64.img` so it's available for the `test-container-aarch64` job.
3. The `deploy-container` stage will need to download `gprofiler_aarch64.img` instead of pulling the image from DockerHub, as it does now.
4. Add the new job `test-container-aarch64` after `build-container-aarch64` in the file :)

.github/workflows/build-container-and-test-aarch64.yml
pfilipko1: I'm not sure whether its aarch64 or newest ubuntu, but packages installed by python3 -m pip are visible only for the root user. Easiest workaround I found (after failing with PYTHONPATH) is adding sudo to the test script. Then, after installation of pip packages there are some pycache dirs with incorrect permissions. This step changes it back.

pfilipko1: I experienced identical problems on our aarch64 ubuntu vm, so I assumed it always occurs with such setup combination.

Jongy (changes requested):
1. Please merge master and resolve conflicts, I merged a change to the docker buildkit usage that I want tested with this PR
2. Please remove `.github/workflows/build-container-and-test-aarch64.yml`

I'll review this tmrw

.github/workflows/build-test-deploy.yml
```diff
@@ -260,48 +262,85 @@ jobs:
...
+    runs-on:
+      - self-hosted
+      - public
+      - ARM64
+
     steps:
-      - name: Checkout Code
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
```
Jongy, line 270: Keep this `fetch-depth`? No reason to remove.

Plus, you're missing one level of indentation that you've removed - please bring it back to match other lists in this YAML file (where the `-` marking each entry is indented by 2 spaces)

.github/workflows/build-test-deploy.yml
```diff
@@ -260,48 +262,85 @@ jobs:
...
-          submodules: true
+    - name: Checkout Code
+      uses: actions/checkout@v3
+      with:
+        submodules: true
 
-      - name: Set up runner requirements
-        run: scripts/setup_runner_requirements.sh
```
Jongy, line 274: If the script is no longer used, please delete it in this PR.

.github/workflows/build-test-deploy.yml
```diff
@@ -260,48 +262,85 @@ jobs:
...
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v2
 
-      - name: Get and verify tag value
```
Jongy, line 276: Please keep this step - don't remove in this PR, it's an unrelated change.

.github/workflows/build-test-deploy.yml
```diff
@@ -260,48 +262,85 @@ jobs:
...
 
-      - name: Login to DockerHub
-        uses: docker/login-action@v1
-        with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+    # the tests need the gprofiler image built (from Dockerfile). I run it separately here, because "docker build" prints the build logs
+    # more nicely. the tests will then be able to use the built image.
+    - name: Build gProfiler image
+      run: ./scripts/build_aarch64_container.sh -t gprofiler_aarch64 --output type=docker
```
Jongy, line 288: Since https://github.com/Granulate/gprofiler/pull/790, the gProfiler image is a blank image that contains the executable.

Please see the job `build-container-x64` - it has
```
    needs:
      - build-executable-x64
```
to make sure the CI runs it after the exe is built, and then it uses the same binary exe built earlier to put it into the image, as done here:
```
      run: ./scripts/build_x86_64_container.sh --skip-exe-build --build-arg EXE_PATH=output/gprofiler_x86_64 -t gprofiler_x86_64
```

Please do the same here - by not passing `--skip-exe-build` you are rebuilding it unnecessarily, and plus, in any case, we'd like them to use the same file exactly.

.github/workflows/build-test-deploy.yml
```diff
@@ -260,48 +262,85 @@ jobs:
...
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
 
-          # build & push Aarch64
-          ./scripts/build_aarch64_container.sh --skip-exe-build --build-arg EXE_PATH=output/gprofiler_aarch64 -t "$AARCH64_IMAGE" --push
+    - name: Checkout Code
+      uses: actions/checkout@v3
+      with:
+        submodules: true
+
+    - name: Download the executable from previous job
```
Jongy, line 320: "Download the image from previous job", no?

.github/workflows/build-test-deploy.yml
```diff
@@ -260,48 +262,85 @@ jobs:
...
+      uses: actions/download-artifact@v2
+      with:
+        name: gprofiler_aarch64.img
+        path: output/
 
+    - name: Import gProfiler image
+      run: docker image load < output/gprofiler_aarch64.img
+
+    - name: Extract resources from gProfiler executable
```
Jongy, line 329: After you follow my suggestion from https://github.com/Granulate/gprofiler/pull/829/files#r1314022332 you can also simplify the code here to use the gProfiler executable instead of image - so it can run the exe and don't need the docker, the mappings etc, just like the x86_64 job.

.github/workflows/build-test-deploy.yml
```diff
@@ -260,48 +262,85 @@ jobs:
...
+      run: sudo docker run -v $PWD/gprofiler/resources:/app/gprofiler/resources gprofiler_aarch64 extract-resources --resources-dest=/app/gprofiler/resources
+
+    - name: setup runner requirements
+      run: ./scripts/setup_runner_requirements.sh
+
+    # TODO: Add docker layer caching when GitHub Actions cache is stabilized and works good with "satackey/action-docker-layer-caching@v0.0.11"
```
Jongy, line 335: This comment is not relevant here, remove pls

.github/workflows/build-test-deploy.yml
```diff
@@ -260,48 +262,85 @@ jobs:
...
+    - name: Run gProfiler tests
+      run: |
+        sudo python3 -m pip install virtualenv
+        sudo virtualenv --python python3.8 aarch64-venv
+        sudo chown -R ${USER} aarch64-venv
+        sudo chmod -R a+rX aarch64-venv
+        source aarch64-venv/bin/activate
+        ./tests/test.sh --ignore=tests/test_executable.py
   deploy-container:
```
Jongy, line 344: Please add a blank before this

.github/workflows/build-test-deploy.yml
```diff
@@ -260,48 +262,85 @@ jobs:
...
+      run: docker image load < output/gprofiler_aarch64.img
+
+    - name: Extract resources from gProfiler executable
+      run: sudo docker run -v $PWD/gprofiler/resources:/app/gprofiler/resources gprofiler_aarch64 extract-resources --resources-dest=/app/gprofiler/resources
+
+    - name: setup runner requirements
+      run: ./scripts/setup_runner_requirements.sh
+
+    # TODO: Add docker layer caching when GitHub Actions cache is stabilized and works good with "satackey/action-docker-layer-caching@v0.0.11"
+    - name: Run gProfiler tests
```
Jongy, line 336: The x86_64 tests run on 3.8, 3.9, 3.10.

I'm actually planning to deprecate 3.8 and 3.9 soon so there's no need to test all 3 - let's focus on 3.10.

Can you use the GH action to install python? i.e
```
    steps:
      - name: Set up Python 3.10
        uses: actions/setup-python@v2
        with:
          python-version: "3.10"
```

Jongy: You know what, let's actually do it with `strategy` like done in the x86_64 tests:
```
    strategy:
      matrix:
        python-version:
          - "3.10"
```
this is extendable for a future where we'll want to add 3.11, and also shows up nicely in the job name (gonna be "test-container-aarch64 (3.10)" like we have for the x86 ones)

Jongy (changes requested):
Gave some comments.

This doesn't include `test-executable-aarch64` but I'm okay with adding it later.

Note that CICD is failing - seems like a Docker issue where we're not using BuiltKit properly (BuildKit recognizes `--platform` and standard Docker doesn't iirc)

.github/workflows/build-test-deploy.yml
roi-granulate: 933546bc4e20470e04d6bdf925072433b8cbe3d5

.github/workflows/build-test-deploy.yml
```diff
@@ -122,11 +121,6 @@ jobs:
      - name: Set up requirements
        run: scripts/setup_runner_requirements.sh
 
-     - name: Get and verify tag value
```
Jongy, line 125: Hmm i'll bring it back so it's consistent with x86_64

.github/workflows/build-test-deploy.yml
```diff
@@ -287,37 +340,83 @@ jobs:
           fetch-depth: 0
           submodules: true
 
-      - name: Set up runner requirements
-        run: scripts/setup_runner_requirements.sh
+      - name: setup runner requirements
+        run: ./scripts/setup_runner_requirements.sh
 
       - name: Get and verify tag value
-        run: ./scripts/verify_tag.sh
```
Jongy, line 294: Need sudo?

scripts/cmp_tags.sh
```diff
@@ -8,7 +8,7 @@ set -euo pipefail
 # Used in CI and checks that last pushed tag is greater than last existing tag.
 # Using python package 'cmp_version' to do the compare work
 
-pip install cmp_version
```
Jongy, line 11: Need it?

Issue #675, opened 2023-02-05
java: Support retrieving "java -version" for profiled processes which are not "java" basename


Jongy:
As part of metadata collection, gProfiler collects the output of `java -version` for profiled Java application. This is doable when the target process is basenamed `java` - in that case we know that we can execute the same path with `-version` argument, and get the version.

However, since our process selection for Java processes is based on processes which have `libjvm.so` loaded (see `JavaProfiler._select_processes_to_profile()`), we stumble upon Java processes which are other executables loading `libjvm.so`. One of them is https://commons.apache.org/proper/commons-daemon/jsvc.html for example. These do not necessarily support `-version`, so we don't do the version check.

### Proposed solutions:
* As suggested [here](https://github.com/Granulate/gprofiler/blob/80eebf84e4ca8c4f0469b34c1edf79cdeecd8d7b/gprofiler/profilers/java.py#L892) - try to get the `java` binary path for the same Java installation, then use `java -version` from that binary.
* Create a whitelist of other basenames and provide specific commandlines to handle them. For example, jsvc per its documentation claims to support `jsvc -version`, so we can just do that. There might be others with different options. I will update this ticket with the different non-`java` I see.

### Sites to fix
* https://github.com/Granulate/gprofiler/blob/80eebf84e4ca8c4f0469b34c1edf79cdeecd8d7b/gprofiler/profilers/java.py#L257
* https://github.com/Granulate/gprofiler/blob/80eebf84e4ca8c4f0469b34c1edf79cdeecd8d7b/gprofiler/profilers/java.py#L892

PR #831, opened 2023-07-26
Support extracting java version for non-java named processes

marcin-ol:
## Related Issue
https://github.com/Granulate/gprofiler/issues/675

## Motivation and Context
Some java processes can be started from a binary that doesn't start with java. We want to support profiling them and provide full metadata including java version.

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -261,9 +261,20 @@ def _get_process_ns_java_path(process: Process) -> Optional[str]:
     """
     Look up path to java executable installed together with this process' libjvm.
     """
+    # This has the benefit of working even if the Java binary was replaced, e.g due to an upgrade.
+    # in that case, the libraries would have been replaced as well, and therefore we're actually checking
+    # the version of the now installed Java, and not the running one.
+    # but since this is used for the "JDK type" check, it's good enough - we don't expect that to change.
+    # this whole check, however, is growing to be too complex, and we should consider other approaches
+    # for it:
+    # 1. purely in async-profiler - before calling any APIs that might harm blacklisted JDKs, we can
+    #    check the JDK type in async-profiler itself.
+    # 2. assume JDK type by the path, e.g the "java" Docker image has
+    #    "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java" which means "OpenJDK". needs to be checked for
+    #    other JDK types.
     java_path: Optional[str] = None
-    nspid = get_process_nspid(process.pid)
     if is_java_basename(process):
+        nspid = get_process_nspid(process.pid)
         java_path = f"/proc/{nspid}/exe"  # it's a symlink and will be resolveable under process' mnt ns
     else:
         libjvm_path: Optional[str] = None
@@ -280,10 +291,11 @@ def _get_process_ns_java_path(process: Process) -> Optional[str]:
             Path(libjvm_dir, "../../../bin/java").resolve(),
         ]
         for p in java_candidate_paths:
+            # don't need resolve_proc_root_links here - paths in /proc/pid/maps are normalized.
             proc_relative_path = Path(f"/proc/{process.pid}/root", p.relative_to("/"))
             if proc_relative_path.exists():
                 if os.access(proc_relative_path, os.X_OK):
-                    java_path = os.path.join(f"/proc/{nspid}/root", p)
+                    java_path = str(p)
                     break
     return java_path
 
@@ -295,24 +307,11 @@ def get_java_version(process: Process, stop_event: Event) -> Optional[str]:
     process_java_path = _get_process_ns_java_path(process)
     if process_java_path is None:
         return None
-    java_path = process_java_path
-
-    # this has the benefit of working even if the Java binary was replaced, e.g due to an upgrade.
-    # in that case, the libraries would have been replaced as well, and therefore we're actually checking
-    # the version of the now installed Java, and not the running one.
-    # but since this is used for the "JDK type" check, it's good enough - we don't expect that to change.
-    # this whole check, however, is growing to be too complex, and we should consider other approaches
-    # for it:
-    # 1. purely in async-profiler - before calling any APIs that might harm blacklisted JDKs, we can
-    #    check the JDK type in async-profiler itself.
-    # 2. assume JDK type by the path, e.g the "java" Docker image has
-    #    "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java" which means "OpenJDK". needs to be checked for
-    #    other JDK types.
 
     def _run_java_version() -> "CompletedProcess[bytes]":
         return run_process(
             [
-                java_path,
+                process_java_path,  # type: ignore # checked to be not-None above
                 "-version",
             ],
             stop_event=stop_event,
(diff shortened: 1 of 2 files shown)


gprofiler/profilers/java.py
```diff
@@ -257,14 +257,45 @@ def is_java_basename(process: Process) -> bool:
...
-
-    nspid = get_process_nspid(process.pid)
+def get_java_version(process: Process, stop_event: Event) -> Optional[str]:
+    # make sure we're able to find "java" binary bundled with process libjvm
+    process_java_path = _get_process_ns_java_path(process)
+    if process_java_path is None:
+        return None
+    java_path = process_java_path
 
     # this has the benefit of working even if the Java binary was replaced, e.g due to an upgrade.
```
Jongy, line 300: This comment is no longer relevant here - it refers to the assignment of `/proc/nspid/exe` which now happens in `_get_process_ns_java_path`. Move it to the relevant site?

gprofiler/profilers/java.py
```diff
@@ -257,14 +257,45 @@ def is_java_basename(process: Process) -> bool:
...
+        if libjvm_path is None:
+            return None
+        libjvm_dir = os.path.dirname(libjvm_path)
+        # support two java layouts - it's either lib/server/../../bin/java or lib/{arch}/server/../../../bin/java:
+        java_candidate_paths = [
+            Path(libjvm_dir, "../../bin/java").resolve(),
+            Path(libjvm_dir, "../../../bin/java").resolve(),
+        ]
+        for p in java_candidate_paths:
+            proc_relative_path = Path(f"/proc/{process.pid}/root", p.relative_to("/"))
```
Jongy, line 283: You need to use `resolve_proc_root_links` here, otherwise this naive join of paths might escape symlinks. Please read that function's docs to understand how to use, and use it here.

gprofiler/profilers/java.py
```diff
@@ -257,14 +257,45 @@ def is_java_basename(process: Process) -> bool:
...
+            if proc_relative_path.exists():
+                if os.access(proc_relative_path, os.X_OK):
+                    java_path = os.path.join(f"/proc/{nspid}/root", p)
```
Jongy, line 286: This doesn't make much sense to me. This path will be accessed inside the target process mount ns - so there's no reason to use `/proc/nspid/root`, as `/proc/nspid/root` is logically equivalent to `/`.

gprofiler/profilers/java.py
```diff
@@ -257,14 +257,45 @@ def is_java_basename(process: Process) -> bool:
...
-    # our callers should check for it.
-    assert is_java_basename(process), f"expected java, found {process!r}"
-
-    nspid = get_process_nspid(process.pid)
+def get_java_version(process: Process, stop_event: Event) -> Optional[str]:
+    # make sure we're able to find "java" binary bundled with process libjvm
+    process_java_path = _get_process_ns_java_path(process)
+    if process_java_path is None:
+        return None
+    java_path = process_java_path
```
Jongy, line 298: Any reason to assign it again? The name you gave it (`process_java_path`) is actually better than the previous name `java_path`, so you can just keep it :)

gprofiler/profilers/java.py
```diff
@@ -312,10 +343,7 @@ def __init__(
         self.java_collect_jvm_flags = java_collect_jvm_flags
 
     def make_application_metadata(self, process: Process) -> Dict[str, Any]:
-        if is_java_basename(process):
-            version = get_java_version(process, self._stop_event)
-        else:
-            version = "not /java"
+        version = get_java_version(process, self._stop_event) or "not /java"
```
Jongy, line 346: Is "not /java" still correct? We do handle "not /java" now. I'd use "/java not found"?

tests/test_java.py
```diff
@@ -670,31 +672,60 @@ def test_java_different_basename(
...
+        application_docker_command=[
+            "bash",
+            "-c",
+            f"{java_notjava_basename} -jar Fibonacci.jar",
+        ],
+    ) as container:
+        logging.warning(container.logs())
+        application_pid = container.attrs["State"]["Pid"]
+        if libjvm_removed:
+            # test extracting java version even if libjvm was replaced
```
Jongy, line 712: Copied from `test_java_deleted_libjvm` right? Let's extract this to a tests util function `simulate_libjvm_delete`. Your code here misses the last `is_libjvm_deleted` assert btw, would be no issue if it was a shared function :)

tests/test_java.py
```diff
@@ -670,31 +672,60 @@ def test_java_different_basename(
...
+        application_docker_image,
+        [],
+        [],
+        application_docker_command=[
+            "bash",
+            "-c",
+            f"{java_notjava_basename} -jar Fibonacci.jar",
+        ],
+    ) as container:
+        logging.warning(container.logs())
```
Jongy, line 709: Debug, I suppose? remove?

tests/test_java.py
```diff
@@ -670,31 +672,60 @@ def test_java_different_basename(
...
+            libjvm_tmp = libjvm + "."
+            shutil.copy(libjvm, libjvm_tmp)
+            os.unlink(libjvm)
+            os.rename(libjvm_tmp, libjvm)
+        with make_java_profiler(
+            profiler_state,
+            duration=1,
+            java_safemode=JAVA_SAFEMODE_ALL,  # explicitly enable, for basename checks
+        ) as profiler:
+            profile = snapshot_pid_profile(profiler, application_pid)
```
Jongy, line 724: Any log line you can catch with caplog here to prove that the `/java` binary found via libjvm relative paths is the one used to extract the version? I'm thinking the log record from the `run_process` maybe?

tests/test_java.py
```diff
@@ -670,31 +672,60 @@ def test_java_different_basename(
...
-                    filter(
-                        lambda r: r.message
-                        == "Non-java basenamed process (cannot get Java version), skipping... (disable"
-                        " --java-safemode=java-extended-version-checks to profile it anyway)",
-                        caplog.records,
-                    )
-                )
-                assert len(log_records) == 1
-                assert os.path.basename(log_record_extra(log_records[0])["exe"]) == java_notjava_basename
+            assert profile.app_metadata is not None
```
Jongy, line 675: You merged the 2 cases but technically the 3rd case (that previously triggered the "non-java basename") is now handled via libjvm. Right?

Jongy (changes requested):
Looks good! Gave some comments.

gprofiler/profilers/java.py
marcin-ol: That function call is not needed here - we're basing search on libjvm path taken from `process.memory_maps()`, which is normalized by system.

gprofiler/profilers/java.py
```diff
@@ -277,12 +272,46 @@ def get_java_version(process: Process, stop_event: Event) -> str:
     # 2. assume JDK type by the path, e.g the "java" Docker image has
     #    "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java" which means "OpenJDK". needs to be checked for
     #    other JDK types.
-    java_path = f"/proc/{nspid}/exe"
+    java_path: Optional[str] = None
```
Jongy, line 275: I think we can ditch the attempt to have a single return-point in this function - remove this var, use `return` wherever assigning to it, and add `return None` last

gprofiler/profilers/java.py
```diff
@@ -277,12 +272,46 @@ def get_java_version(process: Process, stop_event: Event) -> str:
...
+    # make sure we're able to find "java" binary bundled with process libjvm
+    process_java_path = _get_process_ns_java_path(process)
+    if process_java_path is None:
+        return None
 
     def _run_java_version() -> "CompletedProcess[bytes]":
         return run_process(
             [
-                java_path,
+                process_java_path,  # type: ignore # checked to be not-None above
```
Jongy, line 314:
```suggestion
                cast(str, process_java_path),
```
always nicer than `type: ignore`

Jongy (changes requested):
Done reviewing

gprofiler/profilers/java.py
marcin-ol: 👍

gprofiler/profilers/java.py
marcin-ol: 👍

Jongy (approved):

PR #832, opened 2023-07-26
Ensure directories used by gProfiler to host async-profiler DSO are owned by root

Jongy:


d3dave (approved):

Jongy:
I sanity tested that this indeed re-creates `/tmp/gprofiler_tmp` both for gprofiler itself on the host, and for a Java container.

Issue #817, opened 2023-06-27
Identify golang binaries w/o requiring symbols


Jongy:
With https://github.com/Granulate/gprofiler/issues/804 we now collect stripped / not strpped for golang. This is not proving useful because we IDENTFIY golang processes using symbols, so obv, if a program is strpped, it is not considered golang (that stripped my mind).

Our goal with https://github.com/Granulate/gprofiler/issues/804 is still to be able to tell for a golang program if it's stripped or not (because it happens often and then the user needs to rebuild it with symbols). For this purpose I want us to switch identification so it doesn't depend on symbols.

I don't have a concrete idea but I suppose that searching for some magic strings / other details in the ELF might work.

Jongy:
When you implement it, make sure to update the relevant README section [here](https://github.com/Granulate/gprofiler#golang-profiling-options) which claims that symbols are needed for identification.

Jongy:
Before implementing please discuss your suggestions here.

marcin-ol:
Golang binaries, even the stripped ones can be identified by ELF Note section `note.go.buildid`, unless intentionally transformed beyond recognition (e.g.: `strip --remove`).

Found this symbol in binaries built by golang versions 1.5 .. 1.20+:
```console
$ grep . -r --include="elfheaders*" -Hne 'go.buildid'
./elfheaders-1.7.txt:47:  [11] .note.go.buildid
./elfheaders-1.20.txt:49:  [12] .note.go.buildid
./elfheaders-1.16.txt:49:  [12] .note.go.buildid
./elfheaders-1.8.txt:47:  [11] .note.go.buildid
./elfheaders-1.12.txt:47:  [11] .note.go.buildid
./elfheaders-1.5.txt:45:  [10] .note.go.buildid
./elfheaders-1.6.txt:45:  [10] .note.go.buildid
```

Support for this has been added in Golang 1.5. See references.

This works nicely and could be implemented in granulate-utils:
```
[2023-08-03 09:36:17,853] DEBUG: exe=/proc/318897/exe; found note: {'n_name': 'Go', 'n_desc': 'iET6gPjQs8_ercVKmI_5/V4ftLGgOIqqHxmcxKKUT/nXzOVLSE6qHN-pQLyuU-/iET6gPjQs8_ercVKmI_5'}
```

References:
- issue introducing buildid section in golang linker (tagged for 1.5 milestone): https://github.com/golang/go/issues/11048,
[...]

PR #835, opened 2023-08-04
Use and test golang binary identification

marcin-ol:
Use golang binary identification brought to granulate-utils with PR https://github.com/Granulate/granulate-utils/pull/179.

## Description
With new mechanism in granulate-utils it's now possible to identify golang binary looking only at ELF sections. This PR is adopting this mechanism.

## Related Issue
#817 

## How Has This Been Tested?
Added new testcase for a stripped golang binary, proving identification of golang programs is possible.

## Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -429,6 +429,14 @@ def application_docker_capabilities() -> List[str]:
 
 
 @fixture
+def application_docker_command() -> Optional[List[str]]:
+    """
+    Command to issue to application docker as an override.
+    """
+    return None
+
+
+@fixture
 def application_docker_container(
     in_container: bool,
     docker_client: DockerClient,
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -436,13 +444,18 @@ def application_docker_container(
...
     application_docker_mounts: List[docker.types.Mount],
     application_docker_capabilities: List[str],
+    application_docker_command: Optional[List[str]],
 ) -> Iterable[Container]:
     if not in_container:
         yield None
         return
     else:
         with _application_docker_container(
-            docker_client, application_docker_image, application_docker_mounts, application_docker_capabilities
```
Jongy, line 445: Time to use kwargs here, I think

marcin-ol: Keyword-only? Yeah, I agree.
Yet I would leave `docker_client`, `docker_image` as positional, as they will always be mandatory.

Jongy:
I merged https://github.com/Granulate/granulate-utils/pull/179 - let's update the revision here and merge this

Jongy (approved):

PR #836, opened 2023-08-06
Add ELF arch as app metadata

Jongy:
Needs https://github.com/Granulate/granulate-utils/pull/182 and https://github.com/Granulate/granulate-utils/pull/180

### DIFF
--- a/gprofiler/metadata/application_metadata.py
+++ b/gprofiler/metadata/application_metadata.py
@@ -5,7 +5,7 @@
 
 import functools
 from threading import Event, Lock
-from typing import Any, Dict, Optional
+from typing import Any, Callable, Dict, Optional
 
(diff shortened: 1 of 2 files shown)


gprofiler/metadata/application_metadata.py
```diff
@@ -87,4 +88,19 @@ def make_application_metadata(self, process: Process) -> Dict[str, Any]:
             execfn = f"error: {e.__class__.__name__}"
         md["execfn"] = execfn
 
+        try:
```
michelhe, line 91: Consider refactoring the try-except to a function instead of duping it for every call to granulate_utils

Jongy: Hai

michelhe (approved):

michelhe (approved):

PR #837, opened 2023-08-07
Fix arch metadata with Aarch64

Jongy:
We really do need Aarch64 tests in the CI 😅


gprofiler/metadata/application_metadata.py
```diff
@@ -76,7 +76,7 @@ def _wrap_errors(fn: Callable[[Process], str], exception_log: str) -> str:
             except (NoSuchProcess, ZombieProcess):
                 raise  # let caller handle
             except Exception as e:
-                logger.exception(process, pid=process.pid)
+                logger.exception(exception_log, pid=process.pid)
```
Jongy, line 79: I'm saddened that flake8 doesn't catch this

PR #859, opened 2023-11-22
ci: Fix aarch64 container build & additional cleanups

Jongy:


Jongy:
Diff is okay, I'll merge and release a version that'll finally build the container version...

PR #860, opened 2023-11-24
java: Fix regression _find_rw_exec_dir() failing to handle ro dirs

Jongy:
Fixes a regression from https://github.com/Granulate/gprofiler/commit/6360449252793ad59802a3c09fde297646529b7d, and improves the tests
so we'd catch it next time :)

### DIFF
--- a/gprofiler/profilers/java.py
+++ b/gprofiler/profilers/java.py
@@ -2,6 +2,7 @@
 # Copyright (c) Granulate. All rights reserved.
 # Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
 #
+import errno
 import functools
 import json
 import os
@@ -540,7 +541,13 @@ class AsyncProfiledProcess:
             if not is_owned_by_root(full_dir.parent):
                 continue  # the parent needs to be owned by root
 
-            mkdir_owned_root(full_dir)
+            try:
+                mkdir_owned_root(full_dir)
+            except OSError as e:
+                # dir is not r/w, try next one
+                if e.errno == errno.EROFS:
+                    continue
+                raise
 
             if is_rw_exec_dir(full_dir):
                 return str(full_dir)
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -408,3 +408,10 @@ def log_record_extra(r: LogRecord) -> Dict[Any, Any]:
     Gets the "extra" attached to a LogRecord
     """
     return getattr(r, "extra", {})
+
+
+def str_removesuffix(s: str, suffix: str, assert_suffixed: bool = True) -> str:
+    suffixed = s.endswith(suffix)
+    if assert_suffixed:
+        assert suffixed
+    return s[: -len(suffix)] if suffixed else s
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -649,7 +649,8 @@ def python_version(in_container: bool, application_docker_container: Container) 
 
 
 @fixture
-def noexec_tmp_dir(in_container: bool, tmp_path: Path) -> Iterator[str]:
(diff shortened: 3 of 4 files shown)


Jongy:
I will also run one negative test - undo the fix and ensure that the new test fails.

Jongy:
Commenting-out the fixing code in the plugin - the tests fail as expected.
```
tests/test_java.py::test_java_noexec_or_ro_dirs[on_host-noexec] PASSED                                                                                                                                                               [ 25%]
tests/test_java.py::test_java_noexec_or_ro_dirs[on_host-ro] FAILED                                                                                                                                                                   [ 50%]
tests/test_java.py::test_java_noexec_or_ro_dirs[in_container-noexec] PASSED                                                                                                                                                          [ 75%]
tests/test_java.py::test_java_noexec_or_ro_dirs[in_container-ro] FAILED
```

```
  File "/Granulate/gprofiler/gprofiler/profilers/profiler_base.py", line 132, in _wait_for_profiles
    result = future.result()
    return self.__get_result()
    result = self.fn(*self.args, **self.kwargs)                                                                                                                                                                                    [98/1807]
  File "/Granulate/gprofiler/gprofiler/profilers/java.py", line 1211, in _profile_process
    with AsyncProfiledProcess(
  File "/Granulate/gprofiler/gprofiler/profilers/java.py", line 498, in __init__
    self._ap_dir_base = self._find_rw_exec_dir()
  File "/Granulate/gprofiler/gprofiler/profilers/java.py", line 545, in _find_rw_exec_dir
    mkdir_owned_root(full_dir)
  File "/Granulate/gprofiler/gprofiler/utils/fs.py", line 91, in mkdir_owned_root
    os.mkdir(path, mode=mode)
OSError: [Errno 30] Read-only file system: '/proc/1/root/tmp/pytest-of-root/pytest-22/test_java_noexec_or_ro_dirs_on1/tmpfs/gprofiler_tmp'
```

gprofiler/profilers/java.py
```diff
@@ -540,7 +541,13 @@ def _find_rw_exec_dir(self) -> str:
...
                 continue  # the parent needs to be owned by root
 
-            mkdir_owned_root(full_dir)
+            try:
+                mkdir_owned_root(full_dir)
+            except OSError as e:
+                # dir is not r/w, try next one
+                if e.errno == errno.EROFS:
+                    continue
+                raise
```
michelhe, line 550: Exception-based programming is kind weird here, why not simply adding `if not is_rw_exec_dir(full_dir.parent): continue ` at the beginning of the iteration?

tests/utils.py
```diff
@@ -408,3 +408,10 @@ def log_record_extra(r: LogRecord) -> Dict[Any, Any]:
     Gets the "extra" attached to a LogRecord
     """
     return getattr(r, "extra", {})
+
+
+def str_removesuffix(s: str, suffix: str, assert_suffixed: bool = True) -> str:
+    suffixed = s.endswith(suffix)
+    if assert_suffixed:
+        assert suffixed
+    return s[: -len(suffix)] if suffixed else s
```
michelhe, line 417: Are we supporting python<3.9? if not, no need to reimplement its a builtin function after 3.9

Jongy: We do

gprofiler/profilers/java.py
Jongy: Well, the exception-based programming here is because I don't want to "check" if the dir is ro, I want to try and create it, and fail otherwise. `is_rw_exec_dir` is written exception-based too, because it's simpler & more correct to just try and do it. "noexec" for example is not easily tested w/o just TRYING to execute and seeing if it fails

michelhe (approved):

PR #866, opened 2023-12-15
remove SparkSampler

slicklash:
### Description

remove SparkSampler from gProfiler


Jongy (changes requested):
Users are embedding the installation command - they might have embedded it with `--spark-sample-period` or with `--collect-spark-metrics`. I suggest we keep both args, with `help="Deprecated! Removed in version 1.42.0"`, and if any of the args is passed, use `logger.warning` to submit a warning about them not being accepted anymore.

Besides that, looks good, I just want to see the CI passing once (the failure is flaky so let's re-run)

Jongy (approved):

Issue #820, opened 2023-07-04
perf: Upgrade to Linux v6.4


Jongy:
I have an [existing PR to upgrade for v6.3](https://github.com/Granulate/gprofiler/pull/752). Since then, v6.4 was released, so we can use it.
Use branch https://github.com/Granulate/linux/tree/granulate-perf-v6.3 and merge the latest kernel release into it, then test gProfiler with it.

roi-granulate:
perf was upgraded to v6.7 in #888

PR #888, opened 2024-02-22
resource/perf: Upgrade `perf` to v6.7

roi-granulate:
## Checklist:

### DIFF
--- a/scripts/perf_build.sh
+++ b/scripts/perf_build.sh
@@ -6,7 +6,7 @@
 set -euo pipefail
 
 # downloading the zip because the git is very large (this is also very large, but still smaller)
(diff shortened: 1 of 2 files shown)


scripts/perf_build.sh
```diff
@@ -44,8 +44,8 @@ EOF
 
 fi
 
-make -C tools/perf LDFLAGS=-static -j 8 perf
```
roi-granulate, line 47: why was it fixed 8?

Jongy: No particular reason, can use -j

Jongy:
Closes: https://github.com/Granulate/gprofiler/issues/820

Jongy (approved):

PR #890, opened 2024-02-28
Change license to apache 2

motisaroka:
## Description
Changing the license to Apache 2

## Checklist:


roi-granulate:
@motisaroka [this test](https://github.com/Granulate/gprofiler/blob/c3345ebdcf622e9fe6306e474a8c1813924e065a/tests/test_java.py#L1279) fails because the line number changed, it [was line no. 9](https://github.com/Granulate/gprofiler/blob/c3345ebdcf622e9fe6306e474a8c1813924e065a/tests/containers/java/Fibonacci.java#L9) and now it's [line no. 20
](https://github.com/Granulate/gprofiler/blob/cccc80cc7c43468868244be7a4d8ede7d38556fa/tests/containers/java/Fibonacci.java#L20)the line number should be changed [here](https://github.com/Granulate/gprofiler/blob/c3345ebdcf622e9fe6306e474a8c1813924e065a/tests/test_java.py#L1284)
meaning - `function_with_line_numbers = "Fibonacci.fibonacci:9(I)J_[j]"` should now be `function_with_line_numbers = "Fibonacci.fibonacci:20(I)J_[j]"`

deploy/ansible/gprofiler_playbook.yml
```diff
@@ -1,6 +1,17 @@
 #
-# Copyright (c) Granulate. All rights reserved.
-# Licensed under the AGPL3 License. See LICENSE.md in the project root for license information.
+# Copyright (C) 2023 Intel Corporation
```
roi-granulate, line 2: shouldn't it be 2024?

motisaroka: Nope it should be 2023 which is the date Intel acquired us

roi-granulate (changes requested):
note - I believe there are some files that did not include a license, hence - you didn't apply license change upon them.

roi-granulate:
I see that `granulate-utils` already introduce such change, shouldn't this PR also sync `granulate-utils`?

motisaroka:
As discussed I replaced the current copyright with new one matches Apache 2.0 copyrights.
I did not remove/add new copyright from existing files.
Indeed there are some files that has no copyrights at all.
It is true that Granulate utils already changed its license and it is better to change this one ASAP

roi-granulate:
> It is true that Granulate utils already changed its license and it is better to change this one ASAP

I didn't fully understand, should this PR upgrade gProfilers' `granulate-utils`?

motisaroka:
> > It is true that Granulate utils already changed its license and it is better to change this one ASAP
> 
> I didn't fully understand, should this PR upgrade gProfilers' `granulate-utils`, as part of the license change?

gProfielr agent is using the granulate-utils code, they are both must be running the same license, this is why we are upgrading them both

Jongy (commented):
I reviewed, looks ok.

motisaroka:
Year updated to 2022

gprofiler/metadata/application_metadata.py
```diff
@@ -98,9 +98,11 @@ def _wrap_errors(fn: Callable[[Process], str], exception_log: str) -> str:
         # take arch from the executed elf, not the host system, because (although unlikely) it's possible
         # that the process runs a different, emulated architecture.
         md["arch"] = _wrap_errors(
-            lambda p: "error: not supported on Windows"
-            if is_windows()
-            else elf_arch_to_uname_arch(get_elf_arch(f"/proc/{process.pid}/exe")),
+            lambda p: (
```
Jongy, line 101: Makes more sense this way

Jongy (approved):

PR #909, opened 2024-06-10
fix grcp hangs and upgrade grpcio to 1.62.2

slicklash:
### Problem

- `grcp` hangs (broken threading control) when multiple CRI clients are trying to query newly created container with (`thd.cc:160  pthread_create failed: Invalid argument` [source](https://github.com/grpc/grpc/blob/e4b7d46f4305da721d4e29a9936fc91634ddcd23/src/core/lib/gprpp/posix/thd.cc#L160) )

### Solution

use single CRI client

### DIFF
--- a/gprofiler/containers_client.py
+++ b/gprofiler/containers_client.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from threading import Lock
 from typing import Dict, List, Optional, Set
 
 from granulate_utils.containers.client import ContainersClient
@@ -26,15 +27,22 @@ from gprofiler.utils.perf import valid_perf_pid
 logger = get_logger_adapter(__name__)
 
 _containers_client: Optional[ContainersClient] = None
+_containers_client_lock = Lock()
+
+
+def get_containers_client() -> ContainersClient:
+    global _containers_client
+    with _containers_client_lock:
+        if _containers_client is None:
+            _containers_client = ContainersClient()
+        return _containers_client
 
 
 class ContainerNamesClient:
     def __init__(self) -> None:
-        global _containers_client
         try:
-            if _containers_client is None:
-                _containers_client = ContainersClient()
-            logger.info(f"Discovered container runtimes: {_containers_client.get_runtimes()}")
+            self._containers_client: Optional[ContainersClient] = get_containers_client()
+            logger.info(f"Discovered container runtimes: {self._containers_client.get_runtimes()}")
         except NoContainerRuntimesError:
             logger.warning(
                 "Could not find a Docker daemon or CRI-compatible daemon, profiling data will not"
@@ -42,6 +50,7 @@ class ContainerNamesClient:
                 " please open a new issue here:"
                 " https://github.com/Granulate/gprofiler/issues/new"
             )
+            self._containers_client = None
 
         self._pid_to_container_name_cache: Dict[int, str] = {}
         self._current_container_names: Set[str] = set()
@@ -56,7 +65,7 @@ class ContainerNamesClient:
         return list(self._current_container_names)
 
     def get_container_name(self, pid: int) -> str:
-        if _containers_client is None:
+        if self._containers_client is None:
             return ""
 
         if not valid_perf_pid(pid):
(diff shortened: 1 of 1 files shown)


Jongy (changes requested):
Nice find!

Some questions:
1. Any particular reason you deprecated `--databricks-job-name-as-service-name`?
2. The change in `_safely_get_process_container_name` is bad - the 2s is arbitrary (so in production we could still hang in some cases, for all I know), plus it adds unexpected delays in random call sites in gProfiler, and those delays are not necessarily parallelized. Can we understand better why "new containers cause hangs" and add more delicate handling for the particular case, and not something as common as "less than 2s"? Btw, is it not related to the `preexec_fn` issue? I mean, by removing `preexec_fn` and not applying this fix, do we still see it?
3. Our preexec fn was `os.setpgrp` and/or `prctl(PR_SET_PDEATHSIG)`. Does `start_new_session` cover them both?

Jongy:
> start_new_session covers os.setpgrp, it's not possible to set PR_SET_PDEATHSIG without preexec_fn

But we needed the PR_SET_PDEATHSIG :/ If we give up on it, we need to verify that all existing uses of `start_process` are safe even w/o PR_SET_PDEATHSIG.

Jongy:
> > > start_new_session covers os.setpgrp, it's not possible to set PR_SET_PDEATHSIG without preexec_fn
> > 
> > 
> > But we needed the PR_SET_PDEATHSIG :/ If we give up on it, we need to verify that all existing uses of `start_process` are safe even w/o PR_SET_PDEATHSIG.
> 
> Added manual process clean up. SIGKILL is not supported though.

That's a bummer - the SIGKILL use case is not uncommon (e.g: gProfiler getting OOM). I'd like us to think of something that handles SIGKILLs as well.

Jongy:
@slicklash - @d3dave had a cool idea to allow us to continue using PDEATHSIG - we can create a tiny shim program to handle that. For example, instead of:
```
p = subprocess.Popen(["/path/to/py-spy", "arg1", "arg2"], ....)
```

we'll build (as part of gProfiler's CI) a tiny program written in C/C++/Golang (or even bash if bash supports it) that only sets PDEATHSIG on its own PID, and immediately execves its following arguments:
```
p = subprocess.Popen(["/path/to/pdeathsigger", "/path/to/py-spy", "arg1", "arg2"], ....)
```
this way we can get rid of preexec_fn.

As for this tiny program - I'm not sure if bash/sh support this prctl call; if shells support it, since gProfiler already assumes a shell exists, we can use a wrapper shell script for our subprocess calls.
If no shell support, then add a build script and call it from the Dockerfile - a script that'll build this tiny program (whose code should be in this repo as well). You can choose the language but I think C makes sense to make it slim and lean. Build it on the same layer where we build the Python itself (so it's build against same glibc version, etc).

Then update the branch and ping us for re-CR :)

Jongy:
@slicklash please ping @d3dave and me when it's ready for re-review?

tests/test_perf.py
```diff
@@ -144,6 +145,7 @@ def test_perf_comm_change(
     I'm not sure it can be done, i.e is this info even kept anywhere).
     """
     with system_profiler as profiler:
+        time.sleep(2)
```
michelhe, line 148: Why are the sleeps necessary? Is there a way to make the test deterministic instead?

d3dave (changes requested):
Please open a separate PR for preexec_fn and pdeathsigger change as it's independent of grpcio version we are using. Let's keep the CRI client change along with the grpcio version PR.

Jongy:
@slicklash I'm unsure what this PR solves now. gProfiler already creates only a single `ContainersClient` object, only during tests we create more than one (and in `diagnostics.py` mode).

Inside the single `ContainersClient`, we do create several `CriClient`s but only upon startup (as we try v1 and v1alpha2 and check containerd and crio).

Jongy:
> Tests show the problem with implementation of `ContainerNamesClient` itself. There should be only one live CRI instance in multi-threaded context otherwise gRPC hangs. Tests are part of code base and need to be maintained with same care as production code.

I see, so this fixes the `ContainerNamesClient` code and effectively fixes the tests.

gprofiler/containers_client.py
```diff
@@ -25,20 +25,23 @@
...
+
 
 class ContainerNamesClient:
     def __init__(self) -> None:
+        global _containers_client
         try:
-            self._containers_client: Optional[ContainersClient] = ContainersClient()
-            logger.info(f"Discovered container runtimes: {self._containers_client.get_runtimes()}")
+            if _containers_client is None:
+                _containers_client = ContainersClient()
```
Jongy, line 36: Let's extract this to a method `get_containers_client` which will be locked (to avoid any concurrency issues) and maintain the cache.

Then, `ContainerNamesClient` runs `self._containers_client = get_containers_client()`.

Jongy (changes requested):
.

PR #921, opened 2024-09-03
profilers/python_ebpf.py: specifying newly TMPDIR on each instance of PyPerf

roi-granulate:
## Description
cleanup each instance of PyPerf staticx directories on termination.

## Related Issue
We experienced where there were a-lot of staticx tempdirs leftovers.

## Checklist:

### DIFF
--- a/gprofiler/profilers/python_ebpf.py
+++ b/gprofiler/profilers/python_ebpf.py
@@ -188,11 +188,11 @@ class PythonEbpfProfiler(ProfilerBase):
             poll_process(process, self._POLL_TIMEOUT, self._profiler_state.stop_event)
         except TimeoutError:
             process.kill()
-            self._staticx_cleanup()
             raise
         else:
-            self._staticx_cleanup()
             self._check_output(process, self.output_path)
+        finally:
+            self._staticx_cleanup()
 
     def start(self) -> None:
         logger.info("Starting profiling of Python processes with PyPerf")
--- a/gprofiler/profilers/python_ebpf.py
+++ b/gprofiler/profilers/python_ebpf.py
@@ -90,7 +90,7 @@ class PythonEbpfProfiler(ProfilerBase):
         self._pyperf_staticx_tmpdir: Optional[Path] = None
         if os.environ.get("TMPDIR", None) is not None:
             # We want to create a new level of hirerachy in our current staticx tempdir.
-            self._pyperf_staticx_tmpdir = Path(os.environ["TMPDIR"]) / "pyperf_" + random_prefix()
+            self._pyperf_staticx_tmpdir = Path(os.environ["TMPDIR"]) / ("pyperf_" + random_prefix())
 
     @classmethod
     def _check_output(cls, process: Popen, output_path: Path) -> None:
--- a/gprofiler/profilers/python_ebpf.py
+++ b/gprofiler/profilers/python_ebpf.py
@@ -162,7 +162,7 @@ class PythonEbpfProfiler(ProfilerBase):
         """
         We experienced issues with PyPerf's staticx'd directory, so we want to clean it up on termination.
         """
-        assert not self.is_running(), "_cleanup is impossible while PyPerf is running"
+        assert not self.is_running(), "_staticx_cleanup is impossible while PyPerf is running"
         if self._pyperf_staticx_tmpdir is not None and self._pyperf_staticx_tmpdir.exists():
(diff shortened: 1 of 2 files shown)


gprofiler/profilers/python_ebpf.py
```diff
@@ -201,7 +216,7 @@ def start(self) -> None:
         for f in glob.glob(f"{str(self.output_path)}.*"):
             os.unlink(f)
 
-        process = start_process(cmd)
+        process = start_process(cmd, tmpdir=self._pyperf_staticx_tmpdir)
```
d3dave, line 219: Does the directory have to exist prior to starting the process? If the directory does not already exist, does staticx create it?

gprofiler/utils/__init__.py
```diff
@@ -149,6 +153,8 @@ def start_process(
             # explicitly remove our directory from LD_LIBRARY_PATH
             env = env if env is not None else os.environ.copy()
```
roi-granulate, line 154: @Jongy @d3dave looks like when called with `env`, we simply don't pass `TMPDIR`. I suppose it's not the intended behavior, right?

Jongy: Indeed not the intended behavior, although the only process we executed that I expect to make use of TMPDIR is `PyPerf`.

Regardless - it's a bug - you're welcome to fix, but move it to a separate PR please.

roi-granulate: https://github.com/Granulate/gprofiler/pull/922

gprofiler/profilers/python_ebpf.py
```diff
@@ -86,6 +87,10 @@ def __init__(
         self._kernel_offsets: Dict[str, int] = {}
         self._metadata = python.PythonMetadata(self._profiler_state.stop_event)
         self._verbose = verbose
+        self._pyperf_staticx_tmpdir: Optional[Path] = None
+        if os.environ.get("TMPDIR", None) is not None:
+            # We want to create a new level of hirerachy in our current staticx tempdir.
+            self._pyperf_staticx_tmpdir = Path(os.environ["TMPDIR"]) / random_prefix()
```
d3dave, line 93:
```suggestion
            self._pyperf_staticx_tmpdir = Path(os.environ["TMPDIR"]) / "pyperf_" + random_prefix()
```

Jongy:
@d3dave are you able to complete the review this week?

d3dave:
> @d3dave are you able to complete the review this week?

CI is red

roi-granulate:
@Jongy @d3dave I’ll fix CI in a different PR https://github.com/Granulate/gprofiler/pull/924

roi-granulate:
Merged the CI fix from `master`

gprofiler/utils/__init__.py
```diff
@@ -147,9 +151,13 @@ def start_process(
             cmd = [f"{staticx_dir}/.staticx.interp", "--library-path", staticx_dir] + cmd
         else:
             env = env if env is not None else os.environ.copy()
-            # ensure `TMPDIR` env is propagated to the child processes (used by staticx)
-            if "TMPDIR" not in env and "TMPDIR" in os.environ:
+            if tmpdir is not None:
+                tmpdir.mkdir(exist_ok=True)
+                env["TMPDIR"] = tmpdir.as_posix()
+            elif "TMPDIR" not in env and "TMPDIR" in os.environ:
+                # ensure `TMPDIR` env is propagated to the child processes (used by staticx)
```
roi-granulate, line 158: I’m not sure this is the right approach
maybe should we use `tmpdir` always relative to the current env[“TMPDIR”] (e.g. `f”{env[“TMPDIR”]}/{tmpdir}”`)

d3dave: Is there a practical difference? Doesn't the current flow always create the tmpdir passed to this function under $TMPDIR?

d3dave: Ok, but the tmpdir arg, whenever it is passed it is always `Path(os.environ["TMPDIR"]) / ("pyperf_" + random_prefix())` which is under $TMPDIR. So there is no difference because with both approaches the tmpdir is created under $TMPDIR.

roi-granulate: there’s no practical differences… I raised this to suggest what’s the better approach

d3dave (approved):

PR #922, opened 2024-09-11
bugfix: `start_process` ensuring passing of `TMPDIR` when needed

roi-granulate:
## Description
https://github.com/Granulate/gprofiler/pull/921#discussion_r1743316860

## Checklist:

### DIFF
--- a/gprofiler/utils/__init__.py
+++ b/gprofiler/utils/__init__.py
@@ -146,10 +146,11 @@ def start_process(
             # see https://github.com/JonathonReinhart/staticx#run-time-information
             cmd = [f"{staticx_dir}/.staticx.interp", "--library-path", staticx_dir] + cmd
         else:
-            # explicitly remove our directory from LD_LIBRARY_PATH
             env = env if env is not None else os.environ.copy()
+            # ensure `TMPDIR` env is propagated to the child processes (used by staticx)
             if "TMPDIR" not in env and "TMPDIR" in os.environ:
                 env["TMPDIR"] = os.environ["TMPDIR"]
+            # explicitly remove our directory from LD_LIBRARY_PATH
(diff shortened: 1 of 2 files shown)


gprofiler/utils/__init__.py
```diff
@@ -148,7 +148,9 @@ def start_process(
         else:
             # explicitly remove our directory from LD_LIBRARY_PATH
             env = env if env is not None else os.environ.copy()
```
roi-granulate, line 150:
```suggestion
            env = env or os.environ.copy()
```

Jongy: No, prefer explicit `is None`

gprofiler/utils/__init__.py
```diff
@@ -148,7 +148,9 @@ def start_process(
...
-            env.update({"LD_LIBRARY_PATH": ""})
+            if "TMPDIR" not in env and "TMPDIR" in os.environ:
+                env["TMPDIR"] = os.environ.get("TMPDIR")
```
Jongy, line 152:
```suggestion
                env["TMPDIR"] = os.environ["TMPDIR"]`
```
makes it clear that the value can't be `None`.

Jongy (changes requested):
.

gprofiler/utils/__init__.py
```diff
@@ -148,7 +148,9 @@ def start_process(
         else:
             # explicitly remove our directory from LD_LIBRARY_PATH
```
Jongy, line 149: Move the comment to its relevant code line now. And add another comment explaining the TMPDIR ("respect the TMPDIR we were run with in all cases")

Jongy (dismissed):

Jongy (approved):

PR #926, opened 2024-09-23
Refactor into g utils condensed

granulatedekel:
## Checklist:


roi-granulate (changes requested):
Is there a reason wildcard import are preferred here?
if not - let's use explicit imports

PR #931, opened 2024-10-08
add pdeathsigger

slicklash:
### Problem

Sometimes `gprofilter` hangs

### Causes

sometimes `start_process (Popen)` hangs because of `preexec_fn`

> [!CAUTION]
> The **preexec_fn** parameter is NOT SAFE to use in the presence of threads in your application. The child process could deadlock before exec is called. https://docs.python.org/3/library/subprocess.html#popen-constructor

### Solution

replace `preexec_fn` with `start_new_session` flag and introduce `pdeathsigger`

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -215,7 +215,7 @@ def snapshot_pid_profile(profiler: ProfilerInterface, pid: int) -> ProfileData:
         last_snapshot = profiler.snapshot()
         return pid in last_snapshot
 
-    wait_event(timeout=5, stop_event=Event(), condition=has_profile, interval=0.1)
+    wait_event(timeout=5, stop_event=Event(), condition=has_profile, interval=2.0)
     return last_snapshot[pid]
 
 


tests/utils.py
```diff
@@ -205,7 +206,17 @@ def assert_ldd_version_container(container: Container, version: str) -> None:
...
+    last_snapshot = profiler.snapshot()
+
+    def has_profile() -> bool:
+        nonlocal last_snapshot
+        if pid in last_snapshot:
+            return True
+        last_snapshot = profiler.snapshot()
+        return pid in last_snapshot
+
+    wait_event(timeout=5, stop_event=Event(), condition=has_profile, interval=0.1)
```
d3dave, line 218: So this will try to `snapshot()` every 100ms? I don't think we can do snapshots that fast, it is a very heavy operation. If the issue is that pids do not appear fast enough in snapshots (what is the underlying reason 🤔 ? I see it's independent of runtime) then I'd be more comfortable sticking a `sleep(2)` before `profiler.snapshot()` instead of calling `snapshot` multiple times in a row.

PR #932, opened 2024-10-13
PythonProfiler: expose `--python-pyspy-process`

roi-granulate:
## Description
Expose py-spy specific flag that forces gProfiler to profile given processes.

## Checklist:

### DIFF
--- a/gprofiler/profilers/python.py
+++ b/gprofiler/profilers/python.py
@@ -263,7 +263,7 @@ class PySpyProfiler(SpawningProcessProfilerBase):
             return ProfileData(parsed, appid, app_metadata, container_name)
 
     def _select_processes_to_profile(self) -> List[Process]:
-        filtered_procs = []
+        filtered_procs = set()
         if is_windows():
             all_processes = [x for x in pgrep_exe("python")]
         else:
@@ -272,13 +272,14 @@ class PySpyProfiler(SpawningProcessProfilerBase):
         for process in all_processes:
             try:
                 if not self._should_skip_process(process):
-                    filtered_procs.append(process)
+                    filtered_procs.add(process)
             except NoSuchProcess:
                 pass
             except Exception:
                 logger.exception(f"Couldn't add pid {process.pid} to list")
 
-        return filtered_procs + [Process(pid) for pid in self._python_pyspy_process]
+        filtered_procs.update([Process(pid) for pid in self._python_pyspy_process])
+        return list(filtered_procs)
 
     def _should_profile_process(self, process: Process) -> bool:
         return search_proc_maps(process, DETECTED_PYTHON_PROCESSES_REGEX) is not None and not self._should_skip_process(
--- a/gprofiler/profilers/python.py
+++ b/gprofiler/profilers/python.py
@@ -21,6 +21,7 @@ from pathlib import Path
 from subprocess import CompletedProcess
 from typing import Any, Dict, List, Match, Optional, cast
 
+from gprofiler_types import integers_list
 from granulate_utils.linux.elf import get_elf_id
 from granulate_utils.linux.ns import get_process_nspid, run_in_ns
 from granulate_utils.linux.process import (
@@ -345,8 +346,9 @@ class PySpyProfiler(SpawningProcessProfilerBase):
         ProfilerArgument(
             name="--python-pyspy-processes",
             dest="python_pyspy_processes",
-            nargs="+",
-            type=int,
+            action="extend",
+            default=[],
+            type=integers_list,
             help="List of processes (by PID) to profile with py-spy."
             " This option forces gProfiler to profile given processes with py-spy, even if"
             " they are not recognized by gProfiler as Python processes."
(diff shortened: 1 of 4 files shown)


gprofiler/profilers/python.py
```diff
@@ -340,6 +341,13 @@ def _should_skip_process(self, process: Process) -> bool:
             action="store_true",
             help="Enable PyPerf in verbose mode (max verbosity)",
         ),
+        ProfilerArgument(
+            name="--python-pyspy-processes",
+            dest="python_pyspy_processes",
+            nargs="+",
+            type=int,
+            help="List of processes (by PID) to profile with py-spy." "This option forces gProfiler to ",
```
Jongy, line 349: Fix `help`

gprofiler/profilers/python.py
```diff
@@ -275,7 +277,7 @@ def _select_processes_to_profile(self) -> List[Process]:
             except Exception:
                 logger.exception(f"Couldn't add pid {process.pid} to list")
 
-        return filtered_procs
+        return filtered_procs + [Process(pid) for pid in self._python_pyspy_processes]
```
Jongy, line 280: `Process(pid)` will raise if the PID exited, that's okay but let's mention in the `help` that they're expected to be alive forever

gprofiler/profilers/python.py
```diff
@@ -340,6 +342,16 @@ def _should_skip_process(self, process: Process) -> bool:
...
         ),
+        ProfilerArgument(
+            name="--python-pyspy-processes",
+            dest="python_pyspy_processes",
+            nargs="+",
+            type=int,
+            help="List of processes (by PID) to profile with py-spy."
+            "This option forces gProfiler to profile given processes with py-spy, even if"
+            "they are not recognized by gProfiler as Python processes."
+            "Note - gProfiler assumes that the given processes are alive and running forever.",
```
roi-granulate, line 353: I don't like the use of "forever" here
@Jongy what other word could fit?

Jongy:
```suggestion
            "Note - gProfiler assumes that the given processes are kept running as long as gProfiler runs.",
```

gprofiler/profilers/python.py
```diff
@@ -340,6 +342,16 @@ def _should_skip_process(self, process: Process) -> bool:
             action="store_true",
             help="Enable PyPerf in verbose mode (max verbosity)",
         ),
+        ProfilerArgument(
+            name="--python-pyspy-processes",
+            dest="python_pyspy_processes",
+            nargs="+",
+            type=int,
+            help="List of processes (by PID) to profile with py-spy."
+            "This option forces gProfiler to profile given processes with py-spy, even if"
```
Jongy, line 351: Note that spacing is broken - need to add a space before each line

gprofiler/profilers/python.py
```diff
@@ -340,6 +342,16 @@ def _should_skip_process(self, process: Process) -> bool:
             action="store_true",
             help="Enable PyPerf in verbose mode (max verbosity)",
         ),
+        ProfilerArgument(
+            name="--python-pyspy-processes",
+            dest="python_pyspy_processes",
+            nargs="+",
```
grariel, line 348: From argparse docs:
```
'+'. Just like '*', all command-line args present are gathered into a list. Additionally, an error message will be generated if there wasn’t at least one command-line argument present. For example:
```

If I understand correctly, doesn't it require one value?

roi-granulate: It requires one only if the specific argument appears
that is the desired case here I guess (we don't want to support when people specify `--python-pyspy-processes` without PIDs)

Jongy: I think the nicer way is to name the arg singular and use `action="extend"`, so it's provided this way: `--python-pyspy-process xxx --python-pyspy-process yyy`

grariel: Feel free to resolve after finishing with @Jongy's comment

roi-granulate: I preferred the old way

gprofiler/profilers/python.py
```diff
@@ -275,7 +278,7 @@ def _select_processes_to_profile(self) -> List[Process]:
             except Exception:
                 logger.exception(f"Couldn't add pid {process.pid} to list")
 
-        return filtered_procs
+        return filtered_procs + [Process(pid) for pid in self._python_pyspy_process]
```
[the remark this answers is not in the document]
roi-granulate: We're fine with that, as the `help` mentions - we expect processes to live as long as gProfiler runs.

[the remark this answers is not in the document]
roi-granulate: Nice

Jongy (approved):
lgtm

gAmitFrenkel (approved):

gprofiler/profilers/python.py
roi-granulate: @Jongy I resolve this comment because you approve this PR

PR #934, opened 2024-10-21
python: Add option to run py-spy on processes via regex

michelhe:
I need it to test stuff locally where my PIDs are replacing and using `--python-pyspy-process` is not good enough for continously running


roi-granulate:
@michelhe note ci is red

Issue #905, opened 2024-05-31
Run without sudo, even if with limited data?


hsane2001:
Can we allow gprofiler to run without sudo? There are many instances where sudo is not possible and it would be great to get app level stacks even if in a limited manner. Besides perf which would require sudo (although can be made to run otherwise), there are many cases where the code internally uses root access to the filesystem for namespaces and storing intermediate data.

Jongy:
We can definitely make gProfiler run without root. It will require some iterative work of encountering problems, fixing them (by making them handle the lack of permissions gracefully) and continuing.

Things I already have on my mind:
1. Need to remove the actual check for "is root" in `verify_preconditions`.
2. Many sites in gProfiler use `run_in_ns` & access `/proc/pid/ns/` files, which might be inaccessible if you're not root. The use case of running gProfiler w/o root is to profile applications running in the same mount/pid namespace, so all `run_in_ns` interaction is optional and can be made so (for example, gprofiler can have a `run_in_ns_wrapper` that checks if we're root and skips the privileged operation if we're root).
3. `perf` itself can run w/o root - modifying `kernel.perf_event_paranoid` is one thing, in addition I think that we'll need to run perf in a per-process mode (and not `-a` mode). I don't know if Linux allows you to run perf in `-a` mode while underprivileged, but it'd make little sense to me. However, running perf in `-p` mode, targeting PIDs of the same user, makes sense. If `perf` is desired and `-a` will be blocked, we can make gProfiler when runs underprivielged to run `perf record -p` (for example, based on PIDs passed via `--pids` to gProfiler).
4. There might be additional issues - perhaps directories gProfiler tries to write to (we use `/tmp` by default but might fallback to `/opt` which is root-only).

If you're working on eliminating the root requirement, you can write thoughts here about how to handle particular parts being blocked due to underprivilege, and I'll help addressing. I'm also open to a Zoom discussion over it :)

hsane2001:
There has been more requirements to make this happen. Some answers inline:
1. verify_precondition check should be easy to resolve
2. The option of "run_in_ns_wrapper that checks if we're root and skips the privileged operation if we're root" sounds good.
3. Yes, perf can itself run in user mode after applying the sys related changes (one time). "perf -a" also works and is essential in cases where a privileged container can monitor rest of the system or other containers. 
4. The code does have other places where it access root only filesystem and may need modification to a temporary user directory.

pcarella1:
[...]

PR #936, opened 2024-12-05
Run gprofiler without root/sudo

pcarella1:
## Description
This PR adds support for running gprofiler without root/sudo as discussed in issue 905. There are several assumptions and components that I will mention here.

This PR requires a change in the granulate-utils repo that defines the run_in_ns_wrapper function found in PR 265 on that repo https://github.com/Granulate/granulate-utils/pull/265.

Assumptions when running without root: 

1. When running without root, the user must use --pids to select user owned processes only. 
2. The user must direct the log and pids files to a user owned directory (e.g. with --log-file and --pid-file parameters). 
3. The user must set certain system parameters such as kernel.perf_even_paranoid as needed to allow gprofiler to run. 
4. Some of the corner cases that require fallback rw exec directories for POSSIBLE_AP_DIRS may not be resolved. 

Components: 

1. Replaced exit/error when is_root check fails in verify_precondiftions and replaced it with this message: "Not running as root, and therefore functionality is limited. Profile is limted to only processes owned by this user that are passed with --pids. Some additional configuration (e.g. perf_event_paranoid) may be configured to operate without root."
2. Created run_in_ns_wrapper function which bypasses the code to enter name spaces when not root (as we assume we're always in the correct namespace for the processes being profiled)
3. Added a parameter to the pgrep_maps function to ignore permissions errors. Each time a profiler calls this function, it will check if root and if not, will pass "True". 
4. Redirected the default value of TEMPORARY_STORAGE_PATH to the resources directory.
5. Added mkdir_owned_user function which is used in main where the TEMPORARY_STORAGE_PATH creates gprofiler_tmp, so that it doesn't throw an error when we aren't root, but still ensures the directory is owned by the current user. 

Potential issues:
1. Is there anything I should add/change in the message that is displayed when the is_root check in verify_preconditions fails? Also, is print() to stdout correct here?
2. Is it fine to redirect TEMPORARY_STORAGE_PATH to the resources directory even in the default case, or should I add a check to only do this when not root?
3. Do we need to resolve the fallback rw exec directories for POSSIBLE_AP_DIRS? 
4. I've tested this on two systems and it works on both, but on one of them I receive this warning (though gprofiler completes and produces valid results). I discuss this more in the corresponding granulate-utils PR since that is the source of the error. 
	

> 	[2024-12-02 19:59:55,557] WARNING: gprofiler.profilers.java: Failed to enable proc_events listener for exited Java processes (this does not prevent Java profiling)
> 	Traceback (most recent call last):
> 	  File "granulate_utils/linux/proc_events.py", line 222, in start
> 	PermissionError: [Errno 1] Operation not permitted
> 	
> 	The above exception was the direct cause of the following exception:
> 	
> 	Traceback (most recent call last):
> 	  File "gprofiler/profilers/java.py", line 1395, in start
> 	    proc_events.register_exit_callback(self._proc_exit_callback)
> 	  File "granulate_utils/linux/proc_events.py", line 272, in wrapper
> 	  File "granulate_utils/linux/ns.py", line 305, in run_in_ns_wrapper
> 	  File "granulate_utils/linux/ns.py", line 299, in run_in_ns_wrapper
> 	  File "granulate_utils/linux/proc_events.py", line 260, in _start_listener
> 	  File "granulate_utils/linux/proc_events.py", line 225, in start
> PermissionError: This process doesn't have permissions to bind/connect to the process events connector

## Related Issue
https://github.com/Granulate/gprofiler/issues/905

## Motivation and Context
Users of this gprofiler have requested this feature as some cloud instances do not have root access, but still want to profile user owned processes.

## How Has This Been Tested?
I ran stress-ng and targeted gprofiler to the stress-ng pids without sudo. It successfully produced flamegraphs
Sample command line: ./build/x86_64/gprofiler --pids 1421864 -o results/ -d 15 --log-file ./gprofiler.log --pid-file ./gprofiler.pid

I have tested this on x86 using scripts/build_x86_64_executable.sh script. Centos 9 Stream w/ kernel 6.6

Also tested using sudo targeting specific pid(s) and system-wide, and it still works. 

Was not able to run tests/test.sh as it required apt-get/debian environment.

## Checklist:
The code is linted.

I have not updated the README.md doc here. Might need some guidance. 

### DIFF
--- a/gprofiler/utils/__init__.py
+++ b/gprofiler/utils/__init__.py
@@ -39,7 +39,7 @@ import importlib_resources
 import psutil
 from granulate_utils.exceptions import CouldNotAcquireMutex
 from granulate_utils.linux.mutex import try_acquire_mutex
-from granulate_utils.linux.ns import run_in_ns_wrapper
+from granulate_utils.linux.ns import run_in_ns_wrapper, is_root
 from granulate_utils.linux.process import is_kernel_thread, process_exe
 from psutil import Process
 
@@ -62,6 +62,11 @@ from gprofiler.log import get_logger_adapter
 logger = get_logger_adapter(__name__)
 
 GPROFILER_DIRECTORY_NAME = "gprofiler_tmp"
+TEMPORARY_STORAGE_PATH = (
+    f"/tmp/{GPROFILER_DIRECTORY_NAME}"
+    if is_linux()
+    else os.getenv("USERPROFILE", default=os.getcwd()) + f"\\AppData\\Local\\Temp\\{GPROFILER_DIRECTORY_NAME}"
+)
 
 gprofiler_mutex: Optional[socket.socket] = None
 
@@ -75,21 +80,6 @@ def resource_path(relative_path: str = "") -> str:
             return str(path)
     except ImportError as e:
         raise Exception(f"Resource {relative_path!r} not found!") from e
-
-
-TEMPORARY_STORAGE_PATH = (
-    f"{resource_path(GPROFILER_DIRECTORY_NAME)}"
-    if is_linux()
-    else os.getenv("USERPROFILE", default=os.getcwd()) + f"\\AppData\\Local\\Temp\\{GPROFILER_DIRECTORY_NAME}"
-)
-
-
-@lru_cache(maxsize=None)
-def is_root() -> bool:
-    if is_windows():
-        return cast(int, ctypes.windll.shell32.IsUserAnAdmin()) == 1  # type: ignore
-    else:
-        return os.geteuid() == 0
 
 
 libc: Optional[ctypes.CDLL] = None
@@ -353,7 +343,7 @@ else:
         return procs
 
 
-def pgrep_maps(match: str, ignore_permission_errors: bool = False) -> List[Process]:
+def pgrep_maps(match: str) -> List[Process]:
     # this is much faster than iterating over processes' maps with psutil.
     # We use flag -E in grep to support systems where grep is not PCRE
     result = run_process(
@@ -381,7 +371,7 @@ def pgrep_maps(match: str, ignore_permission_errors: bool = False) -> List[Proce
             and (
                 line.endswith(b"/maps: No such file or directory")
                 or line.endswith(b"/maps: No such process")
-                or b"Permission denied" in line
+                or (not is_root() and b"/maps: Permission denied" in line)
             )
         ):
             error_lines.append(line)
--- a/gprofiler/utils/__init__.py
+++ b/gprofiler/utils/__init__.py
@@ -39,7 +39,7 @@ import importlib_resources
 import psutil
 from granulate_utils.exceptions import CouldNotAcquireMutex
 from granulate_utils.linux.mutex import try_acquire_mutex
-from granulate_utils.linux.ns import run_in_ns_wrapper, is_root
+from granulate_utils.linux.ns import is_root, run_in_ns_wrapper
 from granulate_utils.linux.process import is_kernel_thread, process_exe
 from psutil import Process
 
--- a/gprofiler/utils/__init__.py
+++ b/gprofiler/utils/__init__.py
@@ -39,7 +39,7 @@ import importlib_resources
 import psutil
 from granulate_utils.exceptions import CouldNotAcquireMutex
 from granulate_utils.linux.mutex import try_acquire_mutex
-from granulate_utils.linux.ns import run_in_ns
+from granulate_utils.linux.ns import run_in_ns_wrapper
 from granulate_utils.linux.process import is_kernel_thread, process_exe
 from psutil import Process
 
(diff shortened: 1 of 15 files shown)


gprofiler/utils/__init__.py
```diff
@@ -351,7 +353,7 @@ def pgrep_exe(match: str) -> List[Process]:
         return procs
 
 
-def pgrep_maps(match: str) -> List[Process]:
+def pgrep_maps(match: str, ignore_permission_errors: bool = False) -> List[Process]:
```
Jongy, line 356: Let's remove this parameter, and instead - call `is_root` here to decide (instead of having all callers forced to call this and pass as an argument)

gprofiler/utils/fs.py
```diff
@@ -76,6 +76,43 @@ def is_owned_by_root(path: Path) -> bool:
     return statbuf.st_uid == 0 and statbuf.st_gid == 0
 
 
+def is_owned_by_user(path: Path) -> bool:
+    statbuf = path.stat()
+    return statbuf.st_uid == os.geteuid() and statbuf.st_gid == os.getegid()
+
+
+def mkdir_owned_user(path: Union[str, Path], mode: int = 0o755) -> None:
```
Jongy, line 84: This behavior is not needed when we're not root. So - either add a wrapper, which checks if we're root and calls it OR simply `mkdir`s, or if there's 1-2 calls sites, add such `if` directly:
```
if is_root():
   mkdir_owned_root(...)
else:
   mkdir(...)
```

gprofiler/utils/perf.py
```diff
@@ -95,7 +99,7 @@ def discover_appropriate_perf_event(tmp_dir: Path, stop_event: Event) -> Support
                 is_dwarf=False,
                 inject_jit=False,
                 extra_args=current_extra_args,
-                processes_to_profile=None,
+                processes_to_profile=pids,
```
Jongy, line 102: Does it mean anything if we pass only specific PIDs here, permission wise?

Jongy (changes requested):

Jongy:
> Is there anything I should add/change in the message that is displayed when the is_root check in verify_preconditions fails? Also, is print() to stdout correct here?

I prefer that the rootless mode will be opt-in, and not a default in case of non-root user (which allows misconfigurations to continue misusing the profiler).
So a flag like `--rootless` that enabled this behavior & checks that we're not root. And the `verify_preconditions()` function will suggest using `--rootless`.

> Is it fine to redirect TEMPORARY_STORAGE_PATH to the resources directory even in the default case, or should I add a check to only do this when not root?

I left some comments on the PR about this topic.

> Do we need to resolve the fallback rw exec directories for POSSIBLE_AP_DIRS?

Not sure I got you here.

> I've tested this on two systems and it works on both, but on one of them I receive this warning (though gprofiler completes and produces valid results). I discuss this more in the corresponding granulate-utils PR since that is the source of the error.

Yeah, it's fine - as I commented on granulate-utils, proc_events are expected to fail in rootless.

sobolron (changes requested):

gprofiler/utils/fs.py
```diff
@@ -76,6 +77,30 @@ def is_owned_by_root(path: Path) -> bool:
...
+    if path.exists():
+        if not os.access(path, os.W_OK):
+            raise Exception(f"{str(path)} is not writable by current user")
+        return
+
+    try:
+        os.mkdir(path, mode=mode)
+    except FileExistsError:
+        # likely racing with another thread of gprofiler. as long as the directory is the user after all, we're good.
+        pass
```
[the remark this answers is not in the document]
pcarella1: I didn't add anything here because I just copied the code from the mkdir_owned_root function that the wrapper wraps. 
Line 128 - 130 in the commit. 

I can make the change if you think it's necessary. How should I confirm that gprofiler wrote the directory, I can just check against the current user inside the except block like in Line 92 - 95. Should I also make the change in mkdir_owned_root() as well?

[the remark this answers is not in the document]
sobolron: You can just check again for access.. If it's truly a race you will have access now. If it's not a race but something weird happened, you really can't continue.

sobolron:
Hi!
- I've merged the granulate-utils PR, now let's move the ref of the submodule to something on the master branch and I'll approve.
- About documentation - I think we need to create a table that contains the available capabilities of gProfiler, and which are supported and which are not in rootless mode. As this entire table of capabilities doesn't exist yet, I think that it's not really in the scope of this PR, but I highly recommend creating it in a separate one so that feature will be easier to use.
- About testing - The current testing "framework" (scripts 😅 ) doesn't currently support rootless run properly, this will require doing changes to it, and I highly recommend doing this as well, but again - pretty much out of scope IMO.

pcarella1:
> * I've merged the granulate-utils PR, now let's move the ref of the submodule to something on the master branch and I'll approve.

I updated the granulate-utils reference to [dd8aa29](https://github.com/Granulate/granulate-utils/commit/dd8aa29b9d68a05f55e6e1d89dfffab6879c8530)

sobolron (dismissed):

pcarella1:
Added a commit to address the test_not_root checklist failure.

PR #941, opened 2025-03-07
Small fixes of local build of gprofiler executable

dkorlovs:
Added small changes to fix issues that we faced with build locally. It will be useful to have these changes in branch also, for easier development.

## Checklist:

### DIFF
--- a/executable.Dockerfile
+++ b/executable.Dockerfile
@@ -194,12 +194,12 @@ RUN set -e; \
     fi
 RUN set -e; \
     if [ "$(uname -m)" = "aarch64" ]; then \
-        python3 -m pip install --no-cache-dir 'wheel==0.37.0' 'scons==4.2.0'; \
+        python3 -m pip install --trusted-host files.pythonhosted.org --trusted-host pypi.org --no-cache-dir 'wheel==0.37.0' 'scons==4.2.0'; \
     fi
 
 # we want the latest pip
 # hadolint ignore=DL3013
-RUN python3 -m pip install --no-cache-dir --upgrade pip
+RUN python3 -m pip install --trusted-host files.pythonhosted.org --trusted-host pypi.org --no-cache-dir --upgrade pip
 
 FROM ${NODE_PACKAGE_BUILDER_GLIBC} as node-package-builder-glibc
 USER 0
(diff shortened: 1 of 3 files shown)


awherr:
Do you think we can merge this now?

dkorlovs:
> Do you think we can merge this now?

We need to discuss changes with "--trusted-host files.pythonhosted.org --trusted-host pypi.org" in docker file with @mlim19.
But perfspect collection part is ready, and python linter issues fixed.

mlim19:
This patchset combines 2 things: Ron's patch for workaround the proxy issue and our patch for initial perfspect integration which I think not mature yet to merge. And they should be split and merged separately when ready.

Regarding Ron's patch, I believe the pip config change seems more suitable for proxy users which is described at https://stackoverflow.com/questions/25981703/pip-install-fails-with-connection-error-ssl-certificate-verify-failed-certi
The pip config change seems better because we can delegate the workaround adoption to users instead of forcing it to every user. Users don't take the workaround if not needed. I will verify it and update later.

awherr:
I agree that it makes sense to split these up, with the proxy workaround going first. The reasons for asking is that I want to close out the remaining CVEs without disrupting the process. I will work from this branch for now.

dkorlovs:
> This patchset combines 2 things: Ron's patch for workaround the proxy issue and our patch for initial perfspect integration which I think not mature yet to merge. And they should be split and merged separately when ready.
> 
> Regarding Ron's patch, I believe the pip config change seems more suitable for proxy users which is described at https://stackoverflow.com/questions/25981703/pip-install-fails-with-connection-error-ssl-certificate-verify-failed-certi The pip config change seems better because we can delegate the workaround adoption to users instead of forcing it to every user. Users don't take the workaround if not needed. I will verify it and update later.

Let me split changes, I will prepare PR to enable_perfspec_metrics branch with changes related only to perfspect integration.
And after that I'll close this PR.

dkorlovs:
> > This patchset combines 2 things: Ron's patch for workaround the proxy issue and our patch for initial perfspect integration which I think not mature yet to merge. And they should be split and merged separately when ready.
> > Regarding Ron's patch, I believe the pip config change seems more suitable for proxy users which is described at https://stackoverflow.com/questions/25981703/pip-install-fails-with-connection-error-ssl-certificate-verify-failed-certi The pip config change seems better because we can delegate the workaround adoption to users instead of forcing it to every user. Users don't take the workaround if not needed. I will verify it and update later.
> 
> Let me split changes, I will prepare PR to enable_perfspec_metrics branch with changes related only to perfspect integration. And after that I'll close this PR.

PR with changes related only to Perfspect integration:
https://github.com/intel/gprofiler/pull/942

PR #942, opened 2025-03-11
Added an option to gprofiler CLI to use perfspect HW metrics collection and provide the path to it

dkorlovs:
Added an option to gprofiler CLI to use perfspect HW metrics collection and provide the path to it

## Description
(1) Added option "--perfspect-path" to gprofiler CLI.
Without path collection of HW metrics will not start.
(2) Disabled this option not on linux and on aarch64
(3) Fixed Python linter issues

## How Has This Been Tested?
Tested locally with and without option provided:
> sudo ./build/x86_64/gprofiler --perfspect-path ~/.../perfspect/perfspect -o ~/.../gprofiler_results/

Checked .col files after collection:
metrics field in file header with option -
` "metrics": {"cpu_avg": 0.1, "mem_avg": 2.9153846153846152, "cpu_freq": 1.102548528, "cpu_cpi": 2.3034023, "cpu_tma_fe_bound": 40.3853196, "cpu_tma_be_bound": 44.372709, "cpu_tma_bad_spec": 5.5948107, "cpu_tma_retiring": 9.64716058},`
without option -
`"metrics": {"cpu_avg": 0.1, "mem_avg": 2.823076923076923, "cpu_freq": null, "cpu_cpi": null, "cpu_tma_fe_bound": null, "cpu_tma_be_bound": null, "cpu_tma_bad_spec": null, "cpu_tma_retiring": null},`

## Checklist:

### DIFF
--- a/gprofiler/main.py
+++ b/gprofiler/main.py
@@ -181,8 +181,7 @@ class GProfiler:
         self._usage_logger = usage_logger
         if self._collect_metrics:
             self._system_metrics_monitor: SystemMetricsMonitorBase = SystemMetricsMonitor(
-                stop_event = self._profiler_state.stop_event,
-                perfspect_path = perfspect_path
+                stop_event=self._profiler_state.stop_event, perfspect_path=perfspect_path
             )
         else:
             self._system_metrics_monitor = NoopSystemMetricsMonitor()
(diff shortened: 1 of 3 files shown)


mlim19:
When perfspect is not located, should we not populate the hw metrics at all, instead of 'null'?

gprofiler/main.py
```diff
@@ -834,6 +837,17 @@ def parse_cmd_args() -> configargparse.Namespace:
         "The file modification indicates the last snapshot time.",
     )
 
+    if is_linux() and not is_aarch64():
```
mlim19, line 840: We also need the detection for AMD too. I explored the platform module in python but none of the existing functions provide the distinction btw AMD and Intel architecture. I think we should parse "/proc/cpuinfo".

mlim19: Actually, I found a better module to detect it. It's py-cpuinfo https://pypi.org/project/py-cpuinfo/
cpuinfo.get_cpu_info()['vendor_id_raw'] gives the info we need.
>>> cpuinfo.get_cpu_info()['vendor_id_raw']
'GenuineIntel'

PR #944, opened 2025-03-11
Fixes for CVE-2024-6345 setuptools upgraded to version 70.0.0

awherr:
## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ psutil==5.9.8
 requests==2.31.0
 ConfigArgParse==1.3
 distro==1.7.0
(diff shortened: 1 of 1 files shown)


sobolron:
@awherr This build is failing as PyInstaller version is incompatible with the setuptools used.
The issue is that this was only fixed in PyInstaller 6.7.0.
Do we **really** have to upgrade setuptools major version here? (to 70.*)? If so we need to upgrade PyInstaller Major as well and IDK what issues will this raise. I'll try to create a PR with them both.

awherr:
The IPass people really want to see all known CSVs closed, so let's give it a try and see what happens with a major version upgrade.

PR #948, opened 2025-03-18
Upgrade multiple packages for CVE solving

sobolron:
## Description
- Fixes for CVE-2024-6345 setuptools upgraded
- Fixes CVE-2024-3195 by upgrading request
- Fixes [CVE-2024-21503](https://github.com/advisories/GHSA-fj7x-q9j7-g6q6) by upgrading black


awherr:
It makes sense what you have done here, but I'm wondering why all the tests are failing during checks?

sobolron:
> It makes sense what you have done here, but I'm wondering why all the tests are failing during checks?

@awherr 2 Reasons:
- When you looked was after I updated to latest, which in most libraries doesn't support Python3.8 which we officially still support. So I had to downgrade. If we'll decide to remove support from Python 3.8 it'll be possible to upgrade again.
- There were some conflicts between `types-setuptools` and `types-pkg_resources` which was cause because of deprecated packages usage. This caused Mypy checks to fail. Fixed that by removing `types-pkg_resources`. Still waiting for all tests to work.

awherr:
It looks really promising. The executable tests all pass, though it appears to be hung on the container tests. Not sure if these are normally long-running tasks.

PR #962, opened 2025-04-23
Remove granulate.io links due to service shutdown

mlim19:
## Checklist:


skamerintel (commented):
i think we should substitute https://localhost:4433 in place of some of the references to 'a self hosted studio', as the user will want to know what is the direct link to page to get token, etc.

PR #993, opened 2025-08-26
Fix popen leak and handle perf script crash

### LINKED ISSUE #992: CPU profiling on GPU machines
The gProfiler was experiencing segmentation faults (SIGSEGV: 11) when running the perf script command on GPU machines. This caused the entire profiling run to fail with the error:

`
gprofiler.exceptions.CalledProcessError: Command ['/tmp/_MEIdG6uBN/gprofiler/resources/pdeathsigger', '/tmp/some_id/gprofiler/resources/perf', 'script', '-F', '+pid', '-i', '/tmp/gprofiler_tmp/some_id/perf.fp._some_example_id_'] died with <Signals.SIGSEGV: 11>.
`

Root Cause

The issue occurs when the perf binary on GPU machines crashes during the perf script parsing phase. This is a known issue with perf on certain GPU environments where the perf binary can crash when trying to parse performance data, particularly due to:

GPU driver interactions with perf events
Memory issues when processing large perf data files
Incompatibility between certain GPU hardware/drivers and perf's parsing logic

mlim19:
The patch is revised to take care of cleaning up the popen object being still referenced after the associated subprocess gets terminated.

Based on code review, 3 issues are identified:
1. In run_process function, If the assertion does NOT happen, the subprocess is terminated. At this time, I think we should remove the popen object in the global list so that the pipes and file descriptors are cleaned by GC.
- a. With python 3.2 or later, subprocess.Popen supports the context manager protocol for **with** statement, which means __exit__ method is implemented and executed when t the with block ends. Popen.__exit__ calls proc.wait() by default.
- b. At the end of the with block, it looks we ensures the process is terminated by checking the retcode from poll(). https://github.com/intel/gprofiler/blob/63ef4354e2ed12b176983a179218eb4aabf42b8d/gprofiler/utils/__init__.py#L266
2. Some profilers (php, pyperf, perf) call the start_process function directly. So, the above solution for run_process is not helpful. So, we handle the clean up separately here
3. When "perf script" is crashed, we need to add the handling of exception which should fixes #992 

## Checklist:

### DIFF
--- a/gprofiler/profilers/php.py
+++ b/gprofiler/profilers/php.py
@@ -30,7 +30,14 @@ from gprofiler.log import get_logger_adapter
 from gprofiler.profiler_state import ProfilerState
 from gprofiler.profilers.profiler_base import ProfilerBase
 from gprofiler.profilers.registry import ProfilerArgument, register_profiler
-from gprofiler.utils import random_prefix, reap_process, resource_path, start_process, wait_event
+from gprofiler.utils import (
+    random_prefix,
+    reap_process,
+    resource_path,
+    start_process,
+    wait_event,
+    cleanup_process_reference,
+)
 
 logger = get_logger_adapter(__name__)
 # Currently tracing only php-fpm, TODO: support mod_php in apache.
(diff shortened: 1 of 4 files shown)


gprofiler/profilers/php.py
```diff
@@ -218,12 +255,26 @@ def extract_metadata_section(re_expr: Pattern, metadata_line: str) -> str:
         return profiles
 
     def snapshot(self) -> ProcessToProfileData:
+        # Add health check at the beginning
+        if not self._check_process_health():
+            raise Exception("phpspy process has crashed or terminated")
```
dkorlovs, line 260: same question

gprofiler/profilers/php.py
```diff
@@ -130,9 +143,33 @@ def start(self) -> None:
...
-        assert self._process is not None, "profiling not started!"
-        self._process.send_signal(self.dump_signal)
+        if not self._check_process_health():
+            raise Exception("phpspy process is not running")
+
+        try:
+            self._process.send_signal(self.dump_signal)
+        except ProcessLookupError:
+            # Process died between health check and signal send
+            if self._process is not None:
```
dkorlovs, line 168: if you added clean up to check of process health, then better to call check process health everywhere.
but I think we put too much inside of the function, its name says only about checking, not clean up.

I would move "self._process = None" inside of cleanup_process_reference(), so there is 2 options - made it method of class or provide class object as argument.

gprofiler/profilers/php.py
```diff
@@ -130,9 +143,33 @@ def start(self) -> None:
...
+            self._process = None
+            return False
+
+        return True
+
     def _dump(self) -> Path:
-        assert self._process is not None, "profiling not started!"
-        self._process.send_signal(self.dump_signal)
+        if not self._check_process_health():
+            raise Exception("phpspy process is not running")
```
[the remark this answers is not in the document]
mlim19: Revised the logic to streamline the exception handling. Instead of propagating it to callers, handle it with logging

gprofiler/profilers/php.py
mlim19: Simplified the logic with removing redundant checks

dkorlovs (approved):
thank you!

mlim19:
Added a new patch to take care of the cases not handled. After applying the patchset collectively, the overall memory utilization remain same over several hours.
Screenshot 2025-08-28 201708

gprofiler/profilers/python_ebpf.py
```diff
@@ -224,6 +225,7 @@ def start(self) -> None:
             wait_event(self._POLL_TIMEOUT, self._profiler_state.stop_event, lambda: os.path.exists(self.output_path))
         except TimeoutError:
             process.kill()
+            cleanup_process_reference(process)
```
prashantbytesyntax, line 228: we probably need clean up on run time profilers too

prashantbytesyntax (commented):
The fix looks good, there is still memory growth which i think needs handling for runtime profilers and perf output parsing optimization but could be done in future PRs

PR #1000, opened 2025-10-03
Address the slow growth of memory utilization with perf

mlim19:
It seems the perf has known issue regarding memory utilization when perf runs continuously. As it's observed on idle system where there are not many processes running and output file sizes are not huge, it looks the perf has to be restarted if its internal memory usage is beyond a threshold. 
The solution I implement here is to restart the perf collection when the memory utilization growth is more than 100MB (which can be changed depending on use cases, we can also make it as command line argument), compared the first memory RSS size. 

## Checklist:

### DIFF
--- a/container.Dockerfile
+++ b/container.Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine as gprofiler
+FROM alpine AS gprofiler
 
 ARG ARCH
 ARG EXE_PATH=build/${ARCH}/gprofiler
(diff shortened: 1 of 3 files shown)


mlim19:
This PR merge is on hold because Prashant mentioned that https://github.com/intel/gprofiler/pull/1002 resolves most issues related memory usage. We will make a decision whether to merge it or not sometime after merging https://github.com/intel/gprofiler/pull/1002

PR #1047, opened 2026-07-01
Fix security vulnerability: restore ownership validation in mkdir_owned_root

mlim19:
## Summary

  - Fix regression in `mkdir_owned_root()` where ownership validation was bypassed
  - Add ownership validation for non-root case in `mkdir_owned_root_wrapper()`
  - Add symlink checks to prevent symlink-based attacks

## Background

Commit `5285bef7` ("Support java profiling without root privilege in a container #987") accidentally introduced an unconditional `return` statement that bypassed the ownership validation logic:

```python
if path.exists():
    return                                    # <- BUG: makes code below unreachable
    if is_root() and is_owned_by_root(path):  # <- Dead code
        return
    shutil.rmtree(path)                       # <- Never executed
```

This regression created a local privilege escalation vulnerability where a low-privilege attacker could pre-create `/tmp/gprofiler_tmp` before gProfiler runs as root. The ownership check that should remove and recreate non-root-owned directories was bypassed.

## Test plan

- [ ] Verify `mkdir_owned_root()` removes non-root-owned directories when running as root
- [ ] Run existing unit tests
- [ ] Manual test: create `/tmp/gprofiler_tmp` as non-root user, run gProfiler as root, verify directory is recreated as root-owned

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/gprofiler/utils/fs.py
+++ b/gprofiler/utils/fs.py
@@ -78,30 +78,42 @@ def is_owned_by_root(path: Path) -> bool:
     return statbuf.st_uid == 0 and statbuf.st_gid == 0
 
 
+def is_owned_by_current_user(path: Path) -> bool:
+    """Check if path is owned by the current user."""
+    statbuf = path.stat()
+    return statbuf.st_uid == os.getuid()
+
+
 def mkdir_owned_root_wrapper(path: Union[str, Path], mode: int = 0o755) -> None:
     """
-    Ensures a directory exists and writable.
+    Ensures a directory exists and is owned by the current user.
 
-    If the directory exists and is not writable, the function raises.
-    If the directory exists and is not writable, it is left as is.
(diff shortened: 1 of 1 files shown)


dkorlovs (dismissed):

dkorlovs:
@mlim19 do we go through this flow in case of " java profiling without root privilege"? If yes, then correct way to implement it, is to check that user who run gprofiler and owner of directory are the same, i.e. check not only root.

mlim19:
> @mlim19 do we go through this flow in case of " java profiling without root privilege"? If yes, then correct way to implement it, is to check that user who run gprofiler and owner of directory are the same, i.e. check not only root.

Thanks you for the review. I agree with your  point. I updated the change. please take a look

dkorlovs (approved):

